PR #1367, opened 2014-06-17
Fix the bug that 'flatMap' swallows OnErrorNotImplementedException

### LINKED ISSUE #1365: Exception in the inner observable of flatMap not rethrown
Hi all,
I'm trying to figure out, how errors are handled in RxJava (0.19.1). Basically, I'd expect the fundamental rule to be something like: _Any exception thrown within an operator is either handed to the onError handler of subscribe if one is provided, or rethrown if none is provided._

This works for the map operator: 

``` scala
def throwOnB2(prefix: String) = Observable.from(List(1, 2, 3))
  .map(x ⇒ prefix + x.toString)
  .map(s ⇒ if (s == "b2") throw new Exception("Bang!") else s)

throwOnB2("b").subscribe(next ⇒ println(next))
```

throws a `rx.exceptions.OnErrorNotImplementedException: Bang!` as expected.

``` scala
throwOnB2("b").subscribe(
  next ⇒ println(next),
  err ⇒ println(s"error: ${err}"),
  () ⇒ println("completed")
)
```

prints `error: java.lang.Exception: Bang!` as expected.

For nested observables joined using `flatMap`

``` scala
def nested = Observable
  .from(List("a", "b", "c"))
  .flatMap(prefix ⇒ throwOnB2(prefix))

nested
  .subscribe(
    next ⇒ println(next),
    err ⇒ println(s"error: ${err}"),
    () ⇒ println("completed")
  )
```

prints `error: java.lang.Exception: Bang!` as expected,

**But:**

``` scala
nested.subscribe(next ⇒ println(next))
```

prints

```
a1
a2
a3
b1
```

and finishes without throwing any exception. 

Is this the intended behavior? For me it is quite unfortunate, since bugs in my code can easily slip away unnoticed, but perhaps there is something I don't see.

Thank you, best regards

zsxwing:
Fixed #1365
Not sure if any other place will swallow `OnErrorNotImplementedException` or the fatal errors.

### DIFF
--- a/rxjava-core/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/rxjava-core/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -20,6 +20,7 @@ import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import rx.Observable;
 import rx.Observable.Operator;
 import rx.Subscriber;
+import rx.exceptions.Exceptions;
 import rx.observers.SerializedSubscriber;
 import rx.subscriptions.CompositeSubscription;
 
@@ -111,6 +112,7 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
 
         @Override
         public void onError(Throwable e) {
+            Exceptions.throwIfFatal(e);
             if (ONCE_UPDATER.compareAndSet(this, 0, 1)) {
                 parent.onError(e);
             }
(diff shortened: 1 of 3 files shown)


rekhajoshm:
Think the issue still exists, there needs to be clear functionality by which exception can be cleanly trapped and rethrown if needed(user preference) to upper layers.thanks

PR #1618, opened 2014-08-20
Non blocking future to observable implementation.

tbak:


akarnokd:
Hello, here are my observations:

 - I think using the Schedulers.computation for periodic polling is a bit disruptive. Instead, I suggest using the ```GenericScheduledExecutor.INSTANCE``` for this purpose and always schedule the consumption on the scheduler supplied via parameter.
 - In addition, having the constant poll frequencies seems to be problematic: it introduces uncontrollable latency in accepting the resulting value. Perhaps a single poll frequency could be exposed as a parameter.
 - Is there a reason you don't use ```worker.schedule(Action0, long, long, TimeUnit)``` for the periodic scheduling?
 - @benjchristensen should this be "```@Final```", ```@Beta``` or ```@Experimental```; do we want to delay this until 1.1?

The PR is quite old and can't be merged. Once some decisions have been made, Could you perhaps post a new PR based on the most recent 1.x main branch with a single commit containing all changes?

akarnokd:
I'm closing this because not meeting the requirements. Please submit a new PR if you want to pursue the problem.

Issue #1671, opened 2014-09-05
Backpressure: SerializedSubscriber


benjchristensen:
Can we make `SerializedSubscriber` correctly be implemented with a bounded buffer and support reactive pull backpressure?

This should also ensure that `Observable.serialize()` works correctly with backpressure, both bounding its internal queue and chaining through the `Producer.request` calls.

akarnokd:
I don't believe we should add backpressure support to ```SerializedSubscriber```: it is a write-through from the perspective of backpressure: if downstream requests 5, ```SerializedSubscriber``` will request 5. It is not meant to run with backpressure-ignoring upstream and the```onBackpressureXXX``` are there to handle it after.

I vote on closing this issue.

akarnokd:
We use it mainly to serialize out `onNext` vs `onError` | `onCompleted`, for example in `takeUntil`, and not for serializing concurrent `onNext` calls (as far as I can remember).

PR #1774, opened 2014-10-17
Serialized backpressure support proposal.

akarnokd:
For #1671

It took me a while to get it working and I'm not sure what tests to write to check backpressure behavior besides the request 5s from a sequence of 15.

In addition, it was a bit confusing when to call onStart, what exactly the setProducer call supposed to do and what to do when noone calls setProducer from upstream.


akarnokd:
This PR is kind of old and lots of things happened in lock-free world recently so I'll revisit the issue again in a new PR.

Issue #1804, opened 2014-10-28
"Interrupted while waiting for subscription to complete." in 1.0.0-rc8


edenman:
Doesn't happen on rc7.  I can repro 100% of the time.  Happens on both KitKat and Lollipop so I don't think it's specific to any particular Android version.

```
FATAL EXCEPTION: RxCachedThreadScheduler-4
Process: com.mycompany.myapp.development, PID: 4413
java.lang.IllegalStateException: Exception thrown on Scheduler.Worker thread. Add `onError` handling.
        at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:50)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:152)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:265)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
        at java.lang.Thread.run(Thread.java:818)
Caused by: rx.exceptions.OnErrorNotImplementedException: Interrupted while waiting for subscription to complete. 
        at rx.Observable$30.onError(Observable.java:7252)
        at rx.observers.SafeSubscriber._onError(SafeSubscriber.java:127)
        at rx.observers.SafeSubscriber.onError(SafeSubscriber.java:96)
        at rx.internal.operators.OperatorMap$1.onError(OperatorMap.java:48)
        at rx.internal.operators.OperatorMap$1.onNext(OperatorMap.java:56)
        at rx.internal.operators.NotificationLite.accept(NotificationLite.java:150)
        at rx.subjects.SubjectSubscriptionManager$SubjectObserver.emitNext(SubjectSubscriptionManager.java:254)
        at rx.subjects.BehaviorSubject.onNext(BehaviorSubject.java:166)
        at rx.Observable$34.onNext(Observable.java:7437)
        at rx.observers.SafeSubscriber.onNext(SafeSubscriber.java:104)
[...]

PR #1840, opened 2014-11-10
Unsubscribe when thread is interrupted

roman-mazur:
Originated from discussion #1804.

I was considering to make it opt-in, but it would drastically increase number of methods in public API.
What do you think?

### DIFF
--- a/src/main/java/rx/internal/operators/BlockingOperatorLatest.java
+++ b/src/main/java/rx/internal/operators/BlockingOperatorLatest.java
@@ -96,6 +96,7 @@ public final class BlockingOperatorLatest {
                     try {
                         notify.acquire();
                     } catch (InterruptedException ex) {
+                        unsubscribe();
                         Thread.currentThread().interrupt();
                         iNotif = Notification.createOnError(ex);
                         throw Exceptions.propagate(ex);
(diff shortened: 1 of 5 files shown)


src/main/java/rx/internal/operators/BlockingOperatorLatest.java
```diff
@@ -49,7 +50,7 @@ private BlockingOperatorLatest() {
             @Override
             public Iterator<T> iterator() {
                 LatestObserverIterator<T> lio = new LatestObserverIterator<T>();
-                source.materialize().subscribe(lio);
+                lio.subscription = source.materialize().subscribe(lio);
```
akarnokd: This set races with the dereference of L101 and may cause NPE. Besides, since it would be set to the value ```lio``` essentially (plus a transparent SafeSubscriber layer), you don't really need this field. Just call ```this.unsubscribe()``` on L101.

src/main/java/rx/internal/operators/BlockingOperatorNext.java
```diff
@@ -65,6 +66,7 @@ private BlockingOperatorNext() {
         private boolean isNextConsumed = true;
         private Throwable error = null;
         private boolean started = false;
+        private Subscription subscription;
```
akarnokd: This field is also unnecessary. Call observer.unsubscribe() directly on L123.

roman-mazur:
Edited per @akarnokd's remarks and rebased.

akarnokd:
Looks good to me now. Thanks.

PR #1911, opened 2014-12-01
Atomic operations JMH benchmarks

akarnokd:
Created some JMH benchmarks for the typical volatile and atomic operations.

### DIFF
--- a/src/perf/java/rx/internal/AtomicPerf.java
+++ b/src/perf/java/rx/internal/AtomicPerf.java
@@ -265,4 +265,16 @@ public class AtomicPerf {
             bh.consume(VolatileLongFieldState.UPDATER.compareAndSet(state, 1L, 2L));
         }
     }
+    @Benchmark
+    public void atomicIntGetAndSet(AtomicIntState state, Times repeat, Blackhole bh) {
+        for (int i = 0; i < repeat.times; i++) {
+            bh.consume(state.value.getAndSet(i));
+        }
+    }
+    @Benchmark
+    public void atomicLongGetAndSet(AtomicLongState state, Times repeat, Blackhole bh) {
+        for (long i = 0; i < repeat.times; i++) {
+            bh.consume(state.value.getAndSet(i));
+        }
+    }
 }


artem-zinnatullin:
Sorry for coming into the year old thread (I'm reviewing RxJava sources…), but my runs of `./gradlew benchmark` for `OperatorObserveOn` show that switching to plain Atomic fields intead of `volatile` + `AtomicFieldUpdater` give pretty significant performance boost (I'll do more tests tomorrow and probably will open a PR).

I'm not a JMH master, but as far as I understand, tests in this PR are not very good for one big reason — loops. AFAIK loops in benchmark tests are bad because JIT is too smart about them, see http://java-performance.info/jmh/.

If I remove loops from these tests I see that AtomicFieldUpdaters have 25-30% overhead, not 10% (especially in `getAndIncrement` and `incrementAndGet`).

@akarnokd @benjchristensen thoughts? Probably it's a reason to reconsider switch to `volatile` + `AtomicFieldUpdater` instead of `Atomic*`, of course in some (rare) cases it's good because of slightly lower memory usage, but reflection costs too much, especially on Android :(

I guess, @benjchristensen in your usage cases (Netflix) you would like to have faster code than save few bytes of memory. For Android, it will be nice to avoid this reflection too.

akarnokd:
> benchmark tests are bad because JIT is too smart about them

Except when there are atomic operations inside the loop.

> I see that AtomicFieldUpdaters have 25-30% overhead

JIT can sometimes properly eliminate the class-validation in the updater and it should run as fast as Unsafe. Sometimes not, most likely when there are some other implementation of the host class around.

> some (rare) cases it's good because of slightly lower memory usage, but reflection costs too much, especially on Android

Its a tradeoff between portability, memory usage, the cost of validation vs dereference. Do you know a benchmark tool for Android where these cases could be run?

artem-zinnatullin:
>Its a tradeoff between portability, memory usage, the cost of validation vs dereference.

Since RxJava needs a lot of `AtomicInteger`s and `AtomicLong`s and also RxJava should be fast and memory-friendly, what do you think about Object Pools for them? 

It will: 

* Remove some amount of memory allocations
* Remove need in reflection (`AtomicFieldUpdater`)
* Make operators faster 

> Do you know a benchmark tool for Android where these cases could be run?

No :( 

But we can create one similar to JMH :)

akarnokd:
> RxJava should be fast and memory-friendly

Concurrent operators can become faster by padding away heavily mutated counters, but at a cost of higher memory usage and increased class complexity. These latter two become drawbacks with short lived tasks that are run from a tight loop (i.e., benchmarking just().observeOn().subscribe(...) with JMH.

In addition, having a reference to Atomic classes may still end up with false sharing as GC may pack them next to each other or their owner class. Sometimes, even if they are further away, there could be a false-sharing issue with the object header itself. Java 8 Atomic variables *might* get padded, but that's unreliable. What's left is manually assembling classes with padding fields and either using Unsafe (unavailable in Android) or field updaters.

The same is true for our copies of the JCTools queues: class padding may help, element padding may help up to a certain limit.

> what do you think about Object Pools for them

I generally try to avoid them because they can become the new contention point.

> Remove some amount of memory allocations

Allocation is pretty performant in desktop Java and if you don't run short-lived tasks that frequently, it isn't a problem. I saw the tendency for the need of optimizing for short-lived tasks (i.e., just() and Single) and I believe there is an upper bound on how few operations can achieve the functionality without endangering concurrent-safety.

> Remove need in reflection (AtomicFieldUpdater)

I think Android is in disadvantage here; I never had any problems with desktop Java reflection since Java 7. The 2-second long reflective lookup of all methods (NewThreadWorker issue) baffled me. I think the Android platform is so behind that it requires an independent RxJavaAndroid port that specifically knows about such problems.

> Make operators faster

Most benchmarks really test for operator overhead by doing nothing but blackholing source values. Many of these, by using boxed integers, test for GC+operator overhead together (although I think there should be perf test that specifically avoid GC interference).

artem-zinnatullin:
> I think Android is in disadvantage here;

Yeah :(

Though I don't like manual paddings because it's platform specific thing…

Okay, your arguments were enough convincing for me, let's leave `AtomicFieldUpdater` :)

akarnokd:
@artem-zinnatullin I didn't mean to discourage you to do performance imporvements, just saying that there are tradeoffs per platform and per implementation aim.

With `observeOn` we can check for the platform Android and run a different inner subscriber for each. For example, desktop could run a padded version and Android could run with AtomicXXX instances.

There is, however, another effect usually found with async benchmarks and the computation scheduler: each inner iteration may end up on different cores which may trigger more thread migrations and soak up interference from other programs. I've resolved this in 2.x with the introduction of the `Schedulers.single()` which routes to the same underlying thread all the time.

Finally, there is sometimes an effect going on I call emission-grabbing where the thread which requests grabs the emission logic and the whole source-sink ends up on the same thread, boosting the throughput considerably. In terms of the `observeOn` that means padding optimization may not mean too much because there is going to be little-to-none concurrency happening there.

For example, in the range test, I've added a pipelined version where the emission is forced on a different thread than the observation:

image

As you can see, pipelined has 6x less throughput than non-pipelined. Also note that 2.x observeOn is 1  cache line padded (the usual padding is 2x for compensating for adjacent prefetch, but I was a bit sloppy).

That being said, I'm always open to optimizations. Would you like to pursue this further?

artem-zinnatullin:
>I didn't mean to discourage you to do performance imporvements, just saying that there are tradeoffs per platform and per implementation aim.

Yeah, I understand.

>With observeOn we can check for the platform Android and run a different inner subscriber for each. For example, desktop could run a padded version and Android could run with AtomicXXX instances.

This requires JMH-like tool for Android first. I'll work on it in my spare time, feel free to review/comment/follow https://github.com/artem-zinnatullin/droid-jmh

>As you can see, pipelined has 6x less throughput than non-pipelined.

Wow.

>That being said, I'm always open to optimizations. Would you like to pursue this further?

Yep, I'll continue my "investigations" :)

nitsanw:
The CAS vs getAndAdd conclusion is somewhat misleading. From JDK8 getAndAdd is intrinsified into XADD, which scales better than CAS under contention. To see the effect measure with different thread counts.
The volatile read cost is misleading in the sense that side effects of volatile are not measured/discussed. In the single value read there should be no difference. Volatile reads can stop many other memory reads from optimizing.

PR #1955, opened 2014-12-11
OnBackpressureXXX: support for common drain manager & fix for former concurrency bugs

akarnokd:
This is quite a complex operator with lots of cases.

Properties:
1) If there aren't any elements queued up and nothing is requested but terminal event received, emit terminal event and quit.
2) If there are elements in the queue and a terminal flag, and at least the same amount is requested, deliver the events and the terminal event.
3) If more was requested and more became available just after the loop and before the synchronized block, keep looping.
3.a) If more was requested but nothing is available or nothing was requested and something is available: quit and let either the onNext or request do the subsequent drain.
3.b) If elements and termination was produced but not requested: quit and let the request do the drain
3.c) If termination was requested and no elements produced: loop , emit terminal event and quit.

In table form:
```
Available | Terminated | Requested | Action | Reason
   yes         yes          yes       loop    can deliver available
   yes         yes          false     quit    can't deliver available
   yes         no           yes       loop    can deliver available
   yes         no           no        quit    can't deliver available
   no          yes          yes       loop    loop will deliver termination only and quit
   no          yes          no        loop    loop will deliver termination only and quit
   no          no           yes       quit    nothing to deliver
   no          no           no        quit    nothing to deliver
```

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
+++ b/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
@@ -126,7 +126,7 @@ public class OperatorOnBackpressureBuffer<T> implements Operator<T, T> {
         @Override
         public Object poll() {
             Object value = queue.poll();
-            if (capacity != null) {
+            if (capacity != null && value != null) {
                 capacity.incrementAndGet();
             }
             return value;
--- a/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
+++ b/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
@@ -15,7 +15,6 @@
  */
 package rx.internal.operators;
 
-import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
@@ -25,10 +24,9 @@ import rx.Producer;
 import rx.Subscriber;
 import rx.exceptions.MissingBackpressureException;
 import rx.functions.Action0;
+import rx.internal.util.BackpressureDrainManager;
 
 public class OperatorOnBackpressureBuffer<T> implements Operator<T, T> {
-
-    private final NotificationLite<T> on = NotificationLite.instance();
 
     private final Long capacity;
     private final Action0 onOverflow;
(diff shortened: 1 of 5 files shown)


src/main/java/rx/internal/util/BackpressureDrainManager.java
```diff
@@ -0,0 +1,241 @@
+/*
+ * Copyright 2011 David Karnok
```
akarnokd: Wrong copyright, should be Netflix

akarnokd:
Re-trigger travis.

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -52,122 +50,110 @@ public OperatorOnBackpressureBuffer(long capacity, Action0 onOverflow) {
...
+                child.onCompleted();
             }
-        };
-        
-        // if child unsubscribes it should unsubscribe the parent, but not the other way around
-        child.add(parent);
+        }
+        @Override
+        public Object peek() {
+            return queue.peek();
```
zsxwing: Need to add `capacity.incrementAndGet();` here.

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -52,122 +50,110 @@ public OperatorOnBackpressureBuffer(long capacity, Action0 onOverflow) {
...
+        }
+        @Override
+        public Object poll() {
+            return queue.poll();
```
zsxwing: Need to add capacity.incrementAndGet(); here.

zsxwing:
Could you change CRLFs to LFs in BackpressureDrainManager?

src/main/java/rx/internal/util/BackpressureDrainManager.java
```diff
@@ -0,0 +1,238 @@
...
+                        }
+                        if (n == 0) {
+                            break;
+                        }
+                    }
+                    o = a.poll();
+                    if (o == null) {
+                        break;
+                    } else {
+                        if (a.accept(o)) {
```
zsxwing: Looks `a.accept(o)` will always return `false`. Right?

src/main/java/rx/internal/util/BackpressureDrainManager.java
```diff
@@ -0,0 +1,240 @@
...
+     */
+    public BackpressureDrainManager(BackpressureQueueCallback actual) {
+        this.actual = actual;
+    }
+    /**
+     * Checks if a terminal state has been reached.
+     * @return true if a terminal state has been reached
+     */
+    public final boolean isTerminated() {
+        return terminated;
```
zsxwing: `isTerminated`, `terminate`, `terminate` are not used. Could you explain why adding them?

zsxwing: This is really a complex class. So I prefer to keep it as simple as possible.

akarnokd:
> Looks a.accept(o) will always return false. Right?

We can't know. The consumer way want to terminate after a certain item has been accepted.

> isTerminated, terminate, terminate are not used. Could you explain why adding them?

This is a base class, some operators may need to do something after termination and before calling drain.

> This is really a complex class. So I prefer to keep it as simple as possible.

This is an internal base class to help implement backpressure over an abstract queue. The unavoidable complexity is in ```drain()```.

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -52,122 +50,114 @@ public OperatorOnBackpressureBuffer(long capacity, Action0 onOverflow) {
...
-        child.add(parent);
+        }
+        @Override
+        public Object peek() {
+            return queue.peek();
+        }
+        @Override
+        public Object poll() {
+            Object value = queue.poll();
+            if (capacity != null) {
```
zsxwing: Sorry that I missed a case. If we poll `null`, we should not increment `capacity`.

zsxwing:
LGTM now.

akarnokd:
@benjchristensen your review would be welcome.

PR #2091, opened 2014-12-19
Added Operator switchIfEmpty

### LINKED ISSUE #1878: defaultIfEmpty(Observable<T> t)
Went to use `defaultIfEmpty` and found it only accepts a single value .... instead of an `Observable` which means I can't pass in an `Observable.error`.

It would be very helpful to have `defaultIfEmpty(Observable<T> t)`

alexwen:
Like defaultIfEmpty but subscribes to and emits the items in an Observable if the source is empty.

Fixes #1878

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
+++ b/src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
@@ -35,7 +35,9 @@ public class OperatorSwitchIfEmpty<T> implements Observable.Operator<T, T> {
 
     @Override
     public Subscriber<? super T> call(Subscriber<? super T> child) {
-        return new SwitchIfEmptySubscriber(child);
+        final SwitchIfEmptySubscriber parent = new SwitchIfEmptySubscriber(child);
+        child.add(parent);
+        return parent;
     }
 
     private class SwitchIfEmptySubscriber extends Subscriber<T> {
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
```diff
@@ -0,0 +1,126 @@
...
+    private final Observable<T> alternate;
+
+    public OperatorSwitchIfEmpty(Observable<T> alternate) {
+        this.alternate = alternate;
+    }
+
+    @Override
+    public Subscriber<? super T> call(Subscriber<? super T> child) {
+        final SwitchIfEmptySubscriber parent = new SwitchIfEmptySubscriber(child);
+        child.add(parent);
```
akarnokd: The problem with directly adding this to the child is that the child will retain a reference to a now-dead switcher. Instead, I suggest having a ```SerialSubscription```:

```java
SerialSubscription ssub = new SerialSubscription();
SwitchIfEmptySubscriber parent = new SwitchIfEmptySubscriber(child, ssub);
ssub.add(parent);
child.add(ssub);
return parent;
```
The ```subscribeToAlternate``` should now start with:
```java
ssub.set(alternate.unsafeSubscribe(new Subscriber<T>() {
```

PR #2238, opened 2014-12-24
Fix the bug that cache doesn't unsubscribe the source Observable when th...

### LINKED ISSUE #2191: Observable.cache() creates indefinite number of Threads with Schedulers.io()
Hi,

It would appear to me that there's an issue with using when using Observable.cache() and Schedulers.io().  The issue is that an indefinite amount of threads are created as opposed to having thread re-use.  

This can eventually lead the following fatal scenario: "java.lang.OutOfMemoryError: Unable to create new native thread".

I did notice that when using Observable.cache() and Schedulers.from(Executors.newCachedThreadPool()) Rx is able to re-use threads.  

Other than thread-caching configuration, it's not clear to me why these 2 implementations produce wildly different thread caching results and behavior.  I can only assume that this behavior is not intended.  It would be great to hear whether that's the case.

Here's a sample program I've put together that demonstrates the 2 different results.  It's a very simple example that creates multiple observables and ensures that they are subscribed to and that result is read.  I'm using SettableFuture to mock a lag between calls.

Result with Schedulers.from(Executors.newCachedThreadPool()) (last few items displayed, notice the re-use in threads):
Thread[pool-1-thread-18,5,main]
Thread[pool-1-thread-15,5,main]
Thread[pool-1-thread-20,5,main]
Thread[pool-1-thread-7,5,main]
Thread[pool-1-thread-6,5,main]
Thread[pool-1-thread-11,5,main]
Thread[pool-1-thread-17,5,main]
Thread[pool-1-thread-4,5,main]
Thread[pool-1-thread-21,5,main]
Thread[pool-1-thread-1,5,main]
Thread[pool-1-thread-18,5,main]
Thread[pool-1-thread-10,5,main]
Thread[pool-1-thread-7,5,main]
Thread[pool-1-thread-11,5,main]
Thread[pool-1-thread-20,5,main]

Result with Schedulers.io() (last few items displayed, notice that the number is incremental, no re-use):
Thread[RxCachedThreadScheduler-187,5,main]
Thread[RxCachedThreadScheduler-196,5,main]
Thread[RxCachedThreadScheduler-189,5,main]
Thread[RxCachedThreadScheduler-198,5,main]
Thread[RxCachedThreadScheduler-191,5,main]
Thread[RxCachedThreadScheduler-200,5,main]
Thread[RxCachedThreadScheduler-193,5,main]
Thread[RxCachedThreadScheduler-195,5,main]
Thread[RxCachedThreadScheduler-197,5,main]
Thread[RxCachedThreadScheduler-199,5,main]

public class RxIndefiniateThreads {

    static final Scheduler scheduler = Schedulers.from(Executors.newCachedThreadPool());

    // static final Scheduler scheduler = Schedulers.io();
    public static void main(String[] args) throws Exception {
        
        int i = 0;
        final AtomicInteger counter = new AtomicInteger(0);
        while (true) {
            final SettableFuture<Object> first = SettableFuture.create();
            final SettableFuture<Object> second = SettableFuture.create();
            Observable<Object> observe = Async.fromCallable(new Callable<Object>() {

                @Override
                public Object call() throws Exception {
                    System.out.println(Thread.currentThread());
                    return first.get();
                }

            }, scheduler).observeOn(scheduler).flatMap(new Func1<Object, Observable<Object>>() {

                @Override
                public Observable<Object> call(Object t1) {
                    System.out.println(Thread.currentThread());
                    try {
                        return Observable.just((Object) (t1.toString() + " :: " + second.get()));
                    } catch (Exception e) {
                        throw new IllegalStateException("Mock exception for second");
                    }
                }
            }).cache();

            final Future<Object> future = observe.toBlocking().toFuture();
            // Thread.sleep(2);
            i++;
            new Thread(new Runnable(){

                @Override
                public void run() {
                    int count = counter.incrementAndGet();
                    try {
                        Thread.sleep(1);
                        e.printStackTrace();
                    }
                    first.set("done");
                    try {
                        Thread.sleep(1);
                        e.printStackTrace();
                    }
                    second.set("done");
                    try {
                        future.get();
                    } catch (Exception e) {
                        e.printStackTrace();
                    }
                    // System.out.println("doneWith=" + count);
                }
                
            }).start();
            if (i == 100) {
                break;
            }
        }
        System.out.println("done");
    }
    
}

dependencies:
<dependency>
  		<groupId>com.google.guava</groupId>
  		<artifactId>guava</artifactId>
  		<version>18.0</version>
  	</dependency>
  	<dependency>
            <groupId>io.reactivex</groupId>
            <artifactId>rxjava</artifactId>
            <version>1.0.3</version>
        </dependency>
        <dependency>
            <groupId>io.reactivex</groupId>
            <artifactId>rxjava-async-util</artifactId>
            <version>0.21.0</version>
            <exclusions>
                <exclusion>
                    <groupId>io.reactivex</groupId>
                    <artifactId>rxjava</artifactId>
                </exclusion>
            </exclusions>
        </dependency>

zsxwing:
...e source is terminated

Fix #2191

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeCache.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeCache.java
@@ -64,9 +64,10 @@ public final class OnSubscribeCache<T> implements OnSubscribe<T> {
     @Override
     public void call(Subscriber<? super T> s) {
         if (SRC_SUBSCRIBED_UPDATER.compareAndSet(this, 0, 1)) {
-            source.unsafeSubscribe(Subscribers.from(cache));
+            source.subscribe(cache);
             /*
-             * Note that we will never unsubscribe from 'source' as we want to receive and cache all of its values.
+             * Note that we will never unsubscribe from 'source' unless we receive `onCompleted` or `onError`,
+             * as we want to receive and cache all of its values.
              * 
              * This means this should never be used on an infinite or very large sequence, similar to toList().
              */
--- a/src/test/java/rx/internal/operators/OnSubscribeCacheTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeCacheTest.java
@@ -17,6 +17,9 @@ package rx.internal.operators;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 
 import java.util.Arrays;
 import java.util.concurrent.CountDownLatch;
@@ -27,10 +30,10 @@ import org.junit.Test;
 
 import rx.Observable;
 import rx.Subscriber;
+import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.functions.Func1;
 import rx.functions.Func2;
-import rx.internal.operators.OnSubscribeCache;
 import rx.observers.TestSubscriber;
 import rx.schedulers.Schedulers;
 import rx.subjects.AsyncSubject;
@@ -147,5 +150,15 @@ public class OnSubscribeCacheTest {
     @Test
     public void testWithReplaySubjectAndRepeat() {
         testWithCustomSubjectAndRepeat(ReplaySubject.<Integer> create(), 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3);
+    }
+
+    @Test
+    public void testUnsubscribeSource() {
+        Action0 unsubscribe = mock(Action0.class);
+        Observable<Integer> o = Observable.just(1).doOnUnsubscribe(unsubscribe).cache();
+        o.subscribe();
+        o.subscribe();
+        o.subscribe();
+        verify(unsubscribe, times(1)).call();
     }
 }


akarnokd:
I managed to run the original code in #2191 and putting the sleep after the thread.start will use only a handful of io() threads. It confirms my suspicion that the original code simply floods the system with threaded activities and does not give them enough time to finish.

If the upstream completes, it should let any downstream object go (i.e., a completed ReplaySubject will throw away all of its subscribers). I've changed the code to allocate a large object and a memory dump right after the thread has completed shows no object retention. I don't see any issues here so could you elaborate on this issue/change (or on #2455)?

zsxwing:
> If the upstream completes, it should let any downstream object go (i.e., a completed ReplaySubject will throw away all of its subscribers).

I'm confused. This is not a memory leak bug. Instead, it's a bug that `OnSubscribeCache` doesn't unsubscribe the source after `onCompleted` or `onError`. See the unit test `testUnsubscribeSource` in this PR.

Because `unsubscribe` won't be called, the following codes in `CachedThreadScheduler` won't be called:

```Java
        @Override
        public void unsubscribe() {
            if (ONCE_UPDATER.compareAndSet(this, 0, 1)) {
                // unsubscribe should be idempotent, so only do this once
                CachedWorkerPool.INSTANCE.release(threadWorker);
            }
            innerSubscription.unsubscribe();
        }
```

Then, the worker won't be put back into the `expiringWorkerQueue`. That's why Schedulers.io() will create so many threads in #2191.

akarnokd:
Thanks, that makes sense now.

PR #2244, opened 2014-12-24
OperatorTakeLast add check for isUnsubscribed to fast path

davidmoten:
Went looking for a bug that wasn't a bug but found this anyway.  ```OperatorTakeLast``` doesn't check ```subscriber.isUnsubscribed()``` while emitting its queued events so I added a failing unit test and the fix.


zsxwing:
LGTM

PR #2330, opened 2015-01-05
Fixed off-by-one error and value-drop in the window operator

### LINKED ISSUE #2290: window time+size emits only 1 window
This code only emits 1 window. 

```java
        Observable.range(0, 10000).window(1, TimeUnit.SECONDS, 20).flatMap(o -> {
            System.out.println("window");
           return o.count(); 
        }).toBlocking().forEach(System.out::println);
```

There are 2 issues I have with it:

1) Why are there 21 instead of 20 elements in it?
2) Why does this only emit 1 window? Why does this not just behave like `window(20)` since the time never triggers?

akarnokd:
Fixes #2290


akarnokd:
I'll submit a new cleaned version.

PR #2331, opened 2015-01-05
Adds a isInitialised call to check if the BehaviorSubject is initialised

Dorvaryn:
Since the BehaviorSubject can be created without a default value, it could be interesting to be able to know if the subject is initialised (if it holds a value).
Adding this check allows for lazy initialisation of the behavior's value.

### DIFF
--- a/src/main/java/rx/subjects/BehaviorSubject.java
+++ b/src/main/java/rx/subjects/BehaviorSubject.java
@@ -176,4 +176,8 @@ public final class BehaviorSubject<T> extends Subject<T, T> {
     public boolean hasObservers() {
         return state.observers().length > 0;
     }
+
+    public boolean isInitialised() {
+        return state.get() != null;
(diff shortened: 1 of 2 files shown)


akarnokd:
Is there a particular use case you try to solve?

Dorvaryn:
I was looking at the BehaviorSubject to represent the value of some remote objects, since it caches the last value it is perfect to act as a local cache and if a refresh is triggered any observer would get the update.

Having this check would allow to use the empty construction and trigger the remote fetch only if the subject is uninitialised (first access) .

akarnokd:
Please check out #2470 for a set of Subject API changes that includes your is initialized check by the name hasValue.

Dorvaryn:
Closing this since it is addressed in a more complete way in https://github.com/ReactiveX/RxJava/pull/2470

PR #2332, opened 2015-01-05
Operator retry test fix attempt

akarnokd:
Maybe the error function was not emitting events properly.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -16,12 +16,9 @@
 package rx.observers;
 
 import java.util.List;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
-import rx.Notification;
-import rx.Observer;
-import rx.Subscriber;
+import rx.*;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/rx/observers/TestSubscriber.java
```diff
@@ -227,9 +227,9 @@ public void awaitTerminalEvent() {
      * @throws RuntimeException
      *          if the Subscriber is interrupted before the Observable is able to complete
      */
-    public void awaitTerminalEvent(long timeout, TimeUnit unit) {
+    public boolean awaitTerminalEvent(long timeout, TimeUnit unit) {
```
zsxwing: This is a breaking change.

akarnokd: Any suggestion how I can sneak out the await's result?

zsxwing: Throw `TimeoutException`? It's a bit minor breaking change. @benjchristensen Is it fine?

zsxwing: Or add a new method?

akarnokd: I've changed it back and made it throw a RTE with TimeoutException as cause.

PR #2335, opened 2015-01-06
A set of stateless operators that don't need to be instantiated

akarnokd:
for ```lift()``` all the time.


benjchristensen:
That's a nice change. This should improve GC behavior. Have you been able to measure it?

davidmoten:
There's a javadoc typo that has been copy pasted throughout: 'an singleton instance'.

akarnokd:
I haven't measured the GC gain; but I guess your use case that required queue pooling might benefit from another 16 bytes of allocation-reduction.

akarnokd:
The following program prints 240 bytes/instance on head and 225 on this PR; that is ~15 MB of heap saved
```java
public class MergeAlloc {
        MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
        MemoryUsage memHeap = memoryMXBean.getHeapMemoryUsage();
        long initial = memHeap.getUsed();
        
        int m = 1_000_000;
        
        System.out.printf("Starting: %.3f MB%n", initial / 1024.0 / 1024.0);
        
        List<Observable<Integer>> list = new LinkedList<>();
        
        for (int i = 0; i < m; i++) {
            list.add(Observable.merge(Observable.just(1), Observable.just(2)));
        }
        
        memHeap = memoryMXBean.getHeapMemoryUsage();
        long after = memHeap.getUsed();
        System.out.printf("Starting: %.3f MB%n", after / 1024.0 / 1024.0);
        
        System.out.printf("Cost: %d bytes/item", (after - initial) / m);
    }
}
```

akarnokd:
Triggering re-Travis.

PR #2447, opened 2015-01-12
Fail early if a null subscription is added to a CompositeSubscription.

jnlopar:
Otherwise, it'll just fail late when unsubscribing, which is much harder to trace.

I discovered this while writing  a unit test, when I didn't properly mock out a method to return a valid Observable. Seems like it would be more likely to happen in test code than production, but still could be a stumbling block that's hard to track down if there's a bug in app logic adding a null subscription.

### DIFF
--- a/src/main/java/rx/subscriptions/CompositeSubscription.java
+++ b/src/main/java/rx/subscriptions/CompositeSubscription.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2014 Netflix, Inc.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -55,8 +55,8 @@ public final class CompositeSubscription implements Subscription {
      *          the {@link Subscription} to add
      */
     public void add(final Subscription s) {
-        if (s == null) {
-            throw new NullPointerException("Added Subscription cannot be null.");
+        if (s.isUnsubscribed()) {
+            return;
         }
         Subscription unsubscribe = null;
         synchronized (this) {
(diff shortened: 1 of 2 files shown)


src/test/java/rx/subscriptions/CompositeSubscriptionTest.java
```diff
@@ -337,4 +337,11 @@ public void testTryRemoveIfNotIn() {
         
         csub.remove(csub1); // try removing agian
     }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testAddingNullSubscriptionIllegal() {
+      CompositeSubscription csub = new CompositeSubscription();
+      csub.add(null);
```
JakeWharton: Indentation should be four spaces here.

jnlopar: Fixed, thanks!

src/main/java/rx/subscriptions/CompositeSubscription.java
```diff
@@ -55,6 +55,9 @@ public synchronized boolean isUnsubscribed() {
      *          the {@link Subscription} to add
      */
     public void add(final Subscription s) {
+        if (s == null) {
+            throw new IllegalArgumentException("Added Subscription cannot be null.");
```
zsxwing: It's better to use `NullPointerException`. Other null assertions in RxJava throws NPE.

jnlopar: done! :)

akarnokd:
I suggest an alternative to get another effect: adding an already unsubscribed Subscription is essentially no-op and just blocks others adding regular Subscriptions concurrently. How about do an isUnsubscribed check in add:
```java
public void add(Subscription s) {
    if (s.isUnsubscribed()) {
        return;
    }
}
```

Certainly, this will get you NPE. I read somewhere that JIT may completely remove null-checks and simply relying on page faults of address 0. The drawback is that the NPE indicates an error in RxJava now instead of the contract violation. Thoughts?

jnlopar:
Done. Good thought. One concern though, is it possible for an unsubscribed subscription to become subscribed again? I can't think of a case where that happens, but if it could, would this still be desired behavior?

akarnokd:
Unsubscribed is a terminal state and can't be reviewd.

akarnokd:
Thanks. Merging.

PR #2456, opened 2015-01-15
Increase the timeout of testRetryWithBackpressure to 30 seconds

zsxwing:


akarnokd:
The problem is with the fail function not properly working, see: #2332

zsxwing:
Didn't notice it :)

Issue #1919, opened 2014-12-02
(Updated): timeout() operator leaking ScheduledFutureTasks on Scheduler


daschl:
Hi folks,

(update, see below, looks like it's the timeout operator) 

I'm raising a ticket earlier than later, because I suspect we have a leak somewhere in the Scheduler logic. It could by in my code too, but from the characteristics it looks more like something rx internal.

I've been able to generate some OOM scenarios where GC takes all the time to clean things up and it's not able to, which pointed me to a leak. A heap dump showed the leak suspect is the ScheduledFutureTask as part of the computation scheduler:

![screen shot 2014-12-02 at 10 40 28](https://cloud.githubusercontent.com/assets/29086/5260491/b9420760-7a0f-11e4-874f-e8e9ad22a918.png)

You can see a full GC kicking in without actually making old gen smaller:

![screen shot 2014-12-02 at 10 42 50](https://cloud.githubusercontent.com/assets/29086/5260509/044a317e-7a10-11e4-8cc3-b3a17bd1d1f6.png)

The code path uses both .observeOn() once to move it onto a scheduler and also blocks .single() at the very end. So I'll start looking in those places. But maybe @akarnokd or @benjchristensen you have an idea where I can start debugging?

akarnokd:
We've fixed the task retention problem for JDK 6 in 1.0.5. Does the problem still happen for you?

daschl:
@akarnokd oh you did? which changeset was that? :)

akarnokd:
@daschl with #2465. I must add, it only works with ```ScheduledThreadPoolExecutor``` that the JDK is using because the ```ScheduledExecutorService``` interface doesn't have a ```purge``` method.

akarnokd:
I'm closing this due to inactivity. If you have further problems, please don't hesitate to reopen this issue or post a new one.

PR #2465, opened 2015-01-19
ScheduledExecutorService: call purge periodically on JDK 6 to avoid

akarnokd:
cancelled task-retention.

First debated in #1922, see also #1919.

We may want to discuss the naming of system parameters. I chose these so RxJava 2.0 specific properties may be trivially separated:

``` io.reactivex.rxjava.scheduler.jdk6.purge-frequency-millis ```
Specifies the purge frequency in milliseconds. Default is 1000.

``` io.reactivex.rxjava.scheduler.jdk6.purge-force ```
Forces the use of the purge (if set to true) even if the setRemoveOnCancelPolicy is supported. The benefit is that removing cancelled tasks now runs on a different thread so the main pool thread doesn't waste time on them. The drawback is the retention window can be still to large.

Do we have a wiki page where such parameters are listed?

### DIFF
--- a/src/main/java/rx/internal/schedulers/NewThreadWorker.java
+++ b/src/main/java/rx/internal/schedulers/NewThreadWorker.java
@@ -16,10 +16,14 @@
 package rx.internal.schedulers;
 
 import java.lang.reflect.Method;
+import java.util.Iterator;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReference;
 
 import rx.*;
+import rx.exceptions.Exceptions;
 import rx.functions.Action0;
+import rx.internal.util.RxThreadFactory;
 import rx.plugins.*;
 import rx.subscriptions.Subscriptions;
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by akarnokd** [comment]
I did run the unit test by forcing a JDK 6 runtime and seems to work. Merging to allow progress on schedulers.

File: src/main/java/rx/internal/schedulers/NewThreadWorker.java [hunk 1]
```diff
@@ -30,24 +34,111 @@
     private final ScheduledExecutorService executor;
     private final RxJavaSchedulersHook schedulersHook;
     volatile boolean isUnsubscribed;
-
+    /** The purge frequency in milliseconds. */
+    private static final String FREQUENCY_KEY = "io.reactivex.rxjava.scheduler.jdk6.purge-frequency-millis";
```
benjchristensen: These don't match the naming convention using in `RxRingBuffer` with just the `rx` prefix: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/util/RxRingBuffer.java#L267

We should probably stick with that convention since it is already set, so:

`rx.scheduler.jdk6.purge-frequency-millis`
`rx.scheduler.jdk6.purge-force`

akarnokd: Okay.

akarnokd: I'd like to fix it in master but then it conflicts with #2579, which by itself also adds parameters with unconventional naming.

benjchristensen: Why can't we fix this then go work on #2579 which still needs to be reviewed and can be rebased?

akarnokd: Hystrix could use the enhancements in #2579 and since it affects the same file, it is much easier to change the values there in a plain commit than rebasing.

Issue #1897, opened 2014-11-20
Add current value property to BehaviorSubject


dalewking:
Rx 2.2 added a current value property that would come in handy. See discussion here: https://rx.codeplex.com/workitem/8

akarnokd:
Please check out #2470 for a set of Subject API changes that includes your current value.

akarnokd:
Subjects have been enhanced in 1.0.5 with methods that get you their current state.

PR #2470, opened 2015-01-20
Subject state information methods & bounded ReplaySubject termination

akarnokd:
fix.

This PR aims to support the requests in #2331 and #1897 by adding methods to the (final) subject classes that let developers access in-flight state in a snapshot fashion. 

This may be considered safe API change because all subject classes were final already so adding extra methods won't break anyone's code (and we will be careful with our Observable in the future).

  - ```NotificationLite``` is now extended with two missing value checks: isNull and isNext.
  - Each subject has ```hasCompleted```, ```hasThrowable``` and ```getThrowable``` methods, however, I can't add them to ```Subject``` because that would be an incompatible API change.
  - Where applicable, ```getValue``` and ```getThrowable``` return ```null``` instead of throwing exceptions so users are encouraged to call hasXXX methods beforehand.
  - There was a chaining bug in the bounded ReplaySubject: because the terminal value was added after a potential pruning, the node links could get broken and concurrent replays might not have seen the terminal value.
  - Since ```ReplaySubject``` can have multiple values, I've added ```size()```, ```hasAnyValue()``` (isEmpty is taken) and ```getValues()``` methods to make a snapshot of the current buffer contents whether or not the ```ReplaySubject``` has terminated (the usual toList() would wait until the subject has terminated). 
  - The unrelated ```OperatorMergeTest.testConcurrency``` hangs for me for some reason without activity (either a buffer bug or a merge bug is in play there). I've added a timeout so it doesn't stop the other tests.

### DIFF
--- a/src/main/java/rx/internal/operators/NotificationLite.java
+++ b/src/main/java/rx/internal/operators/NotificationLite.java
@@ -177,6 +177,23 @@ public final class NotificationLite<T> {
     }
 
     /**
+     * Indicates whether or not the lite notification represents a wrapped {@code null} {@code onNext} event.
+     * @param n the lite notification
+     * @return {@code true} if {@code n} represents a wrapped {@code null} {@code onNext} event, {@code false} otherwise
+     */
+    public boolean isNull(Object n) {
+        return n == ON_NEXT_NULL_SENTINEL;
+    }
+
+    /**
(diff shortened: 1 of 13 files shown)


akarnokd:
I'm merging this: changes are either internal or annotated with experimental on final classes.

Issue #1702, opened 2014-09-26
TrampolineScheduler NullPointerException


DylanSale:
There is an issue in TrampolineScheduler on 0.20.4 where it can  throw a NullPointerException in InnerCurrentThreadScheduler `enqueue(Action0 action, long execTime)`

``` java
        private Subscription enqueue(Action0 action, long execTime) {
            if (innerSubscription.isUnsubscribed()) {
                return Subscriptions.empty();
            }
            PriorityQueue<TimedAction> queue = QUEUE.get();
            final TimedAction timedAction = new TimedAction(action, execTime, COUNTER_UPDATER.incrementAndGet(TrampolineScheduler.this));
            queue.add(timedAction);

            if (wip.getAndIncrement() == 0) {
                do {
                    queue.poll().action.call();
                } while (wip.decrementAndGet() > 0);
                return Subscriptions.empty();
            } else {
                // queue wasn't empty, a parent is already processing so we just add to the end of the queue
                return Subscriptions.create(new Action0() {

                    @Override
                    public void call() {
                        PriorityQueue<TimedAction> _q = QUEUE.get();
                        if (_q != null) {
                            _q.remove(timedAction);
                        }
                    }

                });
            }
        }
```

The Exception happens on `queue.poll().action.call();`.
From what I can tell, the queue is empty, and poll is returning null.

This is happening inside a nested `observable.redo().timeout().redo()` chain (the `redo().timeout()` happens earlier in the application, it isn't just one after the other like that), if that helps. I'm not in a position to create a simple replication example however.

My current theory is: One of the redos could be unsubscribing, calling `_q.remove`, while the `wip` loop is running (perhaps unsubscribe is happening on another thread?), causing the queue to become empty, and the next loop of the `wip` loop returns `null`. I think the `queue.poll()` call should check for null.

akarnokd:
Hi. Fixed in 1.x branch via #2471. Can you verify the fix works for your case?

PR #2471, opened 2015-01-20
Fixes NPEs reported in ReactiveX#1702 by synchronizing queue.

jnlopar:
Also adds a unit test for regression.

It appears there is a potential race condition if something adds to/removes from the PQ while it's *inside* the 'poll' operation, which is where the exceptions in #1702 seem to have actually come from. Therefore, the initial null check didn't really address the original problem. The test here seems to reliably recreate those conditions.

I considered using a PriorityBlockingQueue instead of synchronized, but since the isEmpty and poll calls should not allow something to interleave between them and access the queue, a synchronized block seemed wiser here.

### DIFF
--- a/src/main/java/rx/schedulers/TrampolineScheduler.java
+++ b/src/main/java/rx/schedulers/TrampolineScheduler.java
@@ -45,12 +45,13 @@ public final class TrampolineScheduler extends Scheduler {
     /* package accessible for unit tests */TrampolineScheduler() {
     }
 
-    private class InnerCurrentThreadScheduler extends Scheduler.Worker implements Subscription {
+    private static class InnerCurrentThreadScheduler extends Scheduler.Worker implements Subscription {
 
+        private static final AtomicIntegerFieldUpdater COUNTER_UPDATER = AtomicIntegerFieldUpdater.newUpdater(InnerCurrentThreadScheduler.class, "counter");
+        volatile int counter;
         private final PriorityBlockingQueue<TimedAction> queue = new PriorityBlockingQueue<TimedAction>();
         private final BooleanSubscription innerSubscription = new BooleanSubscription();
         private final AtomicInteger wip = new AtomicInteger();
-        private final AtomicInteger counter = new AtomicInteger();
 
         @Override
         public Subscription schedule(Action0 action) {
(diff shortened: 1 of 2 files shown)


src/main/java/rx/schedulers/TrampolineScheduler.java
```diff
@@ -71,14 +71,17 @@ private Subscription enqueue(Action0 action, long execTime) {
...
             if (wip.getAndIncrement() == 0) {
                 do {
-                    TimedAction polled = queue.poll();
-                    // check for null as it could have been unsubscribed and removed
-                    if (polled != null) {
-                        polled.action.call();
+                    synchronized (queue) {
+                        if (!queue.isEmpty()) {
+                            TimedAction polled = queue.poll();
+                            polled.action.call();
```
akarnokd: You should not call the action while holding the lock. Just do a synchronized poll and call the action if you got a non-null value.

In addition, I think the ```COUNTER_UPDATER``` and ```counter``` can be moved into the worker itself because it can ensure total order locally and we can avoid a global serialization point.

jnlopar: Good call. I fixed the wrapping. Now I'm just using poll instead of isEmpty, since poll returns null iff isEmpty.

akarnokd:
Almost good. Two things:

  - Since we only use offer/poll, you can use the unbounded PriorityBlockingQueue now
  - Could you move ```counter``` and ```COUNTER_UPDATER``` into the ```InnerCurrentThreadScheduler ``` class, both non-static fields?

akarnokd:
We are using AtomicFieldUpdaters to save on the instance cost of ```AtomicInteger```s. Could you change the counter to use ```AtomicLongFieldUpdater```?

jnlopar:
Done. On a related but side-note, do you know if these field updaters are proguard-safe for obfuscation? They reflect on field name for sure. It might be worth looking into. The fix would probably be implementing the abstract setters and getters, which would obviously be a bit more boilerplate and kinda less fun, but might be worth it if it reduces custom proguard rules required to use rx.

akarnokd:
Thanks for the changes, looks good to me. I don't know about ProGuard.

jnlopar:
Looks like per the proguard manual, it does indeed recognize FieldUpdater declarations, so this should be fine.

JakeWharton:
Wow I did not expect that. How uncharacteristically useful of them! Good to know for the future.

jnlopar:
http://proguard.sourceforge.net/manual/introduction.html

See the 'Reflection' section (no anchor link). That lists the basic reflection methods it automatically detects.

Issue #2192, opened 2014-12-23
non-deterministic test failure: testRetryWithBackpressure


benjchristensen:
rx.internal.operators.OperatorRetryTest > testRetryWithBackpressure FAILED
    java.lang.Exception: test timed out after 10000 milliseconds

akarnokd:
Fixed in master.

PR #2472, opened 2015-01-20
Cleans up and deflakes up testRetryWithBackpressure

jnlopar:
This should solve #2192, since this test has been failing most of the time on travis.


akarnokd:
That test has been fixed and merged already: the error function's backpressure logic was incorrect.

jnlopar:
Cool, closing. Would you be interested if I did a separate cleanup task using the assertions on TestSubscriber instead of mockito verify? It seems cleaner and easier to read, and we can use an actual Observer/TestObserver instead of the mock.

PR #2477, opened 2015-01-21
Backpressure tests fix0121

akarnokd:
- Increased the limits of ```testMergeAsyncThenObserveOn```
 - Fixed ```testOnBackpressureDrop``` because in case there was no drop at all, the last value would be NUM - 1.
 - Changed ```SLOW_PASS_THRU``` to perform non-trivial work that can't be optimized away by JIT.


akarnokd:
Merging this so other PRs failing due these methods can be re-tested.

Issue #2480, opened 2015-01-21
Non deterministic test: OperatorMergeTest.testBackpressureUpstream2


davidmoten:
```OperatorMergeTest.testBackpressureUpstream2``` fails 2 out of 3 times when running the whole suite in Eclipse on an old 2 core laptop. The test passes every time if run singly. 

See also discussion about test isolation #1624

akarnokd:
I usually put a loop around such test methods and get the failure consistently. I'll investigate ```OperatorMergeTest.testBackpressureUpstream2```.

akarnokd:
Could you verify #2482 does work for you?

akarnokd:
I'm closing this due to inactivity. If you have further problems, please don't hesitate to reopen this issue or post a new one.

Issue #2466, opened 2015-01-19
Out of requests in merge


grzechum:
Below test, when run, quickly goes out of requests, merge operator is quite complex and it is difficult to say what exactly causes the problem but it looks like race condition, because each run gives different results.

This may be related to #1941 but scenario is different.

```java
	private final class OnSubscribeImpl implements OnSubscribe<String> {
		private final int id;
		private final Semaphore requested = new Semaphore(0); // back pressure control

		public OnSubscribeImpl(int id) {
			this.id = id;
		}

		@Override
		public void call(Subscriber<? super String> s) {
			s.setProducer(new Producer() {
				@Override
				public void request(long n) {
					System.out.println(id + " - REQUESTED: " + n);
					requested.release((int) n);
				}
			});

			try {
				long counter = 0;
				while (true) {
					requested.acquire();
					System.out.println(id + " - REMAININIG: " + requested.availablePermits());

					s.onNext(id + " - " + (counter++));
				}
			} catch (InterruptedException ex) {
				ex.printStackTrace();
			}
		}
	}

	@Test
	public void testX() throws InterruptedException {

		Observable<String> obs0 = Observable.create(new OnSubscribeImpl(0)).subscribeOn(Schedulers.io());
		Observable<String> obs1 = Observable.create(new OnSubscribeImpl(1)).subscribeOn(Schedulers.io());

		Observable.merge(obs0, obs1).observeOn(Schedulers.computation()).forEach(s -> System.out.println(s));

		Thread.sleep(Long.MAX_VALUE);
	}
```

benjchristensen:
This will require some time to dig into.

grzechum:
Thanks,

I found a workaround for this, which solves the problem for low and medium load.
- remove backpressure control from producer
- add onBackpressureBlock() after subscribeOn()

```java
	private final class OnSubscribeImpl implements OnSubscribe<String> {
		private final int id;

		public OnSubscribeImpl(int id) {
			this.id = id;
		}

		@Override
		public void call(Subscriber<? super String> s) {
			try {
				long counter = 0;
				while (true) {
					s.onNext(id + " - " + (counter++));
				}
			} catch (Exception ex) {
				ex.printStackTrace();
			}
		}
	}

	@Test
	public void testX() throws InterruptedException {

		Observable<?> obs0 = Observable.create(new OnSubscribeImpl(0)).subscribeOn(Schedulers.io())
				.onBackpressureBlock();
[...]

Issue #1941, opened 2014-12-08
OperatorMerge does not request enough


davidmoten:
Merging of async sources using the backpressure path can result in a stalled stream because not enough items are requested of the InnerSubscribers.

The test below fails 6 out of 10 runs.

The situation is this:

Given a merge of two slowish (1/ms) asynchronous sources that downstream has a take(N) where N >256, then 6 out 10 runs ```OperatorMerge``` requests 128 (```RxRingBuffer.SIZE```) emissions from each source but the operator fails to request more than 256 and the stream stalls.

```java
        @Test
	public void testMergeKeepsRequesting() throws InterruptedException {
		final CountDownLatch latch = new CountDownLatch(1);
		Observable.range(1, 2)
		// produce many integers per second
				.flatMap(new Func1<Integer, Observable<Integer>>() {
					@Override
					public Observable<Integer> call(final Integer number) {
						return Observable.range(1, Integer.MAX_VALUE)
						// pause a bit
								.doOnNext(pauseForMs(1))
								// buffer on backpressure
								.onBackpressureBuffer()
								// do in parallel
								.subscribeOn(Schedulers.computation());
					}

				})
				// take a number bigger than 2* RxRingBuffer.SIZE (used by
				// OperatorMerge)
				.take(RxRingBuffer.SIZE * 2 + 1)
				// log count
				.doOnNext(printCount())
				// release latch
				.doOnCompleted(new Action0() {
					@Override
					public void call() {
						latch.countDown();
					}
				}).subscribe();
		assertTrue(latch.await(1, TimeUnit.SECONDS));
	}

        private static Action1<Integer> printCount() {
		return new Action1<Integer>() {
			long count;

			@Override
			public void call(Integer t1) {
				count++;
				System.out.println("count=" + count);
			}
		};
	}

	private static Action1<Integer> pauseForMs(final long time) {
		return new Action1<Integer>() {
			@Override
			public void call(Integer s) {
				try {
					Thread.sleep(time);
				} catch (InterruptedException e) {
					throw new RuntimeException(e);
				}
			}
		};
	}

```

akarnokd:
Merge has been fixed in 1.0.5 so I'm closing this.

PR #2482, opened 2015-01-22
Merge: fixed hangs & missed scalar emissions

akarnokd:
I think I hunted down the bug that was causing missed value emission if scalar source such as ```just()``` was involved. This either manifested as completion without a value or non-completion because the missing scalar value would sit in the queue and prevented onCompleted from draining. (I've also fixed some visibility issues which might had an effect in this.)

In addition, I've discovered a good practice for spuriously failing tests: I create a new ```@Test```method with a loop in it, calling the original test method. This way, the original is not disturbed and the likelihood of failure is increased by the looping test.

Since the whole ```merge```operator is quite complicated, I can't be 100% certain this works despite local tests passing.

See also #2480, #2466 and #1941.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -17,20 +17,13 @@ package rx.internal.operators;
 
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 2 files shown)


akarnokd:
I'm going to merge this so other bugfixes/enhancements can go on.

Issue #1649, opened 2014-08-31
Need inclusive predicate variant of takeUntil/takeWhile


samueltardieu:
It has been two times I've been missing a `takeUntil(filter)` operator on observables. The last time was in cgeo while implementing a low-power mode: I want to receive location updates through an observable and stop as soon as a location arrives with a precision of 20m or less, but I want this location to be returned. Using `takeWhile(!filter)` would drop the matching item.

Of course, this can be worked around, but I was surprised not to get this operator which is the dual of `takeWhile`. Would you consider a contribution adding it?

benjchristensen:
I hit another use case today where I'd really like to have this.

akarnokd:
An experimental ```takeUntil()``` is now available in 1.0.5.

headinthebox:
One way to distinguish the two versions would be `takeUntil` (forwards, then checks) versus `untilTake` (checks, the forwards). Like `do-while` and `while`.

mxklabs:
I second the idea of needing this operator. 

I'd argue this is a basic primitive stream operation that ought not to be composed. Not including it is like Java designers arguing that you don't need a <= operator on integers because you already have < and == operators.

akarnokd:
@mxklabs how urgently do you need this operator?

mxklabs:
@akarnokd I need it now but can probably work around it. I just wanted to agree with the sentiment that it makes sense to have this operator.

mxklabs:
@akarnokd Also, take until (from what I understand from [here](http://reactivex.io/documentation/operators/takeuntil.html)) is not exactly what's requested. It takes two observables and terminates as soon as the second observable emits an item as opposed to an observable and a condition and keep emitting upto and including the first emit where the condition holds.

(Not sure if I am looking at the right documentation, sorry)

PR #2493, opened 2015-01-23
Operator TakeUntil with predicate

akarnokd:
Take values and check the condition after the value is emitted.

For enhancement request #1649.


benjchristensen:
Thanks for adding this! Not convinced on the name though. The 'do' prefix is used to represent side-effects everywhere else. I think the prefix should still be `take`-something.

akarnokd:
The question then remains, should it stop if predicate turns into false or should it stop if predicate turns into true? Because, classically ```do { } while (true)``` and ```repeat until false``` (Pascal).

benjchristensen:
I like it how it's currently implemented ... do until true.

It fits the "takeUntil" term ... "take until a value == predicate".

```java
        System.out.println("------------------------");
        Observable.just(1, 2, 3, 4, 5, 6, 7)
                .doOnEach(System.out::println)
                .takeUntil(i -> i == 3)
                .forEach(System.out::println);

        System.out.println("------------------------");
        Observable.just(1, 2, 3, 4, 5, 6, 7)
                .doOnEach(System.out::println)
                .takeWhile(i -> i <= 3)
                .forEach(System.out::println);
```

That code shows the difference with `takeWhile` and it works well. We emit 1, 2, 3 and then stop.

```
------------------------
[rx.Notification@891cd479 OnNext 1]
1
[rx.Notification@891cd47a OnNext 2]
2
[rx.Notification@891cd47b OnNext 3]
3
------------------------
[rx.Notification@891cd479 OnNext 1]
1
[rx.Notification@891cd47a OnNext 2]
2
[rx.Notification@891cd47b OnNext 3]
3
[rx.Notification@891cd47c OnNext 4]
```

Issue #2291, opened 2014-12-28
Backpressure: flatMap/merge with maxConcurrent


benjchristensen:
The `merge(Observable/Iterable o, int maxConcurrent)` overloads do not support backpressure. They were missed when implementing the normal `merge`. 

This was first brought up in https://github.com/ReactiveX/RxJava/issues/1927#issuecomment-66885487

This code shows the problem:

```java
import static rx.Observable.*;
import rx.Observable;
import rx.schedulers.Schedulers;

public class MergeMaxConcurrent {

    public static void main(String[] args) {
        Observable<Observable<Integer>> asyncWork = range(1, 1000000)
                //      .doOnNext(i -> System.out.println("Value: " + i))
                .doOnRequest(r -> System.out.println("request1 " + r))
                .map(item -> {
                    return just(item)
                            .doOnNext(MergeMaxConcurrent::sleep)
                            .subscribeOn(Schedulers.io());
                })
                .doOnRequest(r -> System.out.println("request2 " + r));
        merge(asyncWork, 10).toBlocking().forEach(System.out::println);
    }

    public static void sleep(int value) {
        try {
            Thread.sleep(1000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
    }
}
```

Related to this, we should have a `flatMap` overload that accepts a `maxConcurrent` parameter so that the `merge(o, maxConcurrent)` can be used like this with backpressure:

```java
        Observable.range(0, Integer.MAX_VALUE).map(i -> "lineOfData-" + i) // simulate the file with many lines
                .buffer(100) // buffer up chunks of work before processing them
                .flatMap(chunk -> {
                    return Observable.just(chunk).subscribeOn(Schedulers.io()) // put each group on a new thread (io scheduler for blocking IO) 
                            .doOnNext(t -> {
                                // do work here
                                try {
                                    Thread.sleep(1000);
                                } catch (Exception e) {
                                }
                                // write out on blocking IO as a side-effect
                                System.out.println("Emitting to IO: " + t);
                            });
                }).toBlocking().forEach(System.out::println);
```

Right now an awkward use of `groupBy` can be used to avoid using `merge` but it is inefficient since it calls the `groupBy` function on every item instead of being able to do the buffering before `flatMap` and just using backpressure to only `merge` the `maxConcurrent` as they are processed.

Here is the `groupBy` approach:

```java
[...]

PR #2540, opened 2015-01-26
Merge with max concurrency now supports backpressure.

akarnokd:
See #2291

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMergeMaxConcurrent.java
+++ b/src/main/java/rx/internal/operators/OperatorMergeMaxConcurrent.java
@@ -15,12 +15,14 @@
  */
 package rx.internal.operators;
 
-import java.util.LinkedList;
-import java.util.Queue;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
-import rx.Observable;
+import java.util.*;
+import java.util.concurrent.atomic.*;
+
+import rx.*;
 import rx.Observable.Operator;
-import rx.Subscriber;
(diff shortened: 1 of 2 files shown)


akarnokd:
I can't be sure, but probably the first Travis CI build timed out because the test used too many threads.

benjchristensen:
Thanks for tackling this. Do you think it's better to keep this separate or to have a single `merge` implementation? I was leaning towards having a single impl since `merge` is so hard to get right, and all the perf and backpressure aspects are needed for both.

akarnokd:
I would have to combine merge, mergeDelayError and mergeMaxConcurrent. Adding error delaying to mmc shouldn't be difficult, but applying the scalar-synchronous-value optimization as well is quite difficult. I'd say let's have them separate for now.

benjchristensen:
Keeping them separate means we'll be doing bug fixes and optimizations separately on each of them, and performance will be misleadingly different for each. I suggest that if we do merge this it is only a short-term thing. 

`mergeDelayError` has already been moved into the `merge` implementation. I did that a while back. So it's just the concurrency limit that needs to be moved.

akarnokd:
It would be great if my other merge PR gets verified so I can base the joint operator on something that's actually working for others.

akarnokd:
I'm merging this for now as combining the two is difficult and I don't want to hold 1.0.5 release back. From then on, a single implementation would be just an internal affair.

benjchristensen:
Agreed.

PR #2543, opened 2015-01-28
OperatorMerge handle request overflow

davidmoten:
Calling 
```java
request(2);
request(Long.MAX_VALUE-1);
```
hangs ```OperatorMerge``` because the internal requested count overflows and becomes negative. 

This PR adds fix and unit test.

I imagine this will be in other operators as well. I'll have a hunt around.


akarnokd:
We had a conflicting merge. Could you rebase this onto the current master?

davidmoten:
rebased (I think).

akarnokd:
Thanks.

PR #2546, opened 2015-01-28
Observable.java - remove unnecessary SuppressWarnings annotations

davidmoten:
a little bit of cleanup. I think these have been hanging around since we were using varargs parameters.


akarnokd:
I think the ```@SuppressWarning``` is necessary because without them, the ```Array.asList``` calls give warnings about generic array creation (in Eclipse at least).

davidmoten:
yep you are right. I used Eclipse with JDK 7 and gave warnings when annotations present. Now am looking at same with Eclipse + JDK 8 and no warnings. I'll close.

iNoles:
Java 7 has another annotation to fix that warning. http://blog.jetbrains.com/idea/2011/04/java-7-safevarargs/

akarnokd:
@iNoles RxJava has to remain compatible with Java 6.

PR #2547, opened 2015-01-28
Warnings cleanup

davidmoten:
more cleanup:

* removed unnecessary imports
* added missing generic types
* added SuppressWarning annotations


akarnokd:
Thanks for the cleanup.

Issue #2545, opened 2015-01-28
Subscriber.request add ignore for negative requests


davidmoten:
So that negative requests don't need to be checked by every ```Operator``` that supports backpressure why don't we add a check at the start of ```Subscriber.request```:

```java
protected final void request(long n) {
        // ignore zero or negative requests
        if (n <=0) return;
        Producer shouldRequest = null;
        synchronized (this) {
            if (p != null) {
                shouldRequest = p;
            } else {
                requested = n;
            }
        }
        // after releasing lock
        if (shouldRequest != null) {
            shouldRequest.request(n);
        }
    }
```

Happy to submit PR if has support.

zsxwing:
Looks a good place to add it. Just a bit comment: Ignore zero but IAE for negative numbers, as per #1956

davidmoten:
I did wonder about that, and makes sense. I'll make a PR.

davidmoten:
Looks to me that n = 0 should be allowed through because it may override a default of Long.MAX_VALUE sometimes (asking an upstream producer to produce nothing as opposed to whatever a default might be). I'll make a PR with the n < 0 check only.

Issue #1956, opened 2014-12-12
request(0) is supposed to work?


zsxwing:
The docs about `request` doesn't say anything requirement about `n`
```Java
public interface Producer {

    /**
     * Request a certain maximum number of items from this Producer. This is a way of requesting backpressure.
     * To disable backpressure, pass {@code Long.MAX_VALUE} to this method.
     *
     * @param n the maximum number of items you want this Producer to produce, or {@code Long.MAX_VALUE} if you
     *          want the Producer to produce items at its own pace
     */
    public void request(long n);

}
```
  
```Java
    /**
     * Request a certain maximum number of emitted items from the Observable this Subscriber is subscribed to.
     * This is a way of requesting backpressure. To disable backpressure, pass {@code Long.MAX_VALUE} to this
     * method.
     *
     * @param n the maximum number of items you want the Observable to emit to the Subscriber at this time, or
     *           {@code Long.MAX_VALUE} if you want the Observable to emit items at its own pace
     */
    protected final void request(long n) {
```
I assume `n` must be greater than 0. But I also notice I may write something like this occasionally:

```
long getTheNextRequestNumber() {
  if (...) {
       ...
       return x; // x>0
  } else {
        ...
       return 0;
   }
}

long n = getTheNextRequestNumber();
request(n);
```
So `request(0)` is fine? If not, I think it's better to document it.

PR #2548, opened 2015-01-28
Subscriber.request should throw exception if negative request made

davidmoten:
As per discussion in #1956 and #2545, making a call to ```Subscriber.request(n)``` with negative n should throw an ```IllegalArgumentException```.

Includes unit test.


akarnokd:
I think the IAE should be mentioned in request's javadoc.

davidmoten:
javadoc added

akarnokd:
Great!

Issue #2544, opened 2015-01-28
Subscriber.OnStart requests should be additive?


davidmoten:
For a ```Subscriber``` without a ```Producer```, calling 
```java
@Override
public void onStart() {
    request(2);
    request(3);
}
```
does not give 5 requests but 3 only. Shouldn't this be additive? (or at least documented that it is not).

akarnokd:
It should be additive and check for overflow. Do you want to submit a new PR or amend #2548?

davidmoten:
I'll submit a new PR, thanks.

Override was probably misleading. A better term would be probably be *prevents* ```requested``` being set to ```Long.MAX_VALUE```.  I was referring to the code below from ```Subscriber``` where if a ```request(0)``` does nothing but return (doesn't set ```requested``` to 0) then ```requested``` can end up as Long.MAX_VALUE. 

```java
 public void setProducer(Producer producer) {
        long toRequest;
        boolean setProducer = false;
        synchronized (this) {
            toRequest = requested;
            p = producer;
            if (op != null) {
                // middle operator ... we pass thru unless a request has been made
                if (toRequest == Long.MIN_VALUE) {
                    // we pass-thru to the next producer as nothing has been requested
                    setProducer = true;
                }

            }
        }
        // do after releasing lock
        if (setProducer) {
            op.setProducer(p);
        } else {
            // we execute the request with whatever has been requested (or Long.MAX_VALUE)
            if (toRequest == Long.MIN_VALUE) {
                p.request(Long.MAX_VALUE);
            } else {
                p.request(toRequest);
            }
        }
    }
```

PR #2550, opened 2015-01-28
Subscriber.onStart requests should be additive (and check for overflow)

davidmoten:
As per discussion in #2544.


akarnokd:
Looks good. Thanks.

benjchristensen:
That looks helpful. Thanks @davidmoten

PR #2551, opened 2015-01-28
Make testFlatMapAsync pass by allowing 3% overflow caused by gc delays

akarnokd:
testFlatMapAsync is non-deterministic because a slight delay in consumption lets the producer thread run quite ahead. The change now measures this over 1000 runs and allows 30 of such to happen at max. Testing on Windows (not so good scheduler), I get 10-20 overruns.


akarnokd:
This is no good. If left plain, the test fails 10% of the time. If looped, it fails ~35% of the time.

PR #2554, opened 2015-01-28
add Dump debug method

isaiah:
Add a shortcut to dump the sequence and display message for complete and error event. Its usage is demonstrated in [Introduction to Rx](http://www.introtorx.com/Content/v1.0.10621.0/07_Aggregation.html). I think it's very convenient for newbies to play with different type of Observables.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -9020,6 +9020,36 @@ public class Observable<T> {
     }
 
     /**
+     * Subscribes to an Observable and provides an Observer that implements functions to print the items the
+     * Observable emits and any error or completion notification it issues. This is mainly for debug purpose.
+     *
+     * @param observed
+     *             a description of the Observable subscribed to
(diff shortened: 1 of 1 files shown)


akarnokd:
I've resisted the urge to add such method to the observable because the dump style changes from place to place.

JakeWharton:
Yeah, agreed. Also on Android we'd want it sending to `android.util.Log` with the string being used as the tag, not in the message.

I think this is one of those things that's so simple that it will be impossible to include because there's so many different things people will want from it.

benjchristensen:
Generally the approach to debugging is left to methods such as `doOnNext`, `donNextComplete`, `doOnError`, `doOnEach`, `doOnSubscribe`, `doOnUnsubscribe`.

We have avoided logging in RxJava for the reasons @akarnokd and @JakeWharton gave. Thank you for providing this but I'm going to close this out and not add it.

isaiah:
Sorry for the spam, I mistook it as part of the C# implementation.

PR #2565, opened 2015-01-29
Obstruction detection in tests.

akarnokd:
To make sure tests don't hang the computation scheduler.


davidmoten:
I like

akarnokd:
I'm merging this to disable ```testFlatMapAsync```; failing +30% of the time on the slightest hiccup.

Issue #2564, opened 2015-01-29
onErrorResumeNext, refCount and retry


amitai-hoze:
Hi, I noticed some behavior I don't fully understand.

When I run this:
```java
Observable<String> interval =
        Observable.interval(1,TimeUnit.SECONDS)
                .doOnSubscribe(() -> System.out.println("subscribe interval"))
                .doOnUnsubscribe(() -> System.out.println("unsubscribe interval"))
                .flatMap(onClickEvent -> Observable.defer(() -> Observable.<String>error(new Exception("Some exception"))))
                .onErrorResumeNext(Observable::error)
                .publish()
                .refCount()
                .doOnSubscribe(() -> System.out.println("subscribe refCount"))
                .doOnUnsubscribe(() -> System.out.println("unsubscribe refCount"));

interval
        .doOnError(throwable -> System.out.println("subscriber 1 error: " + throwable.getMessage()))
        .retry()
        .subscribe(s -> System.out.println("subscriber 1: " + s));
```
I get:
```
subscribe refCount
subscribe interval
unsubscribe interval
subscriber 1 error: Some exception
unsubscribe refCount
subscribe refCount
subscribe interval
unsubscribe interval
subscriber 1 error: Some exception
unsubscribe refCount
...
```
The three dots indicate that it continues infinitely.
Now when I add a second subscriber:
```java
interval
        .doOnError(throwable -> System.out.println("subscriber 2 error: " + throwable.getMessage()))
        .retry()
        .subscribe(s -> System.out.println("subscriber 2: " + s));
```
I get:
```
subscribe refCount
subscribe interval
subscribe refCount
unsubscribe interval
subscriber 1 error: Some exception
unsubscribe refCount
subscribe refCount
subscriber 2 error: Some exception
unsubscribe refCount
subscribe refCount
```
and that's it, it doesn't retry. More specifically, it unsubscribes from interval and does not resubscribe. When I remove the `onErrorResumeNext` it's working fine like this:
```
subscribe refCount
subscribe interval
subscribe refCount
subscriber 1 error: Some exception
unsubscribe refCount
subscribe refCount
subscriber 2 error: Some exception
unsubscribe refCount
subscribe refCount
subscriber 1 error: Some exception
unsubscribe refCount
subscribe refCount
subscriber 2 error: Some exception
unsubscribe refCount
subscribe refCount
...
```
[...]

PR #2567, opened 2015-01-29
RefCount: disconnect all if upstream terminates

akarnokd:
This should fix the problem in #2564. @davidmoten I hope you can review the changes and implications.

The particular issue was caused by the refCount unable to unsubscribe from upstream in case the current client count was greater than 1. If the upstream sent an error and each client had a retry() in its chain, the retry would immediately resubscribe, thus the refCount would see only a 2 -> 1 -> 2 change in sequence as errors are dispatched sequentially. In the change, each client is wrapped and a termination event triggers the refcount to become 0 atomically and unsubscribes the upstream. When each client then call disconnect, that becomes a no-op.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeRefCount.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeRefCount.java
@@ -80,11 +80,8 @@ public final class OnSubscribeRefCount<T> implements OnSubscribe<T> {
             }
         } else {
             try {
-                // handle unsubscribing from the base subscription
-                subscriber.add(disconnect());
-
                 // ready to subscribe to source so do it
-                source.unsafeSubscribe(subscriber);
+                doSubscribe(subscriber, baseSubscription);
             } finally {
                 // release the read lock
                 lock.unlock();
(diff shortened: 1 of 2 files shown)


davidmoten:
Thanks for chasing this @akarnokd. I'll have a look this weekend.

davidmoten:
Looks good @akarnokd. There are a couple of my comments that are unclear (referring to Action1) but I can fix those later if you like. Can we squeeze this into 1.0.5?

akarnokd:
@benjchristensen seems to be eager to release soon, but if your comments are something that can be done within a few minutes, I can do the changes before release.

benjchristensen:
This looks good to me. @akarnokd Should I proceed with it or would you still like to wait for @davidmoten to review?

benjchristensen:
I was working off a stale browser tab so hadn't seen your comments. If both of you are okay with this then I'm ready to merge and release.

akarnokd:
Its getting late here, but we can perhaps give @davidmoten some time while you could review #2592 and  #2493.

benjchristensen:
Okay, that works for me.

akarnokd:
Let's merge this as-is and @davidmoten will post a PR with his changes for 1.0.6. I'll leave it up to you @benjchristensen whether to merge #2493 and #2580 for 1.0.5.

davidmoten:
I was referring to code comments being slightly out of sync with the code
but nothing serious. I think it can wait till another release as I'm not
able to do a PR at the moment. In a few hours can do but happy if release
goes with this fix in the meantime.

PR #2573, opened 2015-01-30
Search plugins via ServiceLoader, if the properties are not defined

carlopellegriniarchijcom:
This PR adds the `ServiceLoader `mechanism to the plugins system.
This way the library could be used by multiple applications in a JavaEE server, and also confers the ability to use managed threads via a custom `RxJavaSchedulersHook`

### DIFF
--- a/src/main/java/rx/plugins/RxJavaPlugins.java
+++ b/src/main/java/rx/plugins/RxJavaPlugins.java
@@ -15,6 +15,8 @@
  */
 package rx.plugins;
 
+import java.util.Iterator;
+import java.util.ServiceLoader;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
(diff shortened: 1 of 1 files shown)


akarnokd:
@benjchristensen ( @abersnaze ?) opinions about this enhancements?

benjchristensen:
I have the same concerns with this as expressed in https://github.com/ReactiveX/RxJava/pull/2643

benjchristensen:
> confers the ability to use managed threads

How is this not already possible with a custom `RxJavaSchedulersHook`?

carlopellegriniarchijcom:
@benjchristensen 

Well, my biggest problem is that  a custom `RxJavaSchedulersHook` can be defined only via system properties, through  `getPluginImplementationViaProperty.

In a shared JVM with multiple classloaders, setting the property affects all the applications deployed, regardless of their classloarer which is often undesirable, hence my use of the `ServiceLoader`. 

I understand your concerns about #2643 but I share @jtulach view on the matter.

Alternatively is there any other way to safely define a `RxJavaSchedulersHook` for a single classloader?

benjchristensen:
> Alternatively is there any other way to safely define a RxJavaSchedulersHook for a single class loader?

Register the hook via code that initializes on startup.

benjchristensen:
> for a single class loader

Based on my experience with class loaders, I'm pretty sure the only way code with statics works is that you must ensure that library is only loaded in the child class loader, never in a parent. 

But if you have that working already with the various statics that get initialized, then within each class loader you could initialize it differently.

carlopellegriniarchijcom:
Yes, my concern dealed with the system properties approach. Registering the hook via code will deal with the single classloader (better said: single application) issue. Going to close the PR.

PR #2574, opened 2015-01-31
SizeEviction test needs to return false 

jnpr-jalandip:
When a replay subject with size and time eviction policy is used then any observer that subscribes after some events are already pushed will not get those older events. 
This is because if an observable is not yet terminated then on subscription by any new observer the check for event validity fails on the Size Eviction policy hence that event is dropped and not pushed to the new subscriber.

Relevant call stack:
   {{{

     TimeOnAdd.call
         BoundedState.replayObserverFromIndexTest
            BoundedState.replayObserverFromIndexTest
                 PairEvictionPolicy.test
    }}}

Test Case: 
  - add test case for this use case, Fails if without changes.


akarnokd:
Good catch. Thanks!

PR #2575, opened 2015-01-31
subscribeOn drops the subscriptions returned from the scheduler.

abersnaze:
@mattrjacobs noticed that the unsubscribe from timeouts of Hystrix commands was not being propagated across the subscribeOn.

https://github.com/Netflix/Hystrix/issues/354


akarnokd:
Looks good.

PR #2579, opened 2015-02-02
Adding ScheduledAction to the official API

akarnokd:
I've copied the internal ```ScheduledAction``` into the public ```rx.schedulers``` package to allow custom ```Scheduler``` implementors in other projects to take advantage of its correct unsubscription management.

I've removed the internal ```ScheduledAction```. Since relying on internal features are discouraged anyway, use places outside RxJava need to be updated. Deprecating it doesn't work because the class is final and the internals now return the new ```rx.schedulers.ScheduledAction``` and would result in ```ClassCastException``` anyway.

I've introduced a system-wide parameter ```io.reactivex.scheduler.interrupt-on-unsubscribe``` to enable interruption globally. In addition, each ```ScheduledAction``` has its own ```setInterruptOnUnsubscribe``` which allows enabling/disabling interrupts on an individual basis (overrides the default from the system-wide parameter above for the instance).

I've also added detailed documentation to it and (while I was in the mood) to the ```NewThreadWorker```. 

There is a slight performance drawback now since setting the current thread id at the beginning of the ```run()``` method has to preserve the interruptible flag thus it has to perform a CAS instead of a 32-bit lazySet. Also note that the instance size change is unavoidable because a reference+int has the same allocation cost as a single long (there is no byte-atomics).


akarnokd:
I'll post a new fresh version of this soon.

benjchristensen:
While you're at it can you rebase all the commits?

akarnokd:
See rebased PR #2592.

PR #2601, opened 2015-02-04
Added common Exceptions.throwIfAny to throw a collection of exceptions

akarnokd:
Tried to be as flexible with the error text as possible.

### DIFF
--- a/src/main/java/rx/exceptions/Exceptions.java
+++ b/src/main/java/rx/exceptions/Exceptions.java
@@ -15,8 +15,7 @@
  */
 package rx.exceptions;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 
 /**
  * @warn javadoc class description missing
@@ -148,5 +147,32 @@ public final class Exceptions {
         }
         return e;
     }
-
+    /**
+     * Throws a single or multiple exceptions contained in the collection, wrapping it into
+     * {@code CompositeException} if necessary.
+     * @param exceptions the collection of exceptions. If null or empty, no exception is thrown.
+     * If the collection contains a single exception, that exception is either thrown as-is or wrapped into a
+     * CompositeException. Multiple exceptions are wrapped into a CompositeException.
+     * @param whileText the circumstance string to be appended to the thrown CompositeException, inserted after
+     * the sentences "Exception" and "Multiple exceptions".
+     */
+    public static void throwIfAny(Collection<? extends Throwable> exceptions, String whileText) {
+        if (exceptions != null && !exceptions.isEmpty()) {
+            if (exceptions.size() == 1) {
+                Throwable t = exceptions.iterator().next();
+                if (t instanceof RuntimeException) {
+                    throw (RuntimeException) t;
+                } else
+                if (t instanceof Error) {
+                    throw (Error) t;
+                } else {
+                    throw new CompositeException(
+                            "Exception" + whileText, exceptions);
+                }
+            } else {
+                throw new CompositeException(
+                        "Multiple exceptions" + whileText, exceptions);
+            }
+        }
+    }
 }
--- a/src/main/java/rx/exceptions/Exceptions.java
+++ b/src/main/java/rx/exceptions/Exceptions.java
@@ -15,8 +15,8 @@
  */
 package rx.exceptions;
 
-import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import rx.annotations.Experimental;
@@ -37,19 +37,6 @@ public final class Exceptions {
      *         value; {@code propagate} does not actually return anything
      */
     public static RuntimeException propagate(Throwable t) {
-        return propagate(t, null);
-    }
-    /**
-     * Convenience method to throw a {@code RuntimeException} and {@code Error} directly
-     * or wrap any other exception type into a {@code RuntimeException} with an optional custom message.
-     * @param t the exception to throw directly or wrapped
-     * @param message the optional custom message to set up the RuntimeException thrown
-     * in case {@code t} is a checked exception.
-     * @return because {@code propagate} itself throws an exception or error, this is a sort of phantom return
-     *         value; {@code propagate} does not actually return anything
-     */
-    @Experimental
-    public static RuntimeException propagate(Throwable t, String message) {
         /*
          * The return type of RuntimeException is a trick for code to be like this:
          * 
@@ -63,10 +50,9 @@ public final class Exceptions {
         } else if (t instanceof Error) {
             throw (Error) t;
         } else {
-            throw new RuntimeException(message, t);
+            throw new RuntimeException(t);
         }
     }
-
     /**
      * Throws a particular {@code Throwable} only if it belongs to a set of "fatal" error varieties. These
      * varieties are as follows:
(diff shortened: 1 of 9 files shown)


src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -148,5 +147,32 @@ public static Throwable getFinalCause(Throwable e) {
...
+            if (exceptions.size() == 1) {
+                Throwable t = exceptions.iterator().next();
+                if (t instanceof RuntimeException) {
+                    throw (RuntimeException) t;
+                } else
+                if (t instanceof Error) {
+                    throw (Error) t;
+                } else {
+                    throw new CompositeException(
+                            "Exception" + whileText, exceptions);
```
jakudlaty: Why single exception (check line 161) is wrapped into composite exception?

src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -148,5 +147,32 @@ public static Throwable getFinalCause(Throwable e) {
...
+                }
+            } else {
+                throw new CompositeException(
+                        "Multiple exceptions" + whileText, exceptions);
+            }
```
jakudlaty: I think it would be easier to read:
```
if (t instanceof RuntimeException) {
    throw (RuntimeException) t
}
if (t instanceof Error) {
    throw (Error) t;
}
throw new CompositeException("Exception" + whileText, exceptions);
```

akarnokd: Because the original places distinguished between single and multiple expections in the message of CompositeException.

benjchristensen: As per the discussion at https://github.com/ReactiveX/RxJava/issues/2602 ... this is an example where @akarnokd and I have slightly different styles, but it's okay. I always have `else if` on the same like, here it is split, but I can still read the code and it works so I choose to not make a big deal out of it. 

Sometimes when @akarnokd and I end up modifying each others code we'll see some lines get modified to personal style if we touch them, but we don't go around explicitly trying to change code just to match a personal preference ... if for no other reason than to not clutter changelogs and diffs.

src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -148,5 +147,32 @@ public static Throwable getFinalCause(Throwable e) {
...
+        if (exceptions != null && !exceptions.isEmpty()) {
+            if (exceptions.size() == 1) {
+                Throwable t = exceptions.iterator().next();
+                if (t instanceof RuntimeException) {
+                    throw (RuntimeException) t;
+                } else
+                if (t instanceof Error) {
+                    throw (Error) t;
+                } else {
+                    throw new CompositeException(
```
benjchristensen: Right here we have a single exception so why do we need to wrap it in `CompositeException`? The `Exceptions.propagate` method could be called here like it was doing in `publish`.

benjchristensen: Thinking more clearly ... why doesn't this entire if-block just delegate to `propagate` since it is replicating that logic?

benjchristensen:
This is a nice change ... just the comments on using `propagate` that I'd like to handle.

src/main/java/rx/exceptions/Exceptions.java
akarnokd: Because it threw CompositeExceptions so far and I couldn't be sure if the message is relevant or not.

benjchristensen: In `OperatorPublish` it didn't, it used `propagate`.

src/main/java/rx/internal/util/SubscriptionList.java
```diff
@@ -106,20 +106,7 @@ private static void unsubscribeFromAll(Collection<Subscription> subscriptions) {
...
             }
         }
-        if (es != null) {
-            if (es.size() == 1) {
-                Throwable t = es.get(0);
-                if (t instanceof RuntimeException) {
-                    throw (RuntimeException) t;
-                } else {
-                    throw new CompositeException(
-                            "Failed to unsubscribe to 1 or more subscriptions.", es);
```
benjchristensen: This looks like a mistake to have been a `CompositeException` since it's not composite, there is only 1. If it was a `RuntimeException` on the other side of the branch we just threw it, so we should just be using `propagate` here. 

Perhaps the reason this was done was to add a better error message? That's something an overloaded `propagate` could add.

src/main/java/rx/internal/util/SubscriptionRandomList.java
```diff
@@ -155,19 +155,6 @@ public void unsubscribe() {
                 es.add(e);
             }
         }
-        if (es != null) {
-            if (es.size() == 1) {
-                Throwable t = es.get(0);
-                if (t instanceof RuntimeException) {
-                    throw (RuntimeException) t;
-                } else {
-                    throw new CompositeException(
```
benjchristensen: Same thing here as above ...

src/main/java/rx/exceptions/Exceptions.java
benjchristensen: All of the subjects used `propagate` as well.

akarnokd:
Updated the code based on the comments above.

src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -27,11 +30,26 @@ private Exceptions() {
...
     /**
-     * @warn javadoc missing
+     * Convenience method to throw a {@code RuntimeException} and {@code Error} directly
+     * or wrap any other exception type into a {@code RuntimeException}.
+     * @param t the exception to throw directly or wrapped
      * @return because {@code propagate} itself throws an exception or error, this is a sort of phantom return
      *         value; {@code propagate} does not actually return anything
      */
     public static RuntimeException propagate(Throwable t) {
+        return propagate(t, null);
```
benjchristensen: Why do we have return values on these when they end up just throwing? I just noticed that. We can't change it, but it's very odd.

src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -148,5 +166,25 @@ public static Throwable getFinalCause(Throwable e) {
...
+     * CompositeException. Multiple exceptions are wrapped into a CompositeException.
+     * @param whileText the circumstance string to be appended to the thrown CompositeException, inserted after
+     * the sentences "Exception" and "Multiple exceptions".
+     */
+    @Experimental
+    public static void throwIfAny(Collection<? extends Throwable> exceptions, String whileText) {
+        if (exceptions != null && !exceptions.isEmpty()) {
+            if (exceptions.size() == 1) {
+                Throwable t = exceptions.iterator().next();
+                throw propagate(t, "Exception" + whileText);
```
benjchristensen: This `throw` does nothing ... is this "phantom" behavior done so that the code compiles to be terminating here, even though it is inside the method that throws? If so that is tricksy!

benjchristensen:
That build failure is concerning ... it hung and timed out after 10 minutes at this test:

> rx.exceptions.ExceptionsTest > testStackOverflowIsThrown STARTED

src/main/java/rx/exceptions/Exceptions.java
akarnokd: It is explained by the javadoc.

akarnokd:
Most likely it died of stackoverflow while attempting to report stackoverflow. I've removed the custom message and manually inlined the propagate method.

PR #2607, opened 2015-02-05
OnSubscribeRefCount - improve comments

davidmoten:
As mentioned in #2567, improve the comments


akarnokd:
Thanks. Can't merge until master is fully resolved because it would just fail.

Issue #2626, opened 2015-02-06
Add flatMap variant with maxConcurrent


benjchristensen:
To allow use of `merge(Observables..., maxConcurrent)` via flatMap we should consider adding a `flatMap(Func, int maxConcurrent)` overload.

benjchristensen:
Example usage:

```java
        Observable.range(0, Integer.MAX_VALUE).map(i -> "lineOfData-" + i) // simulate the file with many lines
                .buffer(100) // buffer up chunks of work before processing them
                .flatMap(chunk -> {
                    return Observable.just(chunk).subscribeOn(Schedulers.io()) // put each group on a new thread (io scheduler for blocking IO) 
                            .doOnNext(t -> {
                                // do work here
                                try {
                                    Thread.sleep(1000);
                                } catch (Exception e) {
                                }
                                // write out on blocking IO as a side-effect
                                System.out.println("Emitting to IO: " + t);
                            });
                }, maxConcurrentValueHere).toBlocking().forEach(System.out::println);
```

akarnokd:
The related PR has been merged. Closing.

PR #2627, opened 2015-02-06
FlatMap overloads with maximum concurrency parameter

akarnokd:
For issue #2626.

There is no point in adding maxConcurrent to flatMapIterable because each iterable is subscribed to sequentially and run to completion.


benjchristensen:
:+1:

akarnokd:
> There is no point in adding maxConcurrent to flatMapIterable because each iterable is subscribed to sequentially and run to completion.

I was wrong with this statement. Depending on the request pattern, reordering may still happen and the downstream may be still slower. Unboundedness of `flatMapIterable` may still cause excessive source Iterable generation that fills in the tracking arrays inside `flatMap`.

Having the `maxConcurrency` parameter is valuable and prevents buffer bloat.

PR #2642, opened 2015-02-09
Fix a potential memory leak in schedulePeriodically

zsxwing:
There is a potential memory leak in `schedulePeriodically` that may keep a reference to `action` after `unsubscribe`. 

Because `mas.set` is called after `schedule`,  it may replace a new Subscription (created in `recursiveAction`) with the old one. Therefore, `unsubscribe` won't be able to unsubscribe the new Subscription and will keep the reference to `action` until the period time elapses.

This PR fixed it by calling `mas.set` before `schedule`.

### DIFF
--- a/src/main/java/rx/Scheduler.java
+++ b/src/main/java/rx/Scheduler.java
@@ -20,7 +20,6 @@ import java.util.concurrent.TimeUnit;
 import rx.functions.Action0;
 import rx.schedulers.Schedulers;
 import rx.subscriptions.MultipleAssignmentSubscription;
-import rx.subscriptions.SerialSubscription;
 
 /**
  * A {@code Scheduler} is an object that schedules units of work. You can find common implementations of this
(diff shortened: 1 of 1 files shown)


src/main/java/rx/Scheduler.java
```diff
@@ -119,11 +120,17 @@ public void call() {
                     if (!mas.isUnsubscribed()) {
                         action.call();
                         long nextTick = startInNanos + (++count * periodInNanos);
-                        mas.set(schedule(this, nextTick - TimeUnit.MILLISECONDS.toNanos(now()), TimeUnit.NANOSECONDS));
+                        SerialSubscription s = new SerialSubscription();
```
akarnokd: I think there is no need to change this here: the initial SerialSubscription can safely overwritted by the result of the schedule as it is not reentrant.

zsxwing: Do you mean if `period` is large, the new Subscription created in the next action won't be able to replace the one here, and if `period` is small, the leak won't be a problem?

Just to make sure we are in the same page.

akarnokd: No. I mean once the action is executed on the worker, each invocation of call is guaranteed to be serial, therefore, there is no need to have another level of SerialSubscription indirection.

akarnokd:
It is functionally mergeable, but you could have used ```MultipleAssignmentSubscription s = ...``` and saved an import.

src/main/java/rx/Scheduler.java
zsxwing: Gotcha. Forgot it...

PR #2644, opened 2015-02-10
Missing Unsafe class yields NoClassDefFoundError

jtulach:
Missing Unsafe class yields NoClassDefFoundError so checking for Exception is not really appropriate. Let's catch all throwables.


benjchristensen:
Thanks for this fix.

PR #2648, opened 2015-02-11
TakeWhile: don't unsubscribe downstream.

### LINKED ISSUE #2647: flatMap doesn't propagate onComplete with unsubscribe of child
The following code hangs forever as it doesn't emit a terminal event:

```java

import rx.Observable;

public class IntervalTakeWhileExample {

        Observable
                .interval(1, TimeUnit.MILLISECONDS)
                .doOnEach(a -> System.out.println("A ==> " + a))
                .takeWhile(aLong -> aLong < 1)
                .doOnEach(b -> System.out.println("B ==> " + b))
                .flatMap(aLong -> {
                    return Observable
                            .timer(30, TimeUnit.MILLISECONDS)
                            .doOnUnsubscribe(() -> System.out.println("unsubscribed timer?"))
                            .doOnEach(c -> System.out.println("C ==> " + c))
                            .map(aLong1 -> aLong1 + aLong);
                })
                .doOnEach(d -> System.out.println("D ==> " + d))
                .toBlocking().forEach(System.out::println);
    }
}
```

This emits:

```
A ==> [rx.Notification@f6adcf11 OnNext 0]
B ==> [rx.Notification@f6adcf11 OnNext 0]
A ==> [rx.Notification@f6adcf12 OnNext 1]
B ==> [rx.Notification@5b2cf4de OnCompleted]
unsubscribed timer?
```

akarnokd:
Fixes #2647 issue with TakeWhile.

### DIFF
--- a/src/main/java/rx/Subscriber.java
+++ b/src/main/java/rx/Subscriber.java
@@ -40,13 +40,24 @@ public abstract class Subscriber<T> implements Observer<T>, Subscription {
     private long requested = Long.MIN_VALUE; // default to not set
 
     protected Subscriber() {
-        this.op = null;
-        this.cs = new SubscriptionList();
+        this(null, false);
     }
 
     protected Subscriber(Subscriber<?> op) {
+        this(op, true);
+    }
+    /**
+     * Construct a subscriber by using the other subscriber for backpressure
+     * and optionally sharing the underlying subscriptions list.
+     * <p>To retain the chaining of subscribers, the caller should add the
+     * created instance to the op via {@code add()}.
+     * 
+     * @param op the other subscriber
+     * @param shareSubscriptions should the subscription list in op shared with this instance?
+     */
+    protected Subscriber(Subscriber<?> op, boolean shareSubscriptions) {
         this.op = op;
-        this.cs = op.cs;
+        this.cs = shareSubscriptions && op != null ? op.cs : new SubscriptionList();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/rx/Subscriber.java
```diff
@@ -40,13 +40,24 @@
...
+    /**
+     * Construct a subscriber by using the other subscriber for backpressure
+     * and optionally sharing the underlying subscriptions list.
+     * <p>To retain the chaining of subscribers, the caller should add the
+     * created instance to the op via {@code add()}.
+     * 
+     * @param op the other subscriber
+     * @param shareSubscriptions should the subscription list in op shared with this instance?
+     */
+    protected Subscriber(Subscriber<?> op, boolean shareSubscriptions) {
```
benjchristensen: Are there other places this constructor should be used where we're doing it manually today?

akarnokd: Yes, plenty of places.

benjchristensen: Thought so ... thanks for the confirmation.

benjchristensen: I just merged this ... but I'm always hesitant when changing something as core as `Subscriber`. Are we ready to support this new method forever. Is it the right signature for all the use cases?

I think it's right, but I've regretted public API decisions before :-)

/cc @zsxwing @abersnaze for more eyes and thought on this.

PR #2649, opened 2015-02-11
ThrottleFirst fix in case of multiple subscriptions

guillaumervls:
Hi,

Before, `lastOnNext` (the last time an event went through the throttling) was initialized at every subscription. This caused issues when dealing with a *hot* observable, where multiple subscriptions (& unsubscriptions) can happen. It should be initialized at operator instantiation instead.

Hope I didn't miss something !

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorThrottleFirst.java
+++ b/src/main/java/rx/internal/operators/OperatorThrottleFirst.java
@@ -27,6 +27,7 @@ public final class OperatorThrottleFirst<T> implements Operator<T, T> {
 
     private final long timeInMilliseconds;
     private final Scheduler scheduler;
+    private long lastOnNext = 0;
 
     public OperatorThrottleFirst(long windowDuration, TimeUnit unit, Scheduler scheduler) {
         this.timeInMilliseconds = unit.toMillis(windowDuration);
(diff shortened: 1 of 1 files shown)


akarnokd:
What issues do you encounter? Doing something mutable at the operator level is usually incorrect.

guillaumervls:
I want to throttle a stream of UI events (like clicks, scrolls), and that every subscriber see the same stream of throttled events (even if they don't subscribe at the same time or if they unsubscribe & resubscribe later), which is not the case if for every subscription lastOnNext is different.

akarnokd:
Then you need to share the throttled stream:

```java
source.throttleFirst(1, TimeUnit.SECONDS).share().subscribe(...)
```

Or if you want to keep the stream alive:
```
PublishSubject ps = PublishSubject.creat();
source.throttleFirst(1, TimeUnit.SECONDS).subscribe(ps);

ps.subscribe();
ps.subscribe();
```

guillaumervls:
Thanks @akarnokd ! I thought about your first solution but it wasn't working, since it's reset when there's no subscriber. But the second one does the job!

Sorry for the misunderstanding and thanks again for the hint!

Issue #2604, opened 2015-02-05
OperatorMerge - RxRingBuffer throws IllegalStateException again


davidmoten:
With 1.0.5, I'm getting an ```IllegalStateException``` from ```OperatorMerge```. This is repeatable for me but involves a big input data set being processed in chunks using flatMap/onSubscribe so I can't give you a quick test case yet. Hopefully just this description is enough otherwise I'll try and distill a test. I think it is happening close to completion.

```
java.lang.IllegalStateException: This instance has been unsubscribed and the queue is no longer usable.
        at rx.internal.util.RxRingBuffer.onNext(RxRingBuffer.java:346)
        at rx.internal.operators.OperatorMerge$InnerSubscriber.enqueue(OperatorMerge.java:721)
        at rx.internal.operators.OperatorMerge$InnerSubscriber.emit(OperatorMerge.java:698)
        at rx.internal.operators.OperatorMerge$InnerSubscriber.onNext(OperatorMerge.java:586)
        at rx.internal.operators.OperatorSubscribeOn$1$1$1.onNext(OperatorSubscribeOn.java:76)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.handleScalarSynchronousObservableWithRequestLimits(OperatorMerge.java:280)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.handleScalarSynchronousObservable(OperatorMerge.java:243)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:176)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:120)
        at rx.internal.operators.OperatorMap$1.onNext(OperatorMap.java:55)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.handleScalarSynchronousObservableWithRequestLimits(OperatorMerge.java:280)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.handleScalarSynchronousObservable(OperatorMerge.java:243)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:176)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:120)
        at rx.internal.operators.OperatorMap$1.onNext(OperatorMap.java:55)
        at rx.internal.operators.NotificationLite.accept(NotificationLite.java:150)
[...]

PR #2759, opened 2015-02-19
Observable.using should use unsafeSubscribe and enable eager disposal

davidmoten:
See #2604 where it was discovered that ```Observable.using``` used ```subscribe``` instead of ```unsafeSubscribe``` which provoked a race condition leading to an IllegalArgumentException from the merge operator.

This PR uses ```unsafeSubscribe``` and adds an overload for ```using``` to optionally dispose of resources eagerly (just before completion or error). The use case for this is a *synchronous* observable where a downstream operation wants to reuse a resource (but because the observable is synchronous the resource cannot get disposed till the downstream completes).

Unit tests included.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -2547,7 +2547,7 @@ public class Observable<T> {
     }
 
     /**
-     * Constructs an Observable that creates a dependent resource object.
+     * Constructs an Observable that creates a dependent resource object which is disposed of on unsubscription.
      * <p>
      * <img width="640" height="400" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/using.png" alt="">
      * <dl>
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OnSubscribeUsing.java
```diff
@@ -29,38 +33,84 @@
...
-            Action1<? super Resource> dispose) {
-        this.resourceFactory = resourceFactory;
-        this.observableFactory = observableFactory;
-        this.dispose = dispose;
-    }
+		try {
+			// create the resource
+			final Resource resource = resourceFactory.call();
+			// create an action that disposes only once
+			final Action0 disposeOnceOnly = createOnceOnlyDisposeAction(resource);
```
akarnokd: I'd have a class that implements both Action0 and subscription to avoid creating too many objects:
```java
static final class DisposeAction<R> extends AtomicBoolean implements Action0, Subscription {
    Action1<? super R> dispose;
    R resource;
    public DisposeAction(Action1<? super R> dispose, R resource) {
        this.dispose = dispose;
        this.resource = resource;
        lazySet(false); // StoreStore barrier
    }
    @Override
    public void call() {
         if (compareAndSet(false, true)) {
            try {
                dispose.call(resource);
            } finally {
                resource = null;
                dispose = null;
            }
         }
    }
    @Override
    public boolean isUnsubscribed() {
        return get();
    }
    public void unsubscribe() {
        call();
    }
}
```

davidmoten:
Thanks @akarnokd, I've made the change and squashed commits.

src/main/java/rx/Observable.java
```diff
@@ -2568,7 +2568,38 @@ public final Boolean call(T first, T second) {
...
+     *            the factory function to create an Observable
+     * @param disposeAction
+     *            the function that will dispose of the resource
+     * @param disposeEagerly
+     *            if true then disposal will happen either on unsubscription or just before emission of 
+     *            a terminal event (onComplete or onError).
+     * @return the Observable whose lifetime controls the lifetime of the dependent resource object
+     * @see <a href="http://reactivex.io/documentation/operators/using.html">ReactiveX operators documentation: Using</a>
+     */
+    public final static <T, Resource> Observable<T> using(
```
akarnokd: Could you add ```@Experimental``` here? In addition, could you replace tabs with 4 spaces in OnSubscribeUsing?

src/main/java/rx/internal/operators/OnSubscribeUsing.java
```diff
@@ -15,52 +15,120 @@
...
 
 /**
  * Constructs an observable sequence that depends on a resource object.
  */
 public final class OnSubscribeUsing<T, Resource> implements OnSubscribe<T> {
 
-    private final Func0<Resource> resourceFactory;
-    private final Func1<? super Resource, ? extends Observable<? extends T>> observableFactory;
-    private final Action1<? super Resource> dispose;
+	private final Func0<Resource> resourceFactory;
```
akarnokd: Could you replace tabs with spaces?

davidmoten:
Replaced tabs with spaces, added @Experimental annotation to new overload and fixed typo in javadoc, squashed commits

benjchristensen:
This seems reasonable to me. I don't use 'using' so trust the conversation you've had to address the necessary functionality. Reading through the other PR this does seem necessary and I agree that 'unsafeSubscribe' is appropriate for this.

Issue #405, opened 2013-09-23
combineWithMostRecent


samuelgruetter:
I'm looking for an operation which does the following:

Whenever Observable o1 emits an item, combine this item with the most recent item of Observable o2.

Illustration:

```
----A-------B------C----->  o1

--0----1-2----3-4-------->  o2

    |       |      |
    V       V      V

  (A,0)   (B,2)  (C,4)
```

I can't find a nice way of doing this.
Can anyone help me?
Or do we need to add a new operation to `rx.Observable`?

staltz:
http://stackoverflow.com/questions/28580490/rxjava-how-to-emulate-withlatestfrom
People are asking for this operator.

Some one please implement it?

JakeWharton:
:+1: I have had 4 separate instances of need of this in the last two weeks that I felt dirty working around!

akarnokd:
I'll do this.

staltz:
:+1: :+1:

akarnokd:
See #2760 for the proposed name and behavior.

JakeWharton:
This issue can be closed.

abersnaze:
The `window` operator looks like an interesting option for building this. The initial drawing looks almost exactly like the marble diagram for the operator.

![window marble diagram]
(https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/window8.png)

abersnaze:
Damn it, I couldn't stop thinking about this all night.

```
package asdf;

import static rx.Observable.zip;
import rx.subjects.PublishSubject;
import rx.subjects.Subject;

public class Main {
    private static class Tuple {
        public final String t;
        public final int i;

        Tuple(String t, int i) {
            this.t = t;
            this.i = i;
        }

        @Override
        public String toString() {
            return t + ":" + i;
        }
    }

    public static void main(String[] args) {
        Subject<String, String> trigger = PublishSubject.create();
        Subject<Integer, Integer> data = PublishSubject.create();

        trigger.publish(trigger_ -> {
            return zip(trigger_, data.window(trigger_).flatMap(window -> window.lastOrDefault(-1)), Tuple::new);
        }).scan((last, curr) -> curr.i == -1 ? new Tuple(curr.t, last.i) : curr).subscribe(System.out::println);

        data.onNext(0);
        trigger.onNext("A");
        data.onNext(1);
        data.onNext(2);
        trigger.onNext("B");
        trigger.onNext("C");
        data.onNext(3);
[...]

PR #2760, opened 2015-02-19
Operator: WithLatestFrom

akarnokd:
Named after RxJS, for issue #405

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -8770,6 +8770,30 @@ public class Observable<T> {
     }
 
     /**
+     * Merges the specified observable sequence into this Observable sequence by using the resultSelector 
+     * function only when the source observable sequence (this instance) produces an element.
+     * <code><pre>
+     * ----A-------B------C----->  o1
+     *
+     * --0----1-2----3-4-------->  o2
+     *
+     *     |       |      |
+     *     V       V      V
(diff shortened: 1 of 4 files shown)


src/main/java/rx/Observable.java
```diff
@@ -8770,6 +8770,30 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
...
+     *
+     * --0----1-2----3-4-------->  o2
+     *
+     *     |       |      |
+     *     V       V      V
+     *
+     *   (A,0)   (B,2)  (C,4)
+     * </pre></code>
+     * @param other the other observable sequence
+     * @param resultSelector the function to call when this Observable emits an element and the other
```
JakeWharton: Naming this "resultSelector" seems off (copy/paste?) since it's more akin to `combineLatest` which uses "combineFunction".

akarnokd: [RxJS](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/withlatestfrom.md) has this so I kept the naming. (Our flatMap and groupJoin uses ```resultSelector``` as well).

JakeWharton:
:+1: Thanks for doing this!

benjchristensen:
Thanks @akarnokd

PR #2761, opened 2015-02-20
Helper method to submit an (interruptible) action to an ExecutorService

akarnokd:
Adds two helper methods to ```Schedulers``` which performs the correct ScheduledAction hookups when interacting with an ExecutorService.

Replacement for #2592

### DIFF
--- a/src/main/java/rx/internal/schedulers/ScheduledAction.java
+++ b/src/main/java/rx/internal/schedulers/ScheduledAction.java
@@ -16,28 +16,48 @@
 package rx.internal.schedulers;
 
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.*;
 
 import rx.Subscription;
 import rx.exceptions.OnErrorNotImplementedException;
 import rx.functions.Action0;
+import rx.internal.util.SubscriptionList;
 import rx.plugins.RxJavaPlugins;
 import rx.subscriptions.CompositeSubscription;
 
 /**
  * A {@code Runnable} that executes an {@code Action0} and can be cancelled. The analog is the
  * {@code Subscriber} in respect of an {@code Observer}.
+ * <p><b>System-wide properties:</b>
+ * <ul>
+ * <li>{@code rx.scheduler.interrupt-on-unsubscribe}
+ * <dd>Use {@code Future.cancel(true)} to interrupt a running action? {@code "true"} (default) or {@code "false"}.</br>
+ * </li>
+ * </ul>
  */
 public final class ScheduledAction extends AtomicReference<Thread> implements Runnable, Subscription {
     /** */
     private static final long serialVersionUID = -3962399486978279857L;
-    final CompositeSubscription cancel;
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/schedulers/ScheduledAction.java
```diff
@@ -16,28 +16,48 @@
...
  */
 public final class ScheduledAction extends AtomicReference<Thread> implements Runnable, Subscription {
     /** */
     private static final long serialVersionUID = -3962399486978279857L;
-    final CompositeSubscription cancel;
+    static final boolean DEFAULT_INTERRUPT_ON_UNSUBSCRIBE;
+    static final String KEY_INTERRUPT_ON_UNSUBSCRIBE = "rx.scheduler.interrupt-on-unsubscribe";
+    static {
+        String interruptOnUnsubscribeValue = System.getProperty(KEY_INTERRUPT_ON_UNSUBSCRIBE);
+        DEFAULT_INTERRUPT_ON_UNSUBSCRIBE = interruptOnUnsubscribeValue == null || "true".equals(interruptOnUnsubscribeValue);
```
benjchristensen: I'm not sure it is ever good for this to be a global default. Haven't we learned that it causes nasty issues in places like event loops? It seems only appropriate for separate threads, like the IO or NewThread schedulers.

src/main/java/rx/schedulers/Schedulers.java
```diff
@@ -136,4 +139,52 @@ public static TestScheduler test() {
...
+     * Submit an Action0 to the specified executor service with the option to interrupt the task
+     * on unsubscription and add it to a parent composite subscription.
+     * @param executor the target executor service
+     * @param action the action to execute
+     * @param parent if not {@code null} the subscription representing the action is added to this composite with logic to remove it
+     *            once the action completes or is unsubscribed.
+     * @param interruptOnUnsubscribe if {@code false}, unsubscribing the task will not interrupt the task if it is running
+     * @return the Subscription representing the scheduled action which is also added to the {@code parent} composite
+     */
+    public static Subscription submitTo(ExecutorService executor, Action0 action, CompositeSubscription parent, boolean interruptOnUnsubscribe) {
```
benjchristensen: Looks good I think. Should we just us 'CompositeSubscription' or does this warrant an interface for subscriptions that support removal?

benjchristensen:
Can you also mark them as Beta? These stand the risk of needing to change after playing with them.

src/main/java/rx/internal/schedulers/ScheduledAction.java
akarnokd: Should I add such parameters to the scheduler implementations directly or have them fixed?

src/main/java/rx/schedulers/Schedulers.java
akarnokd: Certainly, an interface would be more general but I'm afraid it turns the usage places of CompositeSubscription into bimorphic or even megamorphic call sites. Let's leave it as CompositeSubscription for now.

akarnokd:
I'll add beta.

akarnokd:
Can't do this right now as it would conflict with #2767.

akarnokd:
Replaced by #2772.

PR #2763, opened 2015-02-20
Added Subscriber#isSubscribed()

artem-zinnatullin:
How many times have you **read or write** code like this:

```java
return Observable.create(new Observable.OnSubscribe()<SomeType> {
    @Override public SomeType call(Subscriber<? super SomeType> subscriber) {
        // ...
        
        if (!subscriber.isUnsubscribed) { // <-- here is the problem
            subscriber.onNext(next);
        }

        // ...
    } 
});
```

I think it's way more clear to write it like this:
```java
if (subscriber.isSubscribed()) { // <-- 50% more readable
    subscriber.onNext(next);
}
```

My brain does not work as CPU, I don't like to do negation operations, especially when I am reading/writing code, I think, nobody likes it.

This change does not affect any `Subscriber` implementation and 100% backward compatible.

`Better tools -> better products`

### DIFF
--- a/src/main/java/rx/Subscriber.java
+++ b/src/main/java/rx/Subscriber.java
@@ -82,6 +82,15 @@ public abstract class Subscriber<T> implements Observer<T>, Subscription {
     }
 
     /**
+     * Indicates whether this Subscriber subscribed to its list of subscriptions
+     *
+     * @return {@code true} if this Subscriber subscribed to its subscriptions, {@code false} otherwise
+     */
+    public final boolean isSubscribed() {
+        return !cs.isUnsubscribed();
+    }
+
+    /**
      * Indicates whether this Subscriber has unsubscribed from its list of subscriptions.
      * 
      * @return {@code true} if this Subscriber has unsubscribed from its subscriptions, {@code false} otherwise
--- a/src/test/java/rx/SubscriberTest.java
+++ b/src/test/java/rx/SubscriberTest.java
@@ -16,6 +16,7 @@
 package rx;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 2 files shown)


akarnokd:
Unsubscription is a terminal state and it is much easier to reason about it. ```isSubscribed```, however is misleading as it could mean to check if the Subscriber has already been subscribed to an observable. Note also that ```Subscriber``` is not final thus adding any non-private method to it is an API breaking change.

artem-zinnatullin:
@akarnokd I agree about terminal state, in first attempt I named method `isStillSubscribed()`, method naming is up to RxJava maintainers :neutral_face:

And yes, `Subscriber` is not final, but this is not API breaking, if adding method (which is final btw) is API breaking change, I would say that adding new `Operator` for `Observable` is breaking change. Or if it's breaking change, it's still 100% backward compatible

But main question: what do you think about use case from example?

benjchristensen:
This is not a change I want to accept. It is duplicative inverse behavior, but more importantly it is quite normal to check for cancellation/unsubscription.  I disagree with your assertion that "nobody likes it" and that negation is somehow wrong. The whole point of a "Subscription" is to be unsubscribed, hence unsubscription being the boolean check.

benjchristensen:
Closing out as I don't want to modify the Subscriber class for this.

davidmoten:
I sympathise with @artem-zinnatullin in that the most common use case for using that method for me is to write ```!isUnsubscribed()``` and I have always found it unfortunate because of the double negative (definitely less readable for me). On repeated use I got used to it but never actually liked it. I appreciate that the subscription state machine looks like NOT_SUBSCRIBED -> SUBSCRIBED -> UNSUBSCRIBED and that coming up with better language in the api has its difficulties. Perhaps we could revisit for rxjava 2.0.

PR #2769, opened 2015-02-23
OperatorCombineLatest request overflow check

davidmoten:
Use ```BackpressureUtils.getAndAddRequest(requested, n)``` instead of  ```requested.getAndAdd(n)``` so that an overflow check takes place.


akarnokd:
Thanks!

PR #2770, opened 2015-02-24
OperatorOnBackpressureDrop request overflow check

davidmoten:
Use ```BackpressureUtils.getAndAddRequest(requested, n)``` instead of ```requested.getAndAdd(n)``` so that an overflow check takes place. Includes a unit test that failed on original code (but passes with this PR).


akarnokd:
Thanks!

PR #2772, opened 2015-02-25
Helper methods to schedule tasks (non-)interruptibly on an

akarnokd:
ExecutorService. Replaces #2761.

By default, io() and newThread() schedules interruptible tasks, computation() schedules non-interruptible tasks.


benjchristensen:
I think this looks good ... but I'm trying to decouple in my head the API changes versus functional related to interrupt, as those are complicated and nuanced.

Before we proceed with this I'd like to have a clear description somewhere about our logic and thinking around interruption. 

Additionally, let's mark the API additions as Experimental.

akarnokd:
Okay.

akarnokd:
I'm closing this, let's revisit the problem later.

PR #2777, opened 2015-02-26
fix several typos

stevenzwu:


src/main/java/rx/observables/AbstractOnSubscribe.java
```diff
@@ -98,7 +98,7 @@
  * <pre><code>
  * Iterable<T> iterable = ...;
  * AbstractOnSubscribe.create(s -> {
- *   Iterator<T> it = s.state();
+ *   Iterator<T> it = iterable.iterator();
```
akarnokd: There is no typo here, the per-subscriber state of iterator is created on line 108.

benjchristensen:
Can you please rebase these 3 commits into 1?

stevenzwu:
first time does rebase/squash. seems to get it correct with one commit now

akarnokd:
Thanks.

Issue #2774, opened 2015-02-25
ConnectableObservable.connect() returns a null subscription if already connected


lopespm:
Any posterior `connect()` calls to an already connected `ConnectableObservable` return `null`, but I believe the expected behaviour would be for it to return the subscription created for the first connection

davidmoten:
I suspect that this is a documentation issue on the method because the source code seems to be quite deliberate in its treatment of this case (can code be deliberate?). No doubt an API expert will comment soon.

akarnokd:
I think it should return a subscription every time, but all places where ```connect(Action1)``` is implemented, the action is only called on the first subscription: see ```OperatorMulticast``` and ```OperatorPublish```.

davidmoten:
I can have a look at making this happen if someone isn't working on it already.

akarnokd:
@davidmoten I'm working on something else so you can go ahead.

benjchristensen:
Was merged in https://github.com/ReactiveX/RxJava/pull/2779

davidmoten:
```OperatorPublish``` has not been done yet, I'll open up a new issue.

PR #2779, opened 2015-02-28
OperatorMulticast.connect(connection) should not return null

davidmoten:
See discussion in  #2774 

Changes include 

* fix for #2774 just for ```OperatorMulticast``` (will do ```OperatorPublish``` in another PR once have sorted this one out)
* made fields private that had default visibility
* renamed ```OnSubscribeMulticastTest``` to ```OperatorMulticastTest```
* addressed a possible race condition that could provoke an IAE when ```Observable.subscribe(sub)``` is called with sub=null (unsubscription occurs between L144 and L145 on old code).

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMulticast.java
+++ b/src/main/java/rx/internal/operators/OperatorMulticast.java
@@ -38,11 +38,11 @@ import rx.subscriptions.Subscriptions;
  *            the result value type
  */
 public final class OperatorMulticast<T, R> extends ConnectableObservable<R> {
-    private final Observable<? extends T> source;
-    private final Object guard;
-    private final Func0<? extends Subject<? super T, ? extends R>> subjectFactory;
-    private final AtomicReference<Subject<? super T, ? extends R>> connectedSubject;
-    private final List<Subscriber<? super R>> waitingForConnect;
+    final Observable<? extends T> source;
+    final Object guard;
+    final Func0<? extends Subject<? super T, ? extends R>> subjectFactory;
+    final AtomicReference<Subject<? super T, ? extends R>> connectedSubject;
+    final List<Subscriber<? super R>> waitingForConnect;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorMulticast.java
```diff
@@ -38,14 +38,16 @@
  *            the result value type
  */
 public final class OperatorMulticast<T, R> extends ConnectableObservable<R> {
-    final Observable<? extends T> source;
-    final Object guard;
-    final Func0<? extends Subject<? super T, ? extends R>> subjectFactory;
+    private final Observable<? extends T> source;
```
akarnokd: I try to avoid private because the compiler generates accessor methods which slows down the interpreter phase and makes an extra hop while debugging methods.

src/main/java/rx/internal/operators/OperatorMulticast.java
```diff
@@ -106,6 +109,21 @@ public void onNext(T args) {
                         subject.onNext(args);
                     }
                 };
+                guardedSubscription = Subscriptions.create(new Action0() {
+                    @Override
+                    public void call() {
+                        Subscription s;
+                        synchronized (guard) {
+                            s = subscription;
```
akarnokd: I'd check if ```guardedSubscription == this``` to avoid unsubscribing a newer connection, i.e.:

```java
Subscription s1 = src.connect();
Subscription s2 = src.connect();

s1.unsubscribe();

Subscription s3 = src.connect();

s2.unsubscribe();

// src should be still connected
```

davidmoten:
Thanks @akarnokd, sharp-eyed as ever. I used a final  AtomicReference to help with the == this check, is this the idiom you'd use in this situation?

akarnokd:
Thanks!

Issue #2596, opened 2015-02-04
Unexpected behavior of publish() and connect()


FinalVersion:
I have experienced some unexpected behaviors of publish() and connect() in my project. 
Hopefully it's not my misunderstanding since I'm a newbie to Rx.

I will use an Observer that simply print onNext, onError, onCompleted.
```Java
	public static class PrintObserver<T> implements Observer<T> {
		@Override
		public void onCompleted() {
			System.out.println("onCompleted");
		}
		@Override
		public void onError(Throwable e) {
			System.out.println("onError: " + e);
		}
		@Override
		public void onNext(T t) {
			System.out.println("onNext: " + t);
		}
	}
```
The simplified scenarios are as follows:

1) **ConnectableObservable stops emitting without unsubscribing from source.**
```Java
    Observable<Integer> source = Observable.range(0, 1000);
    ConnectableObservable<Integer> pub1 = source.publish();
    ConnectableObservable<Integer> pub2 = pub1.publish();
    pub2.connect();
    //pub2.subscribe();
    pub1.subscribe(new PrintObserver<Integer>());
    pub1.connect();
```
I expected the output to be 'onNext: 0' up to 'onNext: 999', then 'onCompleted', but it stops at onNext: 127 in my test.
But if I uncomment `//pub2.subscribe();`, namely have something subscribe to pub2, it will work as expected. Or else if I do not `pub2.connect()`, it will also work.

The code may looks silly, but it may not if I put some operator which does expensive computations before publishing as pub2. BTW, what I want is to do some intermediate computations only once if pub2 is subscribed by many subscribers or piped to different downstream operators. If there're better ways to do that, please kindly let me know.

2) **`publish()` behaves like `replay()`**
```Java
	try {
		Observable<Long> source = Observable.interval(1, TimeUnit.SECONDS).take(10);
		ConnectableObservable<Long> pub = source.publish();
		pub.connect();

		Thread.sleep(2500);
		
		pub.subscribe(new PrintObserver<Long>());
		
		Thread.sleep(2000);
	} catch (InterruptedException e) {
		e.printStackTrace();
	}
```
I expected that after `pub.connect()`, source will start emitting, and when `pub` is subscribed after 2.5 sec, 'onNext: 0' and 'onNext: 1' will be missed by this subscriber. However, I got them all together at 2.5 sec, and from 3rd second, I got the following onNext: 2 and so on. This looks like a `replay()` behavior to me.

FinalVersion:
[...]

PR #2788, opened 2015-03-02
Fix the bug that 'publish' will cache items when no subscriber

### LINKED ISSUE #2775: Confused by behavior of publish+flatten
RxScala 0.23.1 (that means RxJava 1.0.4)

```scala
import rx.lang.scala.Observable

val o1 = Observable.just(1, 2, 3)
val o2 = Observable.just(10, 20, 30)
val o3 = Observable.just(100, 200, 300)

val roundRobinSource = o1.publish(po1 ⇒ o2.publish(po2 ⇒ o3.publish(po3 ⇒ {
  def oneRound: Observable[Int] = po1.take(1) ++ po2.take(1) ++ po3.take(1)
  Observable.just(oneRound, oneRound, oneRound, oneRound, oneRound).flatten
})))
roundRobinSource.subscribe(println, println)
```

gives me

```
1
10
100
1
20
200
1
30
300
1
1
```

That just bemuses me. Why am I getting an infinite stream of ones for the first observable, but for the others it seems OK. And I'm not even sure that the behavior of others is what I'd expect according to the documentation stating that ```publish``` *Returns a rx.lang.scala.observables.ConnectableObservable, which waits until the connect function is called before it begins emitting items from this rx.lang.scala.Observable to those rx.lang.scala.Observers that have subscribed to*. It probably is OK, it is just that I'm not able to deduce from that documentation that ```publish``` is somehow internally keeping track of which items from the published observable have already been emitted and which not even in case of resubscribing to it multiple times...

Can someone enlighten me on this, please?

zsxwing:
Fixed #2775 and #2596

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorPublish.java
+++ b/src/main/java/rx/internal/operators/OperatorPublish.java
@@ -213,6 +213,10 @@ public class OperatorPublish<T> extends ConnectableObservable<T> {
             return false;
         }
 
+        public synchronized boolean hasNoSubscriber() {
+            return subscribers.length == 0;
+        }
+
         public synchronized void incrementOutstandingAfterFailedEmit() {
             outstandingRequests++;
         }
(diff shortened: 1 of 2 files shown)


benjchristensen:
Thanks @zsxwing It looks like a good change to me. Thread-safety seems fine as it is all within the WIP checks.

Issue #2797, opened 2015-03-04
add backpressure support to operator: buffer(long timespan, TimeUnit unit, int count)


stevenzwu:
Ben mentioned that this operator should be upgraded to support backpressure

akarnokd:
Operators with time parameters are unlikely to work with backpressure beause if the downstream doesn't request the next value, the clock is till ticking and values may get lost unexpectedly.

(I'm too much into RxJava 2.0 now and maybe there is a way to work this out.)

benjchristensen:
This one could possibly be made to work since it has size, but if the timer ticks and there is no requested demand it would then have to emit an error. That probably isn't appropriate for this existing operator because emitting an error is not expected.

Is there a way we could have an overload with count/time and a strategy of what to do if time ticks but no requested demand exists? For example ... drop or emit an error? What could the signature be to make that happen?

akarnokd:
Allow me to think out loud:

##### Requesting 1 at a time

The operator requests ```count``` elements and starts the timer. If the count is reached, the timer is cancelled and the buffer is emitted to downstream. If the timer fires first, the buffer is emitted, but any excess upstream value needs to be buffered. Such excess values will be the first in a buffer for new request, and only ```count - len``` is requested from upstream.

The coordination of emitting at count vs. emitting at timer will be challenging.

##### Requesting N 

The operator requests ```count * N``` values and start the timer. If ```count``` elements are received, the timer is cancelled and the buffer is emitted. If the downstream request > 0 after, a new timer is started and the upstream can continue producing. If the timer fires, it takes the available buffer contents and emits it to downstream. If the downstream request > 0 after this emission from the timer thread, a new timer is started. If the downstream requests N and no further requests are made, the final segment will have the same things to consider as in the request-1 case above.

##### Buffer with timespan, timeshift and size

I don't even try to imagine this scenario yet.

-----

I'll try to prototype such an operator.

akarnokd:
I have reimplemented the operator in #2806. Could you check if it works for you?

akarnokd:
We have discussed that making timed operations backpressure aware requires starting and stopping the timed part, it adds the latency of the requesting operation and skews the timing of the buffers in respect to a wall-clock.

PR #2806, opened 2015-03-05
Operator buffer(time) and buffer(time, size) now support backpressure.

akarnokd:
Added backpressure support to ```buffer(time)``` and ```buffer(time, size)``` non-overlapping variants as requested in #2797.

A request of ```n | n < Long.MAX_VALUE``` is converted to a request of ```n * size``` for the upstream. Every time the buffer's size reaches the specified amount, the timer is "paused". The remaining time for it to fire is calculated and stored so the next time the timer is started again with this remaining time, either by a downstream request 0 -> !0 transition or by the ```onNext``` itself because downstream request is not null.

Note that this start-stop behavior is of millisecond accuracy regardless of the original time unit specified because our scheduler's now() return the time in millisecond resolution.

I've also added some new assert methods and constructors to ```TestSubscriber``` to make state checks more convenient and support a common zero request at subscription directly (so no need to extend the class just for this anymore).

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -3289,7 +3289,7 @@ public class Observable<T> {
      * @see <a href="http://reactivex.io/documentation/operators/buffer.html">ReactiveX operators documentation: Buffer</a>
      */
     public final Observable<List<T>> buffer(long timespan, TimeUnit unit, int count, Scheduler scheduler) {
-        return lift(new OperatorBufferWithTime<T>(timespan, timespan, unit, count, scheduler));
+        return lift(new OperatorBufferWithTimeAndSize<T>(timespan, unit, count, scheduler));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


benjchristensen:
I don't think we should be changing the behavior of 'buffer(time)', since time is specifically using a temporal approach to flow control. 

The 'buffer(size)' variant can definitely have backpressure though as per #2797 

I'm not sure how 'buffer(time, count)' should work. One way of looking at that is that it is a temporal operator wih max size. Another is it is a size operator with timeout. If it is temporal then it shouldn't have backpressure. If it is size with timeout then backpressure makes sense.

akarnokd:
Because how this "stops" the timer and resumes it, the changes introduce some additional timing-related jitter. I'd say buffers and windows with time won't support backpressure at all.

benjchristensen:
Then this PR should not be pursued, correct?

We do need an answer though for the 'buffer(count, timeout)' use case. That one is common in scenarios where it wants backpressure and time is just a timeout. 

It's almost like there are:

- buffer(count, timeout)
- buffer(time, maxSize)

They are basically the same without backpressure, but including backpressure makes them have different behaviors when backpressure occurs.

akarnokd:
Currently in the buffer(time, size), time is a periodic indicator to emit what's in the buffer and start a new one. What you suggest is that restart the timer whenever the buffer is full or timeout happened. I did this at first, but then one or two tests failed because they expect the periodicity mentioned above.

benjchristensen:
Can you separate out the `TestSubscriber` changes?

akarnokd:
Sure.

PR #2807, opened 2015-03-05
Corrected all Java interfaces declarations

artem-zinnatullin:
Just little Java syntax fix for interfaces

I checked all interfaces in RxJava.

@benjchristensen I can add and configure `Checkstyle` gradle plugin for RxJava if you don't mind, mm?

### DIFF
--- a/src/main/java/rx/Observer.java
+++ b/src/main/java/rx/Observer.java
@@ -34,7 +34,7 @@ public interface Observer<T> {
      * <p>
      * The {@link Observable} will not call this method if it calls {@link #onError}.
      */
-    public abstract void onCompleted();
+    void onCompleted();
 
     /**
      * Notifies the Observer that the {@link Observable} has experienced an error condition.
@@ -45,7 +45,7 @@ public interface Observer<T> {
      * @param e
      *          the exception encountered by the Observable
      */
-    public abstract void onError(Throwable e);
+    void onError(Throwable e);
 
     /**
      * Provides the Observer with a new item to observe.
@@ -58,6 +58,6 @@ public interface Observer<T> {
      * @param t
      *          the item emitted by the Observable
      */
-    public abstract void onNext(T t);
+    void onNext(T t);
 
 }
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -98,7 +98,7 @@ public class Observable<T> {
     /**
      * Invoked when Obserable.subscribe is called.
      */
-    public static interface OnSubscribe<T> extends Action1<Subscriber<? super T>> {
+    public interface OnSubscribe<T> extends Action1<Subscriber<? super T>> {
         // cover for generics insanity
     }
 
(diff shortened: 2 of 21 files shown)


akarnokd:
I like the changes. However, we don't have a style guideline because we don't want to enforce one particular coding style on contributors: the functionality is first, the style is secondary.

artem-zinnatullin:
@akarnokd great

About code style: project is actively developing and code style can be problem in future, I think it will be good to have small amount of `Checkstyle` rules integrated with build process

You can ping me if you'll decide to do it with my help

colriot:
I agree on the necessity of `Checkstyle`, but may be it's already not the right time to remove `public` from interface methods. See: https://bugs.openjdk.java.net/browse/JDK-8071453. It's pretty distant future, but why do additional work that sums up to zero in this future.

artem-zinnatullin:
@colriot as I see, they won't break source compatibility and `empty` modifier in interface will be `public abstract` as before, if you would like to add implementation to interface, you'll have to write `default` modifier or `private`, also there is 100% no need in declaring `static interface` it's like writing `extends Object` in each class :smile:

colriot:
@artem-zinnatullin yeah, completely agree on `static`, I was talking about `public` only. Speaking of compatibility, what you've said is true, but I personally prefer either to hide all the implicits if they are unambigous and ubiquitous, or to write them explicitly if there are other "theoretical" options, like `private`. Obviously, all that has sense if the library is targeting Java 8 or 9.

benjchristensen:
> I can add and configure Checkstyle gradle plugin

I have a general aversion to `Checkstyle` as it is generally just unnecessary purely subjective opinions imposed upon people and causing otherwise fine code to fail a build. 

In the 2+ years of this project taking pull requests style has rarely been a topic of discussion or issue. 

What legit problems that affect behavior or safety will Checkstyle solve? 

As for the changes in these pull request, a brief review seems like they are okay but I'll spend more time when I have it to review in depth.

headinthebox:
+1 for Ben.

artem-zinnatullin:
`What legit problems that affect behavior or safety will Checkstyle solve?` None of them, just code style.

I got your point, okay.

P.S. Not a fan of Airbnb, just want to quote their ["nerds" post](http://nerds.airbnb.com/maintaining-quality-scale/):

screen shot 2015-03-07 at 16 48 46

But I can't say, that it applicable to RxJava, code style is good in most of files :+1:

src/main/java/rx/Observer.java
```diff
@@ -34,7 +34,7 @@
      * <p>
      * The {@link Observable} will not call this method if it calls {@link #onError}.
      */
-    public abstract void onCompleted();
+    void onCompleted();
```
zsxwing: This is a breaking change. E.g., RxScala has a Scala Observer wrapper that calling these methods. Removing `public abstract` will break them.

zsxwing:
Could we revert this PR? "static interface" is fine. But other changes removing `public` will break RxScala.

src/main/java/rx/Observer.java
JakeWharton: Both modifiers are implicit and redundant. Run `javap` on the class file and you'll see both are still present.

zsxwing: Thank you. Never noticed they are public by default...

PR #2809, opened 2015-03-05
Fixed takeUntil not unsubscribing from either of the observables in case

akarnokd:
of a terminal condition.

This issue came up in [a group discussion](https://groups.google.com/forum/#!topic/rxjava/sF2hy5sV5ck) (but is unrelated to the actual problem there). The problem was that takeUntil was unsafeSubscribed to, there was no one to terminate either the main or the other observable on a terminal condition, leaving the connection to upstream active indefinitely.


benjchristensen:
Thanks for figuring this out and fixing.

PR #2813, opened 2015-03-09
Proposed AbstractProducer to replace AbstractOnSubscribe

akarnokd:
Proposed replacement for ```AbstractProducer``` which gives access to the requested amount allowing batch or fast-path value emission.

I've skipped the detailed javadoc until the feedback about using this instead of ```AbstractOnSubscribe```.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -14,6 +14,7 @@ package rx;
 
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 
 import rx.annotations.*;
 import rx.exceptions.*;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/Observable.java
```diff
@@ -9305,5 +9306,263 @@ public void call(Subscriber<? super T> observer) {
             });
         }
     }
+    
+    /**
+     * @warn javadoc
+     */
+    @Experimental
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static <T, S> Observable<T> create(Action1<ProducerState<T, S>> onRequest) {
```
benjchristensen: Does this conflict with Groovy and Clojure because of type erasure? I'm concerned that 2 overloads that take functions will collide.

src/main/java/rx/Observable.java
```diff
@@ -9305,5 +9306,263 @@ public void call(Subscriber<? super T> observer) {
...
+        long produced(long n);
+    }
+    
+    /**
+     * Abstract base class to help implement a producer.
+     *
+     * @param <T> the type of the values emitted
+     * @param <S> the associated custom resource type
+     */
+    @Experimental
```
benjchristensen: Why does this abstract class need to be public? It feels like a complicated implementation detail and I'm wary of adding it to the public API.

src/main/java/rx/Observable.java
akarnokd: Quite possible. Any suggestions for a name?

src/main/java/rx/Observable.java
akarnokd: Because pre-Java 8, extending it is easier than writing 1-3 inner classes.

src/test/java/rx/AbstractProducerTest.java
```diff
@@ -0,0 +1,295 @@
...
+public class AbstractProducerTest {
+    @Test
+    public void testJust() {
+        Observable<Integer> source = Observable.create(new Action1<ProducerState<Integer, Void>>() {
+            @Override
+            public void call(ProducerState<Integer, Void> ps) {
+                Subscriber<? super Integer> s = ps.subscriber();
+                s.onNext(1);
+                s.onCompleted();
+                ps.unsubscribe();
```
benjchristensen: Why does this unsubscribe eagerly?

src/main/java/rx/Observable.java
benjchristensen: Hmm, good point.

src/main/java/rx/Observable.java
benjchristensen: Thinking about it ... probably need to play with the code a little more to get a feel for it.

src/test/java/rx/AbstractProducerTest.java
akarnokd: To indicate to the resource tracking there won't be any further activity and the AbstractProducer may free it when returning from call().

jbripley:
I really like the AbstractOnSubscribe helper for making backpressure-aware Observables, but I've avoided writing a RxScala wrapper for it since it still seems to be in a bit of a flux.

Will this refactoring, or something like it, be merged soon or am I better off writing a RxScala wrapper for the existing AbstractOnSubscribe in the short term?

akarnokd:
This is not a refactoring but a potential replacement to it to allow exploiting a case of batched requesting and infinite requests and reduce the infrastructure overhead of ```AbstractOnSubscribe```. I personally like ```AbstractOnSubscribe``` better because you only need to think about producing a one element at a time.

jbripley:
FWIW, I went ahead and made a RxScala wrapper for this. Didn't run into any issues other than the inevitable generics compiler errors (managed to crash scalac without much issue), which are now handled. Looking forward to this getting merged so I can use it my Scala code.

src/test/java/rx/AbstractProducerTest.java
benjchristensen: But shouldn't the unsubscribe get invoked immediately anyways due to the Rx contract and `SafeSubscriber` while will trigger an unsubscribe at the end of onCompleted?

benjchristensen:
Related to https://github.com/ReactiveX/RxJava/issues/3003

Trying to figure out async batching ...

src/test/java/rx/AbstractProducerTest.java
akarnokd: If the child is a SafeSubscriber, but if it is another operator with unsafeSubscribe, then most bets are off.

benjchristensen:
Waiting on proposal from @stealthcode (in https://github.com/ReactiveX/RxJava/issues/3003) to compare options.

akarnokd:
PR #3118 has much lower overhead and apart from slight API concerns it is a better replacement for `AbstractOnSubscribe` than this.

Issue #2810, opened 2015-03-06
OperatorPublish bugs


akarnokd:
##### Reconnect issue

The following test fails because in OperatorPublish, the ```state.getOrigin()``` still holds the subscriber of a previous connection:

```java
@Test
public void testSubscribeAfterDisconnectThenConnect() {
    ConnectableObservable<Integer> source = Observable.just(1).publish();
    
    TestSubscriber<Integer> ts1 = new TestSubscriber<Integer>();
    
    source.subscribe(ts1);
    
    Subscription s = source.connect();
    
    ts1.assertReceivedOnNext(Arrays.asList(1));
    ts1.assertNoErrors();
    ts1.assertTerminalEvent();
    
    TestSubscriber<Integer> ts2 = new TestSubscriber<Integer>();
    
    source.subscribe(ts2);
    
    Subscription s2 = source.connect();
    
    ts2.assertReceivedOnNext(Arrays.asList(1));
    ts2.assertNoErrors();
    ts2.assertTerminalEvent();
    
    System.out.println(s);
    System.out.println(s2);
}
```

However, it works with ```share()```:
```java
@Test
public void testShareReconnect() {
    Observable<Integer> source = Observable.just(1).share();
    
    TestSubscriber<Integer> ts1 = new TestSubscriber<Integer>();
    
    source.subscribe(ts1);
    
    TestSubscriber<Integer> ts2 = new TestSubscriber<Integer>();
    
    source.subscribe(ts2);
    
    ts1.assertNoErrors();
    ts1.assertTerminalEvent();
    ts1.assertReceivedOnNext(Arrays.asList(1));
    
    ts2.assertNoErrors();
    ts2.assertTerminalEvent();
    ts2.assertReceivedOnNext(Arrays.asList(1));
}
```

##### connect() is racy

In addition, the ```OperatorPublish.connect()``` method is inherently racy: two concurrent connect calls may succeed, one kicking out the upstream subscriber of the other.

##### OriginSubscriber.requestMore may overflow 

```originOutstanding.addAndGet()``` may turn negative

##### Retention of subscribers after completion event.

After making all relevant fields and classes package-private, the following test fails:
```java
@Test
public void testNoSubscriberRetentionOnCompleted() {
    OperatorPublish<Integer> source = (OperatorPublish<Integer>)Observable.just(1).publish();
        
    TestSubscriber<Integer> ts1 = new TestSubscriber<Integer>();
        
    source.unsafeSubscribe(ts1);
[...]

Issue #2803, opened 2015-03-05
OperatorPublish connect(connection) should not return null


davidmoten:
This is outstanding work from #2774.  I can have a go at it next week if no-one in a rush.

akarnokd:
If you don't mind some Reactive-Streams style code, you can look at my [OperatorPublish](https://github.com/akarnokd/RxJavaFlow/blob/master/src/main/java/rx/internal/operators/OperatorPublish.java) which definitely doesn't return null on connection, but due to the half-finished project, I can't run all tests relying on it.

davidmoten:
Looks to me that ```OperatorPublish``` needs a good chunk of rework (#2810) and that the ReactiveStreams code could be basis for a rewrite. I'd be happy to have a go at this but haven't much time on my hands this week and anyway it might make more sense for @akarnokd to do it as he's so familiar with the subject already (and would probably do it in the blink of an eye).

akarnokd:
I'll add this to my todo list for this week.

akarnokd:
The operator has been rewritten so it never returns null. Closing.

PR #2814, opened 2015-03-10
Operator Publish full rewrite

akarnokd:
Complete rewrite of ```publish()``` and fixed bugs of #2810 and #2803.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorPublish.java
+++ b/src/main/java/rx/internal/operators/OperatorPublish.java
@@ -15,367 +15,736 @@
  */
 package rx.internal.operators;
 
-import java.util.*;
+import java.util.Queue;
 import java.util.concurrent.atomic.*;
 
 import rx.*;
-import rx.Observable;
-import rx.exceptions.*;
+import rx.exceptions.MissingBackpressureException;
 import rx.functions.*;
-import rx.internal.util.RxRingBuffer;
+import rx.internal.util.*;
+import rx.internal.util.unsafe.*;
 import rx.observables.ConnectableObservable;
 import rx.subscriptions.Subscriptions;
 
-public class OperatorPublish<T> extends ConnectableObservable<T> {
+/**
+ * A connectable observable which shares an underlying source and dispatches source values to subscribers in a backpressure-aware
+ * manner. 
+ * @param <T> the value type
+ */
+public final class OperatorPublish<T> extends ConnectableObservable<T> {
+    /** The source observable. */
     final Observable<? extends T> source;
-    private final RequestHandler<T> requestHandler;
+    /** Holds the current subscriber that is, will be or just was subscribed to the source observable. */
+    final AtomicReference<PublishSubscriber<T>> current;
 
+    /**
+     * Creates a OperatorPublish instance to publish values of the given source observable.
+     * @param source the source observable
+     * @return the connectable observable
+     */
     public static <T> ConnectableObservable<T> create(Observable<? extends T> source) {
-        return new OperatorPublish<T>(source);
+        // the current connection to source needs to be shared between the operator and its onSubscribe call
+        final AtomicReference<PublishSubscriber<T>> curr = new AtomicReference<PublishSubscriber<T>>();
+        OnSubscribe<T> onSubscribe = new OnSubscribe<T>() {
+            @Override
+            public void call(Subscriber<? super T> child) {
+                // concurrent connection/disconnection may change the state, 
+                // we loop to be atomic while the child subscribes
+                for (;;) {
+                    // get the current subscriber-to-source
+                    PublishSubscriber<T> r = curr.get();
+                    // if there isn't one or it is unsubscribed
+                    if (r == null || r.isUnsubscribed()) {
+                        // create a new subscriber to source
+                        PublishSubscriber<T> u = new PublishSubscriber<T>(curr);
+                        // perform extra initialization to avoid 'this' to escape during construction
+                        u.init();
+                        // let's try setting it as the current subscriber-to-source
+                        if (!curr.compareAndSet(r, u)) {
+                            // didn't work, maybe someone else did it or the current subscriber 
+                            // to source has just finished
+                            continue;
+                        }
+                        // we won, let's use it going onwards
+                        r = u;
+                    }
+                    
+                    // create the backpressure-managing producer for this child
+                    InnerProducer<T> inner = new InnerProducer<T>(r, child);
+                    /*
+                     * Try adding it to the current subscriber-to-source, add is atomic in respect 
+                     * to other adds and the termination of the subscriber-to-source.
+                     */
+                    if (!r.add(inner)) {
+                        /*
+                         * The current PublishSubscriber has been terminated, try with a newer one.
+                         */
+                        continue;
+                        /*
+                         * Note: although technically corrent, concurrent disconnects can cause 
+                         * unexpected behavior such as child subscribers never receiving anything 
+                         * (unless connected again). An alternative approach, similar to 
+                         * PublishSubject would be to immediately terminate such child 
+                         * subscribers as well:
+                         * 
+                         * Object term = r.terminalEvent;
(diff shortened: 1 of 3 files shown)


benjchristensen:
Wow. Will have to come back to this one :-) that's a pretty epic fix of my bad code!

benjchristensen:
@davidmoten have you by chance tried this and been able to confirm it fixes your issues?

akarnokd:
Ah, it was easy; you only need to dualize merge :).

davidmoten:
@benjchristensen the original issue is covered by the rewrite, thanks @akarnokd . I've given the code a once over but not thorough, would take me a while and have little time at the moment.

davidmoten:
@akarnokd I have had another brief foray into the code and I have every faith in your ability to create correct code but I'm not having a great time following it in the short slots of time I have between kid juggling. For longevity's sake I'm wondering if this beautiful looking bit of code could be bulked up with comments to lead me through and to lead future maintainers through. Of course if @benjchristensen or @zsxwing or another of the usual suspects think it's clear as day then further comments may be unnecessary.

akarnokd:
Sure.

akarnokd:
I've rebased, squashed and added comments to this PR.

davidmoten:
In `OperatorPublishTest.testPublish` (old code, you made no changes), would be nice to have an assert between the calls to `subscribe` and the call to `connect` just to ensure that subscription has occured but nothing is emitted before the call to `connect`. This may be covered by other tests but not explicitly so I think could be worthwhile addition.

akarnokd:
Thanks! Updated tests accordingly and fixed a case with clients starting with 0 requests.

davidmoten:
Perhaps add a `connect` idempotency test?

akarnokd:
Done.

src/main/java/rx/internal/operators/OperatorPublish.java
```diff
@@ -15,367 +15,736 @@
...
+                        int len = ps.length;
+                        // Let's assume everyone requested the maximum value.
+                        long maxRequested = Long.MAX_VALUE;
+                        // count how many have triggered unsubscription
+                        int unsubscribed = 0;
+                    
+                        // Now find the minimum amount each child-subscriber requested
+                        // since we can only emit that much to all of them without violating
+                        // backpressure constraints
+                        for (InnerProducer<T> ip : ps) {
```
davidmoten: This is one strategy I suppose and another might be request max but buffer responses for those subscribers that want less than max. I can imagine that either way somebody could suffer and I wonder if this strategy should be configurable?

davidmoten:
By the way thanks for the comments in the code, they are just what I need!

src/main/java/rx/internal/operators/OperatorPublish.java
akarnokd: You can only request the common minimum amount from upstream if the front-queue is bounded, even if some child subscribers request more. These may run ahead but then stay idle until the 'slower ones' catch up.

akarnokd:
I'm preparing a JMH benchmark and it seems the current master fails with MissingBackpressureException on simple chains such as:
```java
@Test
public void testObserveOn() {
    ConnectableObservable<Integer> co = Observable.range(0, 1000).publish();
    Observable<Integer> obs = co.observeOn(Schedulers.computation());
    for (int i = 0; i < 100; i++) {
        TestSubscriber<Integer> ts = new TestSubscriber<Integer>();
        obs.subscribe(ts);
        Subscription s = co.connect();
        ts.awaitTerminalEvent(2, TimeUnit.SECONDS);
        ts.assertTerminalEvent();
        ts.assertNoErrors();
        assertEquals(1000, ts.getOnNextEvents().size());
        s.unsubscribe();
    }
}
```
I'll post and merge the PR for the perf test so it can be run independent of this PR.

akarnokd:
I'll post its benchmark result but it takes a looong time. So far, I'm impressed (sync, 1 batch, 1 child, 1M elements):
master: 3.822 ops/s
this: 18.810 ops/s

akarnokd:
[Here is](https://gist.github.com/akarnokd/8e4591bd4441cedb600e) the benchmark result for this PR in [JMH Comparison GUI](https://github.com/akarnokd/jmh-compare-gui) format. This PR outperforms the current master by 1.5 - 4x most of the time. However, some async parameter combination timed out despite the high CPU usage; I'll look into those cases.

src/main/java/rx/internal/operators/OperatorPublish.java
```diff
@@ -15,367 +15,736 @@
...
+                     * Try adding it to the current subscriber-to-source, add is atomic in respect 
+                     * to other adds and the termination of the subscriber-to-source.
+                     */
+                    if (!r.add(inner)) {
+                        /*
+                         * The current PublishSubscriber has been terminated, try with a newer one.
+                         */
+                        continue;
+                        /*
+                         * Note: although technically corrent, concurrent disconnects can cause
```
benjchristensen: By concurrent disconnect, I assume you mean concurrent subscribe/disconnect? If so then yes, the child subscribers may never receive anything if it is not re-connected. But that is exactly correct behavior if that's what happens. 

Am I understanding your comment here correctly?

src/main/java/rx/internal/operators/OperatorPublish.java
```diff
@@ -15,367 +15,736 @@
...
+                                // meantime or hasn't requested anything yet
+                                // this eager behavior will skip unsubscribed children in case
+                                // multiple values are available in the queue
+                                if (ip.get() > 0L) {
+                                    try {
+                                        ip.child.onNext(value);
+                                    } catch (Throwable t) {
+                                        // we bounce back exceptions and kick out the child subscriber
+                                        ip.unsubscribe();
+                                        ip.child.onError(t);
```
benjchristensen: I believe this is correct, but making sure my thinking is right. If the child `onError` itself throws (and breaks the contract) there is nothing we can do except blow up the entire thing, correct?

src/main/java/rx/internal/operators/OperatorPublish.java
```diff
@@ -15,367 +15,736 @@
...
+                                    ip.produced(1);
+                                }
+                            }
+                            // indicate we emitted one element
+                            d++;
+                        }
+                        
+                        // if we did emit at least one element, request more to replenish the queue
+                        if (d > 0) {
+                            request(d);
```
benjchristensen: Glad to see you requesting in batches.

src/main/java/rx/internal/operators/OperatorPublish.java
```diff
@@ -15,367 +15,736 @@
...
+         */
+        static final long UNSUBSCRIBED = Long.MIN_VALUE;
+        /**
+         * Indicates this child has not yet requested any value. We pretend we don't
+         * see such child subscribers in dispatch() to allow other child subscribers who
+         * have requested to make progress. In a concurrent subscription scennario,
+         * one can't be sure when a subscription happens exactly so this virtual shift
+         * should not cause any problems.
+         */
+        static final long NOT_REQUESTED = Long.MIN_VALUE / 2;
```
benjchristensen: This magic number made me smile :)

benjchristensen:
Excellent code. Excellent comments. Very well done @akarnokd. This is a very hard piece of code. I can only imagine how much time you spent on this.

As best as I can tell it's all correct. It's complicated enough I may be missing something, but everything seems correct and matches what I wish I had coded this to be originally!

src/main/java/rx/internal/operators/OperatorPublish.java
akarnokd: Yes, the behavior implemented and what you said is the correct one.

src/main/java/rx/internal/operators/OperatorPublish.java
akarnokd: This is what happens, yes. I'm not sure if we want to go that deep and paranoid about onError and onCompleted.

PR #2816, opened 2015-03-11
Merged merge, mergeDelayError and mergeMaxConcurrent into a single merge

akarnokd:
Added overload to ```mergeDelayError``` to take a ```maxConcurrent``` parameter as well.

```OperatorMapNotification``` had to be fixed because it disregarded backpressure on its ```onError``` and ```onCompleted``` paths: these produce value that may lack of an associated request and have to be buffered. The downside is the extra overhead due to serialization via ```drain()``` instead of write-through, but it is required for correctness.


benjchristensen:
Thanks for tackling this ...

This will take a while to test. It took several days over a couple weeks to get the current implementation to work in our production environments. 

We will need to analyze how this impacts GC behavior and other such things.

benjchristensen:
cc @abersnaze @mattrjacobs to look at

akarnokd:
Closing because it conflicts with #2878 and has the same bugs the newer PR fixes.

PR #2820, opened 2015-03-12
Backpressure for window(size)

akarnokd:
Changes to the ```window(size)``` operator to respect the backpressure on its *outer* Observable: asking for 1 window will request ```size``` values from upstream. 

Backpressure is ignored on the inner Observable for now, partially because the ```BufferUntilSubscriber``` doesn't support it, partially because coordinating the inner requests with the outer requests needs more thoughts. The problem is that the outer has to request at least 1 element from upstream in order to open the window, but the inner subscriber may not want that single element just yet or would request more than the remaining window size and it would trigger new windows whose value it can't receive but would overflow the next window's observers.

```window(size, skip)``` is not changed as I need to think about it more.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorWindowWithSize.java
+++ b/src/main/java/rx/internal/operators/OperatorWindowWithSize.java
@@ -15,18 +15,14 @@
  */
 package rx.internal.operators;
 
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
 
-import rx.Observable;
+import rx.*;
 import rx.Observable.Operator;
-import rx.Subscription;
+import rx.Observable;
+import rx.Observer;
 import rx.functions.Action0;
 import rx.subscriptions.Subscriptions;
-import rx.Observer;
-import rx.Subscriber;
 
 /**
  * Creates windows of values into the source sequence with skip frequency and size bounds.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by DavidMGross** [comment]
Could you change the javadoc for this operator  in Observable.java as well: http://reactivex.io/RxJava/javadoc/rx/Observable.html#window(int)

It currently reads:

Backpressure Support:
This operator does not support backpressure as it uses count to control data flow.

**Comment by akarnokd** [comment]
Sure.

**Comment by benjchristensen** [comment]
Isn't supporting backpressure on the inner effectively the same as 'onBackpressureBuffer'?

**Comment by akarnokd** [comment]
Yes. BufferUntilSubscriber would need to be smartened to do proper backpressure.

**Comment by benjchristensen** [comment]
What has happened recently causing tests to fail? I don't see any recent commits at should do it. Ideas?

**Comment by akarnokd** [comment]
I have no idea. I thought it was my commit for testOnBackpressureDropWithAction but reverting that didn't fix the master. My best guess is that something changed in the Travis-CI environment; perhaps we don't get 3GB memory anymore (on Windows, the tests take up to 1GB of memory).

**Comment by benjchristensen** [comment]
So to confirm, merging this would be adding back pressure on the outer but still need us to come back later and fix the inner?

**Comment by akarnokd** [comment]
Yes.

File: src/main/java/rx/internal/operators/OperatorWindowWithSize.java [hunk 1]
```diff
@@ -78,26 +74,36 @@ public ExactSubscriber(Subscriber<? super Observable<T>> child) {
...
-        }
-
-        @Override
-        public void onStart() {
-            // no backpressure as we are controlling data flow by window size
-            request(Long.MAX_VALUE);
+            child.setProducer(new Producer() {
+                @Override
+                public void request(long n) {
+                    if (n > 0) {
```
benjchristensen: Curious, why not use BackpressureUtils here? Is it because you are just passing the value through rather than maintaining the AtomicLong in this class?

Issue #2811, opened 2015-03-06
RXJava pattern matching


robertroeser:
Hi,

I think a good enhancement to RXJava would be the ability to do pattern matching. Currently to do this you have to do something like a flatmap, and use if/else-if/else statements within the closing function. I'd like to be able to do something like the following:

```
Observable.range(0,5)
	.case(i->i==0, i->{System.out.println("i == 0")})
	.case(i->i==1, i->{System.out.println("i == 1")})
	.default(System.out::println)
```

Would other people find this useful? Since RXJava is functional, what is a good way to do conditional logic in a functional paradigm - pattern matching as mentioned or is there a better construct?

Thanks,
Robert

abersnaze:
The only equivalent to that would be `doOnNext` or 'map'.

```
Observable.range(0, 5)
    .doOnNext(i -> { if (i == 0) System.out.println("i == 0"); })
    .doOnNext(i -> { if (i == 1) System.out.println("i == 1"); })
    .doOnNext(i -> { if (i != 0 || i != 1) System.out.println(i); });
```

I haven't needed anything like that personally but this is how I would envision the API to look like.

```
Observable<T> {
  public <R> SwitchBuilder<T, R> case(Func1<T, Boolean> predicate, Func1<T, R> mapping);
}

SwitchBuilder<T, R> {
  public SwitchBuilder<R> case(Func1<T, Boolean> predicate, Func1<T,R> mapping);
  public Observable<R> defaultCase(Func1<T,R> mapping);
}
```

benjchristensen:
I think the case/default operators may need to be inside another operator so as to ensure default behavior, otherwise each `case` is independent, and `default` would have no knowledge of anything else. Also, the default keyword can't be used. 

```java
Observable.range(0,5).match(s -> {
    return s.case(i->i==0, i->{System.out.println("i == 0")})
                .case(i->i==1, i->{System.out.println("i == 1")})
                .default(System.out::println)
}
```

@headinthebox Is there an approach for pattern matching that works in the Rx model beyond just putting conditional logic inside a `flatMap`?

headinthebox:
I don't see any problem with using flatmap & if statements. Pattern matching has zero to do with functional programming, and as long as the underlying language (Java) does not support complex patterns it is utterly useless to add such complexity to Rx IMHO.

abersnaze:
[...]

PR #2821, opened 2015-03-12
match operator

robertroeser:
Added an match operator that constructs chain of predicates mapped to a function. The predicate will be applied to what an Observable emits, and if the predicate returns true the function mapped to the predicate will be applied.

The match operator creates a MatchBuilder which can either chain subsequent matches or return the Observable back when the defaultMatch method is called. This ensures there is a default match provided.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -9307,9 +9307,7 @@ public class Observable<T> {
     }
 
     /**
-     * Filters items emitted by an Observable by only emitting those that satisfy a specified predicate.
-     *
-     * Match emits a MatchBuilder that lets you chain a series of matches a series of matches together. To finish the chain
+     * Returns a MatchBuilder that lets you chain a series of matches together. To finish the chain
      * matchDefault is called which emits an Observable.
      *
      * @param predicate
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -9362,7 +9362,9 @@ public class Observable<T> {
 
         Observable<T> observable;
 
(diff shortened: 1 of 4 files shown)


akarnokd:
Based on the conversation in #2811, I don't think this is worth adding to the core library.

robertroeser:
There were a couple of assertions being made. The first that pattern is complicated. I took about 4 - 5 hours and wrote this. There are operators in RX today are far more sophisticated.

The other assertion was pattern matching has nothing to do with functional program/useless. I find that point confusing as scala, haskell, ML*, Lisp, Erlang, F#, etc all have pattern matching. The most intuitive and easiest choice would be a construct that matches other functional languages, not having developer to know to use a flatMap and if and elses. 

I think this more intuitive:
```
Observable<Void>
        .create(kafkaConsumer())
        .match(i -> i instanceof Foo.class, processor::foo)
        .match(i -> i instanceof Bar.class, processor::bar)
        .match(i -> i instanceof Baz.class, processor::baz)
        .defaultCase(processor::handleUnknowMessage)
        .map(i -> i.ack())
```

Then this:
```
Observable<Void>
        .create(kafkaConsumer())
        .flatMap(i -> {
            if (i instanceof Foo.class) {
                return processor.foo(i);
            } else if (i instanceof Bar.class) {
                return processor.bar(i);
            } else if (i instance if Baz.class) {
                return processor.baz(i);
            } else {
                return processor.handleUknowMessage(i);
            }
        })
        .map(i->i.ack());
```

headinthebox:
IMHO it makes no sense to duplicate functionality that is already available directly in the language. That adds unnecessary weight.

Also, it is a far stretch to call this pattern matching. Pattern matching binds variables. This is just different syntax for conditionals.

Issue #2826, opened 2015-03-15
Why retry method with predicate only retries 2 times?


fatroom:
Hi, can somebody explain me why following piece of code printing only two times message "Retrying"? And not 5 times as specified in predicate?
BTW if i uncomment doOnError line in this sample - i got 5 times message "Retrying".
Any ideas what's going on here?

```
Observable.just(1)
                .map(value -> { throw new RuntimeException("You shall not pass"); })
//                .doOnError(t -> System.out.println("Handling error"))
                .retry((counter, throwable) -> {
                    System.out.println("Retrying");
                    return counter < 5;
                })
                .toBlocking()
                .single();
```

akarnokd:
The bug has been fixed through #2852. Closing.

PR #2827, opened 2015-03-15
The retry with predicate wasn't behaving correctly when it used unsafeSubscribe

abersnaze:
For #2826
The `ScalarSynchronousObservable` and `OperatorRetryWithPredicate` both assume that everyone else obey the rules so they don't have to.  Bad things happen when they are used together.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorRetryWithPredicate.java
+++ b/src/main/java/rx/internal/operators/OperatorRetryWithPredicate.java
@@ -109,7 +109,7 @@ public final class OperatorRetryWithPredicate<T> implements Observable.Operator<
                         };
                         // register this Subscription (and unsubscribe previous if exists) 
                         serialSubscription.set(subscriber);
-                        o.unsafeSubscribe(subscriber);
+                        o.subscribe(subscriber);
                     }
                 });
             }
--- a/src/test/java/rx/internal/operators/OperatorRetryWithPredicateTest.java
+++ b/src/test/java/rx/internal/operators/OperatorRetryWithPredicateTest.java
@@ -16,11 +16,15 @@
 package rx.internal.operators;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+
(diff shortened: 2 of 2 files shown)


src/main/java/rx/internal/operators/OperatorRetryWithPredicate.java
```diff
@@ -109,7 +109,7 @@ public void onNext(T v) {
                         };
                         // register this Subscription (and unsubscribe previous if exists) 
                         serialSubscription.set(subscriber);
-                        o.unsafeSubscribe(subscriber);
+                        o.subscribe(subscriber);
```
davidmoten: Does replacing the ```unsafeSubscribe``` with ```subscribe``` open us up to an ```IllegalArgumentException``` in ```OperatorMerge``` (see #2604)?

abersnaze: I don't think so because it doesn't reuse the upstream subscriber and it decouples the subscriptions with the `SerialSubscription`.  This probably isn't the most efficient way to solve the problem but it is probably the simplest.

akarnokd: Instead of this, I'd add a done flag to the inner Subscriber to ignore any subsequent events; should be more lightweight.

akarnokd:
To clarify the underlying issue, the problem is that merge captures the error and pushes it downstream, which schedules a resubscription. However, just doesn't know about this case, and happily sends out onCompleted which then reaches the subscriber after the retry operator.

akarnokd:
We've fixed the main build, but this still fails and I don't get why. Could you rebase this to the current master?

akarnokd:
I found [this](http://stackoverflow.com/questions/29324886/retrywhen-operator-never-retries) question on StackOverflow which reflects on a similar problem; the fix I suggested instead of this should solve both problems.

> Instead of this, I'd add a done flag to the inner Subscriber to ignore any subsequent events; should be more lightweight.

akarnokd:
I'm closing this, see #2852 for an extended fix.

PR #2830, opened 2015-03-17
Disable MySQL and PostgreSQL on travis

akarnokd:
Not for merge. Diagnosing build failures.


akarnokd:
Closing. It appears travis won't start the build unless ```sudo: false``` is set.

benjchristensen:
Yes we need sudo:false to use their new infrastructure.

PR #2834, opened 2015-03-18
Figuring out why the build fails

akarnokd:
Not for merge. Trying to find what's causing SIGKILL.


akarnokd:
This is extremely strange. The tests still get killed, but  it succeeded on the ```linux-4``` building machine once.

I've added memory leak detection which calls GC after each test which (tests like 3x slower) appears to not trigger the condition to sigkill the test. Either the build environments are corrupted or a JVM GC bug is triggered.

akarnokd:
The test run out of memory not so gracefully. I'll close this PR after I moved a few fixes out into another PR.

PR #2847, opened 2015-03-27
More execution hook support

abersnaze:
I've made a couple of changes to the where the execution hooks are used.
* Moved the `onCreate()` from the create method to the constructor to capture things going on with other subclasses of `Observable` like `GroupedObservable` and `ScalarSynchronousObservable`.
* Moved the `onLift()` hook outside of the closure so as to capture information about where/when the operators lifted rather than when/where they were subscribed.

I also made a change to the `ScalarSynchronousObservable.get()` have a slow path to allow hooks know when the data is pulled as if it was subscribed to normally.

### DIFF
--- a/src/main/java/rx/internal/util/ScalarSynchronousObservable.java
+++ b/src/main/java/rx/internal/util/ScalarSynchronousObservable.java
@@ -15,9 +15,14 @@
  */
 package rx.internal.util;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+
 import rx.*;
 import rx.Scheduler.Worker;
+import rx.exceptions.Exceptions;
 import rx.functions.Action0;
+import rx.internal.operators.NotificationLite;
 import rx.internal.schedulers.EventLoopsScheduler;
 
 public final class ScalarSynchronousObservable<T> extends Observable<T> {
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/util/ScalarSynchronousObservable.java
```diff
@@ -29,28 +34,105 @@
     private final T t;
 
     protected ScalarSynchronousObservable(final T t) {
-        super(new OnSubscribe<T>() {
+        super(new ScalarSynchronousObservableOnSubscribe<T>(t));
+        this.t = t;
+    }
+    private static final class ScalarSynchronousObservableOnSubscribe<T> implements OnSubscribe<T> {
```
akarnokd: Why would you want to hook the scalar value?

src/main/java/rx/internal/util/ScalarSynchronousObservable.java
```diff
@@ -29,28 +34,105 @@
...
+        // Slow path because the observable execution hooks are in effect.  We
+        // have no idea what the onSubscribe is going to do anymore so this code
+        // is being cautious and doesn't assume that it is going to scalar or
+        // synchronous anymore.  I intentionally also didn't use toBlocking()
+        // or any other operators to avoid adding noise to whatever the hooks
+        // are trying to record.  The subscribe below simulates the full
+        // subscribe/onNext/onComplete/unsubscribe life cycle of this observable.
+
+        // use a latch just in case it is asynchronous
+        final CountDownLatch done = new CountDownLatch(1);
```
akarnokd: This makes merging scalar-hooked values a blocking operation.

abersnaze: `ScalarSynchronousObservable.get()` is already a blocking operation. The instance of check at the top short circuits the complicated logic below.

src/main/java/rx/internal/util/ScalarSynchronousObservable.java
abersnaze: If not than the debug hooks have gaps in the data flow.  It looks like `onNext`s are produced out of nowhere.

abersnaze:
Can we merge this or is there more you want to discus?

akarnokd:
I don't like the changes in the scalar.

abersnaze:
Could you be a little more specific so that I can counter or fix them? @benjchristensen you want to weigh in on this as a tiebreaker?

akarnokd:
The get() is now possibly blocking if hooks are applied and makes any optimization based on the availability of the scalar value non-blockingly mute.

abersnaze:
First `get()` by definition is blocking.  Second the `CountDownLatch` will only be put the thread to sleep if the new `onSubscribe` set by the `RxJavaObservableExecutionHook` changes the implementation to something async which for most part I would assume they don't.  This code can't make that assumption.

To settle this here are some numbers measuring the performance of `flatMap()` of synchronous Observables.  All runs with this command `./gradlew benchmarks '-Pjmh=-f 1 -tu ns -bm avgt -wi 5 -i 5 -r 1 .*flatMapIntPassthruSync.*'`

Before
```
Benchmark                                         (size)   Mode   Samples        Score  Score error    Units
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync         1   avgt         5      206.682        8.275    ns/op
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync      1000   avgt         5    30602.662     1086.577    ns/op
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync   1000000   avgt         5 34499052.273  1042145.398    ns/op
```

With this code change and instanceof check evaluating to fast path.
```
# Run complete. Total time: 00:00:37

Benchmark                                         (size)   Mode   Samples        Score  Score error    Units
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync         1   avgt         5      224.155        5.730    ns/op
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync      1000   avgt         5    42361.367     4907.785    ns/op
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync   1000000   avgt         5 38511919.011 13512306.519    ns/op
```

With this code change and instanceof check forced to the new slow path but still using the original `ScalarSynchronousObservableOnSubscribe`
```
# Run complete. Total time: 00:00:37

Benchmark                                         (size)   Mode   Samples        Score  Score error    Units
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync         1   avgt         5      301.834       27.519    ns/op
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync      1000   avgt         5   125380.888     1180.281    ns/op
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync   1000000   avgt         5 131494888.889  6862333.809    ns/op
```

akarnokd:
I don't see why you say ```get``` is blocking: the original has no locks in it and is probably JIT compiled to a field access; and there was no "definition" not even in the form of a javadoc.  Based on your perf, this change has a cost of about 10% which may negate the direct-scheduler optimization of the scalar observable.

abersnaze:
I mean blocking in that the method can not return until it has a value for `t`.

akarnokd:
I suggest adding a method ```isNonblockingGet()``` which should return true with the original scalar and false otherwise: this way, optimizations expecting get() to be non-blocking can still work and there is no need to alter ```get()```.

abersnaze:
@benjchristensen suggested adding a new method to `RxJavaObservableExecutionHook` for creation of the `ScalarSynchronousObservable` so that the debug hook could override to return an equivalent that wouldn't be optimized out (like `Observable.from(Collections.signleton(item)`)

I haven't thought too much about what the new hook method would like but probably something like 
```
public <T> Observable<T> onCreateEmpty();
public <T> Observable<T> onCreateJust(T item);
```

abersnaze:
I'm going to refactor and rebase this.

PR #2849, opened 2015-03-30
Revert "Corrected all Java interfaces declarations"

zsxwing:
Reverts ReactiveX/RxJava#2807

Revert because they are breaking changes. RxScala needs the `public` modifier so that it can wrap them into Scala classes.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -98,7 +98,7 @@ public class Observable<T> {
     /**
      * Invoked when Observable.subscribe is called.
      */
-    public interface OnSubscribe<T> extends Action1<Subscriber<? super T>> {
+    public static interface OnSubscribe<T> extends Action1<Subscriber<? super T>> {
         // cover for generics insanity
     }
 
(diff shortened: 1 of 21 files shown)


JakeWharton:
:-1:

All interface methods are public and abstract regardless of whether you declare them.
All interfaces are static when nested regardless of whether you declare it.

akarnokd:
Is this some kind of Scala compiler bug?

zsxwing:
> Is this some kind of Scala compiler bug?

No. I thought they were different. But since the byte codes are same, it won't break RxScala.

artem-zinnatullin:
LOL, it just can't be true, many Scala apps depend on Java code without `public abstract` modifiers in interfaces, I think it's a problem of mixing concepts "Interface" in Java and "Trait" in Scala, everything should work without these modifiers

zsxwing:
> LOL, it just can't be true, many Scala apps depend on Java code without public abstract modifiers in interfaces, I think it's a problem of mixing concepts "Interface" in Java and "Trait" in Scala, everything should work without these modifiers

Right. Thank you.

PR #2852, opened 2015-04-02
Change retryWhen to eagerly ignore an error'd source's subsequent events

akarnokd:
Fixes both issue #2826 and [this StackOverflow](http://stackoverflow.com/questions/29324886/retrywhen-operator-never-retries) problem.


benjchristensen:
Should I release 1.0.9 with current 1.x branch?

akarnokd:
Yes.

PR #2855, opened 2015-04-06
toBlocking.single should not request more than needed

davidmoten:
When ```observable.toBlocking().single()``` was called the number requested from ```observable``` was ```Long.MAX_VALUE``` due to ```OperatorSingle``` not overriding the request number.


davidmoten:
I'm going to change tack on this because the problem is really with ```OperatorSingle``` in that it should override requests using ```setProducer```.  The methods in ```BlockingObservable``` that don't use ```OperatorSingle``` should be unaffected. I'll update the PR shortly.

PR #2859, opened 2015-04-07
OperatorDoOnRequest should unsubscribe from upstream

davidmoten:
When ```Observable.doOnRequest``` is used it does not pass unsubscription upstream. This PR fixes that and also adds a unit test of the base functionality for ```doOnRequest``` (no unit tests were present for this operator at all).


davidmoten:
Travis failure is unrelated:

```rx.internal.operators.OperatorRetryTest > testRetryWithBackpressureParallel```

akarnokd:
Thanks.

PR #2860, opened 2015-04-07
OperatorSingle should request exactly what it needs

davidmoten:
I noticed that ```observable.single().subscribe()``` requested ```Long.MAX_VALUE``` from ```observable``` which is more than it needs to come up with one item or an error. This PR sets a ```Producer``` for ```OperatorSingle``` so that it requests no more or less than it needs.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSingle.java
+++ b/src/main/java/rx/internal/operators/OperatorSingle.java
@@ -16,8 +16,10 @@
 package rx.internal.operators;
 
 import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import rx.Observable.Operator;
+import rx.Producer;
 import rx.Subscriber;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorSingle.java
```diff
@@ -44,53 +46,98 @@ private OperatorSingle(boolean hasDefaultValue, final T defaultValue) {
...
-                    isNonEmpty = true;
-                    // Issue: https://github.com/ReactiveX/RxJava/pull/1527
-                    // Because we cache a value and don't emit now, we need to request another one.
-                    request(1);
+            public void request(long n) {
+                if (n > 0) {
+                    // should request at most 2 in total to complete with no
+                    // items, emit the single item or decide that there are
+                    // too many items.
+                    long numToRequest = Math.min(2 - requested.get(), n);
```
akarnokd: Why not request(2) once in this request()? I.e.,
```java
public void request(long n) {
  if (n > 0 && once.compareAndSet(false, true)) {
    parent.requestMore(2);
  }
}
```

davidmoten:
yep, indeed. I got a bit caught up preserving the existing logic. I'll clean it up.

davidmoten:
there's a bit of a cleanup, have another look, thanks.

akarnokd:
Thanks.

PR #2861, opened 2015-04-07
Ensure OperatorZipTest.testZipRace does not fail on slower machines

davidmoten:
The test ```OperatorZipTest.testZipRace``` fails every time on my pretty fast desktop due to not achieving the desired number of iterations within the timeout. It also fails every time on my slow laptop. This PR adds a time limit for the iterations of one second less than the test timeout so that the test will fail if a zip hangs but not if the machine is slow.


akarnokd:
Thanks.

PR #2862, opened 2015-04-08
Cleanup warnings in test source

davidmoten:
Made changes to test source to avoid or suppress these warnings:

* raw types 
* unused imports
* unused variables


akarnokd:
Looks good. Thanks.

PR #2864, opened 2015-04-08
IndexedRingBuffer.getInstance can infer type

davidmoten:
Added type to IndexedRingBuffer.POOL (to get rid of raw types warning) and changed signature of ```IndexedRingBuffer.getInstance``` so that it infers type. Means that 13 ```unchecked``` warnings in the code base can be removed.


akarnokd:
Thanks!

PR #2867, opened 2015-04-09
More warnings cleanup

davidmoten:
* removed unused constructors that were the subject of varargs warnings
* removed unused import

I think that cleans up all java source warnings (not including javadoc issues).


akarnokd:
Thanks.

PR #2871, opened 2015-04-10
Use explicit versioning for gradle-rxjava-project-plugin

davidmoten:
When a release is created the tagged version should build deterministically at any time. The wildcard dependency on ```gradle-rxjava-project-plugin``` in ```build.gradle``` means this will not be the case. This PR sets the version at 2.2.3 (the latest on Maven Central).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
   repositories { jcenter() }
-  dependencies { classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:2.+' }
+  dependencies { classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:2.2.3' }
 }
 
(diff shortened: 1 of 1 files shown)


davidmoten:
I'm curious to the delay on this PR, seems straightforward. @JakeWarton I'm going to quote you here from your [gist ](https://gist.github.com/JakeWharton/2066f5e4f08fbaaa68fd):

>Prevent wildcard versions in your Gradle project. These undermine deterministic and hermetic builds and are generally considered bad practice.

@quidryan Thought I'd check with you here seeing as you have done a lot of gradle build work with RxJava.

benjchristensen:
> I'm curious to the delay on this PR

Just me.

> wildcard versions in your Gradle project

We avoid them in actual project dependencies (of which we have none), but for the Gradle project plugins itself we were okay with it as we controlled both of them and it simplified build fixes.

I'm okay either way. Have you run into issues with the plugin?

PR #2872, opened 2015-04-10
Use the correct accounts for sonatype sync

rspieldenner:


zsxwing:
@benjchristensen could you merge this one before releasing the next version? So that you can test the account.

benjchristensen:
I thought we already had this solved. This changes all 4 encrypted values, why? Because we don't know which one is which?

rspieldenner:
I didn't know which of the 4 was which so deleted all and recreated

PR #2878, opened 2015-04-14
flatMap: fixed scalar-merging.

akarnokd:
FlatMapping over a sequence of scalar observable values while observing them from a different thread did not work: items in the scalar queue where ignored by some terminal checks.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -131,8 +131,8 @@ public class OperatorMerge<T> implements Operator<T, Observable<? extends T>> {
         private volatile RxRingBuffer scalarValueQueue = null;
 
         /* protected by lock on MergeSubscriber instance */
-        private int missedEmitting = 0;
-        private boolean emitLock = false;
+        private boolean missedEmitting;
+        private boolean emitLock;
 
         /**
          * Using synchronized(this) for `emitLock` instead of ReentrantLock or AtomicInteger is faster when there is no contention.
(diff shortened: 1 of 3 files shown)


benjchristensen:
Is this a replacement for your significant rewrite of `merge`? Are you still trying to rewrite it or just incrementally adjust it as done here?

akarnokd:
No, this is still incremental. The full rewrite will only happen for 2.0.

benjchristensen:
Cool, I like that approach better. That context helps me review this. 

Don't spend too much time on the 2.0 rewrite quite yet though until we figure out some of the foundational abstractions and contracts.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -246,63 +246,58 @@ private void handleScalarSynchronousObservable(ScalarSynchronousObservable<? ext
...
+            boolean moreToDrain;
+            try {
+                actual.onNext(value);
+            } finally {
+                moreToDrain = releaseEmitLock();
+            }
+            if (moreToDrain) {
+                drainQueuesIfNeeded();
+            }
+            request(1);
```
benjchristensen: In a future rewrite we can optimize this because it could drain queues and request(n) and then immediately request(1) again.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -246,63 +246,58 @@ private void handleScalarSynchronousObservable(ScalarSynchronousObservable<? ext
...
         }
 
         private void handleScalarSynchronousObservableWithRequestLimits(ScalarSynchronousObservable<? extends T> t) {
+            
+            // if we didn't return above we need to enqueue
+            // enqueue the values for later delivery
```
benjchristensen: These comments are confusing here as what "return above" is it referring to? It's the start of a method.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -246,63 +246,58 @@ private void handleScalarSynchronousObservable(ScalarSynchronousObservable<? ext
...
+            try {
+                getOrCreateScalarValueQueue().onNext(t.get());
```
benjchristensen: This now means it will always pass through a queue which diminishes some of the optimization. This class tries hard to avoid putting values in a queue if it can otherwise emit directly.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: Oh, those comments were copy/pasted from the end of the method ... yes, definitely unclear.

benjchristensen:
I'm not ready to proceed with this PR. I've added some comments above. The key part that concerns me is that we are changing to always route through a queue for scalar values when request(n) < Long.MAX_VALUE is happening.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Not sure what caused it, but if I didn't do this enqueueing, despite the other changes the operator often didn't deliver all scalar values; not to mention, scalar values could get reordered for some reason.

akarnokd:
Apparently a full rewrite can't be avoided.

benjchristensen:
If you're going to pursue a full rewrite, make sure to pay attention not just to the JMH throughput numbers, but also the object allocation rates. 

I expect a rewrite of merge to take some non-trivial time to write, review and test considering the history and attempts made. It is a far more important operator performance-wise than most.

PR #2880, opened 2015-04-15
Use singleton reduction functions in count and countLong

davidmoten:
Every call to Observable.count() and countLong() instantiated a new Func2 which I have now made a singleton in the interests of minimizing GC pressure. 

Can anyone in the know confirm for me that there is GC benefit from this one because I assume a sophisticated enough compiler could replace the new Func2 call with a singleton during its optimizations.

If this one is worthwhile then there are heaps more of these in ```Observable.java``` and possibly elsewhere.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -3630,12 +3630,16 @@ public class Observable<T> {
      * @see #countLong()
      */
     public final Observable<Integer> count() {
-        return reduce(0, new Func2<Integer, T, Integer>() {
+        return reduce(0, CountHolder.INSTANCE);
+    }
+    
+    private static final class CountHolder {
+        static final Func2<Integer, Object, Integer> INSTANCE = new Func2<Integer, Object, Integer>() {
             @Override
-            public final Integer call(Integer t1, T t2) {
-                return t1 + 1;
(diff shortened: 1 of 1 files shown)


akarnokd:
Since we are Java 6+, we can't assume a bunch of optimizations happens so what's left is manual optimization/inlining.

akarnokd:
I'd instead do this directly with a long counting Subscriber and for int, I'd map it to min(count, Integer.MAX_VALUE) at the end saving the duplications.

davidmoten:
I guess this is an opportunity to decide what overflow behaviour we desire. I see three options:

1.  be consistent with Integer/Long overflow behaviour
2. return Integer.MAX_VALUE/Long.MAX_VALUE on overflow
3. throw an error on overflow

My preference is 3.

zsxwing:
I think `3` will be a breaking change so it's not acceptable. `1` is not a big deal since it's consistent with Java, which all Java users should be familiar with it.

zsxwing:
@davidmoten thoughts?

davidmoten:
I'm happy to maintain the existing overflow behaviour so it's not a
breaking change. Given that I think the PR could go in as is.

zsxwing:
Thanks.

PR #2881, opened 2015-04-15
VarArg parameters for the assertReceivedOnNext(T... items)

Unisay:
Syntactic simplification for the `assertReceivedOnNext(List<T> items)`.
Instead of 
```
testSubscriber.assertReceivedOnNext(Arrays.asList(1, 2, 3))
```
(almost all existing calls use Arrays.asList())
 proposed vararg method 
```
testSubscriber.assertReceivedOnNext(1, 2, 3)
```
 looks simplier and nicer.

### DIFF
--- a/src/main/java/rx/observers/TestObserver.java
+++ b/src/main/java/rx/observers/TestObserver.java
@@ -16,6 +16,7 @@
 package rx.observers;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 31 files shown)


akarnokd:
Hi. First of all, this is an incompatible change since the TestSubscriber is a public API and needs to maintain its method signatures. Second, varargs in Java 6 is problematic and we can't use the SafeVarargs annotation: the moment you try to assert an array of list of values, you get the unchecked warning.

Unisay:
@akarnokd, regarding the change incompatibility: I have only introduced new method with varargs, the method with list remains available, so should not break anything.
As for your second point: I see the problem in Java 6. Is compiler warning a blocker?

akarnokd:
You're right, I've misread the diff page. The warning isn't a real blocker but we'd like to avoid much of it as possible.

As a matter of fact, I happen to have a [PR](https://github.com/ReactiveX/RxJava/pull/2806/files) in which the TestSubscriber is also expanded with lots of useful methods.

Regardless, I wouldn't have changed tests to use the new method since we usually don't look at test unless there is a problem with them. In that case or with new tests, the developers are free to use the more convenient methods.

benjchristensen:
@akarnokd I'm trying to figure out if this will cause any of the issues that made us move away from varargs in the past. You seem to have concluded that it won't?

Also I need to make sure the overload won't conflict with dynamic languages like Groovy and Clojure. 

@Unisay Can you add the method without changing all the unit tests so this doesn't affect 30 files that don't need to be changed? Please rebase the change so it is a single commit just touching TestSubscriber.

akarnokd:
I'm not sure about dynamic languages but I'd prefer a shorter ```assertValues(Object... values)```.

benjchristensen:
Handling in https://github.com/ReactiveX/RxJava/pull/2948 without changing the other 30 files.

PR #2882, opened 2015-04-16
OperatorScan - don't call onNext after onError is called

davidmoten:
title says it all

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorScan.java
+++ b/src/main/java/rx/internal/operators/OperatorScan.java
@@ -20,6 +20,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import rx.Observable.Operator;
 import rx.Producer;
 import rx.Subscriber;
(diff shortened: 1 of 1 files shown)


src/main/java/rx/internal/operators/OperatorScan.java
```diff
@@ -104,6 +104,7 @@ public void onNext(T currentValue) {
                         this.value = accumulator.call(this.value, currentValue);
                     } catch (Throwable e) {
                         child.onError(OnErrorThrowable.addValueAsLastCause(e, currentValue));
+                        return;
```
zsxwing: Good catch. Could you also add `Exceptions.throwIfFatal(e);` above `child.onError` like `OperatorMap`?

davidmoten:
Added ```Exceptions.throwIfFatal``` as suggested by @zsxwing

akarnokd:
Thanks.

benjchristensen:
:+1:

Issue #2458, opened 2015-01-15
Lifting into a Subject


abersnaze:
Mildly interesting idea to see if there would be any interest.

```
public class Subject<T, R> extends Observable<R> implements Observer<T> {
    ...
    public <X> Subject<X, R> liftHead(final Operator<? extends T, ? super X> op) { ... }
    public <X> Subject<T, X> liftTail(final Operator<? extends X, ? super R> op) { ... }

    @Override
    public <X> Subject<T, X> lift(final Operator<? extends X, ? super R> op) { return liftTail(op); }
}
```

akarnokd:
What do you want to achieve with this change?

abersnaze:
Up till now all the APIs that want to delegate part of the observable change chain to another piece of code has had to pass in the head as an argument.  With these methods the delegate could return a subject the represents the some segment of the chain without any arguments.

For example `retryWhen()` takes a notification handler `Func1<Observable<Exception>, Observable<?>>`.  With this it would have been possible for `retryWhen` to take a notification handler `Func0<Subject<Exception, ?>`

akarnokd:
If I understand correctly, a chain of Subject<Integer, String> -> Subject<String, Date> -> Subject<Date, Long>, I can hand out each to different parties and observe them at any point?

abersnaze:
Yes.  The `Subject` returned could be reused to `subscribe()` different source observables.  The `Subject` would just be an capsulation of some computation without defining where the data comes from or where it is going to.

benjchristensen:
The encapsulation of computation sounds like the Transformer with Observable.compose.

akarnokd:
Do you still want to pursue this?

benjchristensen:
Punting to 2.0 for bikeshedding.

akarnokd:
See #3975.

akarnokd:
I'm closing this as the PR implementing the proposed feature was not really endorsed by the OP after all. If you have further input on the issue, don't hesitate to reopen this issue or post a new one.

PR #2883, opened 2015-04-16
Proposal: standardized Subject state-peeking methods.

akarnokd:
This PR modifies the ```Subject``` class to host the union of the state-peeking methods of the various ```Subject``` implementations and fixes the inconsistent behavior of ```ReplaySubject.getValues(T[])```.

The changes enable the ```SerializedSubject``` to relay such state-peeking method calls into the wrapped subject and allows future ```Subject``` wrappers to do the same (#2458).

### DIFF
--- a/src/main/java/rx/subjects/AsyncSubject.java
+++ b/src/main/java/rx/subjects/AsyncSubject.java
@@ -15,6 +15,7 @@
  */
 package rx.subjects;
 
+import java.lang.reflect.Array;
 import java.util.*;
 
 import rx.Observer;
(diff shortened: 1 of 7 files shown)


benjchristensen:
Seems to make sense. The additions to `Subject` all should be non-breaking.

benjchristensen:
Due to the Experimental marking of these I'm being a little less stringent in discussing the addition to public APIs, but it seems to me that the precedent set on the BehaviorSubject shows the value of these methods. Additionally, Subjects are the stateful, side-effecting types and these methods seem to make sense for them.

PR #2886, opened 2015-04-17
OnSubscribeRange Producer field should be volatile

davidmoten:
The ```RangeProducer.request``` method may be called from different threads so the mutable field ```index``` that is read and written by this method should be volatile.


akarnokd:
No, because it is encapsulated between the ```BackpressureUtils.getAndAddRequest()``` and ```REQUESTED_UPDATER.addAndGet``` which act as full barriers. Since the emission is only entered by one thread at a time, the above two points create a happens-before relationship and the ```index``` variable is properly published; you don't even need to worry about word tearing on 32bit.

davidmoten:
I see, good info thanks.

PR #2892, opened 2015-04-19
Fix Observable.range race condition

davidmoten:
In `RangeProducer` if the first two requests come concurrently and are both for `Long.MAX_VALUE` then there is a possible race condition where the fast path is started twice thus emitting some or all elements twice.

This PR fixes the race by only allowing the fast path if the current request count is 0 (using ```compareAndSet```).


davidmoten:
If this is right I'll submit another PR for `OnSubscribeFromIterable` which has the same problem.

akarnokd:
Looks good. Thanks.

PR #2895, opened 2015-04-20
Fix Observable.from(Iterable) race condition

davidmoten:
In `IterableProducer` if the first two requests come concurrently and are both for `Long.MAX_VALUE` then there is a possible race condition where the fast path is started twice thus emitting some or all elements twice.

This PR fixes the race by only allowing the fast path if the current request count is 0 (using ```compareAndSet```).


akarnokd:
Thanks!

PR #2897, opened 2015-04-20
Fix for overlapping windows. 

### LINKED ISSUE #2869: Non-deterministic test: OnSubscribeCombineLatestTest.testWithCombineLatestIssue1717
I put ```./gradlew clean test``` into a loop and got a faiure on second run in ```OnSubscribeCombineLatestTest.testWithCombineLatestIssue1717```:

```
java.lang.AssertionError: expected:<2000> but was:<2002>
	at org.junit.Assert.assertEquals(Assert.java:472)
	at org.junit.Assert.assertEquals(Assert.java:456)
	at rx.internal.operators.OnSubscribeCombineLatestTest.testWithCombineLatestIssue1717(OnSubscribeCombineLatestTest.java:852)
```

### LINKED ISSUE #2896: window() emitting overlapping items
I've seen this in 1.07 and 1.09 - using http://reactivex.io/RxJava/javadoc/rx/Observable.html#window(rx.functions.Func0), I am able to observe overlapping items in the output. 

Example test case here:  https://gist.github.com/gjesse/1ae59c32f63d2ad581e8

alexwen:
Source was emitting t multiple times while holding queue.

Fixes #2896

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorWindowWithObservable.java
+++ b/src/main/java/rx/internal/operators/OperatorWindowWithObservable.java
@@ -119,7 +119,7 @@ public final class OperatorWindowWithObservable<T, U> implements Operator<Observ
                 do {
                     drain(localQueue);
                     if (once) {
-                        once = true;
+                        once = false;
                         emitValue(t);
                     }
                     
--- a/src/test/java/rx/internal/operators/OperatorWindowWithObservableTest.java
+++ b/src/test/java/rx/internal/operators/OperatorWindowWithObservableTest.java
@@ -22,6 +22,7 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import org.junit.Test;
@@ -29,6 +30,8 @@ import org.junit.Test;
 import rx.Observable;
 import rx.Observer;
 import rx.exceptions.TestException;
+import rx.functions.Func0;
+import rx.observers.TestSubscriber;
 import rx.subjects.PublishSubject;
 
 public class OperatorWindowWithObservableTest {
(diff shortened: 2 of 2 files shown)


davidmoten:
Could you add the unit test from your gist as well?

davidmoten:
Ah, someone else's gist. Still, as the unit test is there may as well chuck it in.

alexwen:
Thanks for the feedback. Is it kosher to do a sleep in the test? Its the
only way I've been able to get the windows to overlap and reproduce the
condition.

Also, it doesn't seem too reliable to depend upon sleep.

akarnokd:
Good catch! An unit test would be great, but you don't really need concurrency: reentrancy will do just fine:
```java
    @Test
    public void testWindowNoDuplication() {
        final PublishSubject<Integer> source = PublishSubject.create();
        final TestSubscriber<Integer> tsw = new TestSubscriber<Integer>() {
            boolean once;
            @Override
                if (!once) {
                    once = true;
                    source.onNext(2);
                }
                super.onNext(t);
            }
        };
        TestSubscriber<Observable<Integer>> ts = new TestSubscriber<Observable<Integer>>() {
            @Override
            public void onNext(Observable<Integer> t) {
                t.subscribe(tsw);
                super.onNext(t);
            }
        };
        source.window(new Func0<Observable<Object>>() {
            @Override
            public Observable<Object> call() {
                return Observable.never();
            }
        }).subscribe(ts);
        
        source.onNext(1);
        source.onCompleted();
        
        assertEquals(1, ts.getOnNextEvents().size());
        assertEquals(Arrays.asList(1, 2), tsw.getOnNextEvents());
    }
```

alexwen:
Thanks @akarnokd - great test!

akarnokd:
Thanks!

benjchristensen:
Should I push a patch release with this today?

akarnokd:
A verification on #2878 would be great. After that, the patch release can commence.

Issue #2884, opened 2015-04-17
Backpressure and onComplete from an empty observable


davidmoten:
What is the *preferred* behaviour for an observable that supports backpressure but emits nothing when it has been subscribed to with an initial request of 0? Should it report completion as soon as it knows it is empty or wait for a positive request? 
 
I notice that reactive-streams-jvm 2.9 says

>A Subscriber MUST be prepared to receive an onComplete signal with or without a preceding Subscription.request(long n) call.

This allows us to go either way when developing an ```Operator``` for instance but perhaps RxJava wants to proffer a preferred or mandated direction for it's internal operators? @benjchristensen 

FYI @viktorklang

akarnokd:
```from``` and ```range``` already emit ```onCompleted``` if they are essentially empty or just after the last item so I think that should stay as the behavior.

davidmoten:
I don't thinks that's the case for ```from```. Its ```Producer``` does nothing at all if request is 0.

davidmoten:
Same for ```range```.

akarnokd:
@davidmoten you're right, I just checked the code.

viktorklang:
Hi @davidmoten,

since `request` is for `elements` and `complete` and `error` are not elements,
it is intended for these signals to be propagated as soon as they are known.

Let me know if you have any follow-on questions!

davidmoten:
Thanks @viktorklang. Do you think that intention should be a SHOULD addition to the Publisher rules in reactive-streams-jvm?

viktorklang:
Given that 2.9 already provisions for this I think a good compromise (I am not a fan of SHOULDs in general) to include this explanation as a part of https://github.com/reactive-streams/reactive-streams-jvm/issues/177

Would that cover it?

davidmoten:
Sounds fine, thanks.

viktorklang:
@davidmoten Great. I'd like to take the opportunity to encourage you to assist with that Issue, if you have any cycles to spare. :)

davidmoten:
@akarnokd as you suspected originally `range` does emit `onCompleted` eagerly from an empty range. `from` does not emit `onCompleted` eagerly from an empty iterable so I submitted #2899.

akarnokd:
Seems to be fixed. Closing.

PR #2898, opened 2015-04-20
Observable.range - add unit test for eager completion on empty

davidmoten:
As per discussion in #2884, it turns out that `range` does emit `onComplete` even when zero is requested because `Observable.range` does not use `OperatorRange` when the count is zero but rather returns an `Observable.empty()`. This PR adds a unit test confirming this fact.


akarnokd:
Looks good. Thanks!

PR #2899, opened 2015-04-21
Observable.from(empty) to emit onComplete even when 0 requested

davidmoten:
As per discussion in #2884, `Observable.from(iterable)` is preferred to emit `onCompleted` when `iterable` is empty even when nothing has been requested.

This PR adds a check on `iterator.hasNext` before the `Producer` is assigned and emits `onCompleted` immediately if `hasNext` returns false.

Includes unit test that failed on existing code base.


davidmoten:
ah erk I need to rebase

davidmoten:
rebased now

akarnokd:
Thanks!

PR #2901, opened 2015-04-21
Operators toList and toSortedList now support backpressure

akarnokd:
Added support for backpressure in ```toList``` and ```toSortedList```: they emit their buffered content only when downstream actually requested it.

Few other notes:
  - Added overload to ```toSortedList``` taking a ```initialCapacity``` argument which should help reduce the number of times the buffer needs to be resized.
  - The Func2 parameter of ```toSortedList``` is now wrapped once per operator instead of once per subscriber.
  - Using non-final list buffers which are set to null on completion to not hold onto the buffer.
  - ```toSortedList```now returns a modifiable ```ArrayList```: since it is a handoff procedure, there is no value in forcing a non-modification constraint on the emitted list (similar to ```toList``` even before this PR).

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorToObservableList.java
+++ b/src/main/java/rx/internal/operators/OperatorToObservableList.java
@@ -52,10 +52,11 @@ public final class OperatorToObservableList<T> implements Operator<List<T>, T> {
     private OperatorToObservableList() { }
     @Override
     public Subscriber<? super T> call(final Subscriber<? super List<T>> o) {
-        return new Subscriber<T>(o) {
+        final SingleDelayedProducer<List<T>> producer = new SingleDelayedProducer<List<T>>(o);
+        Subscriber<T> result =  new Subscriber<T>() {
 
-            private boolean completed = false;
-            final List<T> list = new LinkedList<T>();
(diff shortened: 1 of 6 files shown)


davidmoten:
Cool. I'll be using that `initialCapacity` and `SingleDelayedProducer` will be useful too.

src/main/java/rx/internal/operators/OperatorToObservableList.java
```diff
@@ -101,6 +107,9 @@ public void onNext(T value) {
             }
 
         };
+        o.add(result);
```
benjchristensen: Why can't `o` just be passed into the `Subscriber` constructor?

benjchristensen:
Good changes.

src/main/java/rx/internal/operators/OperatorToObservableList.java
akarnokd: Because if the result ```Subscriber``` receives its ```Producer``` via ```setProducer```, the default behavior is to call ```setProducer``` on ```op``` set by the constructor and thus can overwrite our ```SingleDelayedProducer```.

PR #2902, opened 2015-04-21
Prevent ExceptionsTest from hanging when testing stack overflow

davidmoten:
On my i5 laptop `ExceptionsTest.testStackOverflowIsThrow` hangs. This PR changes the test so that it confirms that significant stack depth occurs and in another test ensures that the fatal `StackOverflowError` is thrown by `subscribe` when it occurs in `onNext`.

I also removed some unused code from the original test and some extra subscriptions that didn't seem to serve a particular function.


akarnokd:
Looks good. Thanks!

PR #2903, opened 2015-04-21
OperatorPublish - use BackpressureUtils

davidmoten:
Logic is fine, can replace a block of code with a call to `BackpressureUtils.getAndAddRequest`.


akarnokd:
Looks good, but I'd be more glad you could review 2814 instead.

davidmoten:
Ah right, I'll chuck this in preference to your rewrite.

PR #2904, opened 2015-04-22
TakeLast - add request overflow check

davidmoten:
Added request overflow check to `TakeLastQueueProducer` and associated unit test (that failed on existing code base).


akarnokd:
Excellent!

PR #2905, opened 2015-04-22
Use singleton Operators where we can

davidmoten:
To reduce GC pressure use singleton `Operator`s for `Observable.single`, `distinct`, `distinctUntilChanged`,  `onBackpressureBuffer`,  `isEmpty`.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorDistinct.java
+++ b/src/main/java/rx/internal/operators/OperatorDistinct.java
@@ -17,9 +17,11 @@ package rx.internal.operators;
 
 import java.util.HashSet;
 import java.util.Set;
+
 import rx.Observable.Operator;
 import rx.Subscriber;
 import rx.functions.Func1;
+import rx.internal.util.UtilityFunctions;
 
 /**
  * Returns an Observable that emits all distinct items emitted by the source.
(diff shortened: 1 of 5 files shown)


akarnokd:
Could you please do ```never()``` as well?

src/main/java/rx/internal/operators/OperatorDistinct.java
```diff
@@ -29,6 +31,15 @@
  */
 public final class OperatorDistinct<T, U> implements Operator<T, T> {
     final Func1<? super T, ? extends U> keySelector;
+    
+    private static class Holder {
+        static final OperatorDistinct<?,?> INSTANCE = new OperatorDistinct<Object,Object>(UtilityFunctions.<Object>identity());
+    }
+    
+    @SuppressWarnings("unchecked")
```
akarnokd: I'd add a javadoc comment here so it is more apparent the returned instance uses the identity function for comparison.

src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
```diff
@@ -26,6 +27,15 @@
  */
 public final class OperatorDistinctUntilChanged<T, U> implements Operator<T, T> {
     final Func1<? super T, ? extends U> keySelector;
+    
+    private static class Holder {
+        static final OperatorDistinctUntilChanged<?,?> INSTANCE = new OperatorDistinctUntilChanged<Object,Object>(UtilityFunctions.identity());
+    }
+    
+    @SuppressWarnings("unchecked")
```
akarnokd: Same javadoc here as above.

src/main/java/rx/internal/operators/OperatorSingle.java
```diff
@@ -32,7 +32,16 @@
     private final boolean hasDefaultValue;
     private final T defaultValue;
 
-    public OperatorSingle() {
+    private static class Holder {
+        final static OperatorSingle<?> INSTANCE = new OperatorSingle<Object>();
+    }
+    
+    @SuppressWarnings("unchecked")
+    public static <T> OperatorSingle<T> create() {
```
akarnokd: Javadoc would be great here about what defaults the operator will have.

davidmoten:
Updated javadoc and renamed `create` methods to `instance`, added singleton for `never` as well.

davidmoten:
unrelated CI failure. I'll open an issue for it.

akarnokd:
Looks good!

Issue #2908, opened 2015-04-23
Non-deterministic test: RxRingBufferWithoutUnsafeTest.testConcurrency


davidmoten:
Concurrency failure on RxRingBuffer doesn't look good. Hopefully a problem with the test only!
```
rx.internal.util.RxRingBufferWithoutUnsafeTest > testConcurrency FAILED
    java.lang.AssertionError: expected:<6920> but was:<6808>
        at org.junit.Assert.fail(Assert.java:93)
        at org.junit.Assert.failNotEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:128)
        at org.junit.Assert.assertEquals(Assert.java:472)
        at org.junit.Assert.assertEquals(Assert.java:456)
        at rx.internal.util.RxRingBufferWithoutUnsafeTest.testConcurrency(RxRingBufferWithoutUnsafeTest.java:139)
```

akarnokd:
Found the bug, see PR #2909.

akarnokd:
Closed via #2909

PR #2909, opened 2015-04-23
Fix the drainer to check if the queue is empty before quitting.

akarnokd:
Reported in #2908. The bug is on line [113](https://github.com/ReactiveX/RxJava/blob/5ac4448873c09cd806ae73a37727a299cbaee2ba/src/test/java/rx/internal/util/RxRingBufferWithoutUnsafeTest.java#L113). If the drainer is delayed after a poll returning null and the emitter producing & terminating before the getCount(), the queue is non empty but the drainer quits and hence the value difference. It comes up almost always if you put a sleep before L113.


davidmoten:
Pleased to hear just a problem with the test. Thanks for quick resolution @akarnokd

PR #2911, opened 2015-04-23
OperatorPublish benchmark

akarnokd:
Benchmarks the behavior of ```publish()``` in face of number of child subscribers, batching effect in request(), length of the source and whether or not each child observes values on a separate thread.

 [Benchmark results](https://gist.github.com/akarnokd/cccc31312125d8636d3d)

Remarks:
- With no subscribers, the operator seems to fill its queue and then do nothing unless someone subscribes. There is no test to check for this behavior so it might be a bug or a feature.
- In async mode, the benchmark threw ```MissingBackpressureException``` a lot of times which can happen with a single subscriber as well, so it might be some concurrency bug in the operator.

### DIFF
--- a/src/perf/java/rx/operators/OperatorPublishPerf.java
+++ b/src/perf/java/rx/operators/OperatorPublishPerf.java
@@ -1,0 +1,159 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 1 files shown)


benjchristensen:
> The operator seems to fill its queue and then do nothing unless someone subscribes

I remember this being talked about at some point and an agreement that this is wrong behavior. If there are no subscribers it should just ignore/drop the data.

benjchristensen:
Agreed that throwing a `MissingBackpressureException` a lot sounds like a problem.

akarnokd:
There shouldn't be any MBE because every component supports backpressure. There is a request accounting problem somewhere but this is now irrelevant. The new OperatorPublish works without throwing MBE.

Issue #2857, opened 2015-04-06
Performance issues after upgrading from 1.0.4 -> 1.0.8


neerajrj:
Hello,

We are seeing high levels of lock contention after our upgrade to RxJava 108

Apologize for not having a unit test to reproduce this we have a fairly complex system and we are having trouble figuring out which areas to dig deeper to find a reproducible case.

This is a paste from a JMC view. As far as we know nothing should be getting unsubscribed in our application. 
We would appreciate if anyone can shed some light on what kind of behavior would trigger the stack below.
```
Stack Trace	Sample Count	Percentage(%)
java.util.concurrent.locks.LockSupport.unpark(Thread)	1,113	76.233
 java.util.concurrent.locks.AbstractQueuedSynchronizer.unparkSuccessor(AbstractQueuedSynchronizer$Node)	1,113	76.233
      java.util.concurrent.locks.AbstractQueuedSynchronizer.release(int)	1,113	76.233
         java.util.concurrent.locks.ReentrantLock.unlock()	1,113	76.233
            rx.internal.util.SubscriptionList.remove(Subscription)	1,113	76.233
               rx.internal.schedulers.ScheduledAction$Remover2.unsubscribe()	1,113	76.233
                  rx.internal.util.SubscriptionList.unsubscribeFromAll(Collection)	1,113	76.233
                     rx.internal.util.SubscriptionList.unsubscribe()	1,113	76.233
                        rx.internal.schedulers.ScheduledAction.unsubscribe()	1,113	76.233
                           rx.internal.schedulers.ScheduledAction.run()	1,113	76.233
                              java.util.concurrent.Executors$RunnableAdapter.call()	1,113	76.233
                                 java.util.concurrent.FutureTask.run()	1,113	76.233
                                    java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor$ScheduledFutureTask)	1,113	76.233
                                       java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run()	1,113	76.233
                                          java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor$Worker)	1,113	76.233
                                             java.util.concurrent.ThreadPoolExecutor$Worker.run()	1,113	76.233
                                                java.lang.Thread.run()	1,113	76.233

```

akarnokd:
[...]

PR #2912, opened 2015-04-23
Fix the performance degradation due to different schedule execution and

akarnokd:
SubscriptionList.add() and thread unparking.

This PR partially reverts some changes from earlier scheduler optimizations and fixes a case where if multiple concurrent schedule() calls happen, the order in the SubscriptionList might be different from the actual execution order which degrades performance on task termination due to remove() being O(n).

This might be the source of degradation in #2857 as well.

I'll post the ```ComputationSchedulerPerf``` results later.

### DIFF
--- a/src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
+++ b/src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
@@ -117,10 +117,7 @@ public class EventLoopsScheduler extends Scheduler {
             if (isUnsubscribed()) {
                 return Subscriptions.unsubscribed();
             }
-            ScheduledAction s = poolWorker.scheduleActual(action, 0, null);
-            
-            serial.add(s);
-            s.addParent(serial);
+            ScheduledAction s = poolWorker.scheduleActual(action, 0, null, serial);
             
             return s;
         }
--- a/src/main/java/rx/internal/util/SubscriptionList.java
+++ b/src/main/java/rx/internal/util/SubscriptionList.java
@@ -15,12 +15,7 @@
  */
 package rx.internal.util;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.*;
 
 import rx.Subscription;
 import rx.exceptions.Exceptions;
@@ -34,7 +29,6 @@ public final class SubscriptionList implements Subscription {
 
     private LinkedList<Subscription> subscriptions;
     private volatile boolean unsubscribed;
-    private final ReentrantLock lock = new ReentrantLock();
 
     public SubscriptionList() {
     }
@@ -66,8 +60,7 @@ public final class SubscriptionList implements Subscription {
             return;
         }
         if (!unsubscribed) {
-            lock.lock();
-            try {
+            synchronized (this) {
                 if (!unsubscribed) {
                     LinkedList<Subscription> subs = subscriptions;
                     if (subs == null) {
@@ -77,8 +70,6 @@ public final class SubscriptionList implements Subscription {
                     subs.add(s);
                     return;
                 }
-            } finally {
-                lock.unlock();
             }
         }
         // call after leaving the synchronized block so we're not holding a lock while executing this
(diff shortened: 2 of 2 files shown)


akarnokd:
```
Benchmark      (size)        2912   2912 error       1.0.9  1.0.9 error
observeOn           1  198785,487    59066,518  203658,175     2162,208
observeOn          10  153990,002    79263,702  178839,814    17425,416
observeOn         100  150333,481    20749,482  148706,965    15169,371
observeOn        1000   25568,884    24718,878   16911,025     7937,752
observeOn        2000   14914,879    17179,009   10022,748    12460,449
observeOn        3000   11238,118     2400,207   10359,226     2868,300
observeOn        4000    8172,019     1370,653    7084,573     1460,569
observeOn       10000    3491,021      606,603    3412,501      527,199
observeOn      100000     398,684       20,607     400,360       14,743
observeOn     1000000      41,865        1,884      43,887        0,746
subscribeOn         1  209443,261     8593,857  169965,326   104264,868
subscribeOn        10  175152,648    79449,091  151055,057   152837,825
subscribeOn       100  150042,891    84189,199  140177,447   128390,130
subscribeOn      1000   72804,521    29222,089   57929,258    19364,337
subscribeOn      2000   61834,990    14148,419   36040,899     8608,854
subscribeOn      3000   44148,878     4653,713   44982,855     3551,646
subscribeOn      4000   35494,660     3338,209   30343,790    24747,587
subscribeOn     10000   14890,534     6229,533   13920,062     1897,708
subscribeOn    100000    1691,786      230,978    1661,187      141,518
subscribeOn   1000000     193,028       22,564     191,835       16,975
```

There doesn't seem to be any performance degradation, however, small sized measurements are quite hectic for some reason.

src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
```diff
@@ -117,10 +117,7 @@ public Subscription schedule(Action0 action) {
             if (isUnsubscribed()) {
                 return Subscriptions.unsubscribed();
             }
-            ScheduledAction s = poolWorker.scheduleActual(action, 0, null);
-            
-            serial.add(s);
-            s.addParent(serial);
+            ScheduledAction s = poolWorker.scheduleActual(action, 0, null, serial);
```
benjchristensen: If I understand correctly this is just removing work that gets done for us within `scheduleActual`. Am I reading this correctly? Referenced code is here: https://github.com/ReactiveX/RxJava/blob/ecbd27da3492439cf75ea49d43ba473900caea51/src/main/java/rx/internal/schedulers/NewThreadWorker.java#L188

src/main/java/rx/internal/util/SubscriptionList.java
```diff
@@ -34,7 +29,6 @@
 
     private LinkedList<Subscription> subscriptions;
     private volatile boolean unsubscribed;
-    private final ReentrantLock lock = new ReentrantLock();
```
benjchristensen: So is the performance difference related to how the JVM treats `synchronized` vs a `ReentrantLock`?

akarnokd: I've benchmarked this via OperatorPublishPerf and using ```synchronized``` instead of lock gave about 25% more throughput with 1M elements.

src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
akarnokd: The problem here was that ```serial.add()``` happened after the action was scheduled and executed thus if there was some recursive work, those ```serial.add()``` calls could happen before this initial call and thus the order in the list was reversed. Since the removal is O(n) and we expected the list to be in the same order as the executions finish, any disorder increases the remove time and a concurrent lock attempt run out of its spin time and ended up parking. The performance degradation was this increased park/unpark activity. The ```scheduleAction``` overload just adds the action to the ```serial``` before it schedules it thus making sure any recursive schedule happens after it.

src/main/java/rx/internal/util/SubscriptionList.java
benjchristensen: Interesting. The JVM is a complicated piece of black magic.

src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
benjchristensen: Wow, that's an interesting issue. Thanks for the explanation!

PR #2914, opened 2015-04-24
Optimization - use OperatorTakeLastOne for takeLast(1)

davidmoten:
This is an optimization for `takeLast` when called with parameter 1. Using `OperatorTakeLast` carries unnecessary overhead for the `takeLast(1)` case and a decent throughput improvement (x2) for streams of 100 elements or more is seen in the benchmarks below.

`takeLast(1)` is used by the following operators which will also demonstrate a throughput improvement:

* `last`, `lastOrDefault`
* `reduce`
* `collect`
* `count`, `countLong`

Benchmarks comparing using the new `OperatorTakeLastOne` and `OperatorTakeLast`:

```
Benchmark                                                        Mode   Samples        Score  Score error    Units
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLastOne_Few     thrpt         5  2235516.141   129091.019    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLastOne_Many    thrpt         5      103.980        9.233    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLastOne_Some    thrpt         5   984689.481    48560.897    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLast_Few        thrpt         5  2187421.223    93550.379    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLast_Many       thrpt         5       54.575        2.054    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLast_Some       thrpt         5   466892.497     9267.405    ops/s

```

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -7771,7 +7771,12 @@ public class Observable<T> {
      * @see <a href="http://reactivex.io/documentation/operators/takelast.html">ReactiveX operators documentation: TakeLast</a>
      */
     public final Observable<T> takeLast(final int count) {
-        return lift(new OperatorTakeLast<T>(count));
+        if (count == 0)
+            return ignoreElements();
+        else if (count == 1 )
+            return lift(OperatorTakeLastOne.<T>instance());
+        else 
+            return lift(new OperatorTakeLast<T>(count));
     }
 
     /**
--- a/src/main/java/rx/internal/operators/OperatorTakeLastOne.java
+++ b/src/main/java/rx/internal/operators/OperatorTakeLastOne.java
@@ -1,0 +1,173 @@
+package rx.internal.operators;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import rx.Observable.Operator;
+import rx.Producer;
+import rx.Subscriber;
+
+public class OperatorTakeLastOne<T> implements Operator<T, T> {
+
+    private static class Holder {
+        static final OperatorTakeLastOne<Object> INSTANCE = new OperatorTakeLastOne<Object>();
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> OperatorTakeLastOne<T> instance() {
+        return (OperatorTakeLastOne<T>) Holder.INSTANCE;
+    }
+
+    private OperatorTakeLastOne() {
+
+    }
+
+    @Override
+    public Subscriber<? super T> call(Subscriber<? super T> child) {
+        final ParentSubscriber<T> parent = new ParentSubscriber<T>(child);
+        child.setProducer(new Producer() {
+
+            @Override
+            public void request(long n) {
+                parent.requestMore(n);
+            }
+        });
+        child.add(parent);
+        return parent;
+    }
+
+    private static class ParentSubscriber<T> extends Subscriber<T> {
+
+        private final static int NOT_REQUESTED_NOT_COMPLETED = 0;
+        private final static int NOT_REQUESTED_COMPLETED = 1;
+        private final static int REQUESTED_NOT_COMPLETED = 2;
+        private final static int REQUESTED_COMPLETED = 3;
+
+        /*
+         * These are the expected state transitions:
+         * 
+         * NOT_REQUESTED_NOT_COMPLETED   -->   REQUESTED_NOT_COMPLETED 
+         *            |                             | 
+         *            V                             V
+         * NOT_REQUESTED_COMPLETED       -->   REQUESTED_COMPLETED
+         * 
+         * Once at REQUESTED_COMPLETED we emit the last value if one exists
+         */
+
+        // Used as the initial value of last
+        private static final Object ABSENT = new Object();
+
+        // the downstream subscriber
+        private final Subscriber<? super T> child;
+
+        @SuppressWarnings("unchecked")
+        // we can get away with this cast at runtime because of type erasure
+        private T last = (T) ABSENT;
+
+        // holds the current state of the stream so that we can make atomic
+        // updates to it
+        private final AtomicInteger state = new AtomicInteger(NOT_REQUESTED_NOT_COMPLETED);
+
+        ParentSubscriber(Subscriber<? super T> child) {
+            this.child = child;
+        }
+        
+        void requestMore(long n) {
+            if (n > 0) {
+                // CAS loop to atomically change state given that onCompleted()
+                // or another requestMore() may be acting concurrently
+                while (true) {
+                    // read the value of state and then try state transitions
+                    // only if the value of state does not change in the
+                    // meantime (in another requestMore() or onCompleted()). If
+                    // the value has changed and we expect to do a transition
+                    // still then we loop and try again.
+                    final int s = state.get();
+                    if (s == NOT_REQUESTED_NOT_COMPLETED) {
+                        if (state.compareAndSet(NOT_REQUESTED_NOT_COMPLETED,
+                                REQUESTED_NOT_COMPLETED)) {
+                            return;
+                        }
+                    } else if (s == NOT_REQUESTED_COMPLETED) {
+                        if (state.compareAndSet(NOT_REQUESTED_COMPLETED, REQUESTED_COMPLETED)) {
+                            emit();
+                            return;
+                        }
+                    } else
+                        // already requested so we exit
+                        return;
+                }
+            }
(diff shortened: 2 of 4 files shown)


src/main/java/rx/internal/operators/OperatorTakeLastOne.java
```diff
@@ -0,0 +1,162 @@
...
+        // holds the current state of the stream so that we can make atomic updates to it
+        private final AtomicInteger state = new AtomicInteger(
+                NOT_REQUESTED_NOT_COMPLETED);
+
+        ParentSubscriber(Subscriber<? super T> child) {
+            this.child = child;
+        }
+
+        void requestMore(long n) {
+            if (n > 0) {
```
akarnokd: Although not yet available, you could use my SingleDelayedProducer.

src/main/java/rx/internal/operators/OperatorTakeLastOne.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public Subscriber<? super T> call(Subscriber<? super T> child) {
+        final ParentSubscriber<T> parent = new ParentSubscriber<T>(child);
+        child.setProducer(new Producer() {
+
+            @Override
+            public void request(long n) {
+                parent.requestMore(n);
```
akarnokd: I think parent may request Long.MAX_VALUE because it will only hold onto the very last.

src/main/java/rx/internal/operators/OperatorTakeLastOne.java
```diff
@@ -0,0 +1,162 @@
...
+                last = null;
+                return;
+            }
+            // synchronize to ensure that last is safely published
+            synchronized (this) {
+                if (last != ABSENT) {
+                    T t = last;
+                    // release for gc 
+                    last = null;
+                    child.onNext(t);
```
akarnokd: Never call onNext while holding a lock.

src/main/java/rx/internal/operators/OperatorTakeLastOne.java
```diff
@@ -0,0 +1,162 @@
...
+        /**
+         * If not unsubscribed then emits last value and completed to the child subscriber.
+         */
+        private void emit() {
+            if (isUnsubscribed()) {
+                // release for gc
+                last = null;
+                return;
+            }
+            // synchronize to ensure that last is safely published
```
akarnokd: Why would it need safe publishing, emit() is already called after full-barrier CAS? onNext is sequential and setting it to null plainly will become visible to GC anyway.

src/perf/java/rx/operators/OperatorTakeLastOnePerf.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.openjdk.jmh.annotations.Benchmark;
+
+import rx.Observable;
+import rx.internal.operators.OperatorTakeLast;
+import rx.internal.operators.OperatorTakeLastOne;
+
+public class OperatorTakeLastOnePerf {
+
+    private static final int MANY = 1000000;
```
akarnokd: You could use ```@Param({"5", "100", "1000000"})``` instead

src/main/java/rx/internal/operators/OperatorTakeLastOne.java
akarnokd: Since you are calling ```child.onNext``` from ```onCompleted```, you need to bounce back any thrown exceptions because throwing ```onCompleted``` is can't be routed back to ```onError``` and child will not receive the exception.

src/main/java/rx/internal/operators/OperatorTakeLastOne.java
```diff
@@ -0,0 +1,162 @@
...
+                    //read the value of state and then try state transitions  
+                    //only if the value of state does not change in the meantime 
+                    //(in another requestMore() or onCompleted()). If the value 
+                    // has changed and we expect to do a transition still then 
+                    // we loop and try again.
+                    final int s = state.get();
+                    if (s == NOT_REQUESTED_NOT_COMPLETED) {
+                        if (state.compareAndSet(NOT_REQUESTED_NOT_COMPLETED,
+                                REQUESTED_NOT_COMPLETED)) {
+                            request(Long.MAX_VALUE);
```
akarnokd: Since you don't override ```onStart``` it will request MAX_VALUE on subcsription and as such this is no-op and can be removed. If the intention was to request only when downstream actually requests, then I think we can pre-start the upstream and they will meet.

src/main/java/rx/Observable.java
```diff
@@ -7771,7 +7771,10 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
      * @see <a href="http://reactivex.io/documentation/operators/takelast.html">ReactiveX operators documentation: TakeLast</a>
      */
     public final Observable<T> takeLast(final int count) {
-        return lift(new OperatorTakeLast<T>(count));
+        if (count == 1 )
```
akarnokd: Is takeLast(0) meaningful? If so, it can be routed to ```ignoreElements()```. BTW, ```ignoreElements()``` can be optimized by making it singleton and be very simple (i.e., its onNext does nothing) instead of filtering with always false.

davidmoten:
Thanks @akarnokd for the review, I appreciate the lessons!

I've incorporated the changes except for the `ignoreElements` optimization which I'll leave for another PR.

I wrote this up before you submitted the `SingleDelayedProducer` in #2901 but had spotted its usefulness. It's very close to being reusable here, just have to handle the empty stream case which is allowed by `takeLast`. I submitted this anyway because it helps me learn, thanks.

Up to you of course if you want to merge this or wait for #2901 to be merged then get another PR that uses a modified `SingleDelayedProducer` (might have its name changed if it allows single or none).

Updated benchmarks (that's a nifty trick using Params):

```
Benchmark                                                  (size)   Mode   Samples        Score  Score error    Units
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLast            5  thrpt         5  2233452.067    81630.798    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLast          100  thrpt         5   444353.711   224920.714    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLast      1000000  thrpt         5       56.947        2.630    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLastOne         5  thrpt         5  2372577.831    76820.729    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLastOne       100  thrpt         5  1012528.600    75674.064    ops/s
r.o.OperatorTakeLastOnePerf.takeLastOneUsingTakeLastOne   1000000  thrpt         5      108.284        4.965    ops/s
```

src/main/java/rx/internal/operators/OperatorTakeLastOne.java
```diff
@@ -0,0 +1,168 @@
...
+                        return;
+                }
+            }
+        }
+
+        @Override
+        public void onCompleted() {
+            // CAS loop to atomically change state given that requestMore()
+            // may be acting concurrently
+            while (true) {
```
akarnokd: If ```last == ABSENT``` then there is no need to interact with the backpressure logic since a plain onCompleted can be sent before any request. In this case, I'd also not check isUnsubscribed().

src/test/java/rx/internal/operators/OperatorTakeLastOneTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Test
+    public void testLastWithBackpressure() {
+        MySubscriber<Integer> s = new MySubscriber<Integer>(0);
+        Observable.just(1).takeLast(1).subscribe(s);
+        assertEquals(0, s.list.size());
+        s.requestMore(1);
+        assertEquals(1, s.list.size());
+    }
+
```
akarnokd: I'd also add a test for takeLast(0) to make sure it 'delays' the onCompleted.

davidmoten:
Thanks @akarnokd.  I've added your suggestions:
* shortcut to completion when empty
* test for `takeLast(0)`

akarnokd:
Thanks!

benjchristensen:
:+1:

PR #2915, opened 2015-04-25
Observable.ignoreElements - optimize 

davidmoten:
As suggested by @akarnokd, instead of `filter(alwaysFalse)` use a custom operator and a singleton instance of it.

Also snuck in the removal of an unused import from `OperatorTakeLastOneTest.java`.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorIgnoreElements.java
+++ b/src/main/java/rx/internal/operators/OperatorIgnoreElements.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/main/java/rx/internal/operators/OperatorIgnoreElements.java
```diff
@@ -0,0 +1,58 @@
...
+        return (OperatorIgnoreElements<T>) Holder.INSTANCE;
+    }
+
+    private OperatorIgnoreElements() {
+
+    }
+
+    @Override
+    public Subscriber<? super T> call(final Subscriber<? super T> child) {
+        return new Subscriber<T>(child) {
```
akarnokd: This will forward the child's requests and will hang because the child doesn't receive an element and nobody is requesting more. I'd instead write this:
```java
Subscriber<T> result = new Subscribeer<T>() { ... }
child.add(result);
return result;
```

davidmoten:
Ah thanks. Now I understand the difference between those two approaches to connecting the subscribers. Simple really. I might add that bit of info to the javadoc on the Subscriber constructor I used in another PR, I was thinking of adding  bit more doco to Subscriber anyway because it gets stepped through a lot when you debug and it's a bit terse (confusing variable names and stuff).

davidmoten:
changes made including a unit test of the request problem (doesn't hang by the way, just terminates early, without processing all upstream).

akarnokd:
Great, thanks!

PR #2921, opened 2015-04-29
OperatorObserveOn - handle request overflow correctly

davidmoten:
Use `BackpressureUtils.getAndAddRequest` in the `Producer` in `OperatorObserveOn`.

Includes unit test that failed on codebase before this change.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -16,15 +16,22 @@
 package rx.internal.operators;
 
 import java.util.Queue;
-import java.util.concurrent.atomic.*;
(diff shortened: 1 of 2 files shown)


akarnokd:
Looks good, but I don't like what happens in the ```ObserveOnSubscriber``` constructor since the half-constructed ```this``` may escape through the ```setProducer``` call. 

If you could just move the code from the line ```child.add(scheduledUnsubscribe);``` up until the constructor's end into some ```init()``` method and make sure it is called after the object is constucted (near ```return new ObserveOnSubscriber<T>(scheduler, child);```), that would be great.

davidmoten:
Ok, done. Haven't squashed commits because I think worthwhile having as a separate entry in the history.

akarnokd:
Great, thanks!

PR #2923, opened 2015-04-29
OnBackpressureLatest: Non-blocking version of the toBlocking().latest() operator.

akarnokd:
This is essentially the pair of the ```OnBackpressureDrop```. While ```OnBackpressureDrop``` emits the first value requested and then drops the rest, ```OnBackpressureLatest``` drops values but the latest and emits that when requested. One can also think of this operator as a sample with backpressure.

One mentionable property is that since it holds onto the very last value, downstream has to request at least one to receive ```onCompleted()```.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -5368,6 +5368,27 @@ public class Observable<T> {
     }
     
     /**
+     * Instructs an Observable that is emitting items faster than its observer can consume them to 
+     * hold onto the latest value and emit that on request.
+     * <p>
+     * Its behavior is logically equivalent to toBlocking().latest() with the exception that
+     * the downstream is not blocking while requesting more values.
+     * <p>
+     * Note that if the upstream Observable does support backpressure, this operator ignores that capability
+     * and doesn't propagate any backpressure requests from downstream.
+     * <p>
+     * Note that due to the nature of how backpressure requests are propagated through subscribeOn/observeOn,
+     * requesting more than 1 from downstream doesn't guarantee a continuous delivery of onNext events.
+     * @return
+     * @Experimental The behavior of this can change at any time. 
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
+    public final Observable<T> onBackpressureLatest() {
+        return lift(OperatorOnBackpressureLatest.<T>instance());
+    }
+    
+    /**
      * Instructs an Observable to pass control to another Observable rather than invoking
      * {@link Observer#onError onError} if it encounters an error.
      * <p>
--- a/src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
+++ b/src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
@@ -1,0 +1,225 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package rx.internal.operators;
+
+import java.util.concurrent.atomic.*;
+
+import rx.Observable.Operator;
+import rx.*;
+
+/**
+ * An operator which drops all but the last received value in case the downstream
+ * doesn't request more.
+ */
+public final class OperatorOnBackpressureLatest<T> implements Operator<T, T> {
+    /** Holds a singleton instance initialized on class-loading. */
+    static final class Holder {
+        static final OperatorOnBackpressureLatest<Object> INSTANCE = new OperatorOnBackpressureLatest<Object>();
+    }
+    
+    /**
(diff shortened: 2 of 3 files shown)


src/main/java/rx/Observable.java
```diff
@@ -5368,6 +5368,27 @@ public final Boolean call(T t) {
...
+     * Instructs an Observable that is emitting items faster than its observer can consume them to 
+     * hold onto the latest value and emit that on request.
+     * <p>
+     * Its behavior is logically equivalent to toBlocking().latest() with the exception that
+     * the downstream is not blocking while requesting more values.
+     * <p>
+     * Note that if the upstream Observable does support backpressure, this operator ignores that capability
+     * and doesn't propagate any backpressure requests from downstream.
+     * <p>
+     * Note that due to the nature of how backpressure requests are propagated through subscribeOn/observeOn,
```
benjchristensen: This sentence isn't clear to me. I imagine it has something to do with how `request(n)` generally will happen in batches so `onNext` emissions will happen in bursts and drop everything but the latest in between in each burst?

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
```diff
@@ -0,0 +1,225 @@
...
+        static final Object EMPTY = new Object();
+        static final long NOT_REQUESTED = Long.MIN_VALUE / 2;
+        public LatestEmitter(Subscriber<? super T> child) {
+            this.child = child;
+            this.value = new AtomicReference<Object>(EMPTY);
+            this.lazySet(NOT_REQUESTED); // not 
+        }
+        @Override
+        public void request(long n) {
+            if (n >= 0) {
```
benjchristensen: Why doesn't BackpressureUtils apply to this usage?

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
```diff
@@ -0,0 +1,225 @@
...
+                        u = n;
+                    } else {
+                        u = r + n;
+                        if (u < 0) {
+                            u = Long.MAX_VALUE;
+                        }
+                    }
+                    if (compareAndSet(r, u)) {
+                        if (r == NOT_REQUESTED) {
+                            parent.requestMore(Long.MAX_VALUE);
```
benjchristensen: So ... we are tracking the requested value for downstream, but upstream we just request MAX_VALUE the first time we receive a request, correct? Is this why BackpressureUtils doesn't work?

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
```diff
@@ -0,0 +1,225 @@
...
+            done = true;
+            emit();
+        }
+        @Override
+        public void onCompleted() {
+            done = true;
+            emit();
+        }
+        void emit() {
+            synchronized (this) {
```
benjchristensen: Most of the time `onNext` would emit without contention if the consumer is fast. Does lock elision remove the performance impediment of a `synchronized` on every single `onNext` attempt?

Just thinking through the fact that this is a bunch of machinery for something that will typically be put in for dropping data when backpressure happens but should otherwise add negligible overhead.

benjchristensen:
I like this feature. Been wanting something like this for a while.

src/main/java/rx/Observable.java
akarnokd: The reason is that the operator basically pairs requests and emission on a 1:1 basis: if one requests 5, it would behave as if one called request(1) five times and thus giving the window for dropping unawaited values.

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
akarnokd: Because we have the intermediate state of NOT_REQUESTED. You did the same in ```Subscriber```.

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
akarnokd: Partly yes.

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
akarnokd: Using this operator synchronously does trigger lock elision and won't drop any value thus is not useful. Since the emitter-loop is quite simple and the operator is expected to run asynchronously almost always, this can be changed into a queue-drain if so desired. However, since its purpose is to drop elements because the consumer is slow, I don't really see any point in optimizing the emission further and I think a smaller overhead would just drop more values.

benjchristensen: > any point in optimizing the emission further and I think a smaller overhead would just drop more values.

The scenario I'm considering is where this operator is put in place to handle occasional bursts where generally the consumer is fast enough, but occasionally it is slow. This is a common use of `onBackpressureDrop` for us in our stream processor. We would use this in the same way.

src/main/java/rx/internal/operators/OperatorOnBackpressureLatest.java
benjchristensen: BackpressureUtils didn't exist at that point, so I am considering how to consolidate this type of logic as we keep repeating this type of non-trivial code and it is easy to get wrong. 

I'm okay with merging ... we really should spend some time figuring out the core patterns so we can encode the state machine, similar to what BackpressureUtils and AbstractOnSubscribe have started formalizing.

akarnokd: I use this ```NOT_REQUESTED``` only in few places (such as in ```OperatorPublish```) and I'm not yet confident the this request state value exhibits the same behavior and response action every time it needs to be evaluated.

PR #2927, opened 2015-04-30
SubscriptionIndexedRingBuffer - remove redundant check

davidmoten:
Very minor cleanup. A Subscription was being checked twice for a null value before unsubscribe.


akarnokd:
Thanks, but this class won't be in use once my merge rewrite PR #2928 is accepted.

davidmoten:
No worries, I'll close

PR #2928, opened 2015-04-30
Merge and MergeMaxConcurrent unified and rewritten

akarnokd:
I've rewritten ```merge``` and ```mergeMaxConcurrent``` together from scratch.

  - Added experimental ```maxConcurrent``` overloads to the array-merge and the mergeDelayError
  - Fixed ```OperatorMapNotification``` beause it disregarded backpressure on its terminal emissions.
  - While debugging some merge test failures, I've encountered a bug in OperatorPublish with the same underlying logical error (see comment) and fixed it here.
  - Accidentally introduced a few tabs, therefore, I went in and replaced all tabs with spaces.
  - I've changed ```OperatorMergeTest.shouldNotCompleteIfThereArePendingScalarSynchronousEmissionsWhenTheLastInnerSubscriberCompletes``` because it is legal to get onCompleted without any request.


akarnokd:
I haven't benchmarked this because my current, old laptop where I wrote this PR is prone to overheating.

davidmoten:
I'm stoked you've taken this on. Merge has been a thorn in our side for a while. I've only had a cursory look so far, looks great but can you break up the `emitLoop` method, it's super long (>200 lines)?

davidmoten:
Somebody go buy this man a laptop (how about you Netflix)!

akarnokd:
I've updated this PR and added safeguard against onNext before request() and made sure the queues are released if the stream is unsubscribed before completing normally.

akarnokd:
Here is a benchmark comparison on an Intel Celeron 1005M @ 1.9GHz, Windows 7 x64, Java 8u45 (this is not the overheating laptop of mine).

mergeperf

It seems the single-shot overhead is greater in general, but note that the baseline merge has bugs especially with scalar sources: such merges may terminate much earlier and not deliver all values.

```
Benchmark                    (size)      Master     Master err           PR     PR error
flatMapIntPassthruAsync             1  498695,426   259249,193   412488,433   257250,017
flatMapIntPassthruAsync        1000       628,490      325,556      826,485     1030,174
flatMapIntPassthruAsync     1000000         0,372        0,144        0,845        0,266
flatMapIntPassthruSync            1   2413250,773   340583,441  1380370,980   268957,879
flatMapIntPassthruSync         1000     16731,496     5159,943    16374,699     6581,767
flatMapIntPassthruSync      1000000        17,016        4,609       15,445        4,289
flatMapTwoNestedSync              1   1809099,547   314275,657  1074829,348   214620,187
flatMapTwoNestedSync           1000      8908,864     1665,210    12105,668     4147,721
flatMapTwoNestedSync        1000000         8,553        4,492       13,661        3,763
merge1SyncStreamOfN               1   2073223,271   624706,196  1340920,111   321874,624
merge1SyncStreamOfN            1000     15583,667     5675,705    24295,187      741,011
merge1SyncStreamOfN         1000000        16,086        3,177       30,617       14,407
mergeNAsyncStreamsOfN             1     53251,280     1821,052    77537,033    63762,352
mergeNAsyncStreamsOfN          1000         6,258        2,367        7,633        4,073
mergeNSyncStreamsOf1              1   1814390,194   250905,669  1221128,575   523755,940
mergeNSyncStreamsOf1            100    221860,250    69297,969   149696,874    13363,647
mergeNSyncStreamsOf1           1000     20576,165     2416,898    16998,408     2414,499
mergeNSyncStreamsOfN              1   2076685,785   558323,104  1382133,163   931994,488
mergeNSyncStreamsOfN           1000        16,458        3,175       28,144       15,014
mergeTwoAsyncStreamsOfN           1     45570,075     8864,806    62306,327    23355,026
mergeTwoAsyncStreamsOfN        1000      1086,413      260,172     3059,881      554,938
oneStreamOfNthatMergesIn1         1   1894495,101   373234,431  1430870,599   883644,992
oneStreamOfNthatMergesIn1      1000     18433,963     6341,333    18367,707    12874,644
oneStreamOfNthatMergesIn1   1000000        17,750        2,938       17,197        8,896
```

akarnokd:
@davidmoten  Can't really do because I need to return two values: the remaining request and how many main elements to request.

davidmoten:
OK, no worries

akarnokd:
i7 4790, Windows 7 x64, Java 8u45:

image

I've tried my best to close the gap with size = 1 cases, but couldn't figure out why it is that slower: even if I remove the emitter loop because of the trivial case, it is still 10% slower. Perhaps someone else can spot some unnecessary operation I missed.

akarnokd:
Fixed a potential concurrency problem in ```MergeProducer.request()```.

benjchristensen:
@akarnokd I've got this on my list to spend time on ... but it will take several hours to review and some non-trivial time to profile and test, including against our applications. 

Thank you very much for pursuing this.

@davidmoten You mention it's been a thorn in your side. What issues have you been having and are there test cases for them that we can use to assert this fixes them?

davidmoten:
Hi Ben, I haven't got any problems with merge at the moment but I have considered it a thorn in our side because it's such an important operator and yet the code isn't pretty (and has stuff in there now that is disabled like batching because we couldn't get it to work). It's no doubt suffered through RxJava's evolution and a nice clean fresh start from @akarnokd would be a great contribution.

In terms of prettiness, and this might be for @akarnokd to consider in his rewrite, is that there are decisions made in the merge operator about how much to request from each source (the initial and subsequent requests) that I found hard to understand and I suspect would be nice to base on research /perfs or even make customizable. To this end I'd like to see decoupling of this sort of logic into a MergeRequestStrategy class say. I guess my aims would be 

* to be able to customize the requesting behaviour  of the merge operator under backpressure for my use cases
* to optimise the RxJava default request behaviour for the most common use cases (I guess we'd take a wild stab at what they are)

benjchristensen:
> optimise the RxJava default request behaviour for the most common use cases

The choice of 128 was made after testing various different settings: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/util/RxRingBuffer.java#L162

What do you recommend doing differently?

> to be able to customize the requesting behaviour of the merge operator under backpressure for my use cases to increase performance

What do you want to see done?

davidmoten:
128 is the initial request but the requests following were much smaller
from memory (1, 3, 2,1, etc)  and it was hard to figure out what strategy
for those secondary requests was at play and if it had been benchmarked
against other strategies.

davidmoten:
I think I can simplify what I want having thought about it a little bit. I'm happy with 128 as an initial request   and would like further requests to also be for the same number.

benjchristensen:
> would like further requests to also be for the same number.

Ah, that makes sense. I tried that at one point but couldn't get it to work as it got ridiculously complicated and I never went back to solve it. I definitely would like as well.

akarnokd:
While rewatching old Rx videos, I remembered one can implement flatMap in terms of merge and vice versa. I think if merge was be implemented via flatMap, one could save the allocation cost of the lifted map. Then merge == flatMap(o -> o). Any interest in pursuing this?

headinthebox:
That should be easy to check.

benjchristensen:
Perf testing on my MacBook Pro show this PR to be very good in most cases, with small hits in single item usage.

```
Benchmark                                          (size)      Existing            PR    
r.o.OperatorMergePerf.merge1SyncStreamOfN               1   4488176.832   3797359.176  -  
r.o.OperatorMergePerf.merge1SyncStreamOfN            1000     41297.217     66490.338  ++  
r.o.OperatorMergePerf.merge1SyncStreamOfN         1000000        38.634        79.525  ++  
r.o.OperatorMergePerf.mergeNAsyncStreamsOfN             1    102128.738    108055.722    
r.o.OperatorMergePerf.mergeNAsyncStreamsOfN          1000         5.216        13.558  +  
r.o.OperatorMergePerf.mergeNSyncStreamsOf1              1   4390861.330   4299418.343    
r.o.OperatorMergePerf.mergeNSyncStreamsOf1            100    471690.468    599665.596  ++  
r.o.OperatorMergePerf.mergeNSyncStreamsOf1           1000     52613.616     60820.047  + 
r.o.OperatorMergePerf.mergeNSyncStreamsOfN              1   4363837.871   4116988.261  -  
r.o.OperatorMergePerf.mergeNSyncStreamsOfN           1000        38.733        65.728  ++  
r.o.OperatorMergePerf.mergeTwoAsyncStreamsOfN           1     80926.413     86795.725  +  
r.o.OperatorMergePerf.mergeTwoAsyncStreamsOfN        1000      2183.016      6377.939  ++  
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1         1   4622062.299   4015324.624  -  
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1      1000     38870.489     61401.234  ++  
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1   1000000        36.085        63.094  ++  


Benchmark                                          (size)      Existing           PR
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync         1    361714.683   394960.426 +
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync      1000       368.362      422.990 +
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync   1000000         0.331        0.423 +
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync          1   5372884.569  4637892.778 -
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync       1000     38254.799    59231.991 ++
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync    1000000        34.971       59.548 ++
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync            1   3850581.086  3783613.293 
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync         1000     20510.271    34840.826 ++
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync      1000000        19.049       39.653 ++
```

akarnokd:
I've updated the PR to include some scalar merge optimizations:

image

The biggest overhead reduction is in `flatMapIntPassthruSync:1`. Unfortunately, these changes couldn't help much with the `OperatorMergePerf` tests of size 1.

akarnokd:
I've also tried various inlining:
image

  - In ScalarMap, I've inlined the mapping over a scalar.
  - In InlineMap, I've changed the code to calculate `func` in map at construction time for a scalar source (which is incorrect if the `func` depends upon the environment as well)
  - In FixedMerge, I've added checks to merge(Iterable) for a 1 element Iterable.

I measured improvements but either the approach was too special or performance degradation appeared in n > 1 cases.

benjchristensen:
Since you're updating this PR, can you rebase it and remove all the unrelated changes that touched whitespace in various files?

I'm in process of reviewing this code and we'll do a production canary test with our systems at Netflix due to the high impact of these changes.

akarnokd:
Done.

benjchristensen:
Thank you.

@abersnaze Use https://github.com/benjchristensen/RxJava/tree/merge-rewrite-from-2928 as a clean merge of this PR on top of 1.x to production canary this code.

benjchristensen:
I don't see a change in perf results with the updated code.

```
Benchmark                                          (size)      Existing            PR        PR Update
r.o.OperatorMergePerf.merge1SyncStreamOfN               1   4488176.832   3797359.176  -   3793580.516
r.o.OperatorMergePerf.merge1SyncStreamOfN            1000     41297.217     66490.338  ++    67796.814
r.o.OperatorMergePerf.merge1SyncStreamOfN         1000000        38.634        79.525  ++       75.695
r.o.OperatorMergePerf.mergeNAsyncStreamsOfN             1    102128.738    108055.722        94091.453
r.o.OperatorMergePerf.mergeNAsyncStreamsOfN          1000         5.216        13.558  +        13.106
r.o.OperatorMergePerf.mergeNSyncStreamsOf1              1   4390861.330   4299418.343      3710019.659
r.o.OperatorMergePerf.mergeNSyncStreamsOf1            100    471690.468    599665.596  ++   546427.145
r.o.OperatorMergePerf.mergeNSyncStreamsOf1           1000     52613.616     60820.047  +     60663.164
r.o.OperatorMergePerf.mergeNSyncStreamsOfN              1   4363837.871   4116988.261  -   3865008.834 --
r.o.OperatorMergePerf.mergeNSyncStreamsOfN           1000        38.733        65.728  ++       61.194
r.o.OperatorMergePerf.mergeTwoAsyncStreamsOfN           1     80926.413     86795.725  +     78765.125
r.o.OperatorMergePerf.mergeTwoAsyncStreamsOfN        1000      2183.016      6377.939  ++     6060.544
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1         1   4622062.299   4015324.624  -   3935245.004 --
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1      1000     38870.489     61401.234  ++    57938.065
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1   1000000        36.085        63.094  ++       58.667


Benchmark                                          (size)      Existing           PR        Updated PR
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync         1    361714.683   394960.426 +      389851.694
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync      1000       368.362      422.990 +         423.083
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync   1000000         0.331        0.423 +           0.430
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync          1   5372884.569  4637892.778 -     4672089.981
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync       1000     38254.799    59231.991 ++      60356.703
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync    1000000        34.971       59.548 ++         61.814
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync            1   3850581.086  3783613.293       3913069.169
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync         1000     20510.271    34840.826 ++      33852.646
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync      1000000        19.049       39.653 ++         39.805
```

akarnokd:
It appears I've accidentally overwritten the scalarMerge optimization with the 3rd try version when I fixed the whitespaces.

akarnokd:
I've readded the optimization: The results on my i7 4770K, Windows 7 x64, JDK8u45 (the one with ~38M was an i7 4790):

image

benjchristensen:
With that change perf looks good across the board:

```

Benchmark                                          (size)      Existing            PR       Updated PR
r.o.OperatorMergePerf.merge1SyncStreamOfN               1   4488176.832   3797359.176  -   4159885.024 (variability within Score Error)
r.o.OperatorMergePerf.merge1SyncStreamOfN            1000     41297.217     66490.338  ++    68923.081 ++
r.o.OperatorMergePerf.merge1SyncStreamOfN         1000000        38.634        79.525  ++       81.282 ++
r.o.OperatorMergePerf.mergeNAsyncStreamsOfN             1    102128.738    108055.722       106696.532
r.o.OperatorMergePerf.mergeNAsyncStreamsOfN          1000         5.216        13.558  +        13.258 +
r.o.OperatorMergePerf.mergeNSyncStreamsOf1              1   4390861.330   4299418.343      3989522.580 (variability within Score Error)
r.o.OperatorMergePerf.mergeNSyncStreamsOf1            100    471690.468    599665.596  ++   551513.541 ++
r.o.OperatorMergePerf.mergeNSyncStreamsOf1           1000     52613.616     60820.047  +     59881.347 +
r.o.OperatorMergePerf.mergeNSyncStreamsOfN              1   4363837.871   4116988.261  -   3894283.923 (variability within Score Error)
r.o.OperatorMergePerf.mergeNSyncStreamsOfN           1000        38.733        65.728  ++       58.229 ++
r.o.OperatorMergePerf.mergeTwoAsyncStreamsOfN           1     80926.413     86795.725  +     87696.276 +
r.o.OperatorMergePerf.mergeTwoAsyncStreamsOfN        1000      2183.016      6377.939  ++     6114.703 ++
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1         1   4622062.299   4015324.624  -   3858413.960 (variability within Score Error)
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1      1000     38870.489     61401.234  ++    60379.171 ++
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1   1000000        36.085        63.094  ++       60.046 ++


Benchmark                                          (size)      Existing           PR       Updated PR  
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync         1    361714.683   394960.426 +     445826.696 ++
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync      1000       368.362      422.990 +        430.198 + 
r.o.OperatorFlatMapPerf.flatMapIntPassthruAsync   1000000         0.331        0.423 +          0.426 + 
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync          1   5372884.569  4637892.778 -   27905012.198 +++ 
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync       1000     38254.799    59231.991 ++     59401.089 ++ 
r.o.OperatorFlatMapPerf.flatMapIntPassthruSync    1000000        34.971       59.548 ++        59.818 ++ 
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync            1   3850581.086  3783613.293      3673091.400  
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync         1000     20510.271    34840.826 ++     32466.541 ++ 
r.o.OperatorFlatMapPerf.flatMapTwoNestedSync      1000000        19.049       39.653 ++        36.963 ++ 
```

akarnokd:
The problem with these:
```
Benchmark                                          (size)      Existing            PR       Updated PR
r.o.OperatorMergePerf.merge1SyncStreamOfN               1   4488176.832   3797359.176  -   4159885.024 (variability within Score Error)
r.o.OperatorMergePerf.mergeNSyncStreamsOf1              1   4390861.330   4299418.343      3989522.580 (variability within Score Error)
r.o.OperatorMergePerf.mergeNSyncStreamsOfN              1   4363837.871   4116988.261  -   3894283.923 (variability within Score Error)
r.o.OperatorMergePerf.oneStreamOfNthatMergesIn1         1   4622062.299   4015324.624  -   3858413.960 (variability within Score Error)
```

Is that the scalar merge optimization can't apply because they are:
`merge(just(1).map(v -> just(1)))` so the merge() receives an arbitrary Observable instead of a scalar; the merge() hides the first just's scalar nature.

benjchristensen:
I finally spent the hours to review this thoroughly ... very well written code @akarnokd 

I'm going to now add inline comments and questions on the code. Most are clarifications, though there is at least one bug I think I found.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
-    private final boolean delayErrors;
-
     @Override
     public Subscriber<Observable<? extends T>> call(final Subscriber<? super T> child) {
-        return new MergeSubscriber<T>(child, delayErrors);
-
+        MergeSubscriber<T> subscriber = new MergeSubscriber<T>(child, delayErrors, maxConcurrent);
+        MergeProducer<T> producer = new MergeProducer<T>(subscriber);
+        subscriber.producer = producer;
+        producer.lazySet(0); // make sure subscriber.producer is visible
```
benjchristensen: What is this lazySet about? What about visibility would not work if this wasn't done?

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
-         */
+        
         @Override
-        public void onNext(Observable<? extends T> t) {
-            if (t instanceof ScalarSynchronousObservable) {
-                ScalarSynchronousObservable<? extends T> t2 = (ScalarSynchronousObservable<? extends T>)t;
-                handleScalarSynchronousObservable(t2);
-            } else {
-                if (t == null || isUnsubscribed()) {
+        public void request(long n) {
```
benjchristensen: This `request` method, and perhaps this entire `Producer` implementation seems to make sense as part of BackpressureUtils as it is mostly generic. The `subscriber.*` methods that are invoked could be a `BackpressureSubscriber` interface perhaps?

This is code that seems reusable for all of our backpressure aware operators.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+        final Subscriber<? super T> child;
+        final boolean delayErrors;
+        final int maxConcurrent;
+        
+        MergeProducer<T> producer;
+        
+        /** Upstream might not produce scalar values at all so avoid creating a queue for it upfront. */
+        volatile RxRingBuffer scalarQueue;
+        /** Tracks the active subscriptions to sources. */
+        volatile CompositeSubscription subscriptions;
```
benjchristensen: The use of `CompositeSubscription` was specifically avoided before and replaced with `SubscriptionIndexedRingBuffer` to avoid the overhead of `HashSet` when constant add/remove is happening, which is a common behavior for `merge`.

Why did you not continue to use `SubscriptionIndexedRingBuffer`?

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+        static final Object REPORT_COMPLETED = new Object();
+        
+        public MergeSubscriber(Subscriber<? super T> child, boolean delayErrors, int maxConcurrent) {
+            this.child = child;
+            this.delayErrors = delayErrors;
+            this.maxConcurrent = maxConcurrent;
+            this.nl = NotificationLite.instance();
+            this.innerGuard = new Object();
+            this.innerSubscribers = EMPTY;
+            this.request(0);
```
benjchristensen: Curious as to why you are calling this here instead of using `onStart`?

I think this is safe to do from within the constructor due to when this all gets initialized, but on first read it felt wrong for this to happen here. Probably not an issue but would like to understand your thinking for not following the normal pattern of using `onStart`.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
-                request(1);
+            }
+            return c;
+        }
+        
+        /** 
+         * In case the downstream requests less than Long.MAX_VALUE initially, do the standard
+         * queued processing.
+         */
+        void requestDefault() {
```
benjchristensen: Perhaps worth adding docs that this method can only be invoked once at the beginning and that this constraint is ensured by MergeProducer.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+            request(Math.min(maxConcurrent, RxRingBuffer.SIZE));
+        }
+        /**
+         * In case the downstream requests Long.MAX_VALUE, perform a simplified merging without
+         * using queues at all.
+         */
+        void requestAll() {
```
benjchristensen: Perhaps worth adding docs that this method can only be invoked once at the beginning and that this constraint is ensured by MergeProducer.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+            fastMode = true;
+            request(maxConcurrent == Integer.MAX_VALUE ? Long.MAX_VALUE : maxConcurrent);
+        }
+        
+        @Override
+        public void onNext(Observable<? extends T> t) {
+            if (t == null) {
                 return;
+            }
+            if (fastMode) {
```
benjchristensen: This volatile read happens on every single `onNext` for a decision that gets made once at the very beginning. 

I haven't yet figured out a way around it though, but wish there was an option.

It would be preferable if we could have 2 subclasses of `MergeSubscriber` for the two paths, but we don't know at instantiation time which one to use.

I also considered volatile piggybacking, but I don't think there are other volatiles that are guaranteed to be read every time `onNext` is invoked. 

The concern is the number of volatiles being invoked for each `onNext` since that's the hot path and metrics from past testing have shown that to impede performance.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+                        synchronized (this) {
+                            emitting = false;
+                        }
                     }
-                    if (emitted == r) {
-                        // we emitted as many as were requested so stop the forEach loop
-                        return Boolean.FALSE;
+                }
+                // in case there was some activity, continue with the regular emission loop
+                emitLoop();
```
benjchristensen: This line is very hard to reason about and either needs improved documentation, or a refactoring to place it more logically.

This can ONLY be called if `emitted==true` and should only be called if `missed==true`, but that is far from obvious here.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+            } else {
+                // either someone else is emitting or there was no available request
+                // queue up the scalar
+                RxRingBuffer svq = scalarQueue;
+                if (svq == null) {
+                    svq = RxRingBuffer.getSpscInstance();
```
benjchristensen: Am I correct in understanding that this is thread-safe since `tryEmitScalar` will only ever be called synchronously from one `onNext` at a time even though other inner subscribers could concurrently have the `emitting` flag?

It's confusing because the code for `emitting` needs to be synchronized, but this doesn't need to be as it is affected by only the outer Observable, not the concurrently subscribed inner Observables.

Perhaps some explanation in docs as to the thread-safety would be valuable to others reading the code? I had to flip back and forth in code several times to convince myself it was safe.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+            synchronized (this) {
+                if (emitting) {
+                    int s = fastSize;
+                    Object[] o = fastArray;
+                    if (o == null) {
+                        o = new Object[4];
+                        fastArray = o;
+                    } else
+                    if (s == o.length) {
+                        Object[] o2 = new Object[s + (s >> 2)];
```
benjchristensen: This is capable of unbounded queue growth and breaks backpressure.

Here is a simple test:

```java
        Observable.merge(
                Observable.range(0, 10000000).subscribeOn(Schedulers.computation()), 
                Observable.range(10000001, 20000000)
            )
        .subscribe(new Subscriber<Integer>() {

            @Override
            public void onCompleted() {
            }

            @Override
            }

            @Override
                try {
                    Thread.sleep(100); // simulate computation time
                    e.printStackTrace();
                }
                System.out.println(t);
            }
            
        });
```

If I put this code in here after the resize occurs:

```java
System.out.println("resize list = " + o.length);
```

Then I get this output:

```
resize list = 5
resize list = 6
resize list = 7
resize list = 8
resize list = 10
resize list = 12
resize list = 15
resize list = 18
resize list = 22
resize list = 27
resize list = 33
resize list = 41
resize list = 51
resize list = 63
resize list = 78
resize list = 97
resize list = 121
resize list = 151
resize list = 188
resize list = 235
resize list = 293
resize list = 366
resize list = 457
resize list = 571
resize list = 713
resize list = 891
resize list = 1113
resize list = 1391
resize list = 1738
resize list = 2172
resize list = 2715
resize list = 3393
resize list = 4241
resize list = 5301
resize list = 6626
resize list = 8282
resize list = 10352
resize list = 12940
resize list = 16175
resize list = 20218
resize list = 25272
resize list = 31590
resize list = 39487
resize list = 49358
resize list = 61697
resize list = 77121
resize list = 96401
resize list = 120501
resize list = 150626
resize list = 188282
resize list = 235352
resize list = 294190
resize list = 367737
resize list = 459671
resize list = 574588
resize list = 718235
resize list = 897793
resize list = 1122241
resize list = 1402801
```

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
-                boolean sendOnComplete = false;
-                synchronized (this) {
-                    if (!parent) {
-                        wip--;
+                emitting = true;
+            }
+            boolean skipFinal = false;
+            try {
+                child.onNext(value);
+                for (int k = Integer.MAX_VALUE; k >= 0; k--) {
```
benjchristensen: Why the `MAX_VALUE` for-loop instead of `while(true)`?

In an extreme edge case of thread-starvation this seems like it could end the loop while there are still items in the queue. This would then mean the `child.onNext` on the previous line could emit while the queue is not empty for the next thread that comes along and picks it up ... or it could result in the system hanging if a terminal event was left in the queue.

It's an extreme edge case I'm sure, as `Integer.MAX_VALUE` is a large number, but I think it could theoretically happen.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+                if (emitting) {
+                    int s = fastSize;
+                    Object[] o = fastArray;
+                    if (o == null) {
+                        o = new Object[4];
+                        fastArray = o;
+                    } else
+                    if (s == o.length) {
+                        Object[] o2 = new Object[s + (s >> 2)];
+                        System.arraycopy(o, 0, o2, 0, s);
```
benjchristensen: Do we have to do array resizing and copying while we hold the lock? That seems dangerous as we lock all threads while memory IO is happening.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
-        @Override
-        public void onCompleted() {
-            boolean c = false;
+        void fastEmitError() {
             synchronized (this) {
-                completed = true;
-                if (wip == 0) {
-                    c = true;
+                if (emitting) {
+                    Object[] o = new Object[1];
```
benjchristensen: This seems to throw away the current `fastArray` as soon as an error occurs. Doesn't that risk losing data?

The `merge` operator has never dropped data in the event of an error and I don't see a reason to start doing so.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
-                if (wip == 0 && completed) {
-                    sendOnComplete = true;
+                if (emitting) {
+                    int s = fastSize;
+                    Object[] o = fastArray;
+                    if (o == null) {
+                        o = new Object[4];
+                        fastArray = o;
+                    } else
+                    if (s == o.length) {
```
benjchristensen: This array management stuff is repeated at least twice, and should probably be used the same in all three `fastNext/Error/Completed` methods. 

Seems like a good utility to extract out as it is verbose and should always be changed the same in all of them if fixes need to occur.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+                    missed = true;
+                    return;
                 }
+                emitting = true;
             }
-            childrenSubscribers.remove(s.sindex);
-            if (sendOnComplete) {
-                drainAndComplete();
-            }
+            // if we get here, fastArray is always null because fastEmitNext always empties the arrays
```
benjchristensen: It should always empty the arrays for this logic to work, but the `MAX_VALUE` for-loop does all an edge case where data could still be enqueued doesn't it?

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
+                    // check if upstream is done, there are no scalar values and no active inner subscriptions
+                    if (d && (svq == null || svq.isEmpty()) && n == 0) {
+                        Queue<Throwable> e = errors;
+                        if (e == null || e.isEmpty()) {
+                            child.onCompleted();
+                        } else {
+                            reportError();
+                        }
+                        if (svq != null) {
+                            svq.release();
```
benjchristensen: Is this eager release an optimization only? It will also be released via unsubscribe, correct?

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -71,715 +72,878 @@
...
             }
         }
-
-        private void enqueue(T t, boolean complete) {
+    }
+    /**
+     * Subscriber that requests Long.MAX_VALUE from upstream and doesn't queue. 
+     * @param <T>
+     */
+    static final class FastInnerSubscriber<T> extends Subscriber<T> {
```
benjchristensen: I think this optimization can't be used because `merge` itself queues. See the bug report above.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: If child is unsubscribed after this point, the subsequent add() will skip the synchronized block and unsubscribe `subscriber` thus no release takes place and we hand our 'half constructed' subscriber up the chain.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: This is a special case of request method building and would require all sorts of callbacks: `emit`, `requestDefault` and `requestAll` so either 3 Action0 or a base interface. I'm not convinced this pattern will appear somewhere else.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: I don't trust that implementation and current perf tests doesn't seem to be penalized by using `CompositeSubscription`.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: I don't want MergeSubscriber to request anything until the downstream decides to go with limited or unbounded requests. `onStart` happens after hooks have been applied and perhaps synchronization barriers passed. By putting it there, the compiler has the chance to remove  the synchronization block inside request.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Maybe, but we usually don't document the lifecycle intricacies of default and private methods.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Volatile reads are cheap and since it won't change ever after, it will stay in L1. If I do a plain+volatile+plain read (similar to a double-checked locking) that plain variable is not guaranteed to stay in a register and will be promoted to L1 as well, so no win there.

The other concern that the JIT won't be able to do some optimizations due to its barrier nature is not an issue here because the subsequent operations do atomics and `synchronized` which have the same effect.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: It is part of the fast-path recovery here and is called if `emitting == true` and there was a missed true -> false transition just before it. Maybe the diff-view is too distracting in this case. Otherwise, I'm not sure what additional documentation it needs.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Yes, that's correct.

Even if say its thread safe, you have to flip back and forth to verify it.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: This is a tricky situation. The downstream starts with a Long.MAX_VALUE which triggers the unbounded mode which also triggers an unbounded mode with the sources: backpressure is preserved in this sense. However, the asynchrony between the sources will bloat the serialization buffer (the same thing happens with `SerializedObserver`). The solution is to not have the unbounded mode at all which then penalizes sort unbounded merges.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: The `fastEmitNext` is potentially called from multiple threads and only one of them is allowed to resize the buffer. We don't have too many other options: a lock-free version would have to use an atomic array of atomic arrays with an ingress and egress counters which involve at least 4 atomic operations per element and penalizes synchronous operation heavily.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: > Why the MAX_VALUE for-loop instead of while(true)?
To avoid a safepoint poll and thus make a longer lasting emission session more performant in a synchronous case, but since synchronized has the likelihood to bring the thread into a safepoint in an asynchronous case, it doesn't really matter.

> It's an extreme edge case

You are right, we shouldn't take chances with it. I'll add an outer `for(;;)` loop.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: If the merge doesn't delay errors then it has to drop all buffered values because we report errors eagerly in such a case.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Okay.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Yes, I'll make `fastEmitNext` exhaustive.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Yes, but who knows when that's going to happen due to the frequent use of `unsafeSubscribe` in the chains.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: So you say let's drop the unbounded mode completely?

akarnokd: The idea of the unbounded mode is to bypass queueing as much as possible. Naturally, FastInnerSubscriber can request in batches so the serialization buffer doesn't grow unbounded, but what should be the default batch size?

akarnokd: I wouldn't consider this a bug but rather an inconvenience. For shorter streams, the optimization is quite valuable. For longer streams, we could have a batcher operator which changes the request amount to some limited number, just like one would use `observeOn()` without actually introducing asynchrony.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: Still not making sense to me. What synchronized block are you referring to? The `SubscriptionList.add` inside `Subscriber.add` looks safe to me for handling the child unsubscription case. 

Please walk me through it more clearly. If I am struggling to understand this then others surely will as well when trying to understand this code.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: Perf tests typically didn't show it in throughput numbers when I dealt with this issue before. It is object allocation measurements done with JMC and in production apps that show up in GC performance issues in non-trivial apps - which micro-benchmarks are not. 

If you don't trust the implementation please provide concrete reasons why.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: No requests will go upstream until after `onStart` is invoked so the use case you're seeking will be the same, just like everywhere else we do it. That's why `onStart` exists.

What concrete byte code differences or performance tests back up the the claims?

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: > Even if say its thread safe, you have to flip back and forth to verify it.

Yes, of course. But docs then show that this is considered. As it stands it is right next to code that must handle concurrency so this code stands out as being wrong.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: This is not just an inconvenience. It's a bug. It breaks the entire purpose of backpressure by having an unbounded queue. The whole point of `request(n)` backpressure semantics is that things "just work" and there are no unbounded buffers, memory leaks, or dropping data unless a developer actively decides to do any of those things.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: Since when do we report errors eagerly? The only place I'm aware of that decision ever being made is in `observeOn` due to the concurrent nature of thread-hopping. 

We didn't do that with merge before, so why now?

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: I suggest that the solution is to not use a growable array, especially since we shouldn't have an unbounded buffer anyways. This is why we have pooled `RxRingBuffer`.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: It is unbounded down, that doesn't mean unbounded up. The `merge` operator is an async operator, just like `zip` and `observeOn` since it does queueing. This means the `merge` operator itself IS an async operator and thus has buffered, `request(n)` semantics regardless of whether the downstream requests `Long.MAX_VALUE` or not.

The only cases I'm aware of where it doesn't need to buffer are the following where no concurrent emission occurs:

1) When `merge` has only a single stream to merge (not a real merge, so not an interesting case)
2) When `maxConcurrent==1` which is the same as `concat` or effectively the same as usecase 1 above.
3) When each Observable being merged is synchronous, which can't be known by `merge` ahead of time. In this case it again behaves like `concat`. 

Since (3) is the common case to optimize for (ie. many short synchronous streams), we would have to determine this at runtime. Basically we assume that if we never have `missed` occur due to concurrent emission, we stay on the "fast path", but as soon as we see concurrency occur we have to use queues.

This is why the current implementation has an optimization to skip the queues: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorMerge.java#L612 and https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorMerge.java#L686

But it uses the queues as soon as concurrent emissions happen.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: You are right, that line is unnecessary precaution. The there's going to be enough synchronization blocks before the end of the method so there shouldn't be any visibility problem.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Fine, I'll remove the optimization.

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: Fine.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: You're right that there is already lots of synchronization so we won't ever really have an optimal path through merge. It's just the nature of this operator.

> Volatile reads are cheap

Generally yes, but they aren't free. For this case though I agree with you and can't think of a better way.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: I wasn't reading this in diff view. This was an impossible review to do without pulling the code into an IDE and actually working with it.

I know it's part of the recovery path, but it's very non-obvious.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: We don't want to remove optimization, it just needs to be done differently, like the current version has. See comments here: https://github.com/ReactiveX/RxJava/pull/2928#discussion_r32748363

akarnokd: I can't bound that array because I'd need to track who originated each value to ask the proper sender for replenishment, which is essentially what the regular path does.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: Good point. Don't we always do the eager unsubscribe in async operators that could delay the unsubscribe, such as `observeOn`? See here: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorObserveOn.java#L159

Synchronous operators aren't an issue. Only temporal (such as those that delay termination) or async operators. But, if any of them don't correctly optimize like `observeOn` then we would delay releasing.

This eager release is cheap and makes sense, so I'm not suggesting removal of it. Just validating my understanding, and that if all operators were implemented correctly, including with unsafeSubscribe, we should be okay.

src/main/java/rx/internal/operators/OperatorMerge.java
benjchristensen: That's the point of my comments. We have to maintain the backpressure semantics of the "regular path". We can't break backpressure in the name of optimization.

benjchristensen:
Pulling comments on optimization into the main comments rather than the diffs which are getting hidden as code changes.

Modified comment from https://github.com/ReactiveX/RxJava/pull/2928#discussion_r32748363:

----------------

The downstream requesting unbounded data does not necessarily mean unbounded up. The `merge` operator is an async operator, just like `zip` and `observeOn` since it does queueing. This means the `merge` operator itself IS an async operator and thus has buffered, `request(n)` semantics regardless of whether the downstream requests `Long.MAX_VALUE` or not.

The only cases I'm aware of where it doesn't need to buffer are the following where no concurrent emission occurs:

1) When `merge` has only a single stream to merge (not a real merge, so not an interesting case)
2) When `maxConcurrent==1` which is the same as `concat` or effectively the same as usecase 1 above.
3) When each Observable being merged is synchronous, which can't be known by `merge` ahead of time. In this case it again behaves like `concat`. 

Since (3) is the common case to optimize for (ie. many short synchronous streams), we would have to determine this at runtime. Basically we assume that if we never have `missed` occur due to concurrent emission, we stay on the "fast path", but as soon as we see concurrency occur we have to use queues.

This is why the current implementation has an optimization to skip the queues: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorMerge.java#L612 and https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorMerge.java#L686

But it uses the queues as soon as concurrent emissions happen.

So we can't have an unbounded buffer as an optimization, but we can try to skip use of the buffer when we don't have concurrent emissions. As soon as we have concurrent emissions though we must use the buffer. And we must maintain `request(n)` semantics.

akarnokd:
Okay. I'll remove the unbounded paths and add a fast-path to the InnerSubscribers.

However, I found a few cases that needs further decisions:

In `OperatorMergeDelayErrorTest.testErrorDelayed1` expects the source to emit "six" which doesn't happen in the normal path of this PR because I shut the source down. The test indicates the original merge allowed value emissions after error from the same source.

In the same test class in `testMergeSourceWhichDoesntPropagateExceptionBack` if the child subscriber's onNext throws, why do we want to keep emitting the subsequent values if delayErrors == true? If the child crashed while trying to work with the value why do we think the next value won't crash it or find it in an inconsistent state?

The `OperatorFlatMapPerf.flatMapIntPassthruAsync` has a logical error. It runs an async observation but doesn't wait for the completion of it. Basically it floods the system with scheduled tasks. If I add a latched observer, I get ~ 200k ops/s with both 1.x and this PR.

benjchristensen:
> I'll remove the unbounded paths and add a fast-path to the InnerSubscribers.

Thank you.

> The test indicates the original merge allowed value emissions after error from the same source.

That sounds wrong. A given `Observable` can't emit after an `onError` so that test would be invalid. 

> If the child crashed while trying to work with the value why do we think the next value won't crash it or find it in an inconsistent state?

That's a hard one. I can see your point, but `mergeDelayError` asks to "delay errors", so we delay them. If we change that behavior we will almost certainly break usage. This unit test came from user requirements for `mergeDelayError`.

> The OperatorFlatMapPerf.flatMapIntPassthruAsync has a logical error

Then let's fix it in 1.x and rebase this PR on top so we are comparing the right metrics.

akarnokd:
> Then let's fix it in 1.x

I'm busy with the changes to this PR. Could you do this?

benjchristensen:
> Could you do this?

Sure.

akarnokd:
Well, my first try is mostly a disaster:

image

benjchristensen:
:-)

akarnokd:
I did recover most of the performance, but still some cases lag behind:

image

I managed this by adding a batching request to the InnerSubscriber (same logic as in #3030). I think the remaining performance loss is due to requesting from the main source one by one, but I'm not sure if I want to batch those as well because the maxConcurrent nature. The duplicate tryEmit is due to avoid a bimorphic dispatch: if I implement a common base class, I lose ~5% from the throughput values.

I'm finished for today but I'll try to get back to the 1.x performance in the remaining clases.

akarnokd:
It seems the merge is quite sensitive to method structuring: just by refactoring the scalar emission path I got some notable performance back, but still, the single-value merge are a bit slower than 1.x. I think, however, that the green improvements are quite worth it.

image

These are the two 1000 sized tests with their error range:
```
   flatMapIntPassthruSync       1000  thrpt         5    53754,733     4624,880    ops/s
oneStreamOfNthatMergesIn1      1000  thrpt         5    55004,627      772,675    ops/s
```
They never got closer to the 1.x line. At one point, I changed merge() to assume MAX_VALUE requests all the time but it didn't affect the number and I have no idea where the performance leaks. They both merge 1000 scalar values which don't touch any queue, don't account the production and just request another value before they return.

benjchristensen:
Nothing stands out when I do a Flight Recorder profiling ... reviewing code changes.

benjchristensen:
@akarnokd Do the new unit tests in this PR cover the things that were broken in the current merge impl? If not, what other use cases are broken today that this fixes?

Note that I'm still digging into this revised code and exploring various perf tests.

akarnokd:
Yes, there was a potential hang when merging only scalar values and when merging a mixture of scalar and non-scalar values.

benjchristensen:
Were you ever able to reproduce that with a unit test? Do you have that code?

akarnokd:
They are the 'flatMapRangeMixedAsyncLoop' and 'testFastMergeFullScalar' tests.

src/main/java/rx/internal/util/ScalarSynchronousObservable.java
```diff
@@ -117,4 +120,32 @@ public void call() {
...
+    public <R> Observable<R> scalarFlatMap(final Func1<? super T, ? extends Observable<? extends R>> func) {
+        return create(new OnSubscribe<R>() {
+            @Override
+            public void call(final Subscriber<? super R> child) {
+                Observable<? extends R> o = func.call(t);
+                if (o.getClass() == ScalarSynchronousObservable.class) {
+                    child.onNext(((ScalarSynchronousObservable<? extends R>)o).t);
+                    child.onCompleted();
+                } else {
+                    o.subscribe(new Subscriber<R>(child) {
```
benjchristensen: Doesn't this have to be `unsafeSubscribe`?

akarnokd: Right. I've updated & squashed the PR.

benjchristensen:
I did another pass on this and think it's good for merging and building a snapshot for a few days of final testing before release.

I still question the subscribe/unsafeSubscribe thing above, but can discuss that further.

akarnokd:
> I still question the subscribe/unsafeSubscribe thing above, but can discuss that further.

That was a simple oversight, it was already fixed before you merged this.

PR #2929, opened 2015-05-01
OperatorObserveOn onComplete can be emitted despite onError being called

davidmoten:
This is a fix for a race condition in `OperatorObserveOn` where if thread A gets to L164 and thread B starts the pollQueue loop then it will act as if the stream had completed normally instead of with an error. 

The effect is that a stream could appear to complete normally when in fact an error had occurred.

Using two boolean volatiles `completed` and `failed` that as a pair were not atomically updated/read exposed us to this race condition. 

The fix is to use a single volatile integer `status` to represent the  states ACTIVE, COMPLETED, ERRORED to replace `completed` and `failed`.

PR #2907, opened 2015-04-23
Fixed schedule race and task retention with ExecutorScheduler.

akarnokd:
Fixes a race condition with the timed schedule (first potentially overwriting the result of the untimed schedule in mas) and a scheduled task retention problem due to not tracking those.

PR #2930, opened 2015-05-05
OnSubscribeRedo - fix race conditions

### LINKED ISSUE #2863: Non deterministic test: OperatorRetryTest.testRetryWithBackpressureParallel
This fails for me 5 out of 10 times when I run all tests in ```OperatorRetryTest``` in Eclipse on a fast machine.

Here's the stack trace:
```
java.lang.AssertionError: Data content mismatch: {349=[beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, beginningEveryTime, onSuccessOnly]}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:752)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:45)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:15)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:42)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:20)
	at org.junit.internal.runners.statements.FailOnTimeout$StatementThread.run(FailOnTimeout.java:62)
```

davidmoten:
While searching for the cause of #2863 I bumped into this race condition (which doesn't fix #2863):

If a request is made between L238 and L239 then `consumerCapacity` may become `Long.MAX_VALUE` on arriving at L239 in which case we don't wish to decrement it. To fix, used `compareAndSet`.

What is interesting about this fix is that in the test loop of 5000 in `OperatorRetryTest` I see many more occurrences of the failure on average (3 -> 50) presumably because the extra time to perform the  `compareAndSet` action has expanded the window for the race condition causing the failures.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeRedo.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeRedo.java
@@ -35,7 +35,6 @@ import static rx.Observable.create;
 
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicReference;
 
 import rx.Notification;
 import rx.Observable;
@@ -47,13 +46,14 @@ import rx.Subscriber;
 import rx.functions.Action0;
 import rx.functions.Func1;
 import rx.functions.Func2;
+import rx.observers.Subscribers;
 import rx.schedulers.Schedulers;
-import rx.subjects.PublishSubject;
+import rx.subjects.BehaviorSubject;
 import rx.subscriptions.SerialSubscription;
 
 public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
 
-    static final Func1<Observable<? extends Notification<?>>, Observable<?>> REDO_INIFINITE = new Func1<Observable<? extends Notification<?>>, Observable<?>>() {
+    static final Func1<Observable<? extends Notification<?>>, Observable<?>> REDO_INFINITE = new Func1<Observable<? extends Notification<?>>, Observable<?>>() {
         @Override
         public Observable<?> call(Observable<? extends Notification<?>> ts) {
             return ts.map(new Func1<Notification<?>, Notification<?>>() {
@@ -120,7 +120,7 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
     }
 
     public static <T> Observable<T> retry(Observable<T> source) {
-        return retry(source, REDO_INIFINITE);
+       return retry(source, REDO_INFINITE);
     }
 
     public static <T> Observable<T> retry(Observable<T> source, final long count) {
@@ -144,7 +144,7 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
     }
 
     public static <T> Observable<T> repeat(Observable<T> source, Scheduler scheduler) {
-        return repeat(source, REDO_INIFINITE, scheduler);
+        return repeat(source, REDO_INFINITE, scheduler);
     }
 
     public static <T> Observable<T> repeat(Observable<T> source, final long count) {
@@ -172,10 +172,10 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
         return create(new OnSubscribeRedo<T>(source, notificationHandler, false, false, scheduler));
     }
 
-    private Observable<T> source;
+    private final Observable<T> source;
     private final Func1<? super Observable<? extends Notification<?>>, ? extends Observable<?>> controlHandlerFunction;
-    private boolean stopOnComplete;
-    private boolean stopOnError;
+    private final boolean stopOnComplete;
+    private final boolean stopOnError;
     private final Scheduler scheduler;
 
     private OnSubscribeRedo(Observable<T> source, Func1<? super Observable<? extends Notification<?>>, ? extends Observable<?>> f, boolean stopOnComplete, boolean stopOnError,
@@ -189,11 +189,12 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
 
     @Override
     public void call(final Subscriber<? super T> child) {
-        final AtomicBoolean isLocked = new AtomicBoolean(true);
+        
+        // when true is a marker to say we are ready to resubscribe to source
         final AtomicBoolean resumeBoundary = new AtomicBoolean(true);
+        
         // incremented when requests are made, decremented when requests are fulfilled
         final AtomicLong consumerCapacity = new AtomicLong(0l);
-        final AtomicReference<Producer> currentProducer = new AtomicReference<Producer>();
 
         final Scheduler.Worker worker = scheduler.createWorker();
         child.add(worker);
@@ -201,8 +202,18 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
         final SerialSubscription sourceSubscriptions = new SerialSubscription();
         child.add(sourceSubscriptions);
 
-        final PublishSubject<Notification<?>> terminals = PublishSubject.create();
+        // use a subject to receive terminals (onCompleted and onError signals) from 
+        // the source observable. We use a BehaviorSubject because subscribeToSource 
+        // may emit a terminal before the restarts observable (transformed terminals) 
+        // is subscribed
+        final BehaviorSubject<Notification<?>> terminals = BehaviorSubject.create();
+        final Subscriber<Notification<?>> dummySubscriber = Subscribers.empty();
+        // subscribe immediately so the last emission will be replayed to the next 
+        // subscriber (which is the one we care about)
+        terminals.subscribe(dummySubscriber);
 
+        final ProducerArbiter arbiter = new ProducerArbiter();
+        
         final Action0 subscribeToSource = new Action0() {
             @Override
             public void call() {
@@ -212,11 +223,11 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
 
                 Subscriber<T> terminalDelegatingSubscriber = new Subscriber<T>() {
                     boolean done;
+
                     @Override
                     public void onCompleted() {
                         if (!done) {
                             done = true;
-                            currentProducer.set(null);
                             unsubscribe();
                             terminals.onNext(Notification.createOnCompleted());
                         }
@@ -226,7 +237,6 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
                     public void onError(Throwable e) {
                         if (!done) {
                             done = true;
-                            currentProducer.set(null);
                             unsubscribe();
                             terminals.onNext(Notification.createOnError(e));
                         }
@@ -235,20 +245,30 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
                     @Override
                     public void onNext(T v) {
                         if (!done) {
-                            if (consumerCapacity.get() != Long.MAX_VALUE) {
-                                consumerCapacity.decrementAndGet();
-                            }
                             child.onNext(v);
+                            decrementConsumerCapacity();
+                            arbiter.produced(1);
+                        }
+                    }
+
+                    private void decrementConsumerCapacity() {
+                        // use a CAS loop because we don't want to decrement the
+                        // value if it is Long.MAX_VALUE
+                        while (true) {
+                            long cc = consumerCapacity.get();
+                            if (cc != Long.MAX_VALUE) {
+                                if (consumerCapacity.compareAndSet(cc, cc - 1)) {
+                                    break;
+                                }
+                            } else {
+                                break;
+                            }
                         }
                     }
 
                     @Override
                     public void setProducer(Producer producer) {
-                        currentProducer.set(producer);
-                        long c = consumerCapacity.get();
-                        if (c > 0) {
-                            producer.request(c);
-                        }
+                        arbiter.setProducer(producer);
                     }
                 };
                 // new subscription each time so if it unsubscribes itself it does not prevent retries
@@ -278,12 +298,11 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
 
                             @Override
                             public void onNext(Notification<?> t) {
-                                if (t.isOnCompleted() && stopOnComplete)
-                                    child.onCompleted();
-                                else if (t.isOnError() && stopOnError)
-                                    child.onError(t.getThrowable());
-                                else {
-                                    isLocked.set(false);
+                                if (t.isOnCompleted() && stopOnComplete) {
+                                    filteredTerminals.onCompleted();
+                                } else if (t.isOnError() && stopOnError) {
+                                    filteredTerminals.onError(t.getThrowable());
+                                } else {
                                     filteredTerminals.onNext(t);
                                 }
                             }
@@ -313,10 +332,15 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
 
                     @Override
                     public void onNext(Object t) {
-                        if (!isLocked.get() && !child.isUnsubscribed()) {
+                        if (!child.isUnsubscribed()) {
+                            // perform a best endeavours check on consumerCapacity 
+                            // with the intent of only resubscribing immediately 
+                            // if there is outstanding capacity
                             if (consumerCapacity.get() > 0) {
                                 worker.schedule(subscribeToSource);
                             } else {
+                                // set this to true so that on next request
+                                // subscribeToSource will be scheduled
                                 resumeBoundary.compareAndSet(false, true);
                             }
                         }
@@ -334,16 +358,180 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
 
             @Override
             public void request(final long n) {
-                long c = BackpressureUtils.getAndAddRequest(consumerCapacity, n);
-                Producer producer = currentProducer.get();
-                if (producer != null) {
-                    producer.request(n);
-                } else
-                if (c == 0 && resumeBoundary.compareAndSet(true, false)) {
-                    worker.schedule(subscribeToSource);
+                if (n > 0) {
+                    BackpressureUtils.getAndAddRequest(consumerCapacity, n);
+                    arbiter.request(n);
+                    if (resumeBoundary.compareAndSet(true, false))
+                        worker.schedule(subscribeToSource);
                 }
             }
         });
         
+    }
+    
+    /**
+     * Between when the source subscription finishes and the next subscription starts requests may arrive. 
+     * ProducerArbiter keeps track of all requests made and all arriving emissions so that when setProducer 
+     * is called for a new subscription the appropriate number of requests are made to the new producer. 
+     */
+    static final class ProducerArbiter implements Producer {
+        /** Guarded by this. */
+        boolean emitting;
+        /** The current producer. Accessed while emitting. */
+        Producer currentProducer;
+        /** The current requested count. */
+        long requested;
+        
+        long missedRequested;
+        Producer missedProducer;
+        long missedProd;
+        
+        @Override
+        public void request(long n) {
+            if (n <= 0) {
+                return;
+            }
+            Producer mp;
+            long mprod;
+            synchronized (this) {
+                if (emitting) {
+                    missedRequested += n;
+                    return;
+                }
+                emitting = true;
+                mp = missedProducer;
+                mprod = missedProd;
+                
+                missedProducer = null;
+                missedProd = 0L;
+            }
+            
+            boolean skipFinal = false;
+            try {
+                emit(n, mp, mprod);
+                drainLoop();
+                skipFinal = true;
+            } finally {
+                if (!skipFinal) {
+                    synchronized (this) {
+                        emitting = false;
+                    }
+                }
+            }
+        }
+        void setProducer(Producer p) {
+            if (p == null) {
+                throw new NullPointerException();
+            }
+            
+            long mreq;
+            long mprod;
+            synchronized (this) {
+                if (emitting) {
+                    missedProducer = p;
+                    return;
+                }
+                emitting = true;
+                mreq = missedRequested;
+                mprod = missedProd;
+                
+                missedRequested = 0L;
+                missedProd = 0L;
+            }
+            
+            boolean skipFinal = false;
+            try {
+                emit(mreq, p, mprod);
+                drainLoop();
+                skipFinal = true;
+            } finally {
+                if (!skipFinal) {
+                    synchronized (this) {
+                        emitting = false;
+                    }
+                }
+            }
+        }
+        void produced(long n) {
+            if (n <= 0) {
+                throw new IllegalArgumentException(n + " produced?!");
+            }
+            
+            long mreq;
+            Producer mp;
+            synchronized (this) {
+                if (emitting) {
+                    missedProd += n;
+                    return;
+                }
+                emitting = true;
+                mreq = missedRequested;
+                mp = missedProducer;
+                
+                missedRequested = 0L;
+                missedProducer = null;
+            }
+            
+            boolean skipFinal = false;
+            try {
+                emit(mreq, mp, n);
+                drainLoop();
+                skipFinal = true;
+            } finally {
+                if (!skipFinal) {
+                    synchronized (this) {
+                        emitting = false;
+                    }
+                }
+            }
+        }
+        void drainLoop() {
+            for (;;) {
+                long mreq;
+                long mprod;
+                Producer mp;
+                synchronized (this) {
+                    mreq = missedRequested;
+                    mprod = missedProd;
+                    mp = missedProducer;
+                    if (mreq == 0L && mp == null && mprod == 0L) {
+                        emitting = false;
+                        return;
+                    }
+                    missedRequested = 0L;
+                    missedProd = 0L;
+                    missedProducer = null;
+                }
+                emit(mreq, mp, mprod);
+            }
+        }
+        void emit(long mreq, Producer mp, long mprod) {
+            boolean newMp = false;
+            if (mp != null) {
+                newMp = true;
+                currentProducer = mp;
+            } else {
+                mp = currentProducer;
+            }
+            
+            long u = requested + mreq;
+            if (u < 0) {
+                u = Long.MAX_VALUE;
+            } else 
+            if (u != Long.MAX_VALUE) {
+                u -= mprod;
+                if (u < 0) {
+                    throw new IllegalStateException("More produced than requested");
+                }
+            }
+            requested = u;
+            
+            if (mreq > 0 && mp != null) {
+                mp.request(mreq);
+            } else
+            if (newMp && u > 0) {
+                mp.request(u);
+            }
+        }
     }
 }
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -75,15 +75,19 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
         final NotificationLite<T> on = NotificationLite.instance();
 
         final Queue<Object> queue;
-        volatile boolean completed = false;
-        volatile boolean failure = false;
+        
+        // the status of the current stream
+        volatile boolean finished = false;
 
+        @SuppressWarnings("unused")
         volatile long requested = 0;
+        
         @SuppressWarnings("rawtypes")
         static final AtomicLongFieldUpdater<ObserveOnSubscriber> REQUESTED = AtomicLongFieldUpdater.newUpdater(ObserveOnSubscriber.class, "requested");
 
         @SuppressWarnings("unused")
         volatile long counter;
+        
         @SuppressWarnings("rawtypes")
         static final AtomicLongFieldUpdater<ObserveOnSubscriber> COUNTER_UPDATER = AtomicLongFieldUpdater.newUpdater(ObserveOnSubscriber.class, "counter");
 
@@ -127,7 +131,7 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
 
         @Override
         public void onNext(final T t) {
-            if (isUnsubscribed() || completed) {
+            if (isUnsubscribed()) {
                 return;
             }
             if (!queue.offer(on.next(t))) {
@@ -139,30 +143,23 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
 
         @Override
         public void onCompleted() {
-            if (isUnsubscribed() || completed) {
+            if (isUnsubscribed() || finished) {
                 return;
             }
-            if (error != null) {
-                return;
-            }
-            completed = true;
+            finished = true;
             schedule();
         }
 
         @Override
         public void onError(final Throwable e) {
-            if (isUnsubscribed() || completed) {
-                return;
-            }
-            if (error != null) {
+            if (isUnsubscribed() || finished) {
                 return;
             }
             error = e;
             // unsubscribe eagerly since time will pass before the scheduled onError results in an unsubscribe event
             unsubscribe();
-            // mark failure so the polling thread will skip onNext still in the queue
-            completed = true;
-            failure = true;
+            finished = true;
+            // polling thread should skip any onNext still in the queue
             schedule();
         }
 
@@ -185,52 +182,42 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
         void pollQueue() {
             int emitted = 0;
             do {
-                /*
-                 * Set to 1 otherwise it could have grown very large while in the last poll loop
-                 * and then we can end up looping all those times again here before exiting even once we've drained
-                 */
                 counter = 1;
-
-//                middle:
-                while (!scheduledUnsubscribe.isUnsubscribed()) {
-                    if (failure) {
-                        child.onError(error);
-                        return;
-                    } else {
-                        if (requested == 0 && completed && queue.isEmpty()) {
+                long produced = 0;
+                long r = requested;
+                while (!child.isUnsubscribed()) {
+                    Throwable error;
+                    if (finished) {
+                        if ((error = this.error) != null) {
+                            // errors shortcut the queue so 
+                            // release the elements in the queue for gc
+                            queue.clear();
+                            child.onError(error);
+                            return;
+                        } else
+                        if (queue.isEmpty()) {
                             child.onCompleted();
                             return;
                         }
-                        if (REQUESTED.getAndDecrement(this) != 0) {
-                            Object o = queue.poll();
-                            if (o == null) {
-                                if (completed) {
-                                    if (failure) {
-                                        child.onError(error);
-                                    } else {
-                                        child.onCompleted();
-                                    }
-                                    return;
-                                }
-                                // nothing in queue
-                                REQUESTED.incrementAndGet(this);
-                                break;
-                            } else {
-                                if (!on.accept(child, o)) {
-                                    // non-terminal event so let's increment count
-                                    emitted++;
-                                }
-                            }
+                    }
+                    if (r > 0) {
+                        Object o = queue.poll();
+                        if (o != null) {
+                            child.onNext(on.getValue(o));
+                            r--;
+                            emitted++;
+                            produced++;
                         } else {
-                            // we hit the end ... so increment back to 0 again
-                            REQUESTED.incrementAndGet(this);
                             break;
                         }
+                    } else {
+                        break;
                     }
                 }
+                if (produced > 0 && requested != Long.MAX_VALUE) {
+                    REQUESTED.addAndGet(this, -produced);
+                }
             } while (COUNTER_UPDATER.decrementAndGet(this) > 0);
-
-            // request the number of items that we emitted in this poll loop
             if (emitted > 0) {
                 request(emitted);
             }
--- a/src/test/java/rx/internal/operators/OperatorRetryTest.java
+++ b/src/test/java/rx/internal/operators/OperatorRetryTest.java
@@ -395,9 +395,13 @@ public class OperatorRetryTest {
         public void call(final Subscriber<? super String> o) {
             o.setProducer(new Producer() {
                 final AtomicLong req = new AtomicLong();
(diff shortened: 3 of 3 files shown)


src/main/java/rx/internal/operators/OnSubscribeRedo.java
```diff
@@ -235,8 +235,9 @@ public void onError(Throwable e) {
                     @Override
                     public void onNext(T v) {
                         if (!done) {
-                            if (consumerCapacity.get() != Long.MAX_VALUE) {
-                                consumerCapacity.decrementAndGet();
+                            long cap = consumerCapacity.get();
```
akarnokd: Generally, one should decrement the request/capacity after the value has been emitted, because otherwise if the value becomes zero, it may open a window for a concurrent emission that is kicked off by a transition from 0 to some value. Could you see if that helps  with #2863 ?

davidmoten: Unfortunately still fails, now with less frequency in the 5000 attempts as before.

I'll adjust the PR anyway to reflect what you suggest.

davidmoten:
Awesome, for once I'm happy to see a CI failure! Tilll now #5683 hasn't been repeatable except on laptop but now has reappeared in the Travis build.

akarnokd:
If I put in this, I get the missing emission problem:
```java
long cc = consumerCapacity.get();
if (cc < Long.MAX_VALUE) {
    consumerCapacity.compareAndSet(cc, cc - 1);
}

if (rnd.nextDouble() < 0.25) {
    try {
        Thread.sleep(1);
    } catch (InterruptedException ex) {
        ex.printStackTrace();
    }
}

child.onNext(v);
```

If I swap the operations, the test passes for me again.

Could you also update the helper method in the test?

```java
static <T> StringBuilder sequenceFrequency(Iterable<T> it) {
        StringBuilder sb = new StringBuilder();
        
        Object prev = null;
        int cnt = 0;
        
        for (Object curr : it) {
            if (sb.length() > 0) {
                if (!curr.equals(prev)) {
                    if (cnt > 1) {
                        sb.append(" x ").append(cnt);
                        cnt = 1;
                    }
                    sb.append(", ");
                    sb.append(curr);
                } else {
                    cnt++;
                }
            } else {
                sb.append(curr);
                cnt++;
            }
            prev = curr;
        }
        if (cnt > 1) {
            sb.append(" x ").append(cnt);
        }
        
        return sb;
    }
```

davidmoten:
@akarnokd while you were putting your last comment up I synchronized the reads and writes of consumerCapacity as in the attached commit and the failures have stopped. Doesn't explain it yet but seems like consumerCapacity is important to this.

I'll add your changes and revert the synchronization once you've had a quick look.

davidmoten:
I thought as much given our previous conversations and figured that if this is the solution you'd have suggestions for doing it properly. Anyway probably need to figure out the why still I suppose.

src/main/java/rx/internal/operators/OnSubscribeRedo.java
```diff
@@ -235,10 +235,11 @@ public void onError(Throwable e) {
                     @Override
                     public void onNext(T v) {
                         if (!done) {
-                            if (consumerCapacity.get() != Long.MAX_VALUE) {
-                                consumerCapacity.decrementAndGet();
-                            }
                             child.onNext(v);
+                            long cap = consumerCapacity.get();
+                            if (cap != Long.MAX_VALUE) {
+                                consumerCapacity.compareAndSet(cap, cap -1);
```
akarnokd: The problem with this is that if the CAS fails, the 'production' of the value is not accounted and the capacity seems to rise without limit. The options are 1) add a loop that makes sure the decrement actually happens or 2) switch back to decrementAndGet and don't worry about request getting to max concurrently.

davidmoten:
I'm confused by this `resumeBoundary` business, I hope you fare better, off to bed for me.

src/main/java/rx/internal/operators/OnSubscribeRedo.java
davidmoten: yep of course it would need a loop if we went for the CAS option. I can chuck one in, your call.

davidmoten:
I've committed the CAS loop. We can drop it as an optimization later based on perfs if we want or of course now if you think

akarnokd:
Can you devise a benchmark that checks the throughput on a non-throwing source?

If I remember correctly, the resumeBoundary is there because an exception can happen just after the capacity reached zero but we don't immediately resubscribe until an actual request comes in.

davidmoten:
I've added some more comments to the code as my understanding improves. I also have marked three places with //TODO, two of which I think are potential race conditions and the third is just a marker for some of the usual optimizations done in the `request`. 

Another addition in the last commit is a check to ensure that the `consumerCapacity` decrement doesn't go below zero which now provokes a `MissingBackpressureException`.

I'm sure @akarnokd will have the red pen out on this stuff then if any of the race conditions stand scrutiny I'll fix them.

davidmoten:
unrelated test failure `OperatorPublishTest.testObserveOn`. I've raised #2933.

davidmoten:
I've ruled out one race condition (and left comments explaining why all is ok)

akarnokd:
I'd say if the change makes your test failure go away, that should be enough for now. I'd defer the full rewrite to 2.0.

davidmoten:
Righto, I'll concentrate on the fix for the #2863 test failure.  Part of that is probably ruling out surprises like decrementing `consumerCapacity` when it is already 0. My approach in that was to throw a `MissingBackpressureException` because in backpressure mode an operator should not emit more than requested. I would be concerned that this might appear as a breaking change to some though because previously it happily passed them through (with who knows what consequences in the operator). A compromise might be to not decrement when 0 but also not emit an error and then revisit this part of the contract in 2.0. Which way do you think we should go?

davidmoten:
Hmm I just realized that if not emitting MissingBackpressureException then I should decrement otherwise the accounting is stuffed when more requests come through.

davidmoten:
Sorry to say, no progress made. I've protected all reads and writes to `currentProducer` and `consumerCapacity` by a single guard object (without holding a lock on emission or on request or on `worker.schedule` which could be run synchronously using `Schedulers.immediate()` or `Schedulers.trampoline()`) and still get the `OperatorRetryTest.testRetryWithBackpressureParallel` failure on my laptop. I give up for the moment. I'm leaning towards there being something fundamentally wrong that will probably only be fixed by a rewrite.

davidmoten:
I've pared this PR right down (after the unsuccessful hunt for the cause of #2863) so that it addresses just the race conditions that I can see and can confirm.

The changes are:

* The original code to decrement `consumerCapacity` had a race condition, now we use a CAS loop to update
* the `c == 0` check at L342 can cause a stall in the stream because if `request` gets called just before [L320](https://github.com/ReactiveX/RxJava/blob/0a6e26de73c3707dbd9903c166b7be0f6499b6b7/src/main/java/rx/internal/operators/OnSubscribeRedo.java#L320) then no restart will occur in that call and it is possible that no further calls to request will be made after L320 has been performed (a common pattern is a request followed by another request only once emissions have occurred).

#2683 is NOT fixed by these changes.

akarnokd:
I applied some random sleep around the resumeBoundary and these changes worked for me (on top of your changes):
In the method with the locked check:
```java
if (!isLocked.get() && !child.isUnsubscribed()) {
    if (consumerCapacity.get() > 0) {
        worker.schedule(subscribeToSource);
    }
}
```
and in the child's producer:
```java
long c = BackpressureUtils.getAndAddRequest(consumerCapacity, n);
Producer producer = currentProducer.get();
if (producer != null) {
    producer.request(n);
} else
if (c == 0) {
    worker.schedule(subscribeToSource);
}
```

So if there is capacity available, the source will be restarted. If there is no capacity remaining, the first 0 -> n transition will restart the source.

davidmoten:
I'll try it out, thanks! I'll also include a `n>0` check in the producer so that calls of request(0) are no-ops.

davidmoten:
Nice simplification. Still getting `testRetryWithBackpressureParallel` failure on my laptop (first time).

davidmoten:
3 out of 3 failures on laptop, 1 out of 25 failures on fast desktop running all tests in `OperatorRetryTest`. I just run this in a loop:

`./gradlew -q   -Dtest.single=OperatorRetry cleanTest test`

akarnokd:
Apparently, the ```resumeBoundary``` served to prevent unwanted resubscription.
One final try:
wrap the above places into ```synchronized (consumerCapacity)``` plus the contents of ```terminalDelegatingSubscriber.setProducer``` and see if it still fails. If not, start moving out operations from this synchronized block and see which one triggers the failure.

davidmoten:
This is as far as I got. No failures when everything in the synch blocks.  Then started moving out the baddies for synchronization being calls to `producer.request` and `worker.schedule` (because could be a synchronous worker). Everything ok till I pulled `worker.schedule` out of `onNext` synch block. 

```java
                    @Override
                    public void onNext(Object t) {
                        if (!isLocked.get() && !child.isUnsubscribed()) {
                            final boolean scheduleNow;
                            synchronized (consumerCapacity) {
                                if (consumerCapacity.get() > 0) {
                                    scheduleNow = true;
                                } else {
                                    scheduleNow = false;
                                    resumeBoundary.compareAndSet(false, true);
                                }
                            }
                            if (scheduleNow)
                                worker.schedule(subscribeToSource);
                        }
                    }

                    @Override
                    public void setProducer(Producer producer) {
                        producer.request(Long.MAX_VALUE);
                    }
                });
            }
        });

        child.setProducer(new Producer() {

            @Override
            public void request(final long n) {
                final Producer producer;
                final boolean requestNow;
                final boolean scheduleNow;
                synchronized (consumerCapacity) {
                    BackpressureUtils.getAndAddRequest(consumerCapacity, n);
                    producer = currentProducer.get();
                    if (producer != null) {
                        requestNow = true;
                        scheduleNow = false;
                    } else {
                        requestNow = false;
                        scheduleNow = resumeBoundary.compareAndSet(true, false);
                    }
                }
                if (requestNow)
                    producer.request(n);
                else if (scheduleNow)
                    worker.schedule(subscribeToSource);
            }
        })
```

akarnokd:
When the test fails, how many elements are missing from the output. Does it print beginningEveryTime x 256 ?

davidmoten:
Yep:

`java.lang.AssertionError: Data content mismatch: 2315={beginningEveryTime x 256}
        at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:763)
`

akarnokd:
I'm rewriting just the retry(n) variant and see if there is a general logical error with request accounting or just the OnSubscribeRedo has problems.

akarnokd:
[Here is a retry implementation](https://gist.github.com/akarnokd/3b110055414a44d320b9) that serializes out the changing of producers and accounting the capacity. Could you try this implementation?
(Dont forget to change both factory methods in OnSubscribeRedo:
```java
    public static <T> Observable<T> retry(Observable<T> source) {
//        return retry(source, REDO_INIFINITE);
        return retry(source, Long.MAX_VALUE);
    }

    public static <T> Observable<T> retry(Observable<T> source, final long count) {
        if (count < 0)
            throw new IllegalArgumentException("count >= 0 expected");
        if (count == 0)
            return source;
//        return retry(source, new RedoFinite(count));
        return create(new OnSubscribeRetry<T>(source, count));
    }
```

davidmoten:
Cool, thanks. Got this exception:
```
rx.internal.operators.OperatorRetryWithPredicateTest > testJustAndRetry FAILED
    java.util.NoSuchElementException: Sequence contains no elements
        at rx.internal.operators.OperatorSingle$ParentSubscriber.onCompleted(OperatorSingle.java:131)
        at rx.internal.operators.OnSubscribeRetry$1$1.onCompleted(OnSubscribeRetry.java:66)
        at rx.internal.operators.OperatorMap$1.onCompleted(OperatorMap.java:44)
        at rx.internal.util.ScalarSynchronousObservable$1.call(ScalarSynchronousObservable.java:44)
        at rx.internal.util.ScalarSynchronousObservable$1.call(ScalarSynchronousObservable.java:32)
        at rx.Observable$1.call(Observable.java:144)
        at rx.Observable$1.call(Observable.java:136)
        at rx.Observable.unsafeSubscribe(Observable.java:7507)
        at rx.internal.operators.OnSubscribeRetry$1.call(OnSubscribeRetry.java:75)
        at rx.schedulers.TrampolineScheduler$InnerCurrentThreadScheduler.enqueue(TrampolineScheduler.java:80)
        at rx.schedulers.TrampolineScheduler$InnerCurrentThreadScheduler.schedule(TrampolineScheduler.java:59)
        at rx.internal.operators.OnSubscribeRetry.call(OnSubscribeRetry.java:86)
        at rx.internal.operators.OnSubscribeRetry.call(OnSubscribeRetry.java:28)
        at rx.Observable$1.call(Observable.java:144)
        at rx.Observable$1.call(Observable.java:136)
        at rx.Observable.subscribe(Observable.java:7597)
        at rx.observables.BlockingObservable.blockForSingle(BlockingObservable.java:442)
        at rx.observables.BlockingObservable.single(BlockingObservable.java:341)
        at rx.internal.operators.OperatorRetryWithPredicateTest.testJustAndRetry(OperatorRetryWithPredicateTest.java:296)
```

akarnokd:
I went a bit light on the safeguards in OnSubscribeRetry (i.e., not bouncing back any exceptions thrown when calling currentProducer.request()). How does the code work with testRetryWithBackpressureParallel?

akarnokd:
I've [updated the gist](https://gist.github.com/akarnokd/5cb99c6062cea0b23d8b) so it passes the unit tests.

davidmoten:
Good so far. I'll stick the test in a loop for ten minutes or so on the laptop.

davidmoten:
3rd time through:
```
rx.internal.operators.OperatorRetryTest > testRetryWithBackpressureParallel FAILED
    java.lang.AssertionError: Data content mismatch: 3628={beginningEveryTime x 256}
        at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:763)
```

davidmoten:
For your info I tried using `SynchronizedQueue` instead of `SpscArrayQueue` in `OperatorObserveOn` and got the failure on 30th attempt:

```
java.lang.AssertionError: Data content mismatch: 1940={beginningEveryTime x 254}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:763)
```

akarnokd:
There is a race in observeOn in main being fixed by #2929 that might attribute to lost onNexts but can't be sure from the AssertionError.
The simple fix for observeOn is to replace
```java
                            Object o = queue.poll();
                            if (o == null) {
                                if (completed) {
```
with
```java
boolean c = completed;
Object o = queue.poll();
if (o == null) {
    if (c) {
```

This updated test method should print out any additional onCompleted events if they are there:
```java
@Test//(timeout = 15000)
public void testRetryWithBackpressureParallel() throws InterruptedException {
    final int NUM_RETRIES = RxRingBuffer.SIZE * 2;
    int ncpu = Runtime.getRuntime().availableProcessors();
    ExecutorService exec = Executors.newFixedThreadPool(Math.max(ncpu / 2, 2));
    try {
        for (int r = 0; r < 1000; r++) {
            if (r % 10 == 0) {
                System.out.println("testRetryWithBackpressureParallelLoop -> " + r);
            }
        
            final AtomicInteger timeouts = new AtomicInteger();
            final Map<Integer, List<String>> data = new ConcurrentHashMap<Integer, List<String>>();
            
            int m = 5000;
            final CountDownLatch cdl = new CountDownLatch(m);
            for (int i = 0; i < m; i++) {
                final int j = i;
                exec.execute(new Runnable() {
                    @Override
                    public void run() {
                        final AtomicInteger nexts = new AtomicInteger();
                        try {
                            Observable<String> origin = Observable.create(new FuncWithErrors(NUM_RETRIES));
                            TestSubscriber<String> ts = new TestSubscriber<String>();
                            origin.retry()
                            .observeOn(Schedulers.computation()).unsafeSubscribe(ts);
                            ts.awaitTerminalEvent(2500, TimeUnit.MILLISECONDS);
                            List<String> onNextEvents = new ArrayList<String>(ts.getOnNextEvents());
                            if (onNextEvents.size() != NUM_RETRIES + 2) {
                                for (Throwable t : ts.getOnErrorEvents()) {
                                    onNextEvents.add(t.toString());
                                }
                                for (Object o : ts.getOnCompletedEvents()) {
                                    onNextEvents.add("onCompleted");
                                }
                                data.put(j, onNextEvents);
                            }
                        } catch (Throwable t) {
                            timeouts.incrementAndGet();
                            System.out.println(j + " | " + cdl.getCount() + " !!! " + nexts.get());
                        }
                        cdl.countDown();
                    }
                });
            }
            cdl.await();
            assertEquals(0, timeouts.get());
            if (data.size() > 0) {
                fail("Data content mismatch: " + allSequenceFrequency(data));
            }
        }
    } finally {
        exec.shutdown();
    }
}
```

davidmoten:
I've added the suggested changes to this commit but should have mentioned that I tried the fixes from #2929 with your OperatorRetry and had no effect (still get failures).

akarnokd:
What does the failure report as the received events? Could you also check whether replacing the TrampolineScheduler with io scheduler still exhibits the failure?

davidmoten:
29th run on desktop:
```
java.lang.AssertionError: Data content mismatch: 1828={beginningEveryTime x 109}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:759)
```

akarnokd:
I'm running the test on a virtualized ubuntu without CPU virtualization support to emulate a slow machine but it is working so far. What are the specs of your machine which fails the test so often?

davidmoten:
My laptop is Ubuntu 12.04 64-bit i5-430m 1.2GHz java 1.8u45 but the wind has changed this evening and is not failing that test. My desktop Ubuntu 14.04 64-bit Intel® Core™ i7 CPU 920 @ 2.67GHz × 8, java 1.8u40  is where it's failing at the moment.

davidmoten:
The quick fix for observeOn that you mention above still has the race condition that #2929 was reported on so I might paste in #2929 version of OperatorObserveOn and continue testing with that.

davidmoten:
Might #2907 have any bearing on this?

akarnokd:
No, because it is a completely separate scheduler.

I'm running the test on some real debian linux and I the test fails within the first run:
```
4882={}, 4871={}, 4860={}, 4838={beginningEveryTime x 51}, 4849={}
```

davidmoten:
David I don't understand the quick fix in OperatorObserveOn you mentioned above. Is that something new to be added to #2929?  Can you explain it if so?

akarnokd:
Never mind, I didn't notice the queue.isEmpty() check after it.

akarnokd:
I've put in doOnNext() between the operators. In case of a test failure, the emission count after the source and retry() match but are less after observeOn() which makes me think observeOn doesn't pick up the values emitted (and it doesn't request for replacements either).

davidmoten:
The latest commit including #2929 doesn't fail for me on three different machines including the old laptop. 200 runs on i7, 100 runs on laptop and 200 runs on another i7. Did you use the last commit for the failures on the debian machine?

akarnokd:
With #2929 the test fails after 15 runs:

Data: 
  - 4875={Loops:128, Request:0, B/A:[128, 128, 0, 637676, 0]}, 4842={Loops:128, Request:0, B/A:[128, 128, 0, 664205, 0]}, 
  - 4831={beginningEveryTime x 139, Loops:258, Request:10, B/A:[258, 258, 139, 2491693, 1720371]}, 
  - 4853={Loops:128, Request:0, B/A:[128, 128, 0, 670982, 0]}, 4864={Loops:128, Request:0, B/A:[128, 128, 0, 664584, 0]}

Where B/A means [after source count, after retry count, after observeOn count, latest emission after retry in nanoseconds, latest emission after observeOn in nanoseconds]. Both time values are relative to when the entire stream started.

I've completely rewritten the emission loop in observeOn but the test still fails on my linux machine.

akarnokd:
I've increased the timeout and added ```timeInterval``` to the source data and it seems there is a large, seconds-scale hiccup in the stream. Since I'm using SSH, I can't use flight recorder or jvisualvm to see how the threads behave.

davidmoten:
I've lost the ability to induce failures. Tests ran all night without inducing failure. I might try reverting to an earlier version of the test method.

davidmoten:
May as well give the latest commit from here a spin again @akarnokd now that it incorporates the request overflow fix from #2929 I just added. I still can't induce failure so the baton is passed to you!

akarnokd:
So the observeOn changes + my ProducerArbiter works. Great. Now it would be better to remove OnSubscribeRetry and apply its logic on OnSubscribeRedo directly.

davidmoten:
Super news.  So you mean you'll rewrite Redo now? One thing I was worried
about with the original is the possibility of requests getting missed and
the new one accounts for that.

akarnokd:
I meant you apply the ProducerArbiter and the calls to OnSubscribeRedo.

davidmoten:
No worries, can do

davidmoten:
Hmm not quite out of the woods. I've adjusted the code to use `ProducerArbiter` (please look it over) but saw our old friend on  the first run on my i7 desktop which also seemed to lock the machine up for me for a minute. Subsequent runs did not fail (13 runs). The latest commit includes the latest version of `OperatorObserveOn` from #2929.

```
java.lang.AssertionError: Data content mismatch: 1763={beginningEveryTime x 128, onCompleted}, 1767={beginningEveryTime x 21, onCompleted}, 1756={beginningEveryTime x 10, onCompleted}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:759)
```

davidmoten:
I forgot to mention that the last failure also included a timeout failure of `OperatorRetryTest.OperatorRetryTest. testIssue1900`.

akarnokd:
Can you monitor the test with JVisualVM and see if there are thread stalls or excess GC?

davidmoten:
Yeah I'll try that. 28th run on laptop gave 
```
rx.internal.operators.OperatorRetryTest > testRetryWithBackpressureParallel FAILED
    java.lang.AssertionError: Data content mismatch: 3724={onCompleted}, 3720={beginningEveryTime x 17, onCompleted}
        at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:759)
```

davidmoten:
I can see another race condition in `OnSubscribeRedo`. `restarts` must be subscribed to before the first request subscribes to source. I don't think it affects our current test because the `worker` is synchronous but if it was async we would be exposed to it. To fix this in a non-blocking way would require `terminals` to be a slightly different beast than a `PublishSubject` in that it would buffer events till the first subscriber (there should be only one subscriber for `terminals` though the `notificationHandler` could do something weird in theory) and then play all the buffered events to the subscriber. `ReplaySubject` might help out here but I'm not sure what idiom we could use yet. A custom `Subject` would probably be straightforward to build but I'd like to see what ideas you have @akarnokd.

I tried a blocking fix for this using a latch just to play with it but as expected still our normal failure.

Incidentally I can't seem to break `testRetryWithBackpressureParallel` but just looping it in Eclipse. It only fails looping the gradle command line call. This makes monitoring difficult but I could enable gc logging I suppose.

davidmoten:
I think I have addressed the race condition mentioned in the previous comment. By using a `BehaviorSubject` instead of a `PublishSubject` and subscribing with a dummy subscriber immediately after creating then an emission between subscribing to source and subscribing to `restarts` will be replayed on subscribe to `restarts` so it is not lost.

davidmoten:
Failed on run 111 on desktop i7:

```
java.lang.AssertionError: Data content mismatch: 237={beginningEveryTime x 128}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:759)
```

davidmoten:
The race condition I thought had found was what `isLocked` was to prevent. I'm not sure that ignoring a restart instruction because `isLocked` is false is the way to go so I'm going to remove the `isLocked` checks and retest using only the `BehaviorSubject` logic.

davidmoten:
Three hours without failure on three machines (i7,i7 and i5 laptop). Can you try this commit @akarnokd?

davidmoten:
Took 5 hours but got failure on laptop only:
```
java.lang.AssertionError: Data content mismatch: 883={beginningEveryTime x 138}
        at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRet
```

This is going to make testing difficult!

src/main/java/rx/internal/operators/OnSubscribeRedo.java
```diff
@@ -334,16 +346,173 @@ public void setProducer(Producer producer) {
...
+                currentProducer = mp;
+            } else {
+                mp = currentProducer;
+            }
+            
+            long u = requested + mreq;
+            if (u < 0) {
+                u = Long.MAX_VALUE;
+            }
+            u -= mprod;
```
davidmoten: What about not doing this subtraction if u is Long.MAX_VALUE?

davidmoten:
Using jacoco as per #2937 I noticed that the subscriber to `restarts` never completes so never unsubscribes (though `child` does). To achieve this  (and to simplify `OnSubscribeRedo` a little bit) I ensured that all events are routed through to `filteredTerminals` instead of just the `onNext` events.

src/main/java/rx/internal/operators/OnSubscribeRedo.java
akarnokd: Yes. Without it, the retry may land on MAX_VALUE - 1 on restart and not take the fast-path in certain sources.

davidmoten:
I've added the fix to `ProducerArbiter` to not reduce the request if already `Long.MAX_VALUE`. Before applying this fix I ran the test which failed on run 207 but in this test I confirmed that count upstream of  `observeOn` was same as downstream so I think the problem is not with `OperatorSubscribeOn` anymore. I'll rerun now we have a fix in `ProducerArbiter`.

davidmoten:
The fix to `ProducerArbiter` could help the failure because concurrent requests to `FuncWithErrors` one of `Long.MAX_VALUE` and one of less than `Long.MAX_VALUE` could initiate the fast path and the backpressure path simultaneously.

davidmoten:
Added more concurrency protection to `FuncWithErrors` to prevent fast path and backpressure path from running concurrently

davidmoten:
Got this 74th run on laptop:
```
org.mockito.exceptions.verification.VerificationInOrderFailure: 
Verification in order failure:
observer.onNext("beginningEveryTime");
Wanted 257 times:
-> at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressure(OperatorRetryTest.java:702)
But was 240 times:
-> at rx.observers.TestObserver.onNext(TestObserver.java:78)

	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressure(OperatorRetryTest.java:702)
```

davidmoten:
Added another race condition fix for `FuncWithErrors` that prevents the subscriber `o` being passed more signals after a terminal signal has been sent to it.

davidmoten:
54th run on desktop:
```
java.lang.AssertionError: Data content mismatch: 2408={beginningEveryTime x 128}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:764)
```

davidmoten:
despite last commit, 22nd run on desktop:
```
java.lang.AssertionError: Data content mismatch: 4501={beginningEveryTime x 128}
	at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:768)
```

akarnokd:
Try disabling transparent hugepages.

davidmoten:
With the purpose of making the test fail more often?

akarnokd:
In the contrary. Linux transparent hugepages (default: on) is known to introduce latency spikes, maybe so extreme our test times out just by that.

davidmoten:
The catch is that we have a low probability (for the test as it stands) race condition somewhere. If I make the test fail less often then we lose the ability to find that race condition. Which timeout are you thinking  of? The individual timeouts for the tests are commented out, are there scheduler timeouts?

akarnokd:
I think there is no race condition but your linux system gets overwhelmed by the test and it simply times out. This is why I suggested looking at JVisualVM's thread graph to detect some gaps in the execution that can be attributed to the system.

davidmoten:
Inspecting with jvisualvm is  tricky because I have only got failures to happen using the gradle command line which means jvm restarts every thirty seconds or so.  I've dug around pretty much everywhere now and may dig around a bit more in the future but perhaps we have achieved enough with this operator for it to be accepted back into the code base. The hunt certainly unearthed a few race conditions in `retry`, its tests, and in `observeOn` so was pretty useful and the original very frequent failure (for me) is now very infrequent.

I can rebase the commits and remove the updates to OperatorObserveOn from this commit so that it's ready for merge once you and Ben and whoever else has reviewed it. It will be good to get the fixes in #2929 merged as well of course as otherwise there will be some test flakiness on this commit. 

Is that a reasonable plan or would you like to keep digging?

davidmoten:
I've rebased commits. Ready for a hopefully final review.

akarnokd:
I'm quite a bit lost; don't know what worked and what not or did the changes work on your system or not. If the test fails with the outer pool of 1 thread, which should pose the least amount of work, then my suspect is that there is something wrong with Java on linux or with linux itself. Perhaps this is just another case of the [recent futex bug](https://groups.google.com/forum/#!topic/mechanical-sympathy/QbmpZxp6C64).

davidmoten:
The futex stuff is interesting, I'll have a close look at that soon. One thing I'm wondering is should we break the retries when a `java.lang.Error` is signalled? At the moment I assume if one occurs it will be suppressed and may have the side effects we see.

davidmoten:
I did a little test and see that the sort of `Error`s that we care about do stop the retries, so that's not it.

The futex stuff shouldn't be an issue because it turned up in linux kernel 3.14 and I'm running 3.2.0-80 on my laptop and 3.13.0-49 on my desktop. I'll turn off transparent hugepages on my home desktop and see if the error happens.

davidmoten:
Turned off transparent hugepages and failed on 18th run of this command:

```
./gradlew -i -Dtest.single=OperatorRetry cleanTest test
```

The run times are below. You'll notice that the runtime is not significantly larger for the final failing test which may rule out some types of OS behaviour being involved. 

I think we can rule out the futex bug because I run either Ubuntu 12.04 or 14.04:

>Ubuntu 14.04/Debian 8: have the fix for a long time [0] [1]
Ubuntu 12.04/Debian 7: was never affected [3] [2]. Newer enablement stack kernels for Ubuntu has the same fix as [1].
[0] http://kernel.ubuntu.com/git/ubuntu/linux.git/log/?showmsg=1&qt=grep&q=Avoid+taking+the+hb-%3Elock&h=linux-3.13.y
[1] http://kernel.ubuntu.com/git/ubuntu/linux.git/log/?showmsg=1&qt=grep&q=Avoid+taking+the+hb-%3Elock&h=linux-3.16.y
[2] https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/kernel/futex.c?h=linux-3.2.y&id=refs/tags/v3.2.69
[3] http://kernel.ubuntu.com/git/ubuntu/ubuntu-precise.git/tree/kernel/futex.c#n186

```
Total time: 33.624 secs
Total time: 25.215 secs
Total time: 25.204 secs
Total time: 27.108 secs
Total time: 25.532 secs
Total time: 27.042 secs
Total time: 26.061 secs
Total time: 26.199 secs
Total time: 26.177 secs
Total time: 26.294 secs
Total time: 25.974 secs
Total time: 26.123 secs
Total time: 25.983 secs
Total time: 26.475 secs
Total time: 26.321 secs
Total time: 26.268 secs
Total time: 26.459 secs
Total time: 27.076 secs
```

```
rx.internal.operators.OperatorRetryTest > testRetryWithBackpressureParallel FAILED
    java.lang.AssertionError: Data content mismatch: 118={beginningEveryTime x 128}
        at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:768)
```

akarnokd:
I suggest closing this PR and start with a fresh new PR and perhaps a new discussion.

davidmoten:
Continuing issue in #2997.

Issue #1483, opened 2014-07-23
Backpressure: cache


benjchristensen:
The `cache` operator should support downstream backpressure. It should `request(Long.MAX_VALUE)` up but be capable of correctly responding to downstream (similar to `from(Iterable)`).

This will mean we either have to stop using `ReplaySubject` or be able to handle it with `ReplaySubject` in the mix.

The idea here is that once data is cached, it should behave as an `Iterable`. Therefore, the cache really should behave like `onBackpressureBuffer`.

akarnokd:
See #2969

akarnokd:
Cache officially supports backpressure from relelase 1.0.14.

PR #2935, opened 2015-05-06
CachedObservable that supports backpressure and disconnection.

akarnokd:
I've written a ```cache()``` alternative that supports backpressure on its client side and can be disconnected from the upstream.

I've omitted the Experimental annotation for now because I'd like to have some feedback:

  - Should this replace ```cache()``` completely? Note that this implies changing the return type of ```cache()```.
  - Are the method names satisfactory?
  - Should I add state-peeking methods similar to ReplaySubject?

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -3481,6 +3481,47 @@ public class Observable<T> {
     }
 
     /**
+     * Caches the emission of this Observable and replays them to any subscribers.
+     * The returned CachedObservable auto-subscribes to this Observable on the first subscriber and
+     * it allows disconnecting the cache by either triggering an completion or error.
+     * <p>
+     * Note that the underlying buffer is expanded incrementally by 16 elements at once.
+     * <dl>
+     *  <dt><b>Backpressure Support:</b></dt>
+     *  <dd>The operator ignores upstream backpressure support and requests everything but honors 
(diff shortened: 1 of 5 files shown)


benjchristensen:
Related to https://github.com/ReactiveX/RxJava/issues/1483

benjchristensen:
I don't understand why it needs a new `CachedObservable` type. A normal `Observable` type can be returned and support backpressure. I don't see any methods on `CachedObservable` that are required for `cache()` to work correctly. What am I missing?

akarnokd:
The ```cache()``` operators docs say that you have to trade off the ability to unsubscribe, but you don't have to if the ```Observable``` returned can be used for this. Same thing happens in ```StoppableObservable``` of RxJavaAsyncUtils.

benjchristensen:
> trade off the ability to unsubscribe

That seems to break the semantics of what caching is doing. Caching is done for multicasting/replaying. Why should one subscriber unsubscribing early affect a second subscriber so they don't get all the data?

If more control is wanted that can be achieved with `replay` and its many overloads, particularly since it already returns `ConnectableObservable` which offers that ability. 

I agree with the need for the cache to be drained with backpressure support, but I don't think the semantics should change for unsubscribe behavior.

akarnokd:
I'll update the code that cache() will have the same capabilities as before plus backpressure.

PR #2936, opened 2015-05-06
Fix TestSubject bug

jacek-marchwicki:
## Info 

Problem is in TestSubject.onNext()
when is called on current time that is not 0 it will be delayed by current time.
calling TestSubject.onNext() should execute immediately.

## Tests
* TestSubjectTest.testObserverPropagateValueAfterTriggeringActions - represent already working behavior

* TestSubjectTest.testObserverPropagateValueInFutureTimeAfterTriggeringActions - represent a bug that is fixed by this commit

### DIFF
--- a/src/main/java/rx/subjects/TestSubject.java
+++ b/src/main/java/rx/subjects/TestSubject.java
@@ -68,11 +68,11 @@ public final class TestSubject<T> extends Subject<T, T> {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


akarnokd:
The same error is present on ```onError()``` and ```onCompleted()```. Could you fix those as well?

jacek-marchwicki:
@akarnokd  Yep... hold a second

akarnokd:
Looks good. Thanks!

PR #2937, opened 2015-05-07
Add code quality plugins to build.gradle (jacoco, findbugs)

davidmoten:
This is one thing I've missed for a while so I figured out how to get gradle to provide them. This PR enables production of Jacoco coverage reports and Findbugs reports as html in the build directory. It doesn't happen by default, the plugins are run like this (instructions in comments in `build.gradle`):

To run Findbugs:
```./gradlew check```

To run Jacoco test coverage:
```./gradlew test jacocoTestReport```

To run Jacoco on a single test:
```./gradlew -Dtest.single=OperatorRetry test jacocoTestReport```

Then open up the reports in `build/reports`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,10 +7,47 @@ description = 'RxJava: Reactive Extensions for the JVM – a library for composi
 
 apply plugin: 'rxjava-project'
 apply plugin: 'java'
+apply plugin: 'findbugs'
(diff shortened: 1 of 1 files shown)


benjchristensen:
Looks like reasonable additions, and safe since they don't affect normal builds.

Thanks.

PR #2938, opened 2015-05-07
Fixed Observable.combineLatest overflow bug on Android

jacek-marchwicki:
RxRingBuffer size is not a constant and on Android is
less then 128 (16). So it causing silent issues when
there were given 16 < Observers < 128.


akarnokd:
Thanks!

PR #2949, opened 2015-05-14
Remove findbugs+jacoco from build.gradle

davidmoten:
Apologies for the PR that added these to build.gradle. My claim that they were not run by default was wrong (`./gradlew build` runs them). Let's remove till I figure out how to do this properly.


benjchristensen:
:-)

PR #2950, opened 2015-05-15
OperatorGroupBy - check for request overflow and don't decrement when at Long.MAX_VALUE

davidmoten:
Added request overflow check to `OperatorGroupBy` (using `BackpressureUtils`) and prevented `requested` counters from being decremented if they were at `Long.MAX_VALUE`.

Includes unit test that failed on current codebase.


akarnokd:
Thanks!

PR #2951, opened 2015-05-15
OperatorConcat - prevent request overflow and fix race condition

davidmoten:
Used `BackpressureUtils` to prevent request overflow and added an `n > 0` check for requests to ensure that `subscribeNext` is not called concurrently (race condition).

No unit tests written. Let me know if you think worth the effort.


akarnokd:
Could you add your usual unit test for this?

davidmoten:
Sure

davidmoten:
Done, added unit test that failed on original codebase for request overflow. Didn't add a unit test for `n>0` check though (want that too?).

akarnokd:
Not necessary.

akarnokd:
Thanks!

PR #2953, opened 2015-05-15
OperatorObserveOn - reduce object allocations 

davidmoten:
Instead of using a queue of `Notification<T>` we can use a queue of `T` and when `t` is null place a singleton `NULL_SENTINEL` object on the queue (which we get away with because of type erasure).

This means we don't need to wrap every `onNext` arrival as a `Notification` and thus reduce object allocations.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -69,17 +69,18 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
 
     /** Observe through individual queue per observer. */
     private static final class ObserveOnSubscriber<T> extends Subscriber<T> {
+        
+        //represents null in the queue
+        private static final Object NULL_SENTINEL = new Object();
+        
         final Subscriber<? super T> child;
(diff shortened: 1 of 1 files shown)


akarnokd:
I'm a bit surprised by this PR.

  - Saving the size of one reference of the static-valued ```NotificationLite.instace()``` is not much of a gain and you can't be sure how each JVM will pack the particular ```ObserveOnSubscriber``` class.
  - Casting an exaclty Object NULL_SENTINEL reference to T looks fishy.
  - I don't see any performance benefit either because the ```NotificationLite.getValue()``` and ```next()``` does exactly the same.

davidmoten:
Ah that's hilarious, I completely forgot that `NotificationLite` already does this stuff! And no I don't care about the `on` reference. The cast works because T is a generic type and erasure takes care of it. I agree no performance benefit, I'll close.

akarnokd:
> The cast works because T is a generic type and erasure takes care of it.

Yes, true, but I find the pattern itself dangerous: an accidental leak of the Object and now you have ClassCastException somewhere else.

davidmoten:
Yep, makes sense.

PR #2956, opened 2015-05-16
OperatorObserveOn should not request more after child is unsubscribed

davidmoten:
In existing code if unsubscription happens immediately after a `child.onNext` call from `pollQueue` then the number of emissions made in the loop will be requested of upstream despite `child` being unsubscribed.

This PR adds a local variable `isUnsubscribed` and when set to true no further requests will be made.

This PR includes a unit test that failed on the existing codebase.


akarnokd:
Less verbose:
```java
for (;;) {
    if (child.isUnsubscribed()) {
        return;
    }
    // ...
```

davidmoten:
Yeah, that's better. Done

akarnokd:
Thanks!

PR #2961, opened 2015-05-19
fix Amb backpressure bug

davidmoten:
The Amb operator has a backpressure bug:

* if `m` requests were made before the first emission then `m` subscriptions were started on each source observable. We only want once subscription on each.

This PR adds fix code to  `OnSubscribeAmb` and two unit tests.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeAmb.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeAmb.java
@@ -269,6 +269,7 @@ public final class OnSubscribeAmb<T> implements OnSubscribe<T>{
 
         private final Subscriber<? super T> subscriber;
         private final Selection<T> selection;
+        private boolean chosen;
 
         private AmbSubscriber(long requested, Subscriber<? super T> subscriber, Selection<T> selection) {
             this.subscriber = subscriber;
@@ -282,11 +283,11 @@ public final class OnSubscribeAmb<T> implements OnSubscribe<T>{
         }
 
         @Override
-        public void onNext(T args) {
+        public void onNext(T t) {
             if (!isSelected()) {
                 return;
             }
-            subscriber.onNext(args);
+            subscriber.onNext(t);
         }
 
         @Override
@@ -306,12 +307,17 @@ public final class OnSubscribeAmb<T> implements OnSubscribe<T>{
         }
 
         private boolean isSelected() {
+            if (chosen) {
+                return true;
+            }
             if (selection.choice.get() == this) {
                 // fast-path
+                chosen = true;
                 return true;
             } else {
                 if (selection.choice.compareAndSet(null, this)) {
                     selection.unsubscribeOthers(this);
+                    chosen = true;
                     return true;
                 } else {
                     // we lost so unsubscribe ... and force cleanup again due to possible race conditions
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -373,26 +374,46 @@ public void call() {
...
                             selection.unsubscribeOthers(selection.choice.get());
-                            break;
+                            return;
                         }
                         source.unsafeSubscribe(ambSubscriber);
                     }
+                } else {
+                    //subscriptions already happened so propagate the request to all the  
+                    //amb subscribers
+                    for (AmbSubscriber<T> ambSubscriber: selection.ambSubscribers) {
```
akarnokd: This may still run concurrently with the subscription loop and some subscribers won't get the newer requests during that time. The simplest fix is to subscribe all AmbSubscriber in the call() of the operator with 0 initial requests so all subsequent request() can target all or the winner without any additional race. The other way is to get into some complicated serialization logic.

davidmoten:
yeah I like your first suggestion, I'll look at that

davidmoten:
I've moved the subscription to sources loop into the call method as suggested.

src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -379,19 +395,21 @@ public void request(long n) {
...
-                            // Already chose one, the rest can be skipped and we can clean up
-                            selection.unsubscribeOthers(selection.choice.get());
-                            break;
+                    //subscriptions already happened so propagate the request to all the  
+                    //amb subscribers
+                    for (AmbSubscriber<T> ambSubscriber: selection.ambSubscribers) {
+                        if (!ambSubscriber.isUnsubscribed()) {
+                            // make a best endeavours check to not waste requests 
+                            // if first emission has already occurred
+                            if (selection.choice.get() == ambSubscriber) {
```
akarnokd: I'd move ```selection.choice``` into a final variable at the beginning of the method so we don't re-read two fields unnecessarily with all those memory barriers around.

davidmoten:
I've done that and reduced volatile reads a bit too. Given that `selection` and `selection.choice` are both final fields what is the interaction with the memory barriers? (sorry, a bit of a general question this for my education)

akarnokd:
Variables tend to be re-read around volatiles and the JIT *may* be smart enough to optimize it and pull them upfront, but you can't be sure.

src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -357,9 +361,10 @@ public void call(final Subscriber<? super T> subscriber) {
...
             public void call() {
-                if (selection.choice.get() != null) {
+                AmbSubscriber<T> c;
+                if ((c = choice.get()) != null) {
                     // there is a single winner so we unsubscribe it
-                    selection.choice.get().unsubscribe();
+                    c.unsubscribe();
                 } 
                 // if we are racing with others still existing, we'll also unsubscribe them
                 if(!selection.ambSubscribers.isEmpty()) {
```
akarnokd: I'd read ```selection.ambSubscribers``` into a local variable as well.

src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -371,27 +376,47 @@ public void call() {
             }
             
         }));
+        //need to subscribe to all the sources
+        for (Observable<? extends T> source : sources) {
+            if (subscriber.isUnsubscribed()) {
+                return;
+            }
+            AmbSubscriber<T> ambSubscriber = new AmbSubscriber<T>(0, subscriber, selection);
+            selection.ambSubscribers.add(ambSubscriber);
```
akarnokd: Again, ```selection.ambSubscribers``` could be pulled before the loop.

davidmoten:
added `ambSubscribers` final variable

src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -343,10 +344,14 @@ public void unsubscribeOthers(AmbSubscriber<T> notThis) {
...
 
     }
-
-    private final Iterable<? extends Observable<? extends T>> sources;
-    private final Selection<T> selection = new Selection<T>();
-
+    
+    //give default access instead of private as a micro-optimization 
+    //for access from anonymous classes below
+    final Iterable<? extends Observable<? extends T>> sources;
```
akarnokd: I meant read them into a method-local variable.

src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -343,10 +344,14 @@ public void unsubscribeOthers(AmbSubscriber<T> notThis) {
...
+    final Selection<T> selection = new Selection<T>();
+    final AtomicReference<AmbSubscriber<T>> choice = selection.choice;
+    final Collection<AmbSubscriber<T>> ambSubscribers = selection.ambSubscribers;
```
akarnokd: I meant read them into a method-local variable.

davidmoten:
ok, done

akarnokd:
I'd like to review the whole operator but I don't have the time and the focus to do it now. I'll come back to this in a few hours.

zsxwing:
> if m requests were made before the first emission then m subscriptions were started on each source observable. We only want once subscription on each.

Maybe I missed some discussion about this case. So now we support calling `request` in `onStart` multiple times?

davidmoten:
>Maybe I missed some discussion about this case. So now we support calling request in onStart multiple times?

Yes we do, multiple requests in `onStart` are additive but this is about what happens between subscriptions starting and first emission occurring. The new test `testSubscriptionOnlyHappensOnce` demos the problem in that by subscribing to an `amb` of async sources with an `onStart` request of one and then immediately (post subscription) calling `request(1)` we get a bunch of extra subscriptions.

src/test/java/rx/internal/operators/OnSubscribeAmbTest.java
```diff
@@ -219,4 +220,49 @@ public void testBackpressure() {
...
+        };
+        //this aync stream should emit first
+        Observable<Integer> o1 = Observable.just(1).doOnSubscribe(incrementer)
+                .delay(100, TimeUnit.MILLISECONDS).subscribeOn(Schedulers.computation());
+        //this stream emits second
+        Observable<Integer> o2 = Observable.just(1).doOnSubscribe(incrementer)
+                .delay(100, TimeUnit.MILLISECONDS).subscribeOn(Schedulers.computation());
+        TestSubscriber<Integer> ts = new TestSubscriber<Integer>();
+        Observable.amb(o1, o2).subscribe(ts);
+        ts.requestMore(1);
```
zsxwing: This should be forbidden in the real case. Right? `request` should be called in `Subscriber`.

zsxwing: So `testSubscriptionOnlyHappensOnce` is a test case that requesting twice before any `onNext`. Right? The other request is in `subscribe`.

src/test/java/rx/internal/operators/OnSubscribeAmbTest.java
```diff
@@ -219,4 +220,49 @@ public void testBackpressure() {
...
+    public void testSecondaryRequestsPropagatedToChildren() throws InterruptedException {
+        //this aync stream should emit first
+        Observable<Integer> o1 = Observable.from(Arrays.asList(1, 2, 3))
+                .delay(100, TimeUnit.MILLISECONDS).subscribeOn(Schedulers.computation());
+        //this stream emits second
+        Observable<Integer> o2 = Observable.from(Arrays.asList(4, 5, 6))
+                .delay(200, TimeUnit.MILLISECONDS).subscribeOn(Schedulers.computation());
+        TestSubscriber<Integer> ts = new TestSubscriber<Integer>();
+        //before subscription request 1
+        ts.requestMore(1);
```
zsxwing: I'm not sure if this is a valid case. Actually, I think `request` should not be called before `subscribe`.

davidmoten:
Interesting point but I assume async operators downstream of the `amb` could do what we are trying to simulate with `TestSubscriber`.

Saying that requests between between `onStart` and first emission are forbidden is not something I see documented anywhere, let me know if you see it.

src/test/java/rx/internal/operators/OnSubscribeAmbTest.java
davidmoten: that could be the case, I think specifying `onStart` requests to `TestSubscriber` is in another PR @akarnokd?

I could implement a custom `Subscriber` here and put a request for 1 item in the `onStart` but it would have the same result.

zsxwing:
> Saying that requests between between onStart and first emission are forbidden is not something I see documented anywhere, let me know if you see it.

I don't mean it. I mean `request` is `protected` to avoid calling out from `Subscriber`.

src/main/java/rx/internal/operators/OnSubscribeAmb.java
```diff
@@ -343,10 +344,13 @@ public void unsubscribeOthers(AmbSubscriber<T> notThis) {
...
 
     }
-
-    private final Iterable<? extends Observable<? extends T>> sources;
-    private final Selection<T> selection = new Selection<T>();
-
+    
+    //give default access instead of private as a micro-optimization 
+    //for access from anonymous classes below
+    final Iterable<? extends Observable<? extends T>> sources;
```
zsxwing: Cool. Just learnt that accessing a private field of the outer class uses `invokestatic`.

JakeWharton: And generates an accessor method in that outer class, the count of which (methods) is a big deal on Android. So :+1:.

zsxwing:
LGTM for the fix. Just some minor comments for the tests.

davidmoten:
@zsxwing I've updated the test that called `ts.requestMore(1)` before subscription so that it uses the `onStart` method of `TestSubscriber` instead. Thanks for the review.

zsxwing:
LGTM. Let's wait for @akarnokd to take a final look.

akarnokd:
I found a few places for possible optimizations:

  - in ```AmbSubscriber```, you can have a plain ```boolean chosen``` field to avoid the reference comparison:
```
boolean chosen;
...
    private boolean isSelected() {
        if (chosen) {
            return true;
        }
        if (selection.choice.get() == this) {
            // fast-path
            chosen = true;
            return true;
        } else {
            if (selection.choice.compareAndSet(null, this)) {
                selection.unsubscribeOthers(this);
                chosen = true;
                return true;
            } else {
                // we lost so unsubscribe ... and force cleanup again due to possible race conditions
                selection.unsubscribeLosers();
                return false;
            }
        }
    }
```
- There is a race between the unsubscription logic at L360 and the loop at L381: while the loop is busy creating and subscribing ```AmbSubscriber```s, a concurrent unsubscribe will start cancelling that have reached the queue but any subsequent. Possible solutions:
   - Move the creation and enqueueing of ```AmbSubscriber```s at the beginning, then create the unsubscription logic and finally loop and subscribe the ```AmbSubscribers```. This way, the queue will contain all that need to be unsubscribed.
   - Instead of quitting after the ```subscriber.isUnsubscribed()```, drain the queue and unsubscribe its contents, do a similar check after the loop as well to clean up any stragglers.
   - Have a custom subscription manager, similar to what the subjects use that can unsubscribe latecommers and keep the current one-loop approach (I have plans to cover such custom container in my blog soon.)

davidmoten:
I like the optimization for `isSelected`. I've also committed the first approach for the unsub race but I think we can do better as you imply. Might go for the second.

davidmoten:
>Instead of quitting after the subscriber.isUnsubscribed(), drain the queue and unsubscribe its contents, do a similar check after the loop as well to clean up any stragglers.

I've implemented this suggestion @akarnokd. Ready for (another) review, thanks.

akarnokd:
Looks good. Thanks!

PR #2963, opened 2015-05-19
Set of standard producers and updated queue implementations with some

akarnokd:
platform-safe variants.


src/main/java/rx/internal/producers/ProducerArbiter.java
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * Producer that allows changing an underlying producer atomically and correctly resume with the accumulated
+ * requests.
+ */
+public final class ProducerArbiter implements Producer {
+    long requested;
+    Producer currentProducer;
+ 
+    boolean emitting;
```
davidmoten: I think we could spruce up this class with the naming. This class doesn't emit anything, the emit loop actually performs requests so is a request loop (which may induce emissions elsewhere perhaps but I don't think that justifies the use of emit).

src/main/java/rx/internal/producers/ProducerArbiter.java
```diff
@@ -0,0 +1,189 @@
...
+                missedProduced += n;
+                return;
+            }
+            emitting = true;
+        }
+         
+        boolean skipFinal = false;
+        try {
+            long r = requested;
+            long u = r - n;
```
davidmoten: don't reduce requested if already Long.MAX_VALUE?

src/main/java/rx/internal/producers/QueuedProducer.java
```diff
@@ -0,0 +1,187 @@
...
+                        Exceptions.throwIfFatal(ex);
+                        Throwable ex1 = OnErrorThrowable.addValueAsLastCause(ex, v != NULL_SENTINEL ? v : null);
+                        c.onError(ex1);
+                        return;
+                    }
+                    r--;
+                    e++;
+                }
+                 
+                if (e != 0) {
```
davidmoten: && get() != Long.MAX_VALUE

src/main/java/rx/internal/producers/ProducerArbiter.java
akarnokd: It's named after the emitter-loop pattern and  I'd like to keep it named this way so developers can recognize it.

src/main/java/rx/internal/producers/ProducerArbiter.java
akarnokd: Right.

src/main/java/rx/internal/producers/QueuedValueProducer.java
```diff
@@ -0,0 +1,138 @@
...
+                        return;
+                    }
+                    if (c.isUnsubscribed()) {
+                        return;
+                    }
+                    r--;
+                    e++;
+                }
+                 
+                if (e != 0) {
```
davidmoten: && get() != Long.MAX_VALUE

davidmoten: I wonder if we should provide a bit of fast path support so that when the request is `Long.MAX_VALUE` we don't keep asking what its value is from the volatile?

src/main/java/rx/internal/producers/QueuedProducer.java
akarnokd: Thanks.

src/main/java/rx/internal/producers/QueuedValueProducer.java
akarnokd: Because requests and available values run independently and you need to come back for more values even if get() == Long.MAX_VALUE and you'd need to remember you are on a fast-path anyway. I'll update the ```if```.

src/main/java/rx/internal/util/atomic/MpscLinkedAtomicQueue.java
```diff
@@ -0,0 +1,123 @@
...
+ * href="http://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue"> on 1024
+ * Cores</a> by D. Vyukov. The original has been adapted to Java and it's quirks with regards to memory model and
+ * layout:
+ * <ol>
+ * <li>Use XCHG functionality provided by AtomicReference (which is better in JDK 8+).
+ * </ol>
+ * The queue is initialized with a stub node which is set to both the producer and consumer node references. From this
+ * point follow the notes on offer/poll.
+ * 
+ * @author nitsanw
```
benjchristensen: If we are copying this from somewhere I think we should include the URL to where it came from and associated copyright information.

benjchristensen:
This looks like good additions. If I understand correctly there are two things happening here:

1) Abstract Producers to simplify building Observables/Operators
2) New impls of queues that use Atomics rather than Unsafe so they work on Android, Google App Engine, etc?

The only thing I would like changed before merging is to include correct attribution to where the code comes from.

akarnokd:
Sure, I'll add the link. These aren't all the android-compatible variants so I plan to do a separate PR on the remaining and updating all platform-checked places.

benjchristensen:
> I plan to do a separate PR on the remaining and updating all platform-checked places.

Then should the producers and queues be separated into different PRs?

akarnokd:
The default constructor of ```QueuedProducer``` uses ```SpscLinkedQueue``` so at least they need to be in the same PR, but ```MpscLinkedQueue``` can be added after if you wish.

benjchristensen:
Since these don't change any existing code and are all internal, let's merge and move forward.

akarnokd:
Okay.

src/main/java/rx/internal/producers/ProducerArbiter.java
davidmoten: Yeah I noticed the naming pattern from your blogs and I think instead of a *queue drainer* and an *emitter* they are better described as both queue drainers except one is aysnc biased and the other is sync biased.

PR #2969, opened 2015-05-20
Operator cache() now supports backpressure

akarnokd:
I've removed the controversial features and made the backing CachedObservable internal.

Related issue #1483.


abersnaze:
Does it really need to be returning a new type of Observable?  It seems the methods `isConnected()`, `hasObservers()` and `cachedEventCount()` are only used by the unit tests.

akarnokd:
Yes. Otherwise one would need ugly reflection magic to get inside the state of this observable.

abersnaze:
I checked out the branch and took a shot at making it an OnSubscribe function again.
https://github.com/abersnaze/RxJava/tree/OperatorCacheRewrite2

akarnokd:
Cache had originally more functionality, i.e., the ability to terminate it manually, like a disconnect, but was downwoted. I forgot about the `toCached()` methods which are just relics and I'll update this PR.

Otherwise, I don't really see why I shouldn't extend Observable, especially if one day, there will be a real need to support disconnection of cache().

akarnokd:
I've rebased it and I'm going to merge this. If there has concern about it being implemented as a Observable, a PR can always be submitted with the suggested changes.

PR #2973, opened 2015-05-20
observeOn: option to batch-request instead of at end of a drain-run.

akarnokd:
I measured +10% improvement in some cases, but -20% in others:

image

Here is the scan of various batching sizes and number of source elements:

image

The benchmarks fluctuate quite a bit on their own, probably due to frequent biased-lock revocation.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -16,22 +16,15 @@
 package rx.internal.operators;
 
 import java.util.Queue;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
-import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.*;
 
 import rx.Observable.Operator;
-import rx.Producer;
-import rx.Scheduler;
-import rx.Subscriber;
-import rx.Subscription;
+import rx.*;
 import rx.exceptions.MissingBackpressureException;
 import rx.functions.Action0;
-import rx.internal.util.RxRingBuffer;
-import rx.internal.util.SynchronizedQueue;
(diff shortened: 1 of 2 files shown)


abersnaze:
The idea that I had at one point was rather than picking a batch size as a fixed percentage was to compute the batch size as the number of items that can be processed in the latency between the `request` is called and `onNext` is called.  For example if the last time we made a `request` it took 100ms before the first `onNext` came back and the consumer takes 10ms to process one value then the next `request` should be made when there are about (100ms/10ms =) 10 items left in the ring buffer.

The overhead of the book keeping might be too high for many small items.  Maybe a fast path if either the numerator or denominator gets too close to zero.

benjchristensen:
Because it still requests at the end of pollQueue, this seems like it could result in more requests being made. For example, batchSize=50, it emits 55, it would request 50 then 5, correct?

It seems more efficient to just leave it as is. What would be more interesting to me is figuring out a safe way to accumulate small requests across `pollQueue` invocations. So if we end up with 5+3+6+4 we could accumulate that into a single request(18) which would be more efficient. 

We tried that once but caused deadlocks in edge cases. 

Leaving open as this needs more discussion and the perf results are too variable to make a decision right now.

akarnokd:
> Because it still requests at the end of pollQueue, this seems like it could result in more requests being made. For example, batchSize=50, it emits 55, it would request 50 then 5, correct?

Yes, the request outside of the loop turned out to be necessary as I've experienced hangs due to missing requests.

I'm not too keen to pursue this any further, my interest was mostly the performance profile and the effect of batching.

benjchristensen:
Do you think the variable performance results warrant merging this?

akarnokd:
The variance is too high and to inconsistent for my taste so I rather close this PR for now.

PR #2975, opened 2015-05-20
Deprecate and rename two timer overloads to interval

abersnaze:
The existing six methods below didn't make sense and caused some confusion.

```
timer(delay, timeUnit[, scheduler]);  --> emits 0 after delay and completes
timer(delay, period, timeUnit[, scheduler]); --> emits 0 after delay and then i++ after every period forever
interval(period, timeUnit[, schduler); --> emits i++ after every period forever
```

I felt that the middle method acted more like the third method `interval` than first method `timer`.  This PR is to make this change.

```
 timer(delay, timeUnit[, scheduler]);  --> emits 0 after delay and completes
+@Deprecated
 timer(delay, period, timeUnit[, scheduler]); --> emits 0 after delay and then i++ after every period forever
+interval(delay, period, timeUnit[, scheduler]); --> emits 0 after delay and then i++ after every period forever
 interval(period, timeUnit[, schduler); --> emits i++ after every period forever
```

PS: yes, I understand that we can't delete the deprecated timer method.
@davgross and if this PR is merged will have to change the images.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -1225,7 +1225,7 @@ public class Observable<T> {
      * @see <a href="http://reactivex.io/documentation/operators/interval.html">ReactiveX operators documentation: Interval</a>
      */
     public final static Observable<Long> interval(long interval, TimeUnit unit) {
-        return interval(interval, unit, Schedulers.computation());
+        return interval(interval, interval, unit, Schedulers.computation());
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/main/java/rx/Observable.java
```diff
@@ -2441,57 +2497,17 @@ public final Boolean call(T first, T second) {
...
-     * 
-     * @param initialDelay
-     *            the initial delay time to wait before emitting the first value of 0L
-     * @param period
-     *            the period of time between emissions of the subsequent numbers
-     * @param unit
-     *            the time unit for both {@code initialDelay} and {@code period}
-     * @return an Observable that emits a 0L after the {@code initialDelay} and ever increasing numbers after
-     *         each {@code period} of time thereafter
-     * @see <a href="http://reactivex.io/documentation/operators/timer.html">ReactiveX operators documentation: Timer</a>
```
DavidMGross: I'd recommend leaving the documentation in place even when deprecating the method.

src/main/java/rx/Observable.java
```diff
@@ -2441,57 +2497,17 @@ public final Boolean call(T first, T second) {
...
-     *            the initial delay time to wait before emitting the first value of 0L
-     * @param period
-     *            the period of time between emissions of the subsequent numbers
-     * @param unit
-     *            the time unit for both {@code initialDelay} and {@code period}
-     * @param scheduler
-     *            the Scheduler on which the waiting happens and items are emitted
-     * @return an Observable that emits a 0L after the {@code initialDelay} and ever increasing numbers after
-     *         each {@code period} of time thereafter, while running on the given Scheduler
-     * @see <a href="http://reactivex.io/documentation/operators/timer.html">ReactiveX operators documentation: Timer</a>
```
DavidMGross: I'd recommend leaving the documentation in place even when deprecating the method

src/main/java/rx/Observable.java
```diff
@@ -1248,7 +1248,63 @@ public void call(Subscriber<? super Object> t1) {
...
+     * @param initialDelay
+     *            the initial delay time to wait before emitting the first value of 0L
+     * @param period
+     *            the period of time between emissions of the subsequent numbers
+     * @param unit
+     *            the time unit for both {@code initialDelay} and {@code period}
+     * @return an Observable that emits a 0L after the {@code initialDelay} and ever increasing numbers after
+     *         each {@code period} of time thereafter
+     * @see <a href="http://reactivex.io/documentation/operators/timer.html">ReactiveX operators documentation: Timer</a>
+     */
```
DavidMGross: Add a "@since" annotation to indicate that this is a new method that wasn't in 1.0.0
* @since [release #]

src/main/java/rx/Observable.java
```diff
@@ -1248,7 +1248,63 @@ public void call(Subscriber<? super Object> t1) {
...
+     * @param period
+     *            the period of time between emissions of the subsequent numbers
+     * @param unit
+     *            the time unit for both {@code initialDelay} and {@code period}
+     * @param scheduler
+     *            the Scheduler on which the waiting happens and items are emitted
+     * @return an Observable that emits a 0L after the {@code initialDelay} and ever increasing numbers after
+     *         each {@code period} of time thereafter, while running on the given Scheduler
+     * @see <a href="http://reactivex.io/documentation/operators/timer.html">ReactiveX operators documentation: Timer</a>
+     */
```
DavidMGross: Add a "@since" annotation to indicate that this is a new method that wasn't in 1.0.0
* @since [release #]

src/main/java/rx/Observable.java
```diff
@@ -1248,7 +1248,63 @@ public void call(Subscriber<? super Object> t1) {
...
+     *            the initial delay time to wait before emitting the first value of 0L
+     * @param period
+     *            the period of time between emissions of the subsequent numbers
+     * @param unit
+     *            the time unit for both {@code initialDelay} and {@code period}
+     * @param scheduler
+     *            the Scheduler on which the waiting happens and items are emitted
+     * @return an Observable that emits a 0L after the {@code initialDelay} and ever increasing numbers after
+     *         each {@code period} of time thereafter, while running on the given Scheduler
+     * @see <a href="http://reactivex.io/documentation/operators/timer.html">ReactiveX operators documentation: Timer</a>
```
DavidMGross: if this change goes through, I'll move the description over to http://reactivex.io/documentation/operators/interval.html so you can change this @see URL

src/main/java/rx/Observable.java
```diff
@@ -1248,7 +1248,63 @@ public void call(Subscriber<? super Object> t1) {
...
+     * 
+     * @param initialDelay
+     *            the initial delay time to wait before emitting the first value of 0L
+     * @param period
+     *            the period of time between emissions of the subsequent numbers
+     * @param unit
+     *            the time unit for both {@code initialDelay} and {@code period}
+     * @return an Observable that emits a 0L after the {@code initialDelay} and ever increasing numbers after
+     *         each {@code period} of time thereafter
+     * @see <a href="http://reactivex.io/documentation/operators/timer.html">ReactiveX operators documentation: Timer</a>
```
DavidMGross: if this change goes through, I'll move the description over to http://reactivex.io/documentation/operators/interval.html so you can change this @see URL

src/main/java/rx/Observable.java
```diff
@@ -1248,7 +1248,63 @@ public void call(Subscriber<? super Object> t1) {
...
     public final static Observable<Long> interval(long interval, TimeUnit unit, Scheduler scheduler) {
-        return create(new OnSubscribeTimerPeriodically(interval, interval, unit, scheduler));
+        return interval(interval, interval, unit, scheduler);
+    }
+
+    /**
+     * Returns an Observable that emits a {@code 0L} after the {@code initialDelay} and ever increasing numbers
+     * after each {@code period} of time thereafter.
+     * <p>
+     * <img width="640" height="200" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/timer.p.png" alt="">
```
DavidMGross: If this change goes through I will change this image but the image URL will still be "...timer.p.png" so you don't have to change this img tag

src/main/java/rx/Observable.java
```diff
@@ -1248,7 +1248,63 @@ public void call(Subscriber<? super Object> t1) {
...
+     */
+    public final static Observable<Long> interval(long initialDelay, long period, TimeUnit unit) {
+        return interval(initialDelay, period, unit, Schedulers.computation());
+    }
+
+    /**
+     * Returns an Observable that emits a {@code 0L} after the {@code initialDelay} and ever increasing numbers
+     * after each {@code period} of time thereafter, on a specified {@link Scheduler}.
+     * <p>
+     * <img width="640" height="200" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/timer.ps.png" alt="">
```
DavidMGross: If this change goes through I will change this image but the image URL will still be "...timer.p.png" so you don't have to change this img tag

src/main/java/rx/Observable.java
abersnaze: Wouldn't the reinstated javadoc for the deprecated timer method also use the same timer.p.png image?

src/main/java/rx/Observable.java
abersnaze: Wouldn't the reinstated javadoc for the deprecated timer method also use the same timer.ps.png image?

src/main/java/rx/Observable.java
DavidMGross: Yeah, but I don't think that's a big deal. It'd just be another way of highlighting that the behavior of the operator is now available under a more appropriate name.

abersnaze:
Okay, I think I'm done changing the PR unless there are other comments.

akarnokd:
I'm not in favor of these changes.

abersnaze:
any particular reason?

akarnokd:
If it were only a matter of aliasing, I wouldn't see any problem, but you deprecate a historical method just because the name. Every use place need to be changed to avoid deprecation warnings in RxJava and outside it.

abersnaze:
I think it improved clarity of API and is directly related to a bug we had here.  This PR already takes care of migrating the RxJava usages of timer to interval.

akarnokd:
The timer-interval confusion apparently is quite common so I'm going to merge this. Thanks for the changes.

PR #2981, opened 2015-05-23
Add example of dependency for Gradle

Fi5t:


akarnokd:
Too many ```and``` for my taste:

```
Example for Maven:
...
and for Ivy:
...
and for Gradle:
...
```

akarnokd:
I'm closing this due to no response.

PR #2982, opened 2015-05-26
TestSubscriber - add factory methods

davidmoten:
Since we use this so much I've added factory methods so we can infer types. 

Instead of 

```java
TestSubscriber<Integer> ts = new TestSubscriber<Integer>();
```

We can write
```java
TestSubscriber<Integer> ts = TestSubscriber.create();
```

I changed the tests in `OnSubscribeRangeTest` to demo the usage.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -95,6 +95,31 @@ public class TestSubscriber<T> extends Subscriber<T> {
         this(-1);
     }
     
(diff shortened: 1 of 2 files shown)


akarnokd:
Looks handy. @benjchristensen what do you think?

akarnokd:
I'm going to merge these as they come in handy while RxJava requires Java 6 to compile.

benjchristensen:
I'm okay with these additions

Issue #2979, opened 2015-05-22
Subscriber.onStart is called twice on nested unsafeSubscription


ibaca:
This test shows the onStart called twice problem. It is a clone of SubscriberTest.testOnStartCalledOnceViaUnsafeSubscribe but adding the defer operator which calls unsafeSubscribe resulting in a nested call to unsafeSubscribe, the one from the test and the second one from OnSubscribeDefer.
```java
@Test
public void testOnStartCalledOnceViaNestedUnsafeSubscribe() {
    final AtomicInteger c = new AtomicInteger();
    Observable.defer(new Func0<Observable<Integer>>() {
        @Override public Observable<Integer> call() {
            return Observable.just(1, 2, 3, 4).take(2);
        }
    }).unsafeSubscribe(new Subscriber<Integer>() {
        @Override public void onStart() {
            c.incrementAndGet();
            request(1);
        }

        @Override public void onCompleted() { }

        @Override public void onError(Throwable e) { }

        @Override public void onNext(Integer t) {
            request(1);
        }
    });

    assertEquals(1, c.get());
}
```
Two proposed solutions (first one looks ugly, second one adds more code to Subscriber)
1. Add a parameter to subscription calls to transfer the 'onStart called' state.
2. Modify Subscriber so .subscribe() .unsafeSubscribe() calls Subscriber.start() which delegates to the  actual .onStart() protecting for duplicate calls.

akarnokd:
With a little overhead, it is possible to suppress the second ```onStart``` call in defer by wrapping the subscriber with another subscriber that does nothing in its ```onStart```. Would you like to submit a PR?

akarnokd:
In fact, many other operators suffer from the same issue. You can discover them by altering Subscriber and running the standard tests:
```java
final AtomicInteger once = new AtomicInteger();
// ...
public void onStart() {
    if (once.getAndIncrement() > 0) {
        throw new IllegalStateException();
    }
}
```

ibaca:
Closing because has been solved by #2983. Thanks!

PR #2983, opened 2015-05-26
Fixed multiple calls to onStart.

akarnokd:
Started investigating multiple calls to onStart based on #2979 and found a bunch, but since onStart can be overridden, I might have missed cases. In addition, I've found a single place where a producer is set twice on a subscriber but again, many other places may exist.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeDefer.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeDefer.java
@@ -38,7 +38,7 @@ public final class OnSubscribeDefer<T> implements OnSubscribe<T> {
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


ibaca:
(I know that is easy to complain :wink:) This decorator strategy is not very memory friendly, you have this annoying (gwt compatibility perspective) AtomicXxxFieldUpdater in a lot of places to reduce memory usage, and you end up adding this decorators to only disable one method call.

I don't know, what do you think?. Thanks to solve the problem anyway :smile:.

davidmoten:
Yikes! There's a lot of stuffed up backpressure fixed here. I'm hoping we can release soon.

PR #2985, opened 2015-05-27
improve OperatorSerializeTest.testMultiThreadedWithNPEinMiddle 

davidmoten:
to address #2870, have made it much more improbable that the error emission is the last thread to complete by only failing the test if the error emission is the last for `n` repeats. I chose `n` = 3 to keep the runtime low (average <2s).


akarnokd:
Thanks!

PR #2986, opened 2015-05-27
OperatorAll  - implement backpressure and include last value in exception cause

davidmoten:
Implemented backpressure using `SingleDelayedProducer` (thanks @akarnokd) and while I was there ensured that the last value is in the exception cause when the predicate throws an exception.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorAll.java
+++ b/src/main/java/rx/internal/operators/OperatorAll.java
@@ -17,7 +17,10 @@ package rx.internal.operators;
 
 import rx.Observable.Operator;
 import rx.Subscriber;
+import rx.exceptions.Exceptions;
+import rx.exceptions.OnErrorThrowable;
 import rx.functions.Func1;
+import rx.internal.producers.SingleDelayedProducer;
 
 /**
  * Returns an Observable that emits a Boolean that indicates whether all items emitted by an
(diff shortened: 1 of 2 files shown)


src/test/java/rx/internal/operators/OperatorAllTest.java
```diff
@@ -19,12 +19,16 @@
 import static org.mockito.Mockito.*;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.math3.analysis.FunctionUtils;
```
akarnokd: External package reference!
import org.apache.commons.math3.analysis.FunctionUtils;

davidmoten:
oops, fixed.

src/main/java/rx/internal/operators/OperatorAll.java
```diff
@@ -34,16 +37,23 @@ public OperatorAll(Func1<? super T, Boolean> predicate) {
...
+                    return;
+                }
                 if (!result && !done) {
                     done = true;
-                    child.onNext(false);
-                    child.onCompleted();
+                    producer.setValue(false);
                     unsubscribe();
                 } else {
                 	// if we drop values we must replace them upstream as downstream won't receive and request more
```
akarnokd: This case is no longer necessary because the subscriber requests ```Long.MAX_VALUE``` upfront.

davidmoten:
Thanks, I've updated the commit with that change.

akarnokd:
Looks good now. Thanks!

PR #2989, opened 2015-05-28
prevent take() from requesting more than needed

davidmoten:
The `request` method used logic that could overrequest upstream if multiple small requests were made before the next item arrived.

Added a unit test that failed on previous code.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorTake.java
+++ b/src/main/java/rx/internal/operators/OperatorTake.java
@@ -15,6 +15,8 @@
  */
 package rx.internal.operators;
 
+import java.util.concurrent.atomic.AtomicLong;
+
 import rx.Observable.Operator;
 import rx.Producer;
 import rx.Subscriber;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorTake.java
```diff
@@ -78,15 +80,22 @@ public void onNext(T i) {
...
-                            if (n < c) {
-                                producer.request(n);
-                            } else {
-                                producer.request(c);
+                        if (n >0 && !completed) {
+                            // because requests may happen concurrently use a CAS loop to 
+                            // ensure we only request as much as needed, no more no less
+                            while (true) {
+                                long r = requested.get();
+                                long c = Math.min(n, limit - requested.get());
```
akarnokd: Re-read of a volatile value, you need ```limit - r```.

src/main/java/rx/internal/operators/OperatorTake.java
```diff
@@ -78,15 +80,22 @@ public void onNext(T i) {
...
+                                if (requested.compareAndSet(r, r + c)) {
```
akarnokd: if ```c == 0``` there is nothing to request and the loop can quit.

davidmoten:
Thanks @akarnokd. Changes made.

akarnokd:
Great job!

PR #2992, opened 2015-05-30
Fix value rendering in error last cause for primitive types

davidmoten:
When an error occurs some `Operator`s (like `map`) include the value in the exception cause to help debugging. The intent is that for the 9 primitive types (int/Integer, double/Double, byte/Byte, char/Character etc) we render the value using `value.toString` but the test for being a primitive does not work because `.isPrimitive` always returns false (primitives have always been autoboxed by the time they get here).

The fix is to create a lazy loaded singleton set of the primitive classes (Integer.class, Double.class, etc) and check for membership in that set.

Unit tests included for each primitive type.

### DIFF
--- a/src/main/java/rx/exceptions/OnErrorThrowable.java
+++ b/src/main/java/rx/exceptions/OnErrorThrowable.java
@@ -15,6 +15,9 @@
  */
 package rx.exceptions;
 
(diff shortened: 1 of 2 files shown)


akarnokd:
Good catch!

PR #2995, opened 2015-05-31
switchOnNext - ensure initial requests additive and fix request overflow

davidmoten:
`OperatorSwitch` didn't record the total of requests before first inner subscription occurring and had problems with request overflow. 

Includes 4 unit tests that failed on the original code.


akarnokd:
I'd like to review the whole operator (can't do it today) because I have a feeling a complex serialization is required (i.e., a ```ProducerObserverArbiter``` level).

davidmoten:
Yeah I'm not surprised at that, I suspected the same but put off a full head scratch till later.

akarnokd:
I'm merging this as is and perhaps come back later to think about corner cases remaining.

PR #2996, opened 2015-05-31
switchIfEmpty - fix backpressure bug and lost requests

davidmoten:
`OperatorSwitchIfEmpty` suffered from these conditions:

* Backpressure was not enabled for the source observable (as opposed to the altenate) because `setProducer` called `super.setProducer` instead of `child.setProducer`
* Requests could be lost

Bot the of the above problems are solved by using `ProducerArbiter` from @akarnokd.

Included two unit tests that failed on previous code.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
+++ b/src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
@@ -15,9 +15,9 @@
  */
 package rx.internal.operators;
 
-import java.util.concurrent.atomic.AtomicLong;
 
 import rx.*;
+import rx.internal.producers.ProducerArbiter;
 import rx.subscriptions.SerialSubscription;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
```diff
@@ -38,33 +38,35 @@ public OperatorSwitchIfEmpty(Observable<? extends T> alternate) {
         final SwitchIfEmptySubscriber parent = new SwitchIfEmptySubscriber(child, ssub);
         ssub.set(parent);
         child.add(ssub);
+        child.setProducer(new Producer() {
```
akarnokd: I'd instantiate the ProducerArbiter before ```parent``` and supply it to ```SwitchIfEmptySubscriber``` and ```child```, saving on this 'relaying' allocation. (Remark, our life would be much easier if we could implement ```Producer``` in ```Subscriber```, but since ```request()``` is ```protected final```, it is unoverridable :( )

src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
```diff
@@ -81,20 +83,7 @@ private void subscribeToAlternate() {
```
akarnokd: I'd move this anonym inner class into a regular static class to avoid capturing unnecessary references. In addition, I'd first call ```ssub.set``` then call ```unsafeSubscribe```

src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
davidmoten: Yeah life would be easier if `request()` was public. I wouldn't mind seeing it opened up, do you think it worth discussing in an issue?

akarnokd: It's too late for that. We will be careful in RxJava 2.0.

davidmoten:
Made those changes, ta.

davidmoten:
Cleaned up a bit, made the parent subscriber class a static class and improved names

src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
```diff
@@ -77,41 +73,7 @@ public void onCompleted() {
...
-                public void onError(Throwable e) {
-                    child.onError(e);
-                }
-
-                @Override
-                public void onNext(T t) {
-                    child.onNext(t);
-                }
-            }));
+            ssub.set(alternate.unsafeSubscribe(new AlternateSubscriber<T>(child, arbiter)));
```
akarnokd: Instead:
```java
AlternateSubscriber<T> as = new AlternateSubscriber<T>(child, arbiter);
ssub.set(as);
alternate.unsafeSubscribe(as);
```

davidmoten:
Well spotted, thanks. Updated.

src/main/java/rx/internal/operators/OperatorSwitchIfEmpty.java
```diff
@@ -125,4 +89,36 @@ public void onNext(T t) {
...
+
+        @Override
+        public void onError(Throwable e) {
+            child.onError(e);
+        }
+
+        @Override
+        public void onNext(T t) {
+            child.onNext(t);
+            arbiter.produced(1);
```
akarnokd: I think this ```produced(1)``` is unnecessary: sure the arbiter's value will grow indefinitely, but there won't be any further ```setProducer``` on it thus conserving the requested/delivered amount is less important.

davidmoten: The hey-you-delivered-more-than-requested check is useful though

akarnokd: In that case, you need to add this to the other child.onNext() call as well.

davidmoten:
Erk, thanks for noticing that. Updated.

akarnokd:
Great, thanks!

PR #3000, opened 2015-06-02
Replace the Java 7 AssertionError(message, cause) with RuntimeException

zsxwing:
`AssertionError(message, cause)` is a Java 7 API. This PR uses RuntimeException to replace it, just like `assertNoErrors`.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -356,12 +356,14 @@ public class TestSubscriber<T> extends Subscriber<T> {
             throw new AssertionError("No errors");
         } else
         if (err.size() > 1) {
-            // can't use AssertionError because (message, cause) doesn't exist until Java 7
-            throw new RuntimeException("Multiple errors: " + err.size(), new CompositeException(err));
+            AssertionError ae = new AssertionError("Multiple errors: " + err.size());
+            ae.initCause(new CompositeException(err));
+            throw ae;
         } else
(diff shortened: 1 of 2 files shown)


src/main/java/rx/subjects/ReplaySubject.java
```diff
@@ -114,7 +114,7 @@ public void call(SubjectObserver<T> o) {
                 boolean skipFinal = false;
                 try {
                     for (;;) {
-                        int idx = o.index();
+                        int idx = o.<Integer>index();
```
zsxwing: Java 6 compiler is not smart enough to infer `Integer`.

src/main/java/rx/observers/TestSubscriber.java
```diff
@@ -356,10 +356,12 @@ public void assertError(Class<? extends Throwable> clazz) {
             throw new AssertionError("No errors");
         } else
         if (err.size() > 1) {
-            throw new AssertionError("Multiple errors: " + err.size(), new CompositeException(err));
+            // can't use AssertionError because (message, cause) doesn't exist until Java 7
```
akarnokd: You can use `initCause()`
```java
AssertionError ae = new AssertionError(...);
ae.initCause(...)
throw ae;
```

zsxwing: Cool. Updated. I have not updated `assertNoErrors` because changing from RuntimeException to AssertionError is a breaking change.

akarnokd:
Great, thanks!

PR #3001, opened 2015-06-02
use Subscribers.from()

davidmoten:
minor simplification in `OperatorDelayWithSelector` and `OperatorMulticast`


akarnokd:
Looks okay, thanks!

Issue #3006, opened 2015-06-04
Bug in OnSubscribeFromIterable unsubscribed child logic


pelecomepibe:
Hello

The code is missing an additional check for child unsubscription after calling the iterator next method. This is needed when using in conjunction with operators like take, that unsubscribes itself in the onNext method.

Regards

davidmoten:
Sorry, can't see the issue with `take`, can you elaborate? If the child has unsubscribed during the `onNext` then the next loop will check for unsubscription and exit. 

The iterator `next` method could take enough time for unsubscription to have happened in the meantime so we could insert another check for unsubscription before emitting. I'll see what the others think of this as I imagine it would compromise performance a bit for common use cases like using java Collections as sources because each emission would require two volatile reads of `unsubscribed`.

```java
if (o.isUnsubscribed())
    return;
T val = it.next();
if (o.isUnsubscribed()) 
    return;
o.onNext(val);
```
Is this what you were meaning?

pelecomepibe:
The problem arised when using an `Iterable` which produces iterators with a blocking `hasNext`.
Imagine the case where the `hasNext` call would block after retrieving 10 elements (so, in the eleventh call). Chaining a `take`/`limit` of 10 would still block, even after emiting 10 elements.

We did a quick patch:

```
while (it.hasNext() && --numToEmit >= 0) {
                            if (o.isUnsubscribed()) {
                                return;
                            }
                            o.onNext(it.next());
                            if (o.isUnsubscribed()) {
                                return;
                            }

                        }
```

zsxwing:
How about checking `isUnsubscribed` before `hasNext`, such as:

```
                        while (!o.isUnsubscribed() && it.hasNext() && --numToEmit >= 0) {
                            o.onNext(it.next());
                        }
                        if (o.isUnsubscribed()) {
                            return;
                        }
```

davidmoten:
@pelecomepibe yep that does look something to avoid
@zsxwing LGTM

akarnokd:
Thanks for reporting. Fix released in 1.0.12.

PR #3009, opened 2015-06-08
Observable.from(iterable) - ensure it.hasNext() is not called unnecessarily #3006

davidmoten:
This PR is a fix for #3006.

Ensured `it.hasNext()` not called more often than required especially when unsubscription happens before completion.

Added two unit tests (one for fast path and the other for backpressure) that failed on previous codebase.


akarnokd:
You forgot to implement ````remove()``` which is only default in Java 8.

davidmoten:
Ta, fixed.

akarnokd:
Still it doesn't compile:

/home/travis/build/ReactiveX/RxJava/src/test/java/rx/internal/operators/OnSubscribeFromIterableTest.java:233: error: <anonymous rx.internal.operators.OnSubscribeFromIterableTest$4$1> is not abstract and does not override abstract method remove() in Iterator
                return new Iterator<Integer>() {

davidmoten:
woops, there were two of them. Fixed.

zsxwing:
LGTM :+1:

akarnokd:
Thanks!

PR #3010, opened 2015-06-08
Operator flat map rewrite

akarnokd:
An alternative to #2928 by using flatMap(o -> o) == merge() identity.

Benchmark:

image

It is slightly better than #2928 for flatMap() benchmarks and slightly worse than merge()


benjchristensen:
What do the columns represent in that benchmark?

akarnokd:
1.x = current head, 
1st try = rewrite but only with the scalar optimizations, 
3rd try = rewrite+scalar optimizations + unbounded mode optimizations, 
flatMap = merge implemented with flatMap instead the other way around so merge(map(func)) inlines func into merge and saves on a lift() call.

davidmoten:
I made this comment on your tool's site but I think we can't assess whether the benchmarks indicate *significant* change unless the error value is taken into account and I assume these visualizations always ignore the error value (is that right?).

akarnokd:
I'm closing this in favor of the recent changes in #2928.

PR #3011, opened 2015-06-09
Fix TestSubscriber awaitTerminalEvent methods

davidmoten:
Returning if timeout occurred from `TestSubscriber.awaitTerminalEvent` is a feature I wanted and put in this PR. It was a void method and now returns a boolean so this should be a non-breaking change. 

In the process I noticed that `TestSubscriber.awaitTerminalEventAndUnsubscribeOnTimeout` was broken in that it assumed that an exception was thrown on timeout.  Fixed this making use of the boolean return of `TestSubscriber.awaitTerminalEvent`.

Added two unit tests.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -266,12 +266,13 @@ public class TestSubscriber<T> extends Subscriber<T> {
      *          the duration of the timeout
      * @param unit
      *          the units in which {@code timeout} is expressed
+     * @return true if and only if the timeout did not occur
      * @throws RuntimeException
      *          if the Subscriber is interrupted before the Observable is able to complete
      */
-    public void awaitTerminalEvent(long timeout, TimeUnit unit) {
+    public boolean awaitTerminalEvent(long timeout, TimeUnit unit) {
         try {
-            latch.await(timeout, unit);
+            return latch.await(timeout, unit);
         } catch (InterruptedException e) {
             throw new RuntimeException("Interrupted", e);
         }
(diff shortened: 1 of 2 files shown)


akarnokd:
Changing existing method definition is a breaking change (I've also tried this exact change and was rejected).

davidmoten:
Are you referring to  #2594?  That was about throwing exceptions on timeout not the return value. Unfortunately that commit broke `TestSubscriber.awaitTerminalEventAndUnsubscribeOnTimeout`. 

I'd like to get this reviewed because I don't think this PR is a breaking change (void method -> boolean method should affect no-one).

zsxwing:
Maybe we can add new methods and deprecate the new ones.

JakeWharton:
> void method -> boolean method should affect no-one

This is binary incompatible. See: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.15

davidmoten:
Thanks Jake.

benjchristensen:
The binary incompatibility makes this a non-option for merging.

benjchristensen:
Closing out due to incompatibility. An alternate solution can be brought up in a new PR.

ruhkopf:
Hi Ben,
Sorry for commenting on a closed issue, but this came up when searching for testing timeouts with awaitTerminalEvent. Has there been an alternative PR or is this still broken? 

I was looking at the latest 1.x code and don't see it fixed or an alternative method. I can see it was in fact fixed in the 2.x branch.

For 1x, is there a workaround on how to test a timeout? I guess I could use `awaitTerminalEventAndUnsubscribeOnTimeout` in combination with `doOnUnsubscribe()` (as seen in `TestSubscriberTest.testAwaitTerminalEventWithDurationAndUnsubscribeOnTimeout`). But maybe there's an easier way?

PR #3023, opened 2015-06-17
ConnectableObservable autoConnect operator

akarnokd:
There is a need for auto-connection capability for an arbitrary ConnectableObservable similar to how it is possible now to use cache() as an auto-connected unbounded replay.

The new `autoConnect()` method let's specify the number of `Subscriber`s required for the real connection to happen and also offers the ability to receive the connection `Subscription`.

I've also updated the parameter name of `cache(int)` to reflect it is a capacity hint and not a bound.


benjchristensen:
`autoConnect` is a nice additional feature. Like `refCount` except it never disconnects. I particularly like the overload where it auto connects when the specific number of subscribers arrive.

PR #3032, opened 2015-06-18
range() and from() run in fast-path in case request is at least the

akarnokd:
known available amount.

This optimization switches to the fast-path if one requests at least the amount the producer can provide instead of just when requesting MAX_VALUE.

```
1.x
Benchmark                       (size)         Score  Score error
rangeWithBackpressureRequest         1  82865476,500  3050323,928
rangeWithBackpressureRequest      1000    279002,692    11523,399
rangeWithBackpressureRequest   1000000       172,744       17,732
rangeWithoutBackpressure             1  84039644,379  4026176,141
rangeWithoutBackpressure          1000    285125,368     6292,095
rangeWithoutBackpressure       1000000       261,569       28,274

This PR
Benchmark                       (size)         Score  Score error
rangeWithBackpressureRequest         1  86083847,250  3562236,352
rangeWithBackpressureRequest      1000    289972,949    16689,560
rangeWithBackpressureRequest   1000000       278,011       46,335
rangeWithoutBackpressure             1  89601089,634  2260936,315
rangeWithoutBackpressure          1000    294289,152     8461,897
rangeWithoutBackpressure       1000000       272,347       22,644
```

I'm not sure if there is an existing benchmark for `from()` or not.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeFromIterable.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeFromIterable.java
@@ -15,6 +15,7 @@
  */
 package rx.internal.operators;
 
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
@@ -46,8 +47,15 @@ public final class OnSubscribeFromIterable<T> implements OnSubscribe<T> {
         final Iterator<? extends T> it = is.iterator();
         if (!it.hasNext() && !o.isUnsubscribed())
             o.onCompleted();
-        else 
-            o.setProducer(new IterableProducer<T>(o, it));
+        else {
+            long count;
+            if (is instanceof Collection) {
+                count = ((Collection<?>)is).size();
+            } else {
+                count = Long.MAX_VALUE;
+            }
+            o.setProducer(new IterableProducer<T>(o, it, count));
+        }
     }
 
     private static final class IterableProducer<T> implements Producer {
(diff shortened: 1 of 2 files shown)


davidmoten:
Cool. I love how there's always a bit more to squeak out of these classes.

src/main/java/rx/internal/operators/OnSubscribeFromIterable.java
```diff
@@ -46,8 +47,15 @@ public void call(final Subscriber<? super T> o) {
         final Iterator<? extends T> it = is.iterator();
         if (!it.hasNext() && !o.isUnsubscribed())
             o.onCompleted();
-        else 
-            o.setProducer(new IterableProducer<T>(o, it));
+        else {
+            long count;
+            if (is instanceof Collection) {
```
davidmoten: `is` is a weird field name, can you change it to `iterable`?

akarnokd:
I have second thoughts about `from()`. If one hands it a collection which throws `ConcurrentModificationException`, then the optimization is valid: you should use effectively immutable Iterables anyway. 

However, if you iterate a `ConcurrentHashMap` which doesn't throw and is likely to deliver newly put in values, the fast-path is likely to produce more than requested as the hasNext() is a moving target.

We could blacklist a few sources but there isn't a distinctive type for these concurrently-modifiable collection classes (remember, we are dealing with CHM.keySet() and CHM.values() here, not with CHM directly).

akarnokd:
Some more benchmarks:

image

For the drastic drops:
```
flatMapIntPassthruAsync       1  thrpt         5   120125,781   140629,155    ops/s
observeOnNewThread            1  thrpt         5    12582,907     4900,400    ops/s
rangeWithBackpressureRequest  1  thrpt         5 86241753,647 17739600,235    ops/s
```

benjchristensen:
> However, if you iterate a ConcurrentHashMap which doesn't throw and is likely to deliver newly put in values, the fast-path is likely to produce more than requested as the hasNext() is a moving target.

That's a valid point. Leaving this open for further discussion.

benjchristensen:
Perhaps we should just whitelist the ones we know are safe for now? `LinkedList`, `ArrayList` (since they would throw), and `java.util.Collections$UnmodifiableCollection`?

akarnokd:
I'll redo the `range()` optimization separately.

Issue #3027, opened 2015-06-18
takeLast() javadoc suggestion


davidmoten:
@DavidMGross `takeLast` and `takeLastBuffer` methods are inconsistent in the javadoc about their behaviour when less is available than requested. When less is available than requested all are emitted but this behaviour is not clear in the javadoc of all of the variations. Should I leave this to you?

DavidMGross:
You should feel free to change the javadocs as you find errors,
ambiguities, missing information, or as you change the behavior of methods
(e.g. to add backpressure support).  I've done a lot of work with these
javadocs as the tech writer guy, but philosophically I think the front-line
developers ought to consider it part of their responsibility to maintain
their accuracy.

On Wed, Jun 17, 2015 at 10:19 PM, Dave Moten <notifications@github.com>
wrote:

> @DavidMGross <https://github.com/DavidMGross> takeLast and takeLastBuffer
> methods are inconsistent in the javadoc about their behaviour when less is
> available than requested. When less is available than requested all are
> emitted but this behaviour is not clear in the javadoc of all of the
> variations. Should I leave this to you?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ReactiveX/RxJava/issues/3027>.
>



-- 
David M. Gross
PLP Consulting

davidmoten:
no probs

PR #3033, opened 2015-06-18
takeLast() javadoc fixes, standardize parameter names (count instead of num)

davidmoten:
This PR includes these changes:

* standardized parameter names for `take`, `takeLast`, `skip`, `limit` to use `count` instead of `num`
* enhanced the javadoc for `takeLast` so it is clear for all overrides that if the observable contains less than `count` items then all will be emitted.
* improved the exception message thrown by `OperatorTakeLast` when `count` <0

Addresses #3027

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -5141,28 +5141,28 @@ public class Observable<T> {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


akarnokd:
Thanks!

PR #3034, opened 2015-06-19
Instantiate EMPTY lazily

davidmoten:
The singleton `EMPTY` returned by `Observable.empty()` can be instantiated lazily because it is not always required.


akarnokd:
Thanks.

PR #3039, opened 2015-06-22
Window with Observable: fixed unsubscription and behavior

akarnokd:
Fixed unsubscription propagation of two Window variants (boundary, start-end).

In addition, there was a discrepancy reported on [StackOverflow](http://stackoverflow.com/questions/30963993/rxjava-window-buffer-overload-questions-incompatibility-with-rx-net) regarding the factory-boundary version. In Rx.NET, the factory is called whenever the previous boundary observable has produced a value. I.e., instead of a series of `onNexts` from the same `Observable`, it uses a series of `Observable`s with a single `onNext` emission as a boundary indicator.


benjchristensen:
Thank you for tackling this.

PR #3042, opened 2015-06-23
add backpressure support for defaultIfEmpty() 

davidmoten:
When looking at implementing backpressure support for `defaultIfEmpty` the obvious strategy is to reuse `switchIfEmpty` and get rid of `OperatorDefaultIfEmpty`. That's what this PR does.

There is a little bit of overhead with using `Observable.from(Arrays.asList(defaultValue))` as the alternate observable. Might be nice to have a version of `Observable.just(item)` that supports backpressure to use in its place (you've mentioned this before @akarnokd).


akarnokd:
You could just create an Observable inline:

```java
public final Observable<T> defaultIfEmpty(T defaultValue) {
    //if empty switch to an observable that emits defaultValue and supports backpressure
    return switchIfEmpty(Observable.create(s -> {
        s.setProducer(new SingleProducer<>(defaultValue));
    }));
}
```

davidmoten:
Great, can do.  I'd like to see that as its own method. I'll put that change in and open another issue so method names/overrides and desirability can be discussed.

akarnokd:
Thanks!

PR #3043, opened 2015-06-23
TestSubscriber javadoc cleanup

davidmoten:
* standardize on `Asserts that` and `Returns`
* fix tenses and wordings


akarnokd:
Thanks!

Issue #1487, opened 2014-07-23
Backpressure: replay


benjchristensen:
The `replay` operators and `ReplaySubject` should respect backpressure for the downstream while replaying back. Depending on the variant of `replay` (such as unbounded buffering) it could support it even in multicast mode. For the variants that don't do unbounded buffering, then once it finishes replying the cached values, then it can no longer support backpressure (due to multicasting), but it's still good to support it during replay, as that can be a very rapid firehose whereas it could be quite slow once caught up.

akarnokd:
See #3047.

akarnokd:
Done and shipped via 1.0.14.

PR #3047, opened 2015-06-23
Operator replay() now supports backpressure

akarnokd:
I've added backpressure support to the replay() operator(s).

Related issue: #1487


benjchristensen:
Epic refactor. Thanks @akarnokd!

Issue #3046, opened 2015-06-23
NullPointerException at rx.exceptions.CompositeException.getListOfCauses(CompositeException.java:245)


sgonchar:
I use RxAndroid library in my project and I have next exception :

E/AndroidRuntime(27849): Error reporting crash
E/AndroidRuntime(27849): java.lang.NullPointerException
E/AndroidRuntime(27849): at rx.exceptions.CompositeException.getListOfCauses(CompositeException.java:245)
E/AndroidRuntime(27849): at rx.exceptions.CompositeException.getCause(CompositeException.java:100)
E/AndroidRuntime(27849): at android.util.Log.getStackTraceString(Log.java:454)
E/AndroidRuntime(27849): at android.util.Slog.e(Slog.java:151)
E/AndroidRuntime(27849): at com.android.internal.os.RuntimeInit$UncaughtHandler.uncaughtException(RuntimeInit.java:109)
E/AndroidRuntime(27849): at java.lang.ThreadGroup.uncaughtException(ThreadGroup.java:693)
E/AndroidRuntime(27849): at java.lang.ThreadGroup.uncaughtException(ThreadGroup.java:690)
E/AndroidRuntime(27849): at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:66)
E/AndroidRuntime(27849): at android.os.Handler.handleCallback(Handler.java:733)
E/AndroidRuntime(27849): at android.os.Handler.dispatchMessage(Handler.java:95)
E/AndroidRuntime(27849): at android.os.Looper.loop(Looper.java:146)
E/AndroidRuntime(27849): at android.app.ActivityThread.main(ActivityThread.java:5653)
E/AndroidRuntime(27849): at java.lang.reflect.Method.invokeNative(Native Method)
E/AndroidRuntime(27849): at java.lang.reflect.Method.invoke(Method.java:515)
E/AndroidRuntime(27849): at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1291)
E/AndroidRuntime(27849): at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1107)
E/AndroidRuntime(27849): at dalvik.system.NativeStart.main(Native Method)

Somebody know how can I fix it ?
Thank you

akarnokd:
It appears you have a `null` exception added to a composite somewhere. What operators are you using. Do you call `onError(null)` somewhere by accident?

[...]

PR #3048, opened 2015-06-23
CompositeException extra NPE protection

akarnokd:
`CompositeException` won't crash if it received null by some means. A `null` collection or `null` collection item is replaced by a `NullPointerException` instance.

See also: #3046.

### DIFF
--- a/src/main/java/rx/exceptions/CompositeException.java
+++ b/src/main/java/rx/exceptions/CompositeException.java
@@ -49,12 +49,19 @@ public final class CompositeException extends RuntimeException {
     public CompositeException(String messagePrefix, Collection<? extends Throwable> errors) {
         Set<Throwable> deDupedExceptions = new LinkedHashSet<Throwable>();
         List<Throwable> _exceptions = new ArrayList<Throwable>();
-        for (Throwable ex : errors) {
-            if (ex instanceof CompositeException) {
-                deDupedExceptions.addAll(((CompositeException) ex).getExceptions());
-            } else {
-                deDupedExceptions.add(ex);
+        if (errors != null) {
+            for (Throwable ex : errors) {
+                if (ex instanceof CompositeException) {
+                    deDupedExceptions.addAll(((CompositeException) ex).getExceptions());
+                } else 
+                if (ex != null) {
+                    deDupedExceptions.add(ex);
+                } else {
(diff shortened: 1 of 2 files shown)


src/main/java/rx/exceptions/CompositeException.java
```diff
@@ -49,12 +49,19 @@
...
-            if (ex instanceof CompositeException) {
-                deDupedExceptions.addAll(((CompositeException) ex).getExceptions());
-            } else {
-                deDupedExceptions.add(ex);
+        if (errors != null) {
+            for (Throwable ex : errors) {
+                if (ex instanceof CompositeException) {
+                    deDupedExceptions.addAll(((CompositeException) ex).getExceptions());
+                } else 
+                if (ex != null) {
```
zsxwing: Could you move `if (ex != null) {` to the same line of `else`?

src/main/java/rx/exceptions/CompositeException.java
```diff
@@ -49,12 +49,19 @@
...
+                    deDupedExceptions.add(ex);
+                } else {
+                    deDupedExceptions.add(new NullPointerException());
```
zsxwing: How about just throwing `NullPointerException` here? If that causes other issue, could you add some useful exception message, such as, `Should not add null Throwable to CompositeException`? So that people can quickly figure out it's an application bug instead of an RxJava bug.

akarnokd: Like throwing exception from an exception? Since you only have the option to detect null here, throwing an NPE would have the same stacktrace as these new NPEs, but with the exception that they will further 'crash' the application. In my way, these inconsistencies get captured and propagated with the CompositeException itself.

src/main/java/rx/exceptions/CompositeException.java
akarnokd: No. This is my style of coding and there is no style guide for RxJava anyway.

zsxwing: I thought it was unintentional. Never mind. See this style for the first time :)

src/main/java/rx/exceptions/CompositeException.java
zsxwing: I see.

Issue #3038, opened 2015-06-21
Variants of first, single etc. which return Single


tilal6991:
Now that we have the concept of a "Single", I think it would make sense that any operator which returns an observable which is guaranteed to return exactly one item or fail would be a Single. Examples would be operators like single, first, toList, takeFirst, takeLast etc.

I understand that to maintain source and binary compatibility, the existing definitions of these would need to be unchanged so would it be possible to create some sort of conversion mechanism to turn an observable into a Single?

If this would be acceptable I would be happy to try my hand in writing such a conversion mechanism depending on how complex it would have to be. I have been using Rx for a while now but as of yet have not actually written something which interacts with low level Rx stuff (e.g. a custom operator)

Thanks!

benjchristensen:
Yes we will want something to convert, such as `Single.from(Observable o)` or `Observable.toSingle()`.

The `toSingle()` method should behave similarly to `single()` except return a `Single`. 

Awkward, but all we can do during 1.x.

It would be great if you want to contribute this and start the design discussion. Note that public APIs contributions can often go through a lot of back and forth as we discuss them since we are stuck with them forever once we let them in.

PR #3049, opened 2015-06-23
single: add toSingle method to Observable

tilal6991:
* closes ReactiveX/RxJava#3038
* this method allows an observable which is guaranteed to return exactly one
  item to be converted to a Single
* NOTE: the semantics of this function are very similar to that of single
* i.e. errors are passed through, more than one item results in an
  IllegalArgumentException, completion without emission results in a
  NoSuchElementException and exactly one item is passed through the onSuccess
  method of SingleSubscriber

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeSingle.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeSingle.java
@@ -1,0 +1,89 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.operators;
+
+import rx.Observable;
+import rx.Single;
+import rx.SingleSubscriber;
+import rx.Subscriber;
+
+import java.util.NoSuchElementException;
+
+/**
+ * Allows conversion of an Observable to a Single ensuring that exactly one item is emitted - no more and no less.
+ * Also forwards errors as appropriate.
+ */
+public class OnSubscribeSingle<T> implements Single.OnSubscribe<T> {
+
+    private final Observable<T> observable;
+
+    public OnSubscribeSingle(Observable<T> observable) {
+        this.observable = observable;
+    }
+
+    @Override
+    public void call(final SingleSubscriber<? super T> child) {
+        Subscriber<T> parent = new Subscriber<T>() {
+            private boolean emittedTooMany = false;
+            private boolean itemEmitted = false;
+            private T emission = null;
+
+            @Override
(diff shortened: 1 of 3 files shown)


tilal6991:
Would appreciate feedback since this is my first contribution to RxJava! :)

src/main/java/rx/internal/operators/OnSubscribeSingle.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final Observable<T> observable;
+
+    public OnSubscribeSingle(Observable<T> observable) {
+        this.observable = observable;
+    }
+
+    @Override
+    public void call(final SingleSubscriber<? super T> singleSubscriber) {
+        observable.subscribe(new Subscriber<T>() {
```
akarnokd: Unsubscription chain is not maintained here.

tilal6991: Just to double check: does this mean I need to add a few checks like 
                if (singleSubscriber.isUnsubscribed()) {
                    unsubscribe();
                    return;
                }
in the 4 subscriber methods?

akarnokd: See http://akarnokd.blogspot.hu/2015/05/pitfalls-of-operator-implementations.html for a detailed explanation.

tilal6991: Thanks that helps quite a bit in understanding. However, unlike in your post, I cannot pass the child subscriber to the parent since the child has the SingleSubscriber type which isn't accepted by the Subscriber constructor.

akarnokd: But you can add resources to `singleSubscriber`.

akarnokd:
Yes.

src/main/java/rx/internal/operators/OnSubscribeSingle.java
abersnaze: There is also a constructor overload that takes the child Subscriber to automatically chain together unsubscriptions.

benjchristensen:
Looks good. Can you add the license headers to the files please? The block of code to add is shown here: https://github.com/ReactiveX/RxJava/blob/1.x/CONTRIBUTING.md

src/main/java/rx/internal/operators/OnSubscribeSingle.java
tilal6991: As I say though I can't use that overload since the type of the child is SingleSubscriber not Subscriber.

tilal6991:
Have done so.

benjchristensen:
Would you mind rebasing to a single commit?

tilal6991:
Also done :)

benjchristensen:
Thank you @tilal6991 for the contribution and responsiveness on changes.

tilal6991:
Glad to contribute.

I'm planning to look into more tricky issues soon so hopefully this is not the last you'll see of me :)

src/main/java/rx/internal/operators/OnSubscribeSingle.java
```diff
@@ -0,0 +1,89 @@
...
+                    child.onError(new IllegalArgumentException("Observable emitted too many elements"));
+                    unsubscribe();
+                } else {
+                    itemEmitted = true;
+                    emission = t;
+                }
+            }
+        };
+        child.add(parent);
+        observable.subscribe(parent);
```
davidmoten: @benjchristensen Shouldn't this use `unsafeSubscribe`?

PR #3055, opened 2015-06-30
toSingle() should use unsafeSubscribe

davidmoten:
As per my comment in #3049, `Observable.toSingle` should use `unsafeSubscribe` internally.

This PR includes a unit test that failed on previous code.


davidmoten:
Unrelated test failure:

```
testMultiThreadedWithNPEinMiddle FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:92)
        at org.junit.Assert.assertTrue(Assert.java:54)
        at rx.internal.operators.OperatorSerializeTest.testMultiThreadedWithNPEinMiddle(OperatorSerializeTest.java:148)
```

Issue #3037, opened 2015-06-20
Observable<Void> Usage or VoidObservable


NiteshKant:
While working on [RxNetty](https://github.com/ReactiveX/RxNetty), a networking library that layers RxJava on top of netty, I have come across a common case of expressing result of an operation as an `Observable<Void>`. A typical case being the result of a `write` on a `Connection`. The API for write looks like:

`public Observable<Void> write(Observable<T> msgs);`

The return value just represents the asynchronous result of the write and hence does not emit any item, so has the type as `Void`.

This works pretty well, however, it becomes awkward when we start doing a bit more complex operations over this write. eg: If I am writing a TCP client that sends a message on the connection and then reads the response from the server, I start doing things like:

```java
connection.write(Observable.just("hello"))
                 .cast(String.class)
                 .concatWith(connection.getInput())
```
(Actual example [here](https://github.com/ReactiveX/RxNetty/blob/0.5.x/rxnetty-examples/src/main/java/io/reactivex/netty/examples/tcp/echo/EchoClient.java#L81))

In this example, `connection.getInput()` returns an `Observable<T>` emitting items as read on the connection.

and in case, I do not want to wait for write to complete, the code looks like:

```java
connection.write(Observable.just("hello"))
                 .cast(String.class)
                 .mergeWith(connection.getInput())
```

What is awkward in the above example, is the `cast(String.class)`. It is completely safe as the `write` never emits an item (`Void`) so I can merge/concat heterogeneous streams but it is just awkward.

My motivation behind creating this issue is to first see if other folks have come across the same scenario, if so, how have they handled it. Secondly, if this does sound like something worth spending time to make it better, I would like to propose some abstraction on the lines of a `VoidObservable` (for the lack of a better name) which basically provides better constructs around:

- __Concat with a different type__: `write().cast().concatWith()` example above,  where it is expected to continue with another stream of a different type (`String` in the above example)
- __Merge with a different type__: `write().cast().mergeWith()` example above, where it is expected to eagerly subscribe to the second `Observable` but listen only for errors from the `Observable<Void>`

akarnokd:
I see two problems:
## Type inference limitations
[...]

PR #3056, opened 2015-06-30
New experimental operator to switch the types on the completion of empty Observables.

abersnaze:
For the issue #3037 a new operator that let consumer of an empty Observable tie additional work to follow after the completion.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -3909,6 +3909,26 @@ public class Observable<T> {
     }
 
     /**
+     * Returns an Observable that only emits the items emitted by the alternate Observable if the source Observable
+     * is empty. If the source Observable errors or emit any values then an error is propagated and alternate Observable is
+     * never subscribed to.
+     * <p/>
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code switchEmpty} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param alternate
+     *              the alternate Observable to subscribe to if the source does not emit any items
+     * @return  an Observable that only emits the items of an alternate Observable if the source Observable is empty.
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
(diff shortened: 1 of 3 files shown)


davidmoten:
Seems like an ok idea to me but `switchEmpty` and  `switchIfEmpty` are confusing to distinguish though I like them being close together alphabetically for discoverability.

NiteshKant:
aah @abersnaze I did not see that you have the changes here for #3037. I have created a PR #3060  for it too, although it contains an additional operator `mergeError`

NiteshKant:
I implemented `switchEmpty`, as you call it, as a `switchMap().concatWith()`, is there a reason why that does not work?

abersnaze:
I tried to reuse the switchIfEmpty and merge the switchEmpty functionality but couldn't get the generics to work.  I decided to copy and edit because switchIfEmpty looked complicated enough with backpressure support.

```
***************
*** 65,79 ****
  
          @Override
          public void onCompleted() {
!             if (!empty) {
!                 child.onCompleted();
!             } else if (!child.isUnsubscribed()) {
                  subscribeToAlternate();
              }
          }
--- 64,76 ----
  
          @Override
          public void onCompleted() {
!             if (!child.isUnsubscribed()) {
                  subscribeToAlternate();
              }
          }
***************
*** 85,92 ****
  
          @Override
          public void onNext(T t) {
!             empty = false;
!             child.onNext(t);
              arbiter.produced(1);
          }
      }
--- 82,88 ----
  
          @Override
          public void onNext(T t) {
!             child.onError(new RuntimeException("switchEmpty used on a non empty observable. Possible fix is to add .ignoreElements() before .switchEmpty()."));
              arbiter.produced(1);
          }
      }

```

The overhead of the switchMap and concatWith seem pretty high for such a constrained use case.

NiteshKant:
@abersnaze I updated my PR with the things we discussed offline and it now has custom operator implementations, optimized for this usecase where the source `Observable` is never expected to emit any item.

benjchristensen:
Closing out so we can merge these efforts with https://github.com/ReactiveX/RxJava/pull/3060 after agreeing upon naming: https://github.com/ReactiveX/RxJava/issues/3037#issuecomment-121295614

PR #3060, opened 2015-07-02
New operators: `concatEmptyWith` and `mergeEmptyWith`.

NiteshKant:
As discussed in issue #3037, the primary use of these operators is to be applied to `Observable<Void>` so that they can be merged and concatenated with an Observable of a different type.

Both these operators raise an error if the source Observable emits any item.


NiteshKant:
Marked the new operators as `@Experimental`, once the review is complete, I will squash the commits into one.

NiteshKant:
Renamed operators to `mergeEmptyWith` and `concatEmptyWith`. Based on offline discussions with @abersnaze and @benjchristensen these names better suit the intent of these operators.

Also, removed the `swithMap()` + `mergeWith`/`concatWith` implementations with custom operators that are much lighter weight for this case (source never emits an item) vis-a-vis `concat` and `merge`.

benjchristensen:
Waiting on confirmation of names: https://github.com/ReactiveX/RxJava/issues/3037#issuecomment-121295614

src/main/java/rx/Observable.java
```diff
@@ -9468,6 +9468,75 @@ else if (count == 1 )
...
+     * Observable and then emits all items emitted by that Observable. This function does not expect the source
+     * Observable to emit any item, in case, the source Observable, emits any item, an {@link IllegalStateException}
+     * is raised.
+     * <p>
+     *
+     * This is different than {@link #concatWith(Observable)} as it does not expect the source Observable to ever emit
+     * an item. So, this usually is useful for {@code Observable<Void>} and results in cleaner code as opposed to using
+     * a {@link #cast(Class)}, something like:
+     *
+     {@code Observable.<Void>empty().cast(String.class).concatWith(Observable.just("Hello"))}
```
roman-mazur: nit: no `*` before the line

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,193 @@
...
+        public void onNext(R r) {
+            child.onNext(r);
+        }
+    }
+
+    /**
+     * This is a producer implementation that only does the following:
+     * <ul>
+     <li>If the alternate producer has not yet arrived, store the total requested count from downstream.</li>
+     <li>If the alternate producer has arrived, then relay the request demand to it.</li>
```
roman-mazur: nit: no `*` before the line

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,193 @@
...
+            this.emptyProducer = new EmptyProducer();
+            this.alternate = alternate;
+        }
+
+        @Override
+        public void setProducer(final Producer producer) {
+            /*Nothing from this producer will ever be requested as we never expect any items to be emitted from the
+            parent. The only thing w.r.t backpressure that is required is to store the requested count from the
+            child till the time the alternate observable is subscribed. So, this producer is just ignored and the
+            configured EmptyProducer on the child stores any buffered requested items till subscription to alternate.*/
```
abersnaze: How does the subscriber know the producer doesn't emit if we never request it too?

abersnaze: I think if the this should call `producer.request(Long.MAX_VALUE);`.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,193 @@
...
+
+        private final Subscriber<? super R> child;
+        private final SerialSubscription ssub;
+        private final EmptyProducer emptyProducer;
+        private final Observable<? extends R> alternate;
+
+        ParentSubscriber(Subscriber<? super R> child, final SerialSubscription ssub, Observable<? extends R> alternate) {
+            this.child = child;
+            this.ssub = ssub;
+            this.emptyProducer = new EmptyProducer();
```
akarnokd: You should use `ProducerArbiter` instead.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
akarnokd: If the operator uses `ProducerArbiter`, the child request will reach the main source and there is no need for extra requests.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,193 @@
...
+            if (!child.isUnsubscribed()) {
+                AlternateSubscriber as = new AlternateSubscriber(child, emptyProducer);
+                ssub.set(as);
+                alternate.unsafeSubscribe(as);
+            }
+        }
+
+        @Override
+        public void onError(Throwable e) {
+            child.onError(e);
```
akarnokd: I suggest calling `unsubscribe()` here and or in `onNext` to stop a non-empty producer as soon as possible.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,193 @@
...
+     * This is a producer implementation that only does the following:
+     * <ul>
+     <li>If the alternate producer has not yet arrived, store the total requested count from downstream.</li>
+     <li>If the alternate producer has arrived, then relay the request demand to it.</li>
+     </ul>
+     *
+     * Since, this is only applicable to this operator, it does not check for emissions from the source, as the source
+     * is never expected to emit any item. Thus it is "lighter" weight than {@link ProducerArbiter}
+     */
+    private static final class EmptyProducer implements Producer {
```
akarnokd: This class is unnecessary, the functionality is already covered by `ProducerArbiter`.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        @Override
+        public void onCompleted() {
+            // Nothing to do as the child completes with the alternate observable.
+        }
+
+        @Override
+        public void onError(Throwable e) {
+            child.onError(e);
```
akarnokd: Potential concurrent invocation of a onXXX method on an unserialized child. You need to wrap the child with `SerializedSubscriber`. In addition, the error condition in the parent needs to unsubscribe both the main and the alternate source, see the implementation of `takeUntil`.

benjchristensen:
Thanks @akarnokd for the review.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
NiteshKant: Good catch, I would add the `SerializedSubscriber`. I realized the dangling subscription to the `alternateSource` and was fixing it, thanks for pointing it out!

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: I started off with `ProducerArbiter` and then thought that the functionality in `ProducerArbiter` is much larger than what is required here primarily because of the fact that the first producer here would never ever emit an item and hence the use case that `ProducerArbiter` is solving of resuming from an unsatisfied request from downstream, is a potential overhead. I would love to benchmark if there is a difference at all (although by sheer LOC it seems like an overhead) with this approach.

I have put this reasoning behind this class in the javadoc as a disclaimer :)

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: See my reasoning down.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: I see, I was assuming that there will be a `SafeSubscriber` down the line which will unsubscribe post `onError()` and we don't need it here. Is this a general practice we follow in all operators?

src/main/java/rx/Observable.java
NiteshKant: Done

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: Done

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: I have added `producer.request(Long.MAX_VALUE)` to the source `Observable` for both the operators. 
There was a thought to send the `request` to the source _only_ after the subscriber requested anything. Although, this made sense but in this context, it was not making much of a difference but adding complexity of synchronization between `request(n)` and `setProducer()`. The argument against being that for a source that never emits an item, should the terminal events be delayed till `request(n)` or a subscription indicates start of processing.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
NiteshKant: I made the `ChildSubscriber` as `SerializedSubscriber`. 
On thinking more about the `unsubscribe` part, I think I am not yet clear as to why that is a problem. AFAIU, on parent error, it will be propagated to the child as an `onError` after which the `SafeSubscriber` at the end of the chain should `unsubscribe`, which will flow in and unsubscribe from the parent and child. Is my understanding incorrect?

NiteshKant:
Apart from the open discussions, I think I am ready for another round of review. 

The travis build failed because of the test rx.BackpressureTests > testMergeAsyncThenObserveOn failing. I don't think my changes have anything to do with that, is that a flaky test?

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,198 @@
...
+            synchronized (this) {
+                if (0 != missedRequested) {
+                    /*Something was requested from the source Observable, relay that to the new producer*/
+                    requestToAlternate = true;
+                }
+                this.altProducer = altProducer;
+            }
+
+            if (requestToAlternate) {
+                this.altProducer.request(missedRequested);
```
akarnokd: You access `missedRequested` outside of its guard.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
```diff
@@ -0,0 +1,198 @@
...
+                    this.missedRequested = u;
+                } else {
+                    /*If the alternate producer exists, then relay a valid request. The missed requested will be
+                    requested from the alt producer on setProducer()*/
+                    requestToAlternate = true;
+                }
+            }
+
+            if (requestToAlternate) {
+                altProducer.request(requested);
```
akarnokd: You access `altProducer` outside of its guard.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
akarnokd: Generally, if an `onError` or `onCompleted` comes the natural way, we can be fairly sure the upstream also cleaned up. However, if you turn an `onNext` into an `onError` or `onCompleted`, the upstream has no idea you have lost interest in further values and it can take a long time before an unsubscribe arrives from downstream. See the `take()` operator for instance.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        @Override
+        public void onNext(T t) {
+            onError(new IllegalStateException("Merge empty with source emitted an item: " + t));
+        }
+    }
+
+    private final class ChildSubscriber extends SerializedSubscriber<R> {
```
akarnokd: You shouldn't extend `SerializedSubscriber` but use it as a wrapper for another `Subscriber`:

```java
SerializedSubscriber<T> serial = new SerializedSubscriber<T>(child);
```

benjchristensen:
@akarnokd It seems you have open questions/issues on this PR still, is that correct?

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
```diff
@@ -0,0 +1,122 @@
...
+        ChildSubscriber(Subscriber<? super R> delegate) {
+            super(delegate);
+            this.delegate = delegate;
+        }
+
+        @Override
+        public void onCompleted() {
+            synchronized (this) {
+                if (parentCompleted) {
+                    delegate.onCompleted();
```
akarnokd: You shouldn't call onXXX methods while holding a lock.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
```diff
@@ -0,0 +1,122 @@
...
+
+        @Override
+        public void onNext(R r) {
+            delegate.onNext(r);
+        }
+
+        public void parentCompleted() {
+            synchronized (this) {
+                if (childCompleted) {
+                    delegate.onCompleted();
```
akarnokd: You shouldn't call onXXX methods while holding a lock.

akarnokd:
@benjchristensen Found a few problems with this PR in the meantime.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
NiteshKant: duhh .. that was a bad oversight at my end. Thanks for catching it!

NiteshKant:
@akarnokd I would wait for your further comments to make all changes together. So, get the review comments coming, they are good :)

abersnaze:
@NiteshKant is this ready to be merged?

NiteshKant:
I was under the assumption that @akarnokd has more comments. Let me address the comments till now.

akarnokd:
Sorry for not replying. I'll recheck this pr shortly.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
akarnokd: Okay.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
akarnokd: Okay.

akarnokd:
No new comments on this and there are still points to address via code change.

stealthcode:
Hi @NiteshKant. Last call for the release. Can this make it in?

akarnokd:
@stealthcode Please don't be hasty with this PR, there are still problems present here: accessing fields outside their guards and bad use of `SerializedSubscriber`.

NiteshKant:
@stealthcode when are you cutting a release?

NiteshKant:
>>> Please don't be hasty with this PR

I guess we should not be hasty with any PR :) I haven't got around making the changes yet, of course will do before we merge.

stealthcode:
@NiteshKant we'd like to release in the coming week or two but need to know if this is a feature that we should consider for the next minor release (v1.1.0) which will be almost identical to the next patch release v1.0.15. 

@akarnokd I'm not being hasty I am asking for information from my colleague so George and I can plan the release.

NiteshKant:
@stealthcode I will address the comments in a couple of days, it should be able to make it to the release!

stealthcode:
That's great thanks.

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
NiteshKant: Fixed

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
NiteshKant: Fixed

src/main/java/rx/internal/operators/OperatorMergeEmptyWith.java
NiteshKant: Fixed

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: Fixed

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: Fixed

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
NiteshKant: I actually do not understand the need for calling `unsubscribe()` here.

NiteshKant:
I addressed the open comments apart from the suggestion of calling `unsubscribe()` post `onError()`, which I do not understand/agree with.

I see that my rebase caused a bazillion other commits to show here, may be I should close this and issue a new PR?

akarnokd:
Please do create a new and clean PR.

src/main/java/rx/internal/operators/OperatorConcatEmptyWith.java
akarnokd: You are calling onError from onNext and the upstream may not receive an unsubscribe in time and keeps emitting onNexts. Better yet, you should call unsubscribe in onNext to stop the upstream immediately.

NiteshKant: So, this is designed to be concatenating an `Observable<Void>`, which legally should not be emitting any item. Should we be guarding against such illegal sources?

NiteshKant:
@akarnokd created #3430 as a replacement for this one.

Issue #3057, opened 2015-07-01
New Subscribers helper method?


davidmoten:
This sort of code is in a lot of places in the RxJava codebase:

```java
    new Subscriber<T>(s) {
            @Override
            public void onNext(T t) {
                s.onNext(t);
            }
            @Override
            public void onError(Throwable e) {
                s.onError(e);
            }
            @Override
            public void onCompleted() {
                s.onCompleted();
            }
        }
```

I'd like to add a utility method in `Subscribers` that does this. Any suggestions for a name to squeeze into the static namespace of this class? `from` would have been nice but is used by `Subscribers.from(Observer)` and this would be a breaking change if I added `Subscribers.from(Subscriber)` because they do different things and a `Subscriber` is an `Observer`.

Perhaps:

* `Subscribers.fromChained(Subscriber)`
* `Subscribers.passThrough(Subscriber)`
* `Subscribers.chained(Subscriber)`

I'm sure someone will have a nicer suggestion than these.

wujek-srujek:
`Subscribers.wrap(s)`?

davidmoten:
I like `wrap`, thanks @wujek-srujek

davidmoten:
This has been merged

PR #3065, opened 2015-07-05
add Subscribers.wrap

davidmoten:
Add utility method to `Subscribers` to perform this common use case:

Naming briefly discussed in #3057.
 
```java
new Subscriber<T>(subscriber) {

    @Override
    public void onCompleted() {
        subscriber.onCompleted();
    }

    @Override
        subscriber.onError(e);
    }

    @Override
    public void onNext(T t) {
        subscriber.onNext(t);
    }
    
};
```

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeUsing.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeUsing.java
@@ -22,6 +22,7 @@ import rx.*;
 import rx.Observable.OnSubscribe;
 import rx.exceptions.CompositeException;
 import rx.functions.*;
+import rx.observers.Subscribers;
 
 /**
  * Constructs an observable sequence that depends on a resource object.
@@ -68,20 +69,7 @@ public final class OnSubscribeUsing<T, Resource> implements OnSubscribe<T> {
                 observable = source;
             try {
                 // start
-                observable.unsafeSubscribe(new Subscriber<T>(subscriber) {
-                    @Override
-                    public void onNext(T t) {
-                        subscriber.onNext(t);
-                    }
-                    @Override
-                    public void onError(Throwable e) {
-                        subscriber.onError(e);
-                    }
-                    @Override
-                    public void onCompleted() {
-                        subscriber.onCompleted();
-                    }
-                });
+                observable.unsafeSubscribe(Subscribers.wrap(subscriber));
             } catch (Throwable e) {
                 Throwable disposeError = disposeEagerlyIfRequested(disposeOnceOnly);
                 if (disposeError != null)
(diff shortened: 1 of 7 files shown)


src/main/java/rx/internal/operators/OnSubscribeUsing.java
```diff
@@ -68,20 +69,7 @@ public void call(final Subscriber<? super T> subscriber) {
...
-                    @Override
-                    public void onError(Throwable e) {
-                        subscriber.onError(e);
-                    }
-                    @Override
-                    public void onCompleted() {
-                        subscriber.onCompleted();
-                    }
-                });
+                observable.unsafeSubscribe(Subscribers.from(subscriber));
```
benjchristensen: This calls `from` but it should be `wrap`.

benjchristensen:
I'm trying to remind myself what in these cases we can't just pass the subscriber through if we're not actually changing any behavior. Unsubscription will propagate up and down, so it's not decoupling anything. So why must we allocate and wrap in these use cases?

davidmoten:
Double call of onStart was one reason.

src/main/java/rx/internal/operators/OnSubscribeUsing.java
davidmoten: erk, thanks!

davidmoten:
used `wrap` instead of `from` in `OnSubscribeUsing` and rebased.

benjchristensen:
> Double call of onStart was one reason.

Ah right. So `wrap` is really about when we need to decouple or handle nested subscribes.

src/main/java/rx/internal/operators/OperatorDoOnSubscribe.java
```diff
@@ -39,19 +40,6 @@ public OperatorDoOnSubscribe(Action0 subscribe) {
...
-            @Override
-            public void onError(Throwable e) {
-                child.onError(e);
-            }
-            @Override
-            public void onCompleted() {
-                child.onCompleted();
-            }
-        };
+        return Subscribers.wrap(child);
```
benjchristensen: This one is less obvious as to why it needs to be wrapped. The duplicate `onStart` problem seems like something we should fix inside `Subscriber` rather than needing to object allocate and wrap.

src/main/java/rx/internal/operators/OperatorDoOnUnsubscribe.java
```diff
@@ -41,22 +42,6 @@ public OperatorDoOnUnsubscribe(Action0 unsubscribe) {
...
-            @Override
-            public void onError(Throwable e) {
-                child.onError(e);
-            }
-            @Override
-            public void onCompleted() {
-                child.onCompleted();
-            }
-        };
+        return Subscribers.wrap(child);
```
benjchristensen: Same here. The origin of the issue is in `lift`: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L144

Issue #3073, opened 2015-07-13
switchOnNext() can lose requests


davidmoten:
There is a race condition in `OperatorSwitch` that may cause requests to be lost:

1. `onNext` proceeds to just before [L137](https://github.com/ReactiveX/RxJava/blob/d32a1b0ea7ee455a23082cd1a7db80b3adef6d2b/src/main/java/rx/internal/operators/OperatorSwitch.java#L137) 
2. on a different thread `request` proceeds to just before [L112](https://github.com/ReactiveX/RxJava/blob/d32a1b0ea7ee455a23082cd1a7db80b3adef6d2b/src/main/java/rx/internal/operators/OperatorSwitch.java#L112)
3. `onNext` continues and exits method
4. `request` continues and requests from unsubscribed subscriber
5. At this point stream could stall because requests were lost 

Looks like a candidate for `ProducerArbiter` help. I'll work on a PR if no-one else has started on this issue yet (let me know if you have). There may be other race conditions as well that turn up which I'll document in the PR.

akarnokd:
Fixed and released with 1.0.14.

PR #3078, opened 2015-07-14
switchOnNext() - fix lost requests race condition 

davidmoten:
As per #3073 a race condition in `OperatorSwitch` means that requests can be lost. This PR uses `ProducerArbiter` to ensure all unfulfilled requests are carried through to the next `Observable`.

I also changed `InnerSubscriber` to be a static class as opposed to an inner class just to decouple it from surrounding state.


akarnokd:
Thanks.

PR #3079, opened 2015-07-15
fix forEach javadoc

davidmoten:
`Observable.forEach` javadoc suffered from a bit of copy-and-paste and neglected to mention the possible throw of `OnErrorNotImplementedException`.

Also added a couple of unit tests.


akarnokd:
Thanks.

Issue #3076, opened 2015-07-14
1.0.13 Release Candidate


benjchristensen:
Update: After production canary testing on the Netflix API we have gained confidence with `merge` and are now ready to release 1.0.13. We did however choose to rollback changes to `replay()` and `cache()` and will pursue them in the next release. The discussion in the comments below shares the details.

-------

This is a pretty significant release as it has major changes to `merge` ~~and `replay`~~, along with the new `rx.Single` type. 

The `merge` changes are the most sensitive, as they affect almost all applications (either through `merge` directly, or the ubiquitous use of `flatMap`). 

I'm going to wait a few days before releasing to give people a chance to use the [SNAPSHOT](https://github.com/ReactiveX/RxJava/blob/1.x/README.md#binaries). Use the snapshot from a Gradle build like this:

```groovy
repositories {
    maven { url 'https://oss.jfrog.org/libs-snapshot' }
}

dependencies {
    compile 'io.reactivex:rxjava:1.0.13-SNAPSHOT'
}
```

------------

This release has quite a few bug fixes and some new functionality. Items of note are detailed here with the list of changes at the bottom.

##### merge

The `merge` operator went through a major rewrite to fix some edge case bugs in the previous version. This has been sitting for months going through review and performance testing due to the importance and ubiquity of its usage. It is believed this rewrite is now production ready and achieves the goal of being more correct (no known edge cases at this time) while retaining comparable performance and memory usage. 

Special thanks to @akarnokd for this as `merge` is a challenging one to implement. 

##### window fix and behavior change

Unsubscription bugs were fixed in `window`. Along the way it also resulted in a fix to one of the `window` overloads that had a functional discrepancy.

```java
window(Func0<? extends Observable<? extends TClosing>> closingSelector)
```

This is a small behavior change that corrects it. If you use this overload, please review the change to ensure your application is not affected by an assumption of the previously buggy behavior: https://github.com/ReactiveX/RxJava/pull/3039

Note that this behavior change only affects that particular overload while the broader bug fixes affect all `window` overloads. 

##### rx.Single

[...]

Issue #3026, opened 2015-06-18
cache() could pass through items while buffering for replay


davidmoten:
Using 1.0.12 this bit of code indicates that `cache` reads its whole source before emitting anything:

```java
Observable.just(1, 2)
    .delay(1, TimeUnit.SECONDS, Schedulers.immediate())
    .cache()
    .repeat(2)
    .doOnNext(System.out::println)
    .subscribe()
```

Running this we get
```
<wait two seconds>
1
2
1
2
```
I'd like to get:
```
<wait one second>
1
<wait one second>
2
1
2
```
@akarnokd you've done some work on `cache` lately that I haven't looked at closely. Could an improvement like this be incorporated?

akarnokd:
Whit the current head and this code:
```java
Observable.just(1, 2)
        .delay(1, TimeUnit.SECONDS, Schedulers.immediate())
        //.take(2)
        .cache()
        .repeat(2)
        .timeInterval()
        .doOnNext(System.out::println)
        .subscribe();
```

I get
```
TimeInterval [intervalInMilliseconds=1016, value=1]
TimeInterval [intervalInMilliseconds=1001, value=2]
TimeInterval [intervalInMilliseconds=1015, value=1]
TimeInterval [intervalInMilliseconds=0, value=2]
```

The reason for the delay before the second `1` is that delay() delays the `onCompleted()` event by a second as well so the repeat can't resubscribe that early. If I uncomment the `take(2)` before the cache, that's going to trigger an earlier completion.

davidmoten:
Thanks @akarnokd. Pleased to hear it.

akarnokd:
This is now officially fixed in 1.0.14.

PR #3081, opened 2015-07-16
Revert of cache/replay

abersnaze:
#3076. I did revert commit of the PR that added back pressure support to the cache operator and ran a canary.  Our production systems seem stable now even with the new merge.


davidmoten:
That's good news. Not sure what time @akarnokd has for reviewing `.cache` backpressure work. If we can't fix this short term would be nice too see a fix for #3026 go in 1.0.13 if backpressure doesn't make it. Quick summary is in 1.0.12 `.cache` emits nothing until source has completed.

akarnokd:
Can you create an unit test based on the usage pattern? It would be also interesting to verify `replay().autoConnect()` since it ought to be functionally equivalent to `cache()`.

abersnaze:
I have no clue at the moment what the actual code path is causing the lockup. It'll take awhile for me to instrument code to narrow which usage pattern is causing the problem.

benjchristensen:
I believe the issue we're seeing comes from use of `cache()` in Hystrix: https://github.com/Netflix/Hystrix/blob/990394e98ba7472b96d90d7246e785d0966ade9a/hystrix-core/src/main/java/com/netflix/hystrix/AbstractCommand.java#L481

We use Hystrix at very high volume, and `cache()` will be used by a large percentage of those calls. This explains why this impacted our canary so significantly.

abersnaze:
I've reverted both the cache and revert related commits.  The tests and the canary seem to do fine so far.

PR #3082, opened 2015-07-16
Observable.x(ConversionFunc) to allow extensions to Observables

stealthcode:


src/test/java/rx/ObservableTests.java
```diff
@@ -1138,4 +1142,53 @@ public void testSubscribingSubscriberAsObserverMaintainsSubscriptionChain() {
 
         subscriber.assertUnsubscribed();
     }
+    
+    @Test
+    public void testExtend() {
+        final TestSubscriber<Object> subscriber = new TestSubscriber<Object>();
+        final Object value = new Object();
+        Observable.just(value).x(new ConversionFunc<Object,Object>(){
```
akarnokd: This is essentially the same as:
```java
Observable.just(value).to(o -> {
    o.subscribe(subscriber);
    subscriber.assertNoErrors();
    subscriber.assertCompleted();
    subscriber.assertValue(value);
    return subscriber.getOnNextEvents().get(0);
});

File: src/test/java/rx/ObservableTests.java [hunk 2]
```diff
@@ -1138,4 +1142,53 @@ public void testSubscribingSubscriberAsObserverMaintainsSubscriptionChain() {
...
+                subscriber.assertValue(value);
+                return subscriber.getOnNextEvents().get(0);
+            }});
+    }
+
+    @Test
+    public void testOperatorConversion() {
+        final TestSubscriber<Object> subscriber = new TestSubscriber<Object>();
+        final Object value = new Object();
+        Observable.just(value).x(new OperatorConversion<Object,Object>(new Operator<Object, Object>(){
```
akarnokd: This can be achieved with lifting before or in the `to()` operator:
```java
Observable.just(value).to(o -> o.lift(new Operator<Object, Object>() { ... }))
.subscribe(subscriber);
});
```

stealthcode: Per my comments in #2971 I don't think it's advisable to expose the whole Observable type inside the extension. If the operators of Observable are available then you should expect to see flatmaps and groupBys being executed inside a conversion. I do not think that you can achieve the same thing since with `Observable.to` exposes all operators within a conversion step that is meant to subscribe and collect values.

src/test/java/rx/ObservableTests.java
stealthcode: See my other comment regarding exposing the Observable inside the conversion.

stealthcode:
The Travis CI failure is related to an rx backpreasure test

```
rx.BackpressureTests > testMergeAsyncThenObserveOn FAILED
    java.lang.RuntimeException: Unexpected onError events: 1
        at rx.observers.TestSubscriber.assertNoErrors(TestSubscriber.java:263)
        at rx.BackpressureTests.testMergeAsyncThenObserveOn(BackpressureTests.java:138)
        Caused by:
        rx.exceptions.MissingBackpressureException
```

Is this known to be buggy? My changes seem unrelated. It also failed for #3060.

akarnokd:
It didn't fail with `MissingBackpressureException` as far as I know.

stealthcode:
They both failed with missing backpressure.

https://travis-ci.org/ReactiveX/RxJava/builds/71027596#L278

https://travis-ci.org/ReactiveX/RxJava/builds/71831639#L270

akarnokd:
Found the bug in merge(), it was my oversight. Fix posted: #3093.

stealthcode:
Awesome! Thanks @akarnokd

akarnokd:
I've restarted this build and it passes now.

stealthcode:
I've simplified the problem this is trying to address to just adding the `x(ConversionFunc)` method and unit tests to show how we might use this. My apologies to the general populace who aren't familiar with Battlestar Galactica who may be confused by the references made to said series in my unit test.

stealthcode:
I would have liked to have made the static method

```java
private static Subscription Observable.subscribe(Subscriber<? super T>, Observable<T>)
```

be agnostic of the `Observable` and instead have the signature.. 

```java
public static Subscription Observable.subscribe(Subscriber<? super T>, OnSubscribe<T>)
```

This would allow us to reuse this `SafeSubscriber` wrapping and the hooks invocations from other Observables (see test) but then the hooks api would break on...

```java
hook.onSubscribeStart(observable, observable.onSubscribe).call(subscriber);
```

and then it probably wouldn't belong in `Observable.java`...

stealthcode:
Reviewed with @benjchristensen and he agreed to merge this and start testing it out in snapshots.

src/main/java/rx/ConversionFunc.java
```diff
@@ -0,0 +1,21 @@
...
+import rx.annotations.Experimental;
+
+/**
+ * Converts the values emitted by an Observable's OnSubscribe function to a value. 
+ *
+ * @param <T> the type of values to be consumed
+ * @param <R> the return type
+ */
+@Experimental
+public interface ConversionFunc<T, R> {
```
abersnaze: Do we need another type?  Would `Func1<OnSubscribe<T>, R>` be okay

stealthcode: Sure. Updating code now.

stealthcode:
Latest version was rebased onto 1.x and has `ConversionFunc` replaced with `Func1`.

PR #3085, opened 2015-07-16
break tests as approach timeout so that don't fail on slow machines

davidmoten:
These tests were giving me regular failures on my slow laptop:

* `OperatorMergeMaxConcurrentTest.testSimpleOneLessAsync`
* `BackpressureTests.testOnBackpressureDrop`

They both loop repeatedly looking for failure or lockup. If the machine running the tests isn't fast enough then the tests timeout before reaching the desired number of loops. This PR adds a break to the tests when the loop has run for 9 seconds (timeout is 10 seconds).


akarnokd:
Great!

PR #3086, opened 2015-07-16
improve ExecutorSchedulerTest.testOnBackpressureDrop

davidmoten:
This test fails for me occasionally on my slow laptop when full test suite run because work is scheduled for only 1ms in the future. I've decreased the probability of this one failing by changing the schedule to 50ms in the future.


akarnokd:
Thanks.

PR #3088, opened 2015-07-17
Hardened both cache() and replay() against child-thrown exceptions.

akarnokd:
- Made cache() more eager to check for unsubscription in its replay loop
  - Fixed autoConnect() triggering a second call to onStart
  - Added cache() tests to see replay().autoConnect() works equivalently.
  - Changed the way cache() connects to the source, removed unnnecessary synchronization of incoming events.


abersnaze:
Just checked on the canary and it didn't seem to change the slope of the line.

akarnokd:
Could you try this PR and use replay().autoConnect() in Observable.cache()?

PR #3089, opened 2015-07-18
add code quality plugins to build.gradle

davidmoten:
This PR adds plugins for 

* findbugs
* jacoco (code coverage)

The plugins do not run by default and are run as below:

`./gradlew findbugsMain` then open `build/reports/findbugs/main.html`.

`./gradlew test jacocoTestReport` then open `build/reports/jacoco/index.html`

Coverage can be run on a single test like so:

`./gradlew -Dtest.single=OperatorRetry test jacocoTestReport`

All these instructions are in comments in `build.gradle`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,14 +7,52 @@ description = 'RxJava: Reactive Extensions for the JVM – a library for composi
 
 apply plugin: 'rxjava-project'
 apply plugin: 'java'
+apply plugin: 'findbugs'
+apply plugin: 'jacoco'
 
 dependencies {
     testCompile 'junit:junit-dep:4.10'
     testCompile 'org.mockito:mockito-core:1.8.5'
 }
 
+////////////////////////////////////////////////////////////////////
+// to run findbugs: 
+//     ./gradlew findbugsMain
+// then open build/reports/findbugs/main.html
+////////////////////////////////////////////////////////////////////
+
+findbugs {
+  ignoreFailures = false
+  toolVersion = '+'
+  // disable findbugs for default runs of `check` or `build`
+  sourceSets = [] 
+  reportsDir = file("$project.buildDir/reports/findbugs")
+  effort = 'max'
+}
+
+//////////////////////////////////////////////////////////////////
+// to run jacoco: 
+//     ./gradlew test jacocoTestReport
+// to run jacoco on a single test (matches OperatorRetry to OperatorRetryTest in test code base):
+//     ./gradlew -Dtest.single=OperatorRetry test jacocoTestReport
+// then open build/reports/jacoco/index.html
+/////////////////////////////////////////////////////////////////
+
+jacoco {
+    toolVersion = '+'
+    reportsDir = file("$buildDir/customJacocoReportDir")
+}
+
+jacocoTestReport {
+    reports {
+        xml.enabled false
+        csv.enabled false
+        html.destination "${buildDir}/reports/jacoco"
+    }
+}
+
 javadoc {
-    exclude "**/rx/internal/**"
+    exclude '**/rx/internal/**'
 }
 
 // support for snapshot/final releases with the various branches RxJava uses
@@ -28,5 +66,12 @@ if (project.hasProperty('release.useLastTag')) {
 }
 
 test{
-     maxHeapSize = "2g"
+     maxHeapSize = '2g'
 }
+
+tasks.withType(FindBugs) {
+    reports {
+        xml.enabled = false
+        html.enabled = true
+    }
+ }


build.gradle
```diff
@@ -7,12 +7,50 @@ description = 'RxJava: Reactive Extensions for the JVM – a library for composi
...
 }
 
+////////////////////////////////////////////////////////////////////
+// to run findbugs: 
+//     ./gradlew findbugsMain
+// then open build/reports/findbugs/main.html
+////////////////////////////////////////////////////////////////////
+
+findbugs {
+  ignoreFailures = true
```
artem-zinnatullin: But why? There is almost no sense in such tools without enforcing build failure.

build.gradle
```diff
@@ -7,12 +7,50 @@ description = 'RxJava: Reactive Extensions for the JVM – a library for composi
...
+  toolVersion = "+"
```
artem-zinnatullin: Double quotes can be replaced with single quotes (disables string interpolation -> time-- && speed++).

build.gradle
```diff
@@ -7,12 +7,50 @@ description = 'RxJava: Reactive Extensions for the JVM – a library for composi
...
+  // disable findbugs for default runs of `check` or `build`
+  sourceSets = [] 
+  reportsDir = file("$project.buildDir/reports/findbugs")
+  effort = "max"
```
artem-zinnatullin: Double quotes can be replaced with single quotes.

artem-zinnatullin:
I'd also suggest Jacoco-Coverage Plugin: https://github.com/palantir/gradle-jacoco-coverage as addition to Jacoco, with this plugin you can set min coverage level and fail the build if coverage level is not enough, standard Jacoco plugin does not have this feature.

You can find an example in our project StorIO: https://github.com/pushtorefresh/storio/blob/jacoco/gradle/jacoco-android.gradle (will be in master as soon as I write all tests for private constructors and not covered cases :smile:).

build.gradle
davidmoten: Yep thanks, I can make that value false now that I have figured out how to exclude findbugs from default build.

build.gradle
davidmoten: ok, ta

build.gradle
artem-zinnatullin: Please noo, my initial intent — run code quality tools in all regular builds and fail them if one of the tool found some problem, some cases can be suppressed of course, but let's use everything to improve and verify code quality?

In our projects we even trying to enforce compiler warnings as errors (if it's possible, but some kind of warnings can not be suppressed…).

davidmoten: Yep I support that too. Steps would probably be
* get findbugs plugin in so that anyone can run it, not part of default build
* review output from findbugs plugin, configure exclusions
* decide if we want it part of the default build

Setting failure coverage thresholds for jacoco would be trickier as there are a lot of paths in the codebase that are triggered only in quite unusual concurrent circumstances and are tricky to induce in unit tests. I have been missing this sort of tool when zeroing in on particular operators for the purposes of bug finding and contributing new code.

davidmoten:
Updated PR with improvements suggested by @artem-zinnatullin and rebased.

artem-zinnatullin:
LGTM :+1: 

I think I'll try to integrate it with Jacoco-Coverage-Plugin later and enforce min code coverage on the project if you don't mind?

If some cases will be untestable — we will just set up lower min coverage level for such classes (it's possible via this plugin).

davidmoten:
@artem-zinnatullin I don't mind, you could raise an issue to see if the project owners are interested. Running jacoco on the codebase certainly suggests lots of places to contribute tests at the moment.

akarnokd:
The jacoco looks interesting, but I've installed the Eclipse plugin instead. I don't know how to extract the report from Travis and without it, failing the build won't help much.

If someone decides to do a coverage-hunt (which besides me no one seems to have done so far), then he/she can use an IDE plugin or temporarily modify the build.gradle in his/her branch for the duration of the evaluations.

I vote no for this PR.

artem-zinnatullin:
>  I don't know how to extract the report from Travis and without it, failing the build won't help much.

This PR can be improved (by me, for example) to make results visible (only for fails) in the output on the Travis CI and regular builds from the terminal, you will see something like this: 

```
Found the following Jacoco coverage violations
SomeClass.java (280/450 INSTRUCTION coverage < 0.8)
SomeClass.java (52/150 LINE coverage < 0.8)
:rxjava:checkCoverage FAILED
```

The main purpose of code coverage tools — run them on the CI to detect when new or changed code has insufficient coverage.

davidmoten:
I'd like to have a code coverage tool available from the command line without having to deal with IDE plugins (which can be fickle across IDE versions for starters). Wth command line at least people with different IDEs can inspect a common view.

I did not make this PR to impose coverage thresholds but to be able to identify possible problem areas in the code base generally and to help assess modifications (mine and others).

The plugins do not run by default in this PR. Having to transplant the plugins (and keep the build.gradle in sync) into the main line or another branch whenever I want to assess stuff is a giant pain.

CI integration would be great at some point but as this stuff is non-trivial to put into build.gradle (getting stuff to not run by default was a painful experience for example) and offers benefits at no cost to the default build I'd like to see it go in. Let's leave talk of build failures in CI for another PR from @artem-zinnatullin  if he wants to chase it.

davidmoten:
By the way, we are unlikely to see many people go on a coverage hunt without making it easy, this PR makes it easy.

stealthcode:
Can the plugins be added by a gradle init script (using `-I` or `--init-script`)? In this way an individual developer could include a script for coverage or findbugs. This way RxJava doesn't have to support all types of plugins or tools at this moment and theres no way that the build could be impacted.

artem-zinnatullin:
Plugins must be configured so you should also pass some configuration params or even Groovy script and I am pretty sure that nobody will do this.

Perfect scenario for coverage tool integrated into CI builds — PR with new code that does not have enough coverage, imagine `if` statement in the code with test that covers only one branch of this `if`, reviewers might not see that one of the `if` branches is not covered, but coverage tool will detect this and world will be saved.

davidmoten:
@stealthcode that sounds fine to me. As long as the plugins can be activated in an automated fashion from the command line I'm happy. In the meantime the build is not impacted by the current changes and such a modification could occur in a secondary PR.

davidmoten:
@artem-zinnatullin thanks for the examples about the value of code coverage in CI. I'd like to see that usage pursued in a separate issue/PR pending merge of this PR just so we can get this base functionality in for people to use manually. One step at a time, I suppose.

artem-zinnatullin:
Sure, I'll submit a separate PR if this one will be merged.

stealthcode:
The RxJava project doesn't have an opinion on code coverage tools. However I think it's useful to have standard files or documented procedures so that contributors can easily opt-in to running test coverage reports or other build tools. Would you be in favor of changing this PR to move the necessary bits to a `./tools/coverage-jacoco.gradle` and `./tools/findbugs.gradle`? In this way they could be run with a couple `-I` options and the project can maintain an ambivalence with respect to build tools. Thoughts?

davidmoten:
Yep I'm happy with that @stealthcode. I'll see if it works and update the PR.

stealthcode:
I'm going to close this issue but I encourage you to submit a PR to add the mentioned init files. Thanks

Issue #3095, opened 2015-07-21
Observable.take() on a BehaviorSubject swallows "fatal" errors.


mikebaum:
It seems that if your subscriber throws a "fatal" exception when subscribing to a BehaviorSubject chained with just, the fatal there error is swallowed and nothing is logged. See the following example:

```
    public static void main( String[] args )
    {
        BehaviorSubject<String> strings = BehaviorSubject.create( "First" );
        
        RxJavaPlugins.getInstance().registerErrorHandler( new RxJavaErrorHandler()
        {
            public void handleError( Throwable throwable )
            {
                System.err.println( throwable );
            };
        } );
        
        strings.take( 1 ).subscribe( string -> { throw new NoClassDefFoundError(); },
                                     System.err::println );
    }
```

If the error thrown from the subscribe function is anything but a "fatal" error as determined by `Exceptions.throwIfFatal(Throwable)` an error is properly propagated.

akarnokd:
This is a strange situation: take sets its completed flag before the emission and when the exception is bounced back, it is not propagated any further. Looks like a bug.

mikebaum:
@akarnokd Thanks for the quick bug fix!

akarnokd:
The bug is fixed and is now part of 1.0.14. Thanks for reporting.

PR #3101, opened 2015-07-23
Fix take swallowing exception if thrown by exactly the nth onNext call to it.

akarnokd:
Fixes the case uncovered by #3095.


src/main/java/rx/internal/operators/OperatorTake.java
```diff
@@ -56,20 +57,27 @@ public void onCompleted() {
...
                 }
             }
 
             @Override
             public void onNext(T i) {
                 if (!isUnsubscribed()) {
-                    if (++count >= limit) {
-                        completed = true;
-                    }
+                    boolean c = ++count >= limit;
```
artem-zinnatullin: Maybe comment or better variable name? Also, it can be `final`.

benjchristensen:
Odd bug. Thanks for the fix.

Issue #3098, opened 2015-07-23
materialize() and dematerialize() backpressure support


davidmoten:
Just a note that I'm working on PRs for backpressure support for `materialize()` and `dematerialize()`. Let me know if you are already working on this.

The problems are these
* `materialize` can emit one more than requested (the completion or  error notification)
* `dematerialize` on a source of N items should emit N-1 items but requires a request of N to also report the completion event

akarnokd:
Solved via #3103

PR #3103, opened 2015-07-24
materialize() - add backpressure support

davidmoten:
As mentioned in #3098 the existing version of  `materialize` could deliver one more event than requested being the termination event (completion or error).

This PR ensures that a termination event is buffered till requested.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMaterialize.java
+++ b/src/main/java/rx/internal/operators/OperatorMaterialize.java
@@ -15,8 +15,11 @@
  */
 package rx.internal.operators;
 
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+
 import rx.Notification;
 import rx.Observable.Operator;
+import rx.Producer;
 import rx.Subscriber;
 import rx.plugins.RxJavaPlugins;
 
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorMaterialize.java
```diff
@@ -29,41 +32,147 @@
...
-            public void onError(Throwable e) {
-                RxJavaPlugins.getInstance().getErrorHandler().handleError(e);
-                child.onNext(Notification.<T> createOnError(e));
-                child.onCompleted();
+    private static class ParentSubscriber<T> extends Subscriber<T> {
+
+        private final Subscriber<? super Notification<T>> child;
+
+        // guarded by this
+        private Notification<T> terminalNotification;
```
akarnokd: This can be made volatile and there is no need to synchronize to read its value.

src/main/java/rx/internal/operators/OperatorMaterialize.java
```diff
@@ -29,41 +32,147 @@
...
+        private void drain() {
+            synchronized (this) {
+                if (busy) {
+                    // set flag to force extra loop if drain loop running
+                    missed = true;
+                    return;
+                } else {
+                    // is our turn to run drain loop so set missed to false
+                    // so we can detect missed drain calls
+                    missed = false;
```
akarnokd: If not busy, missed is always false (due to exhaustive drain) so no need to set it here.

davidmoten:
Thanks @akarnokd, I've updated with those fixes and rebased.

davidmoten:
A quick question about synchronized blocks vs using volatile variables in general. It would also be possible to atomically change a pair like `busy` and `missed` using an immutable wrapping object `BusyAndMissed` and storing it in an `AtomicReference<BusyAndMissed>` and use spinning to modify. It could involve some GC pressure if rapidly changing of course. Have you any experience of advantages/disadvantages with this approach?

akarnokd:
That approach is wasteful. You only need an AtomicInteger and its increment/decrement methods. 0 means not busy, 1+ means busy and 2+ means there is still work to be done. The pair of booleans + synchronized, however, plays into the hands of lock elision and biased locking which affects our benchmarks. If it were up to me, I'd use atomics everywhere so we avoid the chance of unnecessary thread suspension due to a lock not available around 2 bytes.

davidmoten:
Thanks, quite right for that pair of objects, was probably a bad example. So under circumstances where we are getting significant lock elision and biased locking is it also the case that the volatile read and write of `terminalNotification` is just like a non-volatile read and write? I left it as non-volatile to favour lock elision and biased locking in the first place.

akarnokd:
That is a terminal event that happens once, but when the lock optimizations don't happen, you may get a hefty delay. Paying a few dozen nanoseconds for a single volatile write is way better.

davidmoten:
Makes sense, thanks.

akarnokd:
Thanks!

PR #3104, opened 2015-07-24
Fix SynchronizedQueue.equals

davidmoten:
Courtesy of findBugs, `SynchronizedQueue.equals()` is improperly implemented. I used Eclipse to generate a new `equals()` method based on the `list` field.

The existing code would even have failed to return true when testing queue equality with itself.

I don't think this is causing problems anywhere but this PR will ensure that it doesn't cause a problem in the future.


akarnokd:
SynchronizedQueue should be replaced with one of the JCTools' atomic-based bounded queues instead.

akarnokd:
Thanks.

Issue #3106, opened 2015-07-24
OperatorOnErrorFlatMap - can we remove?


davidmoten:
The class `OperatorOnErrorFlatMap` is not used and I see bugs in its implementation anyway (unsubscribe of `child` does not unsubscribe `resume` subscription). Can we remove?

benjchristensen:
Yes

PR #3108, opened 2015-07-24
remove OperatorOnErrorFlatMap because unused

davidmoten:
as per issue #3106, this PR removes `OperatorOnErrorFlatMap` because it is unused.


akarnokd:
Thanks.

PR #3117, opened 2015-07-27
Operator replay() now supports backpressure (again)

akarnokd:
And applied the hardenings as well.


akarnokd:
I've merged this and cache() so people can look at the code and try out the snapshot versions.

Issue #3003, opened 2015-06-02
API for Creating Safe Back Pressure Observables


stealthcode:
# The Problem:

Creating Back-Pressure Enabled Observables that produce data from a non-blocking source is hard and our current APIs for creating arbitrary observables do not lend towards creating observables that respect the back pressure rules. Users have a choice between calling one of the Observable static method to generate standard case observables using well known patterns or they can attempt to implement a responsible observable out of an `OnSubscribe<T>`. I will define a "responsible Observable" to be an observable that does the following:
- calls `subscriber.onNext(t)` only when the `subscriberCapacity > 0` where `subscriberCapacity` is the sum of the integer arguments for all invocations of `producer.request(n)` minus the number of calls made to `subscriber.onNext(t)`.
- calls `subscriber.onCompleted()` or `subscriber.onError(e)` only once.
- does not call `subscriber.onNext(t)` concurrently.

The onSubscribe lifecycle of a responsible observable should fit within the following lifecycle.
1. The `onSubscribe.call(Subscriber<T>)` method is invoked by the `Observable.subscribe(Subscriber<T>)` method.
2. The OnSubscribe func creates a `Producer` via the `subscriber.setProducer(Producer<T>)` method. At some point in the (potentially distant) future the subscriber may call `producer.request(n)`. This signals to the observable that capacity is allocated  in the subscriber.
   3a. Values are produced via a call to `subscriber.onNext(t)`.
   3b. At some point a terminal event is emitted via `onCompleted` or `onError`.

This allows for many different variations of control flow given that onNexts could possibly be scheduled, data could be pre-fetched, excess data could be batched or dropped, etc...
# Proposal

With this ground work in mind I would like to start a discussion about a general framework that correctly models the various use cases, particularly with respects to the following requirements:
- capable of paginating data across multiple requests
- batching requests
- hot and cold data sources
- async and blocking onNexts
## Naive Implementation

User provides callbacks that the framework will call between each of the steps above. That is
1. Observable calls `onSubscribe.call(s)`
2. The framework's onSubscribe calls: `S dataState = onSubscribeGenerator.call()`
3. Next the onSubscribe calls: `subscriber.setProducer(p)` setting a framework Producer
4. At some point the request is received by our producer `p.request(n)`
5. Then the framework will call the user provided request callback and store the new resultant `dataState` for future requests. `dataState = onRequestCallback.call(dataState, n, s)`

[...]

PR #3118, opened 2015-07-29
Implementing the SyncOnSubscribe

stealthcode:
This is useful for creating an observable that reads from a data source in a synchronous manner. This is a rewrite of the `AbstractOnSubscribe` experimental class based on the ideas in #3003.

### DIFF
--- a/src/main/java/rx/observables/SyncOnSubscribe.java
+++ b/src/main/java/rx/observables/SyncOnSubscribe.java
@@ -1,0 +1,465 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package rx.observables;
+
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicLong;
+
+import rx.Observable.OnSubscribe;
+import rx.Observer;
+import rx.Producer;
+import rx.Subscriber;
+import rx.Subscription;
+import rx.annotations.Experimental;
+import rx.functions.Action0;
+import rx.functions.Action1;
+import rx.functions.Action2;
+import rx.functions.Func0;
+import rx.functions.Func2;
+import rx.internal.operators.BackpressureUtils;
+import rx.plugins.RxJavaPlugins;
+
+/**
+ * A utility class to create {@code OnSubscribe<T>} functions that respond correctly to back
+ * pressure requests from subscribers. This is an improvement over
+ * {@link rx.Observable#create(OnSubscribe) Observable.create(OnSubscribe)} which does not provide
+ * any means of managing back pressure requests out-of-the-box.
+ *
+ * @param <S>
+ *            the type of the user-define state used in {@link #generateState() generateState(S)} ,
+ *            {@link #next(Object, Subscriber) next(S, Subscriber)}, and
+ *            {@link #onUnsubscribe(Object) onUnsubscribe(S)}.
+ * @param <T>
+ *            the type of {@code Subscribers} that will be compatible with {@code this}.
+ */
+@Experimental
+public abstract class SyncOnSubscribe<S, T> implements OnSubscribe<T> {
+    
+    /* (non-Javadoc)
+     * @see rx.functions.Action1#call(java.lang.Object)
+     */
+    @Override
+    public final void call(final Subscriber<? super T> subscriber) {
+        S state = generateState();
+        SubscriptionProducer<S, T> p = new SubscriptionProducer<S, T>(subscriber, this, state);
+        subscriber.add(p);
+        subscriber.setProducer(p);
+    }
+
+    /**
+     * Executed once when subscribed to by a subscriber (via {@link OnSubscribe#call(Subscriber)})
+     * to produce a state value. This value is passed into {@link #next(Object, Observer) next(S
+     * state, Observer <T> observer)} on the first iteration. Subsequent iterations of {@code next}
+     * will receive the state returned by the previous invocation of {@code next}.
+     * 
+     * @return the initial state value
+     */
+    protected abstract S generateState();
+
+    /**
+     * Called to produce data to the downstream subscribers. To emit data to a downstream subscriber
+     * call {@code observer.onNext(t)}. To signal an error condition call
+     * {@code observer.onError(throwable)} or throw an Exception. To signal the end of a data stream
+     * call {@code 
+     * observer.onCompleted()}. Implementations of this method must follow the following rules.
+     * 
+     * <ul>
+     * <li>Must not call {@code observer.onNext(t)} more than 1 time per invocation.</li>
+     * <li>Must not call {@code observer.onNext(t)} concurrently.</li>
+     * </ul>
+     * 
+     * The value returned from an invocation of this method will be passed in as the {@code state}
+     * argument of the next invocation of this method.
+     * 
+     * @param state
+     *            the state value (from {@link #generateState()} on the first invocation or the
+     *            previous invocation of this method.
+     * @param observer
+     *            the observer of data emitted by
+     * @return the next iteration's state value
+     */
+    protected abstract S next(S state, Observer<? super T> observer);
+
+    /**
+     * Clean up behavior that is executed after the downstream subscriber's subscription is
+     * unsubscribed. This method will be invoked exactly once.
+     * 
+     * @param state
+     *            the last state value prior from {@link #generateState()} or
+     *            {@link #next(Object, Observer) next(S, Observer&lt;T&gt;)} before unsubscribe.
+     */
+    protected void onUnsubscribe(S state) {
+
+    }
+
+    /**
+     * Generates a synchronous {@link SyncOnSubscribe} that calls the provided {@code next} function
+     * to generate data to downstream subscribers.
+     * 
+     * @param generator
+     *            generates the initial state value (see {@link #generateState()})
+     * @param next
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @return an OnSubscribe that emits data in a protocol compatible with back-pressure.
+     */
+    @Experimental
+    public static <S, T> OnSubscribe<T> createSingleState(Func0<? extends S> generator, 
+            final Action2<? super S, ? super Observer<? super T>> next) {
+        Func2<S, ? super Observer<? super T>, S> nextFunc = new Func2<S, Observer<? super T>, S>() {
+            @Override
+            public S call(S state, Observer<? super T> subscriber) {
+                next.call(state, subscriber);
+                return state;
+            }
+        };
+        return new SyncOnSubscribeImpl<S, T>(generator, nextFunc);
+    }
+
+    /**
+     * Generates a synchronous {@link SyncOnSubscribe} that calls the provided {@code next} function
+     * to generate data to downstream subscribers.
+     * 
+     * This overload creates a SyncOnSubscribe without an explicit clean up step.
+     * 
+     * @param generator
+     *            generates the initial state value (see {@link #generateState()})
+     * @param next
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @param onUnsubscribe
+     *            clean up behavior (see {@link #onUnsubscribe(Object) onUnsubscribe(S)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
+    public static <S, T> OnSubscribe<T> createSingleState(Func0<? extends S> generator, 
+            final Action2<? super S, ? super Observer<? super T>> next, 
+            final Action1<? super S> onUnsubscribe) {
+        Func2<S, Observer<? super T>, S> nextFunc = new Func2<S, Observer<? super T>, S>() {
+            @Override
+            public S call(S state, Observer<? super T> subscriber) {
+                next.call(state, subscriber);
+                return state;
+            }
+        };
+        return new SyncOnSubscribeImpl<S, T>(generator, nextFunc, onUnsubscribe);
+    }
+
+    /**
+     * Generates a synchronous {@link SyncOnSubscribe} that calls the provided {@code next} function
+     * to generate data to downstream subscribers.
+     * 
+     * @param generator
+     *            generates the initial state value (see {@link #generateState()})
+     * @param next
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @param onUnsubscribe
+     *            clean up behavior (see {@link #onUnsubscribe(Object) onUnsubscribe(S)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
+    public static <S, T> OnSubscribe<T> createStateful(Func0<? extends S> generator, 
+            Func2<? super S, ? super Observer<? super T>, ? extends S> next, 
+            Action1<? super S> onUnsubscribe) {
+        return new SyncOnSubscribeImpl<S, T>(generator, next, onUnsubscribe);
+    }
+    
+    /**
+     * Generates a synchronous {@link SyncOnSubscribe} that calls the provided {@code next} function
+     * to generate data to downstream subscribers.
+     * 
+     * @param generator
+     *            generates the initial state value (see {@link #generateState()})
+     * @param next
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
+    public static <S, T> OnSubscribe<T> createStateful(Func0<? extends S> generator, 
+            Func2<? super S, ? super Observer<? super T>, ? extends S> next) {
+        return new SyncOnSubscribeImpl<S, T>(generator, next);
+    }
+
+    /**
+     * Generates a synchronous {@link SyncOnSubscribe} that calls the provided {@code next} function
+     * to generate data to downstream subscribers.
+     * 
+     * This overload creates a "state-less" SyncOnSubscribe which does not have an explicit state
+     * value. This should be used when the {@code next} function closes over it's state.
+     * 
+     * @param next
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
+    public static <T> OnSubscribe<T> createStateless(final Action1<? super Observer<? super T>> next) {
+        Func2<Void, Observer<? super T>, Void> nextFunc = new Func2<Void, Observer<? super T>, Void>() {
+            @Override
+            public Void call(Void state, Observer<? super T> subscriber) {
+                next.call(subscriber);
+                return state;
+            }
+        };
+        return new SyncOnSubscribeImpl<Void, T>(nextFunc);
+    }
+
+    /**
+     * Generates a synchronous {@link SyncOnSubscribe} that calls the provided {@code next} function
+     * to generate data to downstream subscribers.
+     * 
+     * This overload creates a "state-less" SyncOnSubscribe which does not have an explicit state
+     * value. This should be used when the {@code next} function closes over it's state.
+     * 
+     * @param next
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @param onUnsubscribe
+     *            clean up behavior (see {@link #onUnsubscribe(Object) onUnsubscribe(S)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
+    public static <T> OnSubscribe<T> createStateless(final Action1<? super Observer<? super T>> next, 
+            final Action0 onUnsubscribe) {
+        Func2<Void, Observer<? super T>, Void> nextFunc = new Func2<Void, Observer<? super T>, Void>() {
+            @Override
+            public Void call(Void state, Observer<? super T> subscriber) {
+                next.call(subscriber);
+                return null;
+            }
+        };
+        Action1<? super Void> wrappedOnUnsubscribe = new Action1<Void>(){
+            @Override
+            public void call(Void t) {
+                onUnsubscribe.call();
+            }};
+        return new SyncOnSubscribeImpl<Void, T>(nextFunc, wrappedOnUnsubscribe);
+    }
+
+    /**
+     * An implementation of SyncOnSubscribe that delegates
+     * {@link SyncOnSubscribe#next(Object, Subscriber)}, {@link SyncOnSubscribe#generateState()},
+     * and {@link SyncOnSubscribe#onUnsubscribe(Object)} to provided functions/closures.
+     *
+     * @param <S>
+     *            the type of the user-defined state
+     * @param <T>
+     *            the type of compatible Subscribers
+     */
+    private static final class SyncOnSubscribeImpl<S, T> extends SyncOnSubscribe<S, T> {
+        private final Func0<? extends S> generator;
+        private final Func2<? super S, ? super Observer<? super T>, ? extends S> next;
+        private final Action1<? super S> onUnsubscribe;
+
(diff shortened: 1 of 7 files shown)


akarnokd:
You can't delete `AbstractOnSubscribe` because it is part of the public API and even though marked as experimental, such components can be deleted only at a major release such as 1.1.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,292 @@
...
+
+	/**
+	 * Contains the producer loop that reacts to downstream requests of work.
+	 *
+	 * @param <T>
+	 *            the type of compatible Subscribers
+	 */
+	private static final class SubscriptionProducer<T> implements Producer {
+		private final AtomicLong requestCount = new AtomicLong(0);
+		private final AtomicLong sentCount = new AtomicLong(0);
```
akarnokd: sentCount appears to be accessed only in the emission loop so there is no need for an `AtomicLong` here or any other type besides a boolean since only 1 onNext is allowed ber callback.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	@Override
+	public final void call(final Subscriber<? super T> subscriber) {
+		state = generateState();
+		final SyncOnSubscribe<S, T> self = this;
+		final Subscription s = Subscriptions.create(new Action0() {
+			@Override
+			public void call() {
+				self.onUnsubscribe(state);
```
akarnokd: This allows unsubscribing the state while an onNext is in progress which requires extra care for the state object.

src/test/java/rx/observables/SyncOnSubscribeTest.java
```diff
@@ -0,0 +1,661 @@
...
+    @Test
+    public void testInfiniteRequestSome() {
+        final int finalCount = 4000;
+        final int start = 0;
+        SyncOnSubscribe<Integer, Integer> aos = new SyncOnSubscribe<Integer, Integer>() {
+        	@Override
+        	protected Integer generateState() {
+        		return start;
+        	}
+			@Override
```
akarnokd: Please make sure you don't have tab characters in the source.

src/main/java/rx/observables/SyncOnSubscribe.java
abersnaze: We'ren't sure how much synchronization we needed to insure that a bad implementation didn't call onNext concurrently.  The idea was we could use something like this on line 255
```
long old = sendCount.get();
...
subscriber.onNext(value);
if (!sentCount.compareAndSet(old, old+1)) {
    throw new IllegalStateException("concurrent onNext detected");
}
```
instead of `incrementAndGet()` to detect that something bad was happening as soon as possible.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,292 @@
...
+					if (!hasTerminated) {
+						hasTerminated = true;
+						subscriber.onError(e);
+					}
+				}
+
+				@Override
+				public void onNext(T value) {
+					if (!hasTerminated) {
+						if (sentCount.incrementAndGet() > 1)
```
abersnaze: Do any of the test onNext more than once?

benjchristensen:
Experimental APIs can be removed or change at any time, as per the README:

> @Experimental

> APIs marked with the @Experimental annotation at the class or method level will almost certainly change. They can be modified in any way, or even removed, at any time. You should not use or rely on them in any production code. They are purely to allow broad testing and feedback.

akarnokd:
I would have expected to make such APIs @Deprecated for a few releases before completely removing it. Besides, this PR just gives an alternative to another @Experimental class and could live side-by-side. Then users could decide/vote which to keep at the next major release.

stealthcode:
@akarnokd you have a good point that the efficacy of both implementations should be compared however I do not think that we should expose multiple alternatives in the public api. This ambiguous api leads to confusion and questions about the unexplained differences between one or the other implementation. The pull request should not be the place for debate either (since this should be about issues with THIS implementation). This comparison and debate was what I was hoping to get out of opening issue #3003. Would you like to open a separate issue for comparing the 2 implementations?

src/test/java/rx/observables/SyncOnSubscribeTest.java
stealthcode: Sorry I updated my ide and it overrode my tabs/spaces config.

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: Thanks, I have implemented work stealing to process the onUnsubscribe after the final next invocation.

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: Thanks. I've integrated this into my latest code.

stealthcode:
The latest commit adds work stealing to manage unsubscribe to avoid concurrent modification of `state` while running `onUnsubscribe` concurrently with `next` (and a test to exercise), fixes issues with tabs/spaces, and adds javadocs.

stealthcode:
The latest commit makes a few changes. 

* Added the `AbstractOnSubscribe` back into the code base. The `@Experimental` annotation states clearly that either of these implementations can be removed at any time so it's not necessary for anything to be removed now. It is up to the consumer when they chose to use a potentially unstable api. 
* Added JMH tests to compare performance against `OnSubscribeFromIterable`. The performance is impacted in astounding ways by the additional complexity of a cleanup function. This requires object initialization which the `OnSubscribeFromIterable` does not require. 
* General optimizations. Removing a volatile and reorganizing the request loop.

davidmoten:
@stealthcode can you give us more detail on the JMH comparison?

stealthcode:
Sure, sorry for leaving that out. Here's the JMH report output

```
$ ./gradlew clean benchmarks '-Pjmh=-f 1 -tu s -bm thrpt -wi 5 -i 5 -r 1 .*SyncOnSubscribePerf.*'
```

```
Benchmark                                            (size)   Mode   Samples        Score  Score error    Units
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe          1  thrpt         5 10345304.158   167899.030    ops/s
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe2         1  thrpt         5 10315525.503   519004.209    ops/s
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe2      1000  thrpt         5 10080651.937  1249022.786    ops/s
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe2   1000000  thrpt         5  9676927.741   799435.391    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable                 1  thrpt         5 245301608.512 25693364.239    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2                1  thrpt         5 261726979.345 17122836.564    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2             1000  thrpt         5 256818323.006 17992719.065    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2          1000000  thrpt         5 259483656.124 12863102.025    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe              1  thrpt         5 12770780.794   146195.400    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2             1  thrpt         5 12306631.098  2072808.101    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2          1000  thrpt         5 12326658.468   682619.251    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2       1000000  thrpt         5 12519473.508   365884.932    ops/s
```

Under Flight Recorder analysis, both the AOS and SOS suffer from the cost of creating a `SubscriptionList` and allocating memory where as a the `OnSubscribeFromIterable` does not require adding an unsubscribe action so doesn't need to allocate. I plan on testing some changes to get around this bottle neck. Will submit a pull request soon. Hopefully this upcoming change will be a private change and not have to wait until 2.x.

src/perf/java/rx/observables/SyncOnSubscribePerf.java
```diff
@@ -0,0 +1,111 @@
...
+                else
+                    subscriber.onCompleted();
+                return null;
+            }};
+    }
+    
+    @Benchmark
+//  @Group("single")
+    public void benchSyncOnSubscribe(final SingleInput input) {
+        createSyncOnSubscribe(input.iterator).call(input.newSubscriber());
```
akarnokd: The problem with this is that the iterator gets consumed once and you then benchmark how fast an empty iterator comes to completion through the operator.

stealthcode: Oh I see Thanks @akarnokd ! I'm correcting and re-running now.

stealthcode:
Here are the updated benchmarks. This paints a very different picture than last time. The performance of the `SyncOnSubscribe` is now ~25% slower than `OnSubscribeFromIterable` for small iterators and ~17% slower for large iterators. The bottle neck seems to be in the `SyncOnSubscribe$SubscriptionProducer.request(long)` method.

```
# Run complete. Total time: 00:02:32

Benchmark                                            (size)   Mode   Samples        Score  Score error    Units
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe          1  thrpt         5  8057762.938   350532.503    ops/s
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe2         1  thrpt         5  8150650.148   174362.683    ops/s
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe2      1000  thrpt         5    41560.383      737.346    ops/s
r.o.SyncOnSubscribePerf.benchAbstractOnSubscribe2   1000000  thrpt         5       41.237        8.422    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable                 1  thrpt         5 19134647.666   420991.146    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2                1  thrpt         5 18720840.805   327775.879    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2             1000  thrpt         5   183561.551     5244.489    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2          1000000  thrpt         5      184.511        4.020    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe              1  thrpt         5 11358568.468   298560.809    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2             1  thrpt         5 11367669.868   133428.078    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2          1000  thrpt         5   132173.869     2670.427    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2       1000000  thrpt         5      130.224        3.425    ops/s
```

akarnokd:
Here is a version where the integer stream is freshly created:

image

(i7 4770K, Windows 7 x64, Java 8u51)

AbstractOnSubscribe (AOS) has the highest overhead because of resource checks on each emission

AbstractProducer (AP from #2813) offers amortized resource check overhead due to the inner loop but this loop was not particularly optimized with fast-path and/or amortized request accounting.

FromIterable (FI) doesn't care about resources and has fast-path.

SyncOnSubscribe (SO) in this PR is close but 33% behind FI. I've made [some optimizations](https://gist.github.com/akarnokd/a2669817708827350663) to the fast-path and got a much better result for longer sources:

```
13969673,045   263281,521    ops/s
  237194,622     1475,414    ops/s
     226,248       11,892    ops/s
```

One must be careful around volatile reads because it forces the re-read of instance variables after them.

stealthcode:
Nice, I'm merging your gist. This optimization makes the fast path on-par with FromIterable in the long tail (still however ~25% slower in the 1's - 10's range).

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,458 @@
...
+ * @param <T>
+ *            the type {@code Subscribers} that will be compatible with {@code this}.
+ */
+@Experimental
+public abstract class SyncOnSubscribe<S, T> implements OnSubscribe<T> {
+    /**
+     * Used to thread the state synchronously to every iteration of the
+     * {@link #next(Object, Subscriber) next} function.
+     */
+    private S state;
```
akarnokd: The `state` should be per subscriber and not shared between them.

stealthcode:
I spoke too soon. In running the unit tests there are now problems with the diff applied. I'll try to incorporate what I can and test to make sure the changes are more performant.

akarnokd:
Sorry, I didn't run any test so my changes were only valid for the perf usage and not general. [Here is](https://gist.github.com/akarnokd/dffd7500981968a845d3) a version which passes the tests and has the following performance characteristics:

```
Benchmark                     (size)   Mode   Samples        Score  Score error    Units
benchSyncOnSubscribe2Fresh         1  thrpt         5 18314010,215    78882,829    ops/s
benchSyncOnSubscribe2Fresh      1000  thrpt         5   223973,659     2605,564    ops/s
benchSyncOnSubscribe2Fresh   1000000  thrpt         5      228,057        1,400    ops/s
```

For size = 1, I think the extra allocation of the facade Subscriber makes it twice the overhead. The onNext safeguard also costs ~15%:

```
Benchmark                     (size)   Mode   Samples        Score  Score error    Units
benchSyncOnSubscribe2Fresh         1  thrpt         5 18356265,548   512130,369    ops/s
benchSyncOnSubscribe2Fresh      1000  thrpt         5   268613,838     7801,208    ops/s
benchSyncOnSubscribe2Fresh   1000000  thrpt         5      274,654        7,522    ops/s
```

stealthcode:
@akarnokd your changes are pretty solid. I can't argue with results. Here are the benchmarks for your changes alone.

```
Benchmark                                        (size)   Mode   Samples        Score  Score error    Units
r.o.SyncOnSubscribePerf.benchFromIterable2            1  thrpt         5 17907742.680  1458787.396    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2         1000  thrpt         5   181328.434     3485.931    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2      1000000  thrpt         5      181.733        9.492    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2         1  thrpt         5 14414752.664   572789.234    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2      1000  thrpt         5   182578.431    21732.491    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2   1000000  thrpt         5      186.404        6.581    ops/s
```

I implemented a small api change to expose a setter on an unsubscribe `Action1<S>`. With this change we can check if there's no unsubscribe (and avoid a `SubscriptionList` allocation). 
```
Benchmark                                        (size)   Mode   Samples        Score  Score error    Units
r.o.SyncOnSubscribePerf.benchFromIterable2            1  thrpt         5 17832090.877  3436213.149    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2         1000  thrpt         5   180643.547    10720.476    ops/s
r.o.SyncOnSubscribePerf.benchFromIterable2      1000000  thrpt         5      179.589       20.317    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2         1  thrpt         5 17809008.594   478600.206    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2      1000  thrpt         5   181643.455    45353.810    ops/s
r.o.SyncOnSubscribePerf.benchSyncOnSubscribe2   1000000  thrpt         5      183.044        7.280    ops/s
```

With this change we are now faster than `OnSubscribeFromIterable` in the short and long cases (assuming we don't have to cleanup with unsubscribe).

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,499 @@
...
+    @Experimental
+    public static <S, T> SyncOnSubscribe<S, T> createWithState(Func0<S> generator, final Action2<S, Subscriber<? super T>> next) {
+        Func2<S, Subscriber<? super T>, S> nextFunc = new Func2<S, Subscriber<? super T>, S>() {
+            @Override
+            public S call(S state, Subscriber<? super T> subscriber) {
+                next.call(state, subscriber);
+                return state;
+            }
+        };
+        Action1<S> onUnsubscribe = new Action1<S>() {
```
akarnokd: This will trigger the addition of the unsubscription action in `call()`.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,499 @@
...
+            onUnsubscribe.call(state);
+    }
+    
+    /**
+     * Adds an action to execute after the subscription has been unsubscribed from.
+     * 
+     * @param onUnsubscribe action to execute.
+     * @return this
+     */
+    public SyncOnSubscribe<S, T> onUnsubscribe(Action1<S> onUnsubscribe) {
```
akarnokd: I'm not sure about this being part of the API.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,499 @@
...
+    @Experimental
+    public static <S, T> SyncOnSubscribe<S, T> createWithMutableState(Func0<S> generator, final Func2<S, Subscriber<? super T>, S> next) {
+        Func2<S, Subscriber<? super T>, S> nextFunc = new Func2<S, Subscriber<? super T>, S>() {
+            @Override
+            public S call(S state, Subscriber<? super T> subscriber) {
+                next.call(state, subscriber);
+                return state;
+            }
+        };
+        Action1<S> onUnsubscribe = new Action1<S>() {
```
akarnokd: Again the resource-less optimization won't work with this.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,499 @@
...
+        Action1<Void> wrappedOnUnsubscribe = new Action1<Void>() {
+            @Override
+            public void call(Void t) {
+                onUnsubscribe.call();
+            }
+        };
+        return new LambdaOnSubscribe<Void, T>(generator, nextFunc, wrappedOnUnsubscribe);
+    }
+
+    private static final Action1<Void> VOID_ON_UNSUBSCRIBE = new Action1<Void>() {
```
akarnokd: `null` onUnsubscribe is allowed so no need for this (and would cause optimal path not taken).

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,499 @@
...
+     * {@link SyncOnSubscribe#onUnsubscribe(Object) onUnsubscribe(S)}.
+     * 
+     * @param state
+     *            the value produced by {@link #generateState()} on the first execution or the value
+     *            returned from the previous execution of {@code next(Object, Subscriber)}
+     * @param subscriber
+     *            down stream data subscriber to receive onNext, onCompleted, or onError
+     * @return the state value to be received by the next execution of this function
+     */
+    protected abstract S next(S state, Subscriber<? super T> subscriber);
```
akarnokd: Have you considered exposing only the `Observer` here? Do you want the `next` to add resources to the subscriber or check for its unsubscribed state?

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,499 @@
...
+        public void unsubscribe() {
+            if (UNSUBSCRIBED.compareAndSet(this, 0, 1) && get() == 0L) {
+                // it's safe to process terminal behavior
+                parent.onUnsubscribe(state);
+            }
+        }
+
+        @Override
+        public void request(long n) {
+            if (n > 0 && BackpressureUtils.getAndAddRequest(this, n) == 0L) {
```
akarnokd: A CAS here is slightly faster here (similar to what fromIterable uses):
```java
            if (n == Long.MAX_VALUE && compareAndSet(0, Long.MAX_VALUE)) {
                fastpath();
            } else
            if (n > 0 && BackpressureUtils.getAndAddRequest(this, n) == 0L) {
                slowPath(n);
            }
```

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: I must have missed some changes where I cleaned these in the copy pasting between our forks. I'll fix them. Thanks.

src/perf/java/rx/jmh/InputWithIncrementingInteger.java
```diff
@@ -40,46 +41,42 @@
     @Setup
     public void setup(final Blackhole bh) {
         this.bh = bh;
-        observable = Observable.range(0, getSize());
+        final Integer size = getSize();
```
akarnokd: Why `Integer` here?

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: I agree with this type signature. Good call.

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: I tried this and it wound up slower by 3% than leaving it as is.

src/perf/java/rx/jmh/InputWithIncrementingInteger.java
stealthcode: No particularly good reason. Corrected.

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: It would be much cleaner to bake into the implementation however in order to detect when one is provided at runtime and skip the `SubscriptionList` allocation we have to have a flag of some sort (which we get in the case of a `LambdaOnSubscribe`).

akarnokd:
> With this change we are now faster than OnSubscribeFromIterable

See #3137 :)

src/main/java/rx/observables/SyncOnSubscribe.java
akarnokd: Instead of this, `onUnsubscribe(S state)` could return a Subscription/Action0 or null, then make sure it reaches the proper places but not through an instance variable of `SyncOnSubscribe`.

stealthcode: That is actually implemented via the static method calls (see `SyncOnSubscribe#create...` overloads that do not have an onUnsubscribe). Using this method to set an `onUnsubscribe` is not the primary way to create a `SyncOnSubscribe`. This method provides a way to extend and still take advantage of this fast path. Normal usage would be to use the static methods. I'll see what the team thinks of this method. We might also consider converting this into a builder of an `OnSubscribe`.

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -0,0 +1,468 @@
...
+     *            produces data to the downstream subscriber (see {@link #next(Object, Subscriber)
+     *            next(S, Subscriber)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
+    public static <S, T> SyncOnSubscribe<S, T> createWithState(Func0<S> generator, final Action2<S, Observer<? super T>> next) {
+        Func2<S, Observer<? super T>, S> nextFunc = new Func2<S, Observer<? super T>, S>() {
+            @Override
+            public S call(S state, Observer<? super T> subscriber) {
```
davidmoten: Can we use parameter `observer` rather than `subscriber` now that the type has changed. This is in a few places.

stealthcode:
Highlights of the recent changes includes:
* removing the weird `onUnsubscribe(Action1)` method on `SyncOnSubscribe`.
* added new tests and fixed generics issues (not enough `?`s).
* added `Blackhole.consumeCpu()` to perf tests in order to force the micro benchmarks to artificially reduce volatile write contention.

I'm now happy with this pull request and believe it's ready to be merged.

akarnokd:
Looks good to me now, although I don't see how one could implement never() with it or a source that stops emitting but doesn't terminate.

stealthcode:
There is a test that implements never() behavior however it does not terminate where as the Observable.never() instance would simply terminate without setting a producer.

stealthcode:
Javadocs need some work here but i think that's the only thing remaining.

akarnokd:
This test times out:
```java
    @Test(timeout = 1000)
    public void testNever2() {
        OnSubscribe<Integer> os = SyncOnSubscribe.createStateless(new Action1<Observer<? super Integer>>() {
            @Override
            public void call(Observer<? super Integer> subscriber) {
                
            }});
        Observable.create(os).subscribe();
    }
```

stealthcode:
Yes it does and I think that it should remain this way. It should be documented in the contract of using the `SyncOnSubscribe` that your next function will not terminate if you do not eventually call the `observer.onCompleted()` or `observer.onError(e)`. Its similar to an `Observable.create` call with an `OnSubscribe` that synchronously loops over each `n` for all `requeqst(n)`. This would only terminate when... 

* the consumer is done (request 0)
* the producer declares that its done (via terminal event)
* an error occurs or subscription is otherwise unsubscribed (take)

akarnokd:
You could also extend the API and use some `ObserverEx` that implements `Observer` plus a `stop()` method. Then document that if you whish to not emit any further items but don't want to terminate the stream, call stop().

stealthcode:
I cannot see a valid use case of `Observable.never()` aside from merging in with an asynchronous hot observable such as `Observable.combineLatest(someFuncThatMayReturnNever(foo), hotObs)`. It seems like a tool to force merging observables in ways that obviate the need to merge at all. Are there use cases that I am missing?

That said I would rather not corrupt the api with a one-off concept of `stop()` without good reason. It's important that we get the API correct (even in @experimental).

stealthcode:
Javadoc is now updated.

abersnaze:
:+1: reviewed the code; looks good.

Issue #1914, opened 2014-12-01
Thread Interruption on Unsubscribe


benjchristensen:
A few issues recently have been related to Thread interruption and brings into question why we interrupt the thread when we unsubscribe a scheduled action (via a `Scheduler`). 

- #1913 Computation Scheduler gets interrupted thousands of times per second?
- #1898 Scheduled action no interrupt
- #1832 Fix Take Early Unsubscription Causing Interrupts
- #1804 "Interrupted while waiting for subscription to complete." in 1.0.0-rc8

The code that causes this is from https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/subscriptions/Subscriptions.java#L102 that wraps a `Subscription` around a `Future`. This in turn is used by `ScheduledAction` which is used when scheduling work via a `Scheduler`. 

Issue #1898 is working to not cause an interrupt after successful termination, but I want to question why we ever interrupt as default behavior. 

I don't think much thought went into choosing to call `cancel(true)` instead of `cancel(false)` and I'd like to discuss changing this to not interrupt. 

The implication of interrupting are difficult to manage well and add a lot of complexity. 

The only time interrupting seems wanted is if unsubscribing early to a `ScheduledAction` doing blocking IO. Otherwise the `Subscription` itself is the cancellation token that work should be checking (most code doesn't properly use `Thread.isInterrupted()` anyways so it rarely helps). 

If an `Observable` is doing blocking IO it seems it should opt itself into doing Thread interruption. Perhaps we should figure out an idiomatic approach to that for opting into that behavior without requiring interruption all the time.

akarnokd:
I have a ```ScheduledAction``` enhancements that will support customization of the interrupt behavior. See #2592.

benjchristensen:
Resolved in https://github.com/ReactiveX/RxJava/pull/2579

vinc3m1:
Following #2579 to #2592 to #2761 to #2772, it doesn't look like this is actually resolved, is it @akarnokd?

akarnokd:
None of the customizations made it and it seems Hystrix solved it internally. There were a few changes made to ScheduledAction so it doesn't call cancel(true) when cancelling from the same thread.

PR #3120, opened 2015-07-29
No InterruptedException with synchronous BlockingObservable

ypresto:
I'm using Observable for backport of Java 8's java.util.stream (i.e. [no more loops](http://www.deadcoderising.com/java-8-no-more-loops/)) for Android project.

```java
List<Foo> list2 = Observable.from(list)
        .map(...)
        .filter(...)
        .toList().toBlocking().single()
```

But it sometimes emits InterruptedException at BlockingObservable.
https://github.com/ReactiveX/RxJava/issues/1804#issuecomment-61396523

As the BlocingObservable is placed in map() of another observable with `subscribeOn(Schedulers.io())`, and it is unsubscribed from main thread, perhaps Future.cancel(true) is called on unsubscribing. (#1914)

This PR allows BlockingObsevable not to be interrupted when source observable emits synchronously, by checking current latch or queue state before awaiting for them.

### DIFF
--- a/src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
+++ b/src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
@@ -93,6 +93,10 @@ public final class BlockingOperatorToIterator {
 
             private Notification<? extends T> take() {
                 try {
+                    Notification<? extends T> poll = notifications.poll();
+                    if (poll != null) {
+                        return poll;
+                    }
                     return notifications.take();
                 } catch (InterruptedException e) {
                     subscription.unsubscribe();
--- a/src/main/java/rx/observables/BlockingObservable.java
+++ b/src/main/java/rx/observables/BlockingObservable.java
@@ -123,17 +123,7 @@ public final class BlockingObservable<T> {
                 onNext.call(args);
             }
         });
-        // block until the subscription completes and then return
-        try {
-            latch.await();
-        } catch (InterruptedException e) {
-            subscription.unsubscribe();
-            // set the interrupted flag again so callers can still get it
-            // for more information see https://github.com/ReactiveX/RxJava/pull/147#issuecomment-13624780
-            Thread.currentThread().interrupt();
-            // using Runtime so it is not checked
-            throw new RuntimeException("Interrupted while waiting for subscription to complete.", e);
-        }
+        awaitForComplete(latch, subscription);
 
         if (exceptionFromOnError.get() != null) {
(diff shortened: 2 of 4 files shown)


akarnokd:
So what happens is that you use some blocking calls inside an observable sequence and it gets interrupted because of the underlying task is cancelled? 

I believe it is an indication that the flow can be assembled non-blockingly instead. I've been fiddling with the idea that once RxJava scheduler threads become identifiable, `toBlocking()` would throw/onError if it is attempted to use it inside such schedulers.

ypresto:
> So what happens is that you use some blocking calls inside an observable sequence and it gets interrupted because of the underlying task is cancelled?

Yes. But in my case, blocking call is inside of **another (synchronous) method** and I don't want to make it non-blocking, for testability and convenience.

For example, in instagram-like app, there are entities to represent photo/movie and comment.
Both entities has entity translation logic and they can be written in for-loop or `toList().toBlocking().single()`.

```java
# MediaFileModel
public MediaFileEntity entityFromDaoEntity(MediaFileDao dao, List<CommentDao> commentDaoList) {
    List<CommentEntity> comments = Observable.from(commentDaoList);
            .map(mCommentModel::toDaoEntity)
            .filter(not(mCommentModel::isDeleted))
            .toList().toBlocking().single(); // called from where..?
    MediaFileEntity mediaFile = new MediaFileEntity();
    mediaFile.setUuid(dao.getUuid());
    mediaFile.setComments(comments);
    ...
    return mediaFile;
}
```

```java
# MediaFileStore
public Observable<List<MediaFileEntity>> allEntities() {
    return queryAll() // async operation
            .map(daoList -> {
                    Observable.from(daoList)
                            .map(mMediaFileModel::entityFromDaoEntity)
                            .toList()
            );
}
```

I think so too that asynchronous (with blocking wait) operation should be assembled non-blockingly.
But I think it is just convenient if nesting of synchronous (without blocking wait) operation is allowed. :)

akarnokd:
The changes look okay to me.

ypresto:
Thanks! :tada: :tada:

ypresto:
For your (who found this PR by searching) information:

RxJava is quite slow for blocking loop operation.
Especially for zip(), it has significant overhead of waiting for async (and even it is sync actually) observables.

So it'll better to use Java 8 Stream API or its backport for sync stream operations.
[Lightweight-Stream-API](https://github.com/aNNiMON/Lightweight-Stream-API) has only <500 methods and also suitable for Android.

PR #3121, opened 2015-07-29
Improve performance of NewThreadWorker, disable search for setRemoveOnCancelPolicy() on Android API < 21

### LINKED ISSUE #3119: NewThreadWorker.tryEnableCancelPolicy doing costly reflection on Android
I was  analyzing startup time in the NY Times Android app and started method profiling on startup using Android Device Monitor.  Total time from the beginning of the application class to end of onCreate for first activity is roughly  2.2seconds. Diving deeper I was able to observe that ```NewThreadWorker.tryEnableCancelPolicy```
 was taking 1200ms to execute with the offending line being  
```for (Method m : exec.getClass().getMethods())``` (1017ms)
Diving deeper shows a call to 
```CollectionUtils.removeDuplicates (992 ms)```  
which will call 
```collection.sort``` (719ms)
 & 
``` reflect.compare ```(259ms).

tryEnableCancelPolicy has the following comment: 
```/** 
     * Tries to enable the Java 7+ setRemoveOnCancelPolicy.
     * <p>{@code public} visibility reason: called from other package(s) within RxJava.
     * If the method returns false, the {@link #registerExecutor(ScheduledThreadPoolExecutor)} may
     * be called to enable the backup option of purging the executors.
     * @param exec the executor to call setRemoveOnCaneclPolicy if available.
     * @return true if the policy was successfully enabled 
     */ 
```

I tried creating a Scheduler from an Executor but still hit the offending code.  Is there a way to avoid this code or fix the large performance hit that it is causing?

artem-zinnatullin:
Resolves #3119.

This PR adds methods for detecting Android API Version and disables `NewThreadWorker.tryEnableCancelPolicy()` on Android API < 21 which has significant performance cost on Android.

PR also improves performance of `NewThreadWorker.tryEnableCancelPolicy()` via caching Reflection.

### DIFF
--- a/src/main/java/rx/internal/schedulers/NewThreadWorker.java
+++ b/src/main/java/rx/internal/schedulers/NewThreadWorker.java
@@ -27,6 +27,8 @@ import rx.internal.util.*;
 import rx.plugins.*;
 import rx.subscriptions.*;
 
+import static rx.internal.util.PlatformDependent.ANDROID_API_VERSION_IS_NOT_ANDROID;
+
 /**
  * @warn class description missing
  */
@@ -39,8 +41,7 @@ public class NewThreadWorker extends Scheduler.Worker implements Subscription {
     /** Force the use of purge (true/false). */
     private static final String PURGE_FORCE_KEY = "rx.scheduler.jdk6.purge-force";
     private static final String PURGE_THREAD_PREFIX = "RxSchedulerPurge-";
-    /** Forces the use of purge even if setRemoveOnCancelPolicy is available. */
-    private static final boolean PURGE_FORCE;
+    private static final boolean SHOULD_TRY_ENABLE_CANCEL_POLICY;
     /** The purge frequency in milliseconds. */
     public static final int PURGE_FREQUENCY;
     private static final ConcurrentHashMap<ScheduledThreadPoolExecutor, ScheduledThreadPoolExecutor> EXECUTORS;
@@ -48,8 +49,17 @@ public class NewThreadWorker extends Scheduler.Worker implements Subscription {
     static {
         EXECUTORS = new ConcurrentHashMap<ScheduledThreadPoolExecutor, ScheduledThreadPoolExecutor>();
         PURGE = new AtomicReference<ScheduledExecutorService>();
-        PURGE_FORCE = Boolean.getBoolean(PURGE_FORCE_KEY);
         PURGE_FREQUENCY = Integer.getInteger(FREQUENCY_KEY, 1000);
+
+        // Forces the use of purge even if setRemoveOnCancelPolicy is available
+        final boolean purgeForce = Boolean.getBoolean(PURGE_FORCE_KEY);
+
+        final int androidApiVersion = PlatformDependent.getAndroidApiVersion();
+
+        // According to http://developer.android.com/reference/java/util/concurrent/ScheduledThreadPoolExecutor.html#setRemoveOnCancelPolicy(boolean)
+        // setRemoveOnCancelPolicy available since Android API 21
+        SHOULD_TRY_ENABLE_CANCEL_POLICY = !purgeForce
+                && (androidApiVersion == ANDROID_API_VERSION_IS_NOT_ANDROID || androidApiVersion >= 21);
     }
     /** 
      * Registers the given executor service and starts the purge thread if not already started. 
@@ -85,6 +95,7 @@ public class NewThreadWorker extends Scheduler.Worker implements Subscription {
     public static void deregisterExecutor(ScheduledExecutorService service) {
         EXECUTORS.remove(service);
     }
+
     /** Purges each registered executor and eagerly evicts shutdown executors. */
     static void purgeExecutors() {
         try {
@@ -102,31 +113,88 @@ public class NewThreadWorker extends Scheduler.Worker implements Subscription {
             RxJavaPlugins.getInstance().getErrorHandler().handleError(t);
         }
     }
-    
-    /** 
+
+    /**
+     * Improves performance of {@link #tryEnableCancelPolicy(ScheduledExecutorService)}.
+     * Also, it works even for inheritance: {@link Method} of base class can be invoked on the instance of child class.
+     */
+    private static volatile Object cachedSetRemoveOnCancelPolicyMethod;
+
+    /**
+     * Possible value of {@link #cachedSetRemoveOnCancelPolicyMethod} which means that cancel policy is not supported.
+     */
+     private static final Object SET_REMOVE_ON_CANCEL_POLICY_METHOD_NOT_SUPPORTED = new Object();
+
+    /**
      * Tries to enable the Java 7+ setRemoveOnCancelPolicy.
      * <p>{@code public} visibility reason: called from other package(s) within RxJava.
      * If the method returns false, the {@link #registerExecutor(ScheduledThreadPoolExecutor)} may
      * be called to enable the backup option of purging the executors.
-     * @param exec the executor to call setRemoveOnCaneclPolicy if available.
+     * @param executor the executor to call setRemoveOnCaneclPolicy if available.
      * @return true if the policy was successfully enabled 
      */
-    public static boolean tryEnableCancelPolicy(ScheduledExecutorService exec) {
-        if (!PURGE_FORCE) {
-            for (Method m : exec.getClass().getMethods()) {
-                if (m.getName().equals("setRemoveOnCancelPolicy")
-                        && m.getParameterTypes().length == 1
-                        && m.getParameterTypes()[0] == Boolean.TYPE) {
-                    try {
-                        m.invoke(exec, true);
-                        return true;
-                    } catch (Exception ex) {
-                        RxJavaPlugins.getInstance().getErrorHandler().handleError(ex);
-                    }
+    public static boolean tryEnableCancelPolicy(ScheduledExecutorService executor) {
+        if (SHOULD_TRY_ENABLE_CANCEL_POLICY) {
+            final boolean isInstanceOfScheduledThreadPoolExecutor = executor instanceof ScheduledThreadPoolExecutor;
+
+            final Method methodToCall;
+
+            if (isInstanceOfScheduledThreadPoolExecutor) {
+                final Object localSetRemoveOnCancelPolicyMethod = cachedSetRemoveOnCancelPolicyMethod;
+
+                if (localSetRemoveOnCancelPolicyMethod == SET_REMOVE_ON_CANCEL_POLICY_METHOD_NOT_SUPPORTED) {
+                    return false;
+                }
+
+                if (localSetRemoveOnCancelPolicyMethod == null) {
+                    Method method = findSetRemoveOnCancelPolicyMethod(executor);
+
+                    cachedSetRemoveOnCancelPolicyMethod = method != null
+                            ? method
+                            : SET_REMOVE_ON_CANCEL_POLICY_METHOD_NOT_SUPPORTED;
+
+                    methodToCall = method;
+                } else {
+                    methodToCall = (Method) localSetRemoveOnCancelPolicyMethod;
+                }
+            } else {
+                methodToCall = findSetRemoveOnCancelPolicyMethod(executor);
+            }
+
+            if (methodToCall != null) {
+                try {
+                    methodToCall.invoke(executor, true);
+                    return true;
+                } catch (Exception e) {
+                    RxJavaPlugins.getInstance().getErrorHandler().handleError(e);
                 }
             }
         }
+
         return false;
+    }
+
+    /**
+     * Tries to find {@code "setRemoveOnCancelPolicy(boolean)"} method in the class of passed executor.
+     *
+     * @param executor whose class will be used to search for required method.
+     * @return {@code "setRemoveOnCancelPolicy(boolean)"} {@link Method}
+     * or {@code null} if required {@link Method} was not found.
+     */
+    static Method findSetRemoveOnCancelPolicyMethod(ScheduledExecutorService executor) {
+        // The reason for the loop is to avoid NoSuchMethodException being thrown on JDK 6
+        // which is more costly than looping through ~70 methods.
+        for (final Method method : executor.getClass().getMethods()) {
+            if (method.getName().equals("setRemoveOnCancelPolicy")) {
+                final Class<?>[] parameterTypes = method.getParameterTypes();
+
+                if (parameterTypes.length == 1 && parameterTypes[0] == Boolean.TYPE) {
+                    return method;
+                }
+            }
+        }
+
+        return null;
     }
     
     /* package */
--- a/src/main/java/rx/internal/util/PlatformDependent.java
+++ b/src/main/java/rx/internal/util/PlatformDependent.java
@@ -20,31 +20,55 @@ import java.security.PrivilegedAction;
 
 /**
  * Allow platform dependent logic such as checks for Android.
- * 
+ *
  * Modeled after Netty with some code copy/pasted from: https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/PlatformDependent.java
  */
 public final class PlatformDependent {
 
-    private static final boolean IS_ANDROID = isAndroid0();
+    /**
+     * Possible value of {@link #getAndroidApiVersion()} which means that the current platform is not Android.
+     */
+    public static final int ANDROID_API_VERSION_IS_NOT_ANDROID = 0;
+
+    private static final int ANDROID_API_VERSION = resolveAndroidApiVersion();
+
+    private static final boolean IS_ANDROID = ANDROID_API_VERSION != ANDROID_API_VERSION_IS_NOT_ANDROID;
 
     /**
-     * Returns {@code true} if and only if the current platform is Android
+     * Returns {@code true} if and only if the current platform is Android.
      */
     public static boolean isAndroid() {
         return IS_ANDROID;
     }
 
-    private static boolean isAndroid0() {
-        boolean android;
-        try {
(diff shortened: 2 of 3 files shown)


src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -50,6 +53,16 @@
         PURGE = new AtomicReference<ScheduledExecutorService>();
         PURGE_FORCE = Boolean.getBoolean(PURGE_FORCE_KEY);
         PURGE_FREQUENCY = Integer.getInteger(FREQUENCY_KEY, 1000);
+
+        // Cached to prevent method call
+        IS_ANDROID = PlatformDependent.isAndroid();
```
JakeWharton: This is redundant with the other checks

src/main/java/rx/internal/util/PlatformDependent.java
```diff
@@ -20,13 +20,19 @@
...
+ *
  * Modeled after Netty with some code copy/pasted from: https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/PlatformDependent.java
  */
 public final class PlatformDependent {
 
     private static final boolean IS_ANDROID = isAndroid0();
 
+    // Will be null if current platform is not Android
+    // Or if Android API Version can not be resolved.
+    private static final Integer ANDROID_API_VERSION = IS_ANDROID
```
JakeWharton: Checking `IS_ANDROID` is redundant.

src/main/java/rx/internal/util/PlatformDependent.java
```diff
@@ -20,13 +20,19 @@
 
 /**
  * Allow platform dependent logic such as checks for Android.
- * 
+ *
  * Modeled after Netty with some code copy/pasted from: https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/PlatformDependent.java
  */
 public final class PlatformDependent {
 
     private static final boolean IS_ANDROID = isAndroid0();
```
JakeWharton: Change to `ANDROID_API_VERSION != null`

src/main/java/rx/internal/util/PlatformDependent.java
```diff
@@ -20,13 +20,19 @@
 
 /**
  * Allow platform dependent logic such as checks for Android.
- * 
+ *
```
artem-zinnatullin: Uh, sorry for this, IDEA… I can fix it if you want.

src/main/java/rx/internal/util/PlatformDependent.java
JakeWharton: And delete `isAndroid0()`, there's no reason to have so many mechanisms for checking

artem-zinnatullin:
Checked on JVM (Java 8), Android 4.1.2 (API 16), Android 5.1.1 (API 22).

src/main/java/rx/internal/util/PlatformDependent.java
artem-zinnatullin: Ok, agree.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
artem-zinnatullin: I don't think so, we trying to enable cancel policy if current platform is not Android or if it's Android and API >= 21, that's why I have 2 separate booleans.

Or maybe I just don't see easier way.

JakeWharton: You are checking if the API is `null` which is equivalent (and thus redundant) with this boolean.

src/main/java/rx/internal/util/PlatformDependent.java
```diff
@@ -20,31 +20,49 @@
...
  * Modeled after Netty with some code copy/pasted from: https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/PlatformDependent.java
  */
 public final class PlatformDependent {
 
-    private static final boolean IS_ANDROID = isAndroid0();
+    // Will be null if current platform is not Android
+    // Or if Android API Version can not be resolved.
+    private static final Integer ANDROID_API_VERSION = resolveAndroidApiVersion();
```
akarnokd: Is there an API version 0? If not, we could use `int` here so there is no need for unboxing and null checking elsewhere.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
artem-zinnatullin: Yep, but then I use `IS_ANDROID` on line 127. I've improved `IS_ANDROID` initialization.

src/main/java/rx/internal/util/PlatformDependent.java
artem-zinnatullin: Nope, there is no API version 0 as far as I know, we can also use `-1` as Not-Android-Platform value and even create a constant.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -50,6 +53,15 @@
         PURGE = new AtomicReference<ScheduledExecutorService>();
         PURGE_FORCE = Boolean.getBoolean(PURGE_FORCE_KEY);
         PURGE_FREQUENCY = Integer.getInteger(FREQUENCY_KEY, 1000);
+
+        final Integer androidApiVersion = PlatformDependent.getAndroidApiVersion();
+
+        // Cached to prevent method call
+        IS_ANDROID = androidApiVersion != null;
```
JakeWharton: You can still eliminate this variable altogether. The same check is included in the other.

src/main/java/rx/internal/util/PlatformDependent.java
JakeWharton: Nope. Great idea! RxJava doesn't support anything below API 9 anyway.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -112,7 +124,7 @@ static void purgeExecutors() {
      * @return true if the policy was successfully enabled 
      */
     public static boolean tryEnableCancelPolicy(ScheduledExecutorService exec) {
-        if (!PURGE_FORCE) {
+        if (!PURGE_FORCE && (!IS_ANDROID || IS_ANDROID_AND_SUPPORTS_SET_REMOVE_ON_CANCEL_POLICY)) {
```
JakeWharton: What i'm saying is, just make this 
```java
if (!PURGE_FORCE && IS_ANDROID_AND_SUPPORTS_SET_REMOVE_ON_CANCEL_POLICY)
```
to avoid the redundancy of having both variables and checking the same condition twice on the true case.

artem-zinnatullin: Maybe I'm not understanding something because it's 10 pm and I'm still in the office... 

But your code will always return `false` on regular JVM because `IS_ANDROID_AND_SUPPORTS_SET_REMOVE_ON_CANCEL_POLICY` will be `false`.

JakeWharton: Should be an OR and a not then.

JakeWharton: Sorry my boolean logic is all screwed up because I have my own version of this in my head.

My point was and remain, both of these are constant booleans which means whatever logic you want to apply to them can be done once. This yields a single static field which encapsulates both conditions (not Android or API 22+).

akarnokd: I'd change the entire method as well to do some caching and avoid looping at all if applied to the default `ScheduledThreadPoolExecutor` that is returned from `Executors.newScheduledThreadPool()`:

```java
private static volatile Object setOnCancelPolicyMethod;
private static final Object SET_ON_CANCEL_NOT_SUPPORTED = new Object();

public static boolean tryEnableCancelPolicy(ScheduledExecutorService exec) {
    if (!PURGE_FORCE && NOT_ANDROID_OR_SUPPORTS_SET_ON_CANCEL_POLICY) {
        final boolean executorPool = exec instanceof ScheduledThreadPoolExecutor;
        Method methodToCall;
        if (executorPool) {
            final Object localSetOnCancelPolicyMethod = setOnCancelPolicyMethod;
            if (localSetOnCancelPolicyMethod == SET_ON_CANCEL_NOT_SUPPORTED) {
                return false;
            }
            if (localSetOnCancelPolicyMethod == null) {
                Method m = findSetOnCancelPolicyMethod(exec);
                
                setOnCancelPolicyMethod = m != null ? m : SET_ON_CANCEL_NOT_SUPPORTED;
                
                methodToCall = m;
            } else {
                methodToCall = (Method)localSetOnCancelPolicyMethod;
            }
        } else {
            methodToCall = findSetOnCancelPolicyMethod(exec);
        }
        
        if (methodToCall != null) {
            try {
                methodToCall.invoke(exec, true);
                return true;
            } catch (Exception ex) {
                RxJavaPlugins.getInstance().getErrorHandler().handleError(ex);
            }
        }

    }
    return false;
}

private static Method findSetOnCancelPolicyMethod(ScheduledExecutorService exec) {
    for (final Method m : exec.getClass().getMethods()) {
        if (m.getName().equals("setRemoveOnCancelPolicy")
                && m.getParameterTypes().length == 1
                && m.getParameterTypes()[0] == Boolean.TYPE) {
            return m;
        }
    }
    return null;
}
```

artem-zinnatullin: @JakeWharton done! Take a look at `SHOULD_TRY_ENABLE_CANCEL_POLICY`.

@akarnokd done! Rewritten your code, added some tests and checked that it handles inheritance for `ScheduledExecutorService` (I was afraid that cached method won't work for child classes).

And tested again on JVM (Java 8), Android API 19 and Android API 22.

I should say that I noticed performance improvement in production Android app after these change for Android API < 21, hope it'll be in RxJava 1.0.14.

artem-zinnatullin: Also, I used `int` for Android API Version. And as far as I saw via debugging this method called in many cases, so this PR should improve performance of RxJava in many apps (backends and clients).

src/main/java/rx/internal/util/PlatformDependent.java
```diff
@@ -20,31 +20,56 @@
...
- * 
+ *
  * Modeled after Netty with some code copy/pasted from: https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/PlatformDependent.java
  */
 public final class PlatformDependent {
 
-    private static final boolean IS_ANDROID = isAndroid0();
+    /**
+     * Constant value of Android API Version which means that current platform is not Android.
+     * We could use {@link Integer} and {@code null}, but we want to avoid boxing/unboxing.
```
JakeWharton: Second sentence is not useful to retain.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -48,8 +49,17 @@
     static {
         EXECUTORS = new ConcurrentHashMap<ScheduledThreadPoolExecutor, ScheduledThreadPoolExecutor>();
         PURGE = new AtomicReference<ScheduledExecutorService>();
-        PURGE_FORCE = Boolean.getBoolean(PURGE_FORCE_KEY);
         PURGE_FREQUENCY = Integer.getInteger(FREQUENCY_KEY, 1000);
+
+        // Forces the use of purge even if setRemoveOnCancelPolicy is available
+        final boolean purgeForce = Boolean.getBoolean(PURGE_FORCE_KEY); // Compiler should inline this
```
JakeWharton: I don't understand what this trailing comment is telling me. Why is it here? Speculation about compiler optimizations make for bad comments. Especially since this is a local field in a static block that's only ever run once where perf doesn't _really_ matter.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -48,8 +49,17 @@
...
+
+        // According to http://developer.android.com/reference/java/util/concurrent/ScheduledThreadPoolExecutor.html#setRemoveOnCancelPolicy(boolean)
+        // setRemoveOnCancelPolicy available since Android API 21
+        final int androidApiVersion = PlatformDependent.getAndroidApiVersion();
+
+        SHOULD_TRY_ENABLE_CANCEL_POLICY = !purgeForce
+                        && (androidApiVersion == ANDROID_API_VERSION_IS_NOT_ANDROID || androidApiVersion >= 21);
```
JakeWharton: Four space indent is already killing me on this whole project, but continuations should be double that which is 8. This is 16.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -48,8 +49,17 @@
...
         EXECUTORS = new ConcurrentHashMap<ScheduledThreadPoolExecutor, ScheduledThreadPoolExecutor>();
         PURGE = new AtomicReference<ScheduledExecutorService>();
-        PURGE_FORCE = Boolean.getBoolean(PURGE_FORCE_KEY);
         PURGE_FREQUENCY = Integer.getInteger(FREQUENCY_KEY, 1000);
+
+        // Forces the use of purge even if setRemoveOnCancelPolicy is available
+        final boolean purgeForce = Boolean.getBoolean(PURGE_FORCE_KEY); // Compiler should inline this
+
+        // According to http://developer.android.com/reference/java/util/concurrent/ScheduledThreadPoolExecutor.html#setRemoveOnCancelPolicy(boolean)
+        // setRemoveOnCancelPolicy available since Android API 21
```
JakeWharton: These comments belong on the static field initialization where the comparison against 21 is actually made.

src/test/java/rx/internal/schedulers/NewThreadWorkerTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    // This test requires Java 7 for compilation
+    @Test
+    public void tryEnableCancelPolicyShouldInvokeMethodOnExecutor() {
+        ScheduledThreadPoolExecutor executor = spy(new ScheduledThreadPoolExecutor(1));
+
+        boolean result = NewThreadWorker.tryEnableCancelPolicy(executor);
+
+        assertTrue(result);
+        verify(executor).setRemoveOnCancelPolicy(true);
```
akarnokd: This will fail on a Java 6 build but I'm not sure if people actually build RxJava from source on Java 6. What I would do is subclass ScheduledThreadPoolExecutor and add the setRemoveOnCancelPolicy(boolean) method. For Java 7+, this will count as a simple override.

src/test/java/rx/internal/schedulers/NewThreadWorkerTest.java
```diff
@@ -0,0 +1,103 @@
...
+        ScheduledExecutorService executor = mock(ScheduledExecutorService.class);
+
+        boolean result = NewThreadWorker.tryEnableCancelPolicy(executor);
+
+        assertFalse(result);
+        verifyZeroInteractions(executor);
+    }
+
+    @Test
+    public void tryEnableCancelPolicyShouldBeSkipped() throws NoSuchFieldException, IllegalAccessException {
```
akarnokd: This looks rather dangerous. What I'd do is to have a package-private method `tryEnableCancelPolicy(ScheduledExecutor executor, boolean tryEnable)` so the test can inject the necessary value.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
artem-zinnatullin: +1, my bad. Would be nice to switch to 2 spaces instead of 4.

src/test/java/rx/internal/schedulers/NewThreadWorkerTest.java
artem-zinnatullin: Nice workaround!

src/test/java/rx/internal/schedulers/NewThreadWorkerTest.java
artem-zinnatullin: Hm, a separate method just for injecting value which can completely disable it… Not sure I am following this (on Android we have some limits on the quantity of methods in the app, this will increment it just for test purposes), let's just remove this test for now? 

Or I can remove `private` modifier from `SHOULD_TRY_ENABLE_CANCEL_POLICY` and make this test more "static", reflection will only change the value of this field.

akarnokd: I'd rather not have this test for now. All other changes appear to be okay. If you update the PR, I'll merge it.

artem-zinnatullin:
Fixed all comments, removed test with reflection, added more javadoc and comments, tests are now compilable on JDK 6 (didn't check but they should work fine). @akarnokd, @JakeWharton PTAL

src/main/java/rx/internal/util/PlatformDependent.java
```diff
@@ -20,31 +20,55 @@
...
+     * can not be resolved or if current platform is not Android.
+     */
+    public static int getAndroidApiVersion() {
+        return ANDROID_API_VERSION;
+    }
+
+    /**
+     * Resolves version of Android API.
+     *
+     * @return version of Android API or {@code null} if version can not be resolved.
```
akarnokd: This no longer returns null but `ANDROID_API_VERSION_IS_NOT_ANDROID`.

artem-zinnatullin: Fixed.

akarnokd:
Great, thanks!

PR #3123, opened 2015-07-29
Remove redundant type parameter in EmptyAction

jacek-rzrz:
There is one too many.


akarnokd:
Looks simple enough but since this involves removing a type argument from a private class which is only assigned to a public interface of his, I'm not certain if this would or wouldn't cause binary compatibility problems.

akarnokd:
Haven't found anything that indicates binary incompatibility due to this change, thanks!

PR #3126, opened 2015-07-31
Allow to use null as error for TestSubscriber.assertError()

### LINKED ISSUE #3125: Subscriber onErrror not called if exception is null
rxjava version 1.0.13

Subscriber onErrror not called if exception passed to onError is null.

 @Override
           //this is method in subscriber
}

If this is called onError is not called
 subscriber.onError(null); 
if works if
 subscriber.onError(new Exception());

artem-zinnatullin:
Resolves #3125 via test, makes `TestSubscriber.assertError()` little bit faster via caching first error, and allows to use `TestSubscriber.assertError(null)`.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -404,20 +404,29 @@ public class TestSubscriber<T> extends Subscriber<T> {
      */
     @Experimental
     public void assertError(Throwable throwable) {
-        List<Throwable> err = testObserver.getOnErrorEvents();
-        if (err.size() == 0) {
+        final List<Throwable> errorEvents = testObserver.getOnErrorEvents();
+
+        if (errorEvents.isEmpty()) {
             throw new AssertionError("No errors");
-        } else
-        if (err.size() > 1) {
-            AssertionError ae = new AssertionError("Multiple errors: " + err.size());
-            ae.initCause(new CompositeException(err));
-            throw ae;
-        } else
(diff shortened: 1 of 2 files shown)


akarnokd:
Generally, you are not supposed to call onError with null and not expect one so the NPE in the test in this case, in my opinion, is acceptable as a failure state.

artem-zinnatullin:
Yeah, I agree, but if we'll deny `null` as error we will break backward compatibility because it was allowed in previous versions of RxJava :(

akarnokd:
I don't know. Are there really users that send null through onError and test for it with this assert?

artem-zinnatullin:
According to #3125 there is at least one user of `onError(null)`, but personally, I'd like to deny it. Though it might be interpreted as breaking change in the API I'd just call it "bugfix", add check to the `TestSubscriber` and write tests that asserts that `null` will break the execution :)

If you agree, I'll update this PR and we will deny `null`s for errors.

akarnokd:
`onError(null)` is definitely a bad practice and reactive-streams explicitly forbids it, so should we.

benjchristensen:
Agreed, `onError(null)` is invalid. This is supposed to represent throwing exceptions, and we can not "throw null".

artem-zinnatullin:
Oh, I wanted to change this PR and add javadoc about `null` into the `TestSubscriber.assertError()` and `Observer.onError()` but you closed this PR… :)

I'll do it in a separate PR.

PR #3135, opened 2015-08-06
Renaming rx.Subscriber#request(long) to requestFromProducer(long)

stealthcode:
Right now its not possible to both extend `rx.Subscriber` and implement `rx.Producer`. See test below.

```java
public class Foo<T> extends Subscriber<T> implements Producer {

    @Override
    public void onCompleted() {
    }

    @Override
    }

    @Override
    public void onNext(T t) {
    }

    @Override
    public void request(long n) {
    }
}
```

Currently this blows up because `protected rx.Subscriber#request(long)` collides with `public rx.Producer#request(long)`.

### DIFF
--- a/src/main/java/rx/Subscriber.java
+++ b/src/main/java/rx/Subscriber.java
@@ -137,7 +137,7 @@ public abstract class Subscriber<T> implements Observer<T>, Subscription {
      * @throws IllegalArgumentException
      *             if {@code n} is negative
      */
-    protected final void request(long n) {
+    protected final void requestFromProducer(long n) {
         if (n < 0) {
             throw new IllegalArgumentException("number requested cannot be negative: " + n);
         } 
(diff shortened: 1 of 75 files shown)


akarnokd:
This is a drastic public API change which breaks anyone relying on the protected request() method. I fear the name clash won't be resolved in 1.x line. ReaciveStreams luckily doesn't suffer from this. Otherwise, its usually not much saving because if there is request tracking involved, either the specific producer extends AtomicLong or has an AtomicLong instance field.

stealthcode:
This affects anybody writing a class that extends `rx.Subscriber` so this certainly would not be a patch. This could prevent the need for an unnecessary object initialization cost for separate internal subscribers and producers. I think it should be considered.

davidmoten:
It would be good for changes like this to marked for consideration at the next breaking change API release. Perhaps we could use a label to track these (say `2.x`).

akarnokd:
I don't think this won't be much of a problem in 2.0 since it is going to be based on reactive-streams (or j.u.c.Flow) *interfaces* and if there is going to be a default implementation for a `Subscriber` (i.e., Observer), that won't be used by the operators so it can have a method `protected void request(n)`.

I vote for closing this PR.

stealthcode:
We are going to address this as an issue with 2.x label.

PR #3138, opened 2015-08-09
Range overhead reduction.

akarnokd:
Applied some refactorings and local variable usage to reduce the overhead.

Few observations:
  - Having too many local variables may cause register spill, even on x64 which makes some `size` benchmark faster while other slower.
  - The observeOn benchmarks are quite hectic because of receiving thread migration caused by the round-robin worker assignment. It affects the benchmarks with 1 or 1000 elements in the stream.
  - Note that the previous `OperatorRangePerf` `size = 1` measured the speed of `just` due to the optimization of `range()`. The updated perf now instantiates the `OnSubscribeRange`.
  - Note that the `observeOn` benchmark with `size = 1` run the `just()` as well.

Benchmark comparison (i7 4770K, Windows 7 x64, Java 8u51)
image

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeRange.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeRange.java
@@ -15,11 +15,10 @@
  */
 package rx.internal.operators;
 
-import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.AtomicLong;
 
+import rx.*;
 import rx.Observable.OnSubscribe;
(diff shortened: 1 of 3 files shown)


akarnokd:
Updated the PR to fix the lack of widening in the fast path if the end is Integer.MAX_VALUE by adding 1L (the slow path adds the long idx value which does this widening).

akarnokd:
I'm merging this; the values are better for the thread-stable range perf and observeOn is simply too hectic with the round-robin core usage.

Issue #3145, opened 2015-08-10
Expanding Community Involvement & Contribution


benjchristensen:
Earlier this year we expanded the committers of RxJava to include non-Netflix engineers. I'd like to now further explore how we can broaden community involvement and lessen dependence on me. The recent ["RxJava Truck Factor" issue](https://github.com/ReactiveX/RxJava/issues/3133) was a reminder of this topic and pushed several of us inside Netflix to raise the priority on tackling this. 

First, a note on how great the community has been thus far! We have had great [contributors](https://github.com/ReactiveX/RxJava/graphs/contributors), several having contributed significant time, effort and skill and without whom RxJava would not be as good as it is. Many who are not shown with "code contributions" have helped immensely in answering questions on Twitter, StackOverflow, Google Groups, blogging, speaking at meetups and conferences, or helping build reactivex.io. In short, the community around RxJava is already great.

However, I'd like to achieve a few things:

1) Make it possible for others to fulfill the "release engineer" role I'm still playing.
2) Ensure there are no irreplaceable people on the project. 
3) Improve the reliability for getting code reviews and questions answered in a reasonable time frame.
4) Improve the release cadence so it is more reliable, and less dependent on me getting bursts of time. 
5) Make it easier for community to get involved by pointing to where they can help in ways other than writing code.
6) Find a way to recognize people for involvement in the many important areas other than lines of code committed. 

A few ideas on making these happen include:

- More engineers from Netflix with time formally allocated. I will bootstrap them and ensure they have everything needed to do releases without me.
- More comprehensive CONTRIBUTING.md (Rust has a good example: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md)
- Ask community to help with StackOverflow, Google Groups, Issue/PR Triage, Twitter (and perhaps recognize people somehow who do these things?)

A key thing I want people to understand is that writing code is actually the easiest part of maintaining a project like this. The lifecycle is far harder and more time consuming. This includes:

- code reviews
- answering questions
- deciding whether a change should be made or not
- discussing design decisions and deciding which tradeoffs to make
- balancing change with stability
- writing documentation
- debugging issues 
- replicating bugs with unit tests
- answering more questions
- tutorials, blogs, speaking, training, etc

All who are interested in RxJava, what ideas do you have for helping us improve and achieve the above? What else should we pursue and improve?

benjchristensen:
[...]

PR #3147, opened 2015-08-10
Remove unnecessary static modifier

artem-zinnatullin:
P.S. 
@benjchristensen @akarnokd sorry for large amount of simple pull requests, I'm doing full code review of RxJava sources http://artemzin.com/blog/rxjava-code-review-part-1/ and I'll create PRs and issues during this review.

I'll try to group changes by themes so it will be easy to review them. Looks like this adventure is going to be part of #3145!

### DIFF
--- a/src/main/java/rx/Notification.java
+++ b/src/main/java/rx/Notification.java
@@ -162,7 +162,7 @@ public final class Notification<T> {
         }
     }
 
-    public static enum Kind {
+    public enum Kind {
         OnNext, OnError, OnCompleted
     }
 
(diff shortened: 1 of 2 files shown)


akarnokd:
If it were only me, I'd accept cosmetic changes such as this, but history tells me not to.

artem-zinnatullin:
@akarnokd why? Sad that `javac` is not strict enough and allows such things. We can add `checkstyle` with only java-lang rules and remove all these things from the code, but last time I suggested to add `checkstyle` @benjchristensen was not very happy…

I understand that RxJava is not "my project" and I can't just come and force you to use "my codestyle" (which is actually just standard java codestyle), so if you don't want to accept such changes — it's okay, I'll stop doing such PRs and will report only major issues.

akarnokd:
I like this PR too much, I'm going to merge it ;)

benjchristensen:
> I like this PR too much, I'm going to merge it ;)

:-)

PR #3151, opened 2015-08-13
Add semantic-named override for unsubscribe-action subscription.

### LINKED ISSUE #2967: Rename method Subscriptions.create for semantic clarity.
The `Subscriptions.create` method is extremely useful but poorly named. Is there a reason that it does not have a more semantic name which emphasizes the nature of the action being for unsubscribe rather than on subscribe?

When you read code like this:
```java
subscription.add(Subscriptions.create(() -> runnable.run()));
```
it is not at all apparent that this code is for unsubscription.

Something along the lines of `unsubscribeAction`, `withUnsubscribeAction`, or `doOnUnsubscribe` would add a ton of clarity to the readability of this code.
```java
subscription.add(Subscriptions.unsubscribeAction(() -> runnable.run()));
subscription.add(Subscriptions.withUnsubscribeAction(() -> runnable.run()));
subscription.add(Subscriptions.doOnUnsubscribe(() -> runnable.run()));
```
Now I realize we can't just rename this method, so I am proposing it as an addition along with deprecation of the original.

JakeWharton:
Closes #2967.

### DIFF
--- a/src/main/java/rx/internal/operators/BufferUntilSubscriber.java
+++ b/src/main/java/rx/internal/operators/BufferUntilSubscriber.java
@@ -88,7 +88,7 @@ public final class BufferUntilSubscriber<T> extends Subject<T, T> {
         @Override
         public void call(final Subscriber<? super T> s) {
             if (state.casObserverRef(null, s)) {
-                s.add(Subscriptions.create(new Action0() {
+                s.add(Subscriptions.doOnUnsubscribe(new Action0() {
                     @SuppressWarnings("unchecked")
                     @Override
                     public void call() {
--- a/src/main/java/rx/internal/operators/OnSubscribeAmb.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeAmb.java
@@ -364,7 +364,7 @@ public final class OnSubscribeAmb<T> implements OnSubscribe<T>{
     public void call(final Subscriber<? super T> subscriber) {
         
         //setup unsubscription of all the subscribers to the sources
-        subscriber.add(Subscriptions.create(new Action0() {
+        subscriber.add(Subscriptions.doOnUnsubscribe(new Action0() {
 
             @Override
             public void call() {
--- a/src/main/java/rx/internal/operators/OnSubscribeRefCount.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeRefCount.java
@@ -146,7 +146,7 @@ public final class OnSubscribeRefCount<T> implements OnSubscribe<T> {
     }
 
     private Subscription disconnect(final CompositeSubscription current) {
-        return Subscriptions.create(new Action0() {
+        return Subscriptions.doOnUnsubscribe(new Action0() {
             @Override
             public void call() {
                 lock.lock();
--- a/src/main/java/rx/internal/operators/OnSubscribeToObservableFuture.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeToObservableFuture.java
@@ -58,7 +58,7 @@ public final class OnSubscribeToObservableFuture {
 
         @Override
         public void call(Subscriber<? super T> subscriber) {
-            subscriber.add(Subscriptions.create(new Action0() {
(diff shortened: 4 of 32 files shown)


JakeWharton:
Bikeshed away the name!

JakeWharton:
I think I'm leaning towards `withUnsubscribeAction` actually...

stevecstian:
Should we add `BooleanSubscription#createWithUnsubscribeAction`?

JakeWharton:
Oh! I didn't realize it was in the public API let alone has duplicates of these methods. I would think that we'd deprecate both of its static factory methods and the entire class itself as there's no reason (that I can see) for accessing it directly vs the methods on `Subscriptions`.

akarnokd:
I prefer `from` or `wrap`.

JakeWharton:
Both of those seem equally as ambiguous. Creating a subscription _from_ and action and _wrapping_ an action to create a subscription offer no insight into their behavior. One has to have seen this method in use before and/or have looked at the javadoc (and in both of those cases, explicitly remembered the behavior).

stevecstian:
builder?

stevecstian:
`create(UnsubscribeAction)`?

stealthcode:
@stevecstian the method itself returns a `Subscription` so changing it to return a builder or have different semantics would be much more significant than renaming. The method is marked as `@Beta` which means it's possible to change but we should avoid a method signature change unless we have a good reason. We could create a builder by introducing a new static method and this change would not break this method's semantics. 

I like `Subscriptions.doOnUnsubscribe(Action0)` or `Subscriptions.onUnsubscribe(Action0)`.

stealthcode:
@benjchristensen can you provide some clarification on the changes that would be acceptable for methods labeled `@Beta`? This clarification may be necessary for future issues.

stevecstian:
Having `UnsubscribeAction extends Action0` and `Subscriptions.create(UnsubscribeAction)` seems like overkill...  `Subscriptions.doOnUnsubscribe(Action0)` is looking good to me.

@JakeWharton thoughts?

JakeWharton:
I am comfortable with that. That was one of my suggestions from #2967.

JakeWharton:
What about the deprecation of associated methods on `BooleanSubscription`? I don't see a reason to have both sets which duplicate the exact same API functionality.

stevecstian:
@JakeWharton if we deprecate `BooleanSubscription.create(Action0)` then we need to change `BooleanSubscription(Action0)` from `private` to default level constructor, right?

JakeWharton:
I would:
* Deprecate `BooleanSubscription.create()` and `BooleanSubscription.create(Action0)` methods with pointers to `Subscriptions`'s static methods.
* Deprecate the `BooleanSubscription` type with instructions to use `Subscriptions`'s static methods.
* Change `BooleanSubscription`'s constructor to package-private.
* Reference the constructor directly from `Subscriptions.empty()`, `Subscriptions.create(Action0)`, and `Subscriptions.doOnUnsubsribe(Action0)`.

JakeWharton:
Also:
* Deprecate `BooleanSubscription`'s no-argument constructor with a reference to `Subscriptions.empty()`

stevecstian:
@JakeWharton :+1:

benjchristensen:
> can you provide some clarification on the changes that would be acceptable for methods labeled @Beta? This clarification may be necessary for future issues.

It is stated at https://github.com/ReactiveX/RxJava#beta:

> APIs marked with the `@Beta` annotation at the class or method level are subject to change. They can be modified in any way, or even removed, at any time. If your code is a library itself (i.e. it is used on the CLASSPATH of users outside your own control), you should not use beta APIs, unless you repackage them (e.g. using ProGuard, shading, etc).

Thus, we have higher confidence in `@Beta` APIs than we do `@Experimental`, but both can change at any time, even in a minor release.

abersnaze:
Seeing how `Subscriptions.create(Action0)` is already part of the public API I'm inclined to leave this wart alone in 1.x. I recommend closing this PR and opening a new one against 2.x branch.

JakeWharton:
Sounds good to me.

PR #3154, opened 2015-08-14
Add Observable.fromCallable() as a companion for Observable.defer()

artem-zinnatullin:
Yep, this is a new operator.

Motivation? `Observable.defer()` requires function that returns `Observable<T>` when usually we don't want to create `Observable<T>` manually, we just want to defer execution of some function.

Like this:

```java
Observable
  .fromCallable(() -> someFunc())
  .subscribeOn(...)
  ...
```

Instead of this:

```java
Observable
  .defer(() -> Observable.just(someFunc()))
  .subscribeOn(...)
  ...
```

And more important case with deferring code that throws checked exceptions:
```java
Observable
  .fromCallable(() -> {
    Value value = ...;
    // some code
    return value;
  })
 .subscribeOn(...)
 ...
```

Instead of this:

```java
Observable
  .defer(() -> {
    try {
      Value value = ...;
       // some code that throws checked exceptions
      return Observable.just(value);
    } catch (Exception e) {
      return Observable.error(e);
    }
  })
 .subscribeOn(...)
 ...
```

I'd use name `defer` but both methods will have same type erasure (`Func0<Observable>` and `Func0<T>`), so I had to use a different name.

Questions:

1. Useful or not? For me — deferring some function call is a common task.
2. Naming.

If the decision about this operator will be positive — I'll add javadoc to this PR and then create separate PR for `Single.fromCallable()`.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -1251,6 +1251,29 @@ public class Observable<T> {
     }
 
     /**
+     * Returns an Observable that invokes passed function and emits its result for each new Observer that subscribes.
+     * <p>
+     * Allows you to defer execution of passed function until Observer subscribes to the Observable.
+     * It makes passed function "lazy".
+     * Result of the function invocation will be emitted by the Observable.
+     * <dl>
+     *   <dt><b>Scheduler:</b></dt>
+     *   <dd>{@code fromCallable} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param func
+     *         function which execution should be deferred, it will be invoked when Observer will subscribe to the Observable
+     * @param <T>
+     *         the type of the item emitted by the Observable
+     * @return an Observable whose {@link Observer}s' subscriptions trigger an invocation of the given function
+     * @see #defer(Func0)
+     */
+    @Experimental
+    public static <T> Observable<T> fromCallable(Callable<? extends T> func) {
+        return create(new OnSubscribeFromCallable<T>(func));
+    }
+
+    /**
      * Returns an Observable that emits a sequential number every specified interval of time.
      * <p>
      * <img width="640" height="195" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/interval.png" alt="">
--- a/src/main/java/rx/internal/operators/OnSubscribeFromCallable.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeFromCallable.java
@@ -1,0 +1,38 @@
+package rx.internal.operators;
+
+import rx.Observable;
+import rx.Subscriber;
+import rx.exceptions.Exceptions;
+import rx.internal.producers.SingleDelayedProducer;
+
+import java.util.concurrent.Callable;
+
+/**
+ * Do not invoke the function until an Observer subscribes; Invokes function on each
+ * subscription.
+ * <p>
+ * Pass {@code fromCallable} a function, and {@code fromCallable} will call this function to emit result of invocation
+ * afresh each time a new Observer subscribes.
+ */
+public final class OnSubscribeFromCallable<T> implements Observable.OnSubscribe<T> {
+
+    private final Callable<? extends T> resultFactory;
+
+    public OnSubscribeFromCallable(Callable<? extends T> resultFactory) {
+        this.resultFactory = resultFactory;
+    }
+
+    @Override
+    public void call(Subscriber<? super T> subscriber) {
+        final SingleDelayedProducer<T> singleDelayedProducer = new SingleDelayedProducer<T>(subscriber);
+
+        subscriber.setProducer(singleDelayedProducer);
+
+        try {
+            singleDelayedProducer.setValue(resultFactory.call());
+        } catch (Throwable t) {
+            Exceptions.throwIfFatal(t);
+            subscriber.onError(t);
+        }
+    }
+}
--- a/src/test/java/rx/internal/operators/OnSubscribeFromCallableTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeFromCallableTest.java
@@ -1,0 +1,140 @@
+package rx.internal.operators;
+
+import org.junit.Test;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import rx.Observable;
+import rx.Observer;
+import rx.Subscription;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+
+import static org.mockito.Mockito.*;
+import static rx.schedulers.Schedulers.computation;
+
+public class OnSubscribeFromCallableTest {
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void shouldNotInvokeFuncUntilSubscription() throws Exception {
+        Callable<Object> func = mock(Callable.class);
+
+        when(func.call()).thenReturn(new Object());
+
+        Observable<Object> fromCallableObservable = Observable.fromCallable(func);
+
+        verifyZeroInteractions(func);
+
+        fromCallableObservable.subscribe();
+
+        verify(func).call();
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void shouldCallOnNextAndOnCompleted() throws Exception {
+        Callable<String> func = mock(Callable.class);
+
+        when(func.call()).thenReturn("test_value");
+
+        Observable<String> fromCallableObservable = Observable.fromCallable(func);
+
+        Observer<String> observer = mock(Observer.class);
+
+        fromCallableObservable.subscribe(observer);
+
+        verify(observer).onNext("test_value");
+        verify(observer).onCompleted();
+        verify(observer, never()).onError(any(Throwable.class));
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void shouldCallOnError() throws Exception {
+        Callable<Object> func = mock(Callable.class);
+
+        Throwable throwable = new IllegalStateException("Test exception");
+        when(func.call()).thenThrow(throwable);
+
+        Observable<Object> fromCallableObservable = Observable.fromCallable(func);
+
+        Observer<Object> observer = mock(Observer.class);
+
+        fromCallableObservable.subscribe(observer);
+
+        verify(observer, never()).onNext(anyObject());
+        verify(observer, never()).onCompleted();
+        verify(observer).onError(throwable);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void shouldNotDeliverResultIfSubscriberUnsubscribedBeforeEmission() throws Exception {
+        Callable<String> func = mock(Callable.class);
+
+        final CountDownLatch funcLatch = new CountDownLatch(1);
+        final CountDownLatch observerLatch = new CountDownLatch(1);
+
+        when(func.call()).thenAnswer(new Answer<String>() {
+            @Override
+            public String answer(InvocationOnMock invocation) throws Throwable {
+                observerLatch.countDown();
+
+                try {
+                    funcLatch.await();
+                } catch (InterruptedException e) {
+                    // It's okay, unsubscription causes Thread interruption
+
+                    // Restoring interruption status of the Thread
+                    Thread.currentThread().interrupt();
+                }
+
+                return "should_not_be_delivered";
+            }
+        });
+
+        Observable<String> fromCallableObservable = Observable.fromCallable(func);
+
+        Observer<String> observer = mock(Observer.class);
+
+        Subscription subscription = fromCallableObservable
+                .subscribeOn(computation())
+                .subscribe(observer);
+
+        // Wait until func will be invoked
+        observerLatch.await();
+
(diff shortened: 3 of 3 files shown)


artem-zinnatullin:
Also, I see that for Rx newcomers (some of my coworkers, for example) is not very clear why they need to return `Observable` if they just want to defer some code.

davidmoten:
Sounds ok to me. A name using the `defer` prefix might be nice for discoverability purposes.

artem-zinnatullin:
We can call it `from()`, I guess it will be nice to write `Observable.from(() - > someFunc())` and even better with method `Observable.from(this::someFunc())`!

akarnokd:
We can't call it from because of potential conflict in dynamic languages. I'd prefer start or startAsync .

src/main/java/rx/Observable.java
```diff
@@ -1069,6 +1069,10 @@ public void call(Subscriber<? super R> o) {
         return create(new OnSubscribeDefer<T>(observableFactory));
     }
 
+    public static <T> Observable<T> lazy(Func0<T> lazyFunc) {
```
akarnokd: Func0<? extends T>

src/main/java/rx/internal/operators/OnSubscribeLazy.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public OnSubscribeLazy(Func0<? extends T> resultFactory) {
+        this.resultFactory = resultFactory;
+    }
+
+    @Override
+    public void call(Subscriber<? super T> subscriber) {
+        try {
+            subscriber.onNext(resultFactory.call());
+        } catch (Throwable t) {
```
akarnokd: I'd split this and check for isUnsubscribed before calling onNext

src/main/java/rx/internal/operators/OnSubscribeLazy.java
```diff
@@ -0,0 +1,26 @@
...
+            subscriber.onError(t);
```
davidmoten: An enhancement that we use elsewhere like in `OperatorMap` is `Exceptions.throwIfFatal(e);` just before this line

src/test/java/rx/internal/operators/OnSubscribeLazyTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        Observer<Object> observer = mock(Observer.class);
+
+        lazyObservable.subscribe(observer);
+
+        verify(observer, never()).onNext(anyObject());
+        verify(observer, never()).onCompleted();
+        verify(observer).onError(throwable);
+    }
+}
```
davidmoten: add unsubscribe tests

src/main/java/rx/internal/operators/OnSubscribeLazy.java
artem-zinnatullin: Okay, 
```java
T result = resultFactory.call();

if (!subscriber.unsubscribed()) {
  subscriber.onNext(result);
}
```

vs

```java
if (!subscriber.isUnsubscribed()) {
  subscriber.onNext(resultFactory.call());
}
```

What variant should we use? (anyway I'll notice this in javadoc, but we should decide if we want to execute `lazyFunc` even if `subscriber` already unsubscribed or we want to prevent execution)

artem-zinnatullin:
>We can't call it from because of potential conflict in dynamic languages. I'd prefer start or startAsync .

`Observable.start()` will be confusing because actually it won't start until you subscribe to it.

`Observable.startAsync()` is also confusing and it can be non-async if you won't add `subscribeOn()`.

If we can not use `defer()` I'd prefer `lazy()` mm? cc @benjchristensen

src/main/java/rx/internal/operators/OnSubscribeLazy.java
akarnokd: Use SingleDelayedProducer, get the value  in a try-catch, set it on the producer via setValue.

abersnaze:
Whatever the name is (`deferJust(Func0<T>)`?) I would like the implementation to return a subclass of ScalarSynchronousObservable that overrides the `T get()` method with a call to the closure passed in.

src/test/java/rx/internal/operators/OnSubscribeLazyTest.java
artem-zinnatullin: I've added one, can you please suggest more ideas for the unsubscribe tests?

src/main/java/rx/internal/operators/OnSubscribeLazy.java
artem-zinnatullin: Ah, sure, added!

src/main/java/rx/internal/operators/OnSubscribeLazy.java
artem-zinnatullin: `SingleDelayedProducer` is pretty nice, thanks for the suggestion (I think I'll use it for some of my own operators).

Switched to it.

src/main/java/rx/Observable.java
artem-zinnatullin: Fixed

artem-zinnatullin:
Operator is ready for next review!

> Sounds ok to me. A name using the defer prefix might be nice for discoverability purposes.

Another naming suggestion: `deferFunc()`, it will be obvious for those who used `defer()`, but it will be strange a little because `defer` is not `deferObservable`…

artem-zinnatullin:
@abersnaze 

>  I would like the implementation to return a subclass of ScalarSynchronousObservable

Not sure about this, because I don't know what variant will be better (efficient, maintainable, etc) `SingleDelayedProducer` or `ScalarSynchronousObservable`, @akarnokd I'd like to hear your opinion on this if you don't mind.

abersnaze:
There is special logic in `merge()` [link to code] (https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorMerge.java#L227) to unwrap `ScalarSynchronousObservable` by invoking the `get()` method directly and avoiding all the overhead of creating a `Subscriber` and `SingleDelayedProducer` to just get the value.  This should make `flatMap` of these very fast.

akarnokd:
You can't return the scalar observable because the value has to be ready at assembly time and be constant. Any side effect would only happen once as well.

src/main/java/rx/Observable.java
```diff
@@ -1069,6 +1069,11 @@ public void call(Subscriber<? super R> o) {
         return create(new OnSubscribeDefer<T>(observableFactory));
     }
 
+    @Experimental
+    public static <T> Observable<T> lazy(Func0<? extends T> lazyFunc) {
```
akarnokd: I'd consider Callable to allow IO operations without the need to do try-catch. An overload won't work due to dynamic languages.

artem-zinnatullin: Yeah, I've thought about this too, I'll take a look what I can do after work.

abersnaze:
@akarnokd I'll try to send in an example of what I mean but I don't see any reason why `T get()` can't call a `Func0<T>`.  It would still be Scalar and Synchronous but there isn't anything that says that it has to be constant.

akarnokd:
Instead of modifying ScalarSynchronousObservable, you could add a new case to the type check in merge.

abersnaze:
check out https://github.com/abersnaze/RxJava/commit/18d2cd51c9080c19e17007b6d0a2d6aca974a505

akarnokd:
There are several problems with your proposal
  - increased likelihood of bimorphic/megamorphic dispatch at the call sites of get()
  - get() can now throw which is unexpected at the call sites
  - calling the func might carry side effects and calling it during assembly time is incorrect

src/main/java/rx/Observable.java
artem-zinnatullin: Done!

akarnokd:
:+1:

artem-zinnatullin:
Switched to `Callable` (available in JDK since 1.5, so it's okay), so now you can defer function that potentialy throws checked exception!

I hope `Observable.lazy()` (or whatever name it will has) will be the easiest way to create `Observable` that should emit one item only after `subscribe()` (which is pretty common Rx task).

So, I guess it's time to select name for this operator and add javadoc.

stealthcode:
I would also like to see the observables produced by `lazy` take advantage of a path similar to ScalarSynchronousObservable (statically allocated and statically available data). I have an idea that I'd like to explore more and I'll open another pull request that enables this for both `Observable.defer` and `Observable.lazy` (assuming this gets merged).

artem-zinnatullin:
Added javadoc and squashed!

P.S. Today I have used `Observable.defer()` ~20 times and each time I could use `Observable.lazy()` and save some lines of code, would be nice to see it in 1.0.15!

staltz:
All Observables are lazy. Why can't you just use Observable.create()?

akarnokd:
@staltz That would conflict with `create(OnSubscribe)` in dynamic languages.

My new name proposals:

  - `deferJust`
  - `deferSingle`
  - `deferScalar`

dimsuz:
`deferSingle` sounds nice to me.
Also thought of `deferValue`.

artem-zinnatullin:
>All Observables are lazy.

It's not true, in Java (and all JVM langs that I know) argument of a method computed before actual invocation of this method.

For example if you have method `Value computeValue() { ... }` and you want to make an Observable from it, you can use `Observable.just(computeValue())` but the problem is that `computeValue()` will be computed in the thread that called `Observable.just()` and not on the Scheduler that you specify via `subscribeOn()`.

So, in my opinion `just()` and `from` are not "lazy".

`deferJust` looks good. `deferSingle` may conflict with `Single` for RxJava users.

headinthebox:
Sorry to be late into the game here, but it seems you really want an overload of `just` that has side-effects, i.e. reruns a factory function every time you subscribe to produce the value. I think the name should reflect that.

I am surprised this is a common scenario since it feel super narrow. Personally I think the longer `defer` code conveys the intent perfectly. Plus it also works if you want to do generate more than one value

```Java
Observable
  .defer(() -> {
    Value value0, value1, ...valuen-1 = ...;
    // some code
    return Observable.just(value0, ..., valuen-1);
  })
 .subscribeOn(...)
```
Anyway, just my 2 cents.

artem-zinnatullin:
Renamed operator from `lazy()` to `deferJust()`.

>I am surprised this is a common scenario since it feel super narrow. Personally I think the longer defer code conveys the intent perfectly.

Surprise surprise — nobody likes boilerplate. Also, if you need to defer function that throws checked exception via `defer()` you have to deal with `try/catch`, `deferJust()` handles it for you.

There is even post from @dlew [Deferring Observable code until subscription in RxJava](http://blog.danlew.net/2015/07/23/deferring-observable-code-until-subscription-in-rxjava/) which, I guess, makes this problem more visible.

headinthebox:
You have just solved the boilerplate for one very particular case that does not generalize to the other overloads of `just`, and increased the concept count. But if it makes you happy, it can't be that bad.

dlew:
I'm not a fan. I prefer composability. You can easily compose existing operators to recreate `deferJust`. They're in the name.

akarnokd:
I called this operator `fromCallable` in 2.x where the `from` operators are suffixed by the type to be very specific and to avoid overload resolution problems.

artem-zinnatullin:
Nice and understandable name :+1: Renamed to `fromCallable()` to make a switch from v1 to v2 easier.

artem-zinnatullin:
@akarnokd @benjchristensen is there anything that blocks this PR?

akarnokd:
I'm totally in favor of this addition so we can start promoting it instead of `create` with a single result so commonly found. We are still waiting for the second approve from another member.

artem-zinnatullin:
>we can start promoting it instead of create with a single result so commonly found.

Yes!

I see some confusion when newbies hear: "To defer some code and turn it into `Observable` you need to use `Observable.create()` (which is pretty tricky and boilerplate) or use create `Observable` via `just()` or `from()`, etc and defer it via `Observable.defer()`". 

I literally can read "wtf"s in their eyes when they ask — "So, I need to create `Observable` to defer an `Observable`, what?", this thing must be easier :)

stealthcode:
:-1: this is a less general use case than an `Observable.just()` or a `defer` that returns a `just` and it does not benefit from the merge optimizations.

akarnokd:
We can always optimize later.

stealthcode:
I see the compelling usability need (easy to type, simple to find) so I see that users would adopt it but it would detract from the optimized path. I don't think this will perform well if its adopted in the Netflix ecosystem. Many other users here have raised objections as this use case can be handled with existing operators and that composition would skip the merge ring buffer allocation/draining.

akarnokd:
That's not how the merge optimizations work.

stealthcode:
Oh, you're right. Merge can't optimize either path.

stealthcode:
My objections were on faulty assumptions. I take back my -1. :+1:

akarnokd:
Great!

jenzz:
Great addition! Thanks :+1:

benjchristensen:
> so we can start promoting it instead of create with a single result so
commonly found.

That is generally why `Single` exists. It still has its place on
`Observable`, but if just single responses are being done, such as for RPC,
then `Single` makes more sense.

Issue #2464, opened 2015-01-17
`Observable.using` sliently ignores exception of dispose action


ghost:
Code:

```java
import rx.Observable;
import rx.functions.Action0;
import rx.functions.Action1;
import rx.functions.Func0;
import rx.functions.Func1;

class Main {

    public static void main(String[] args) {
        Observable.using(
                new Func0<String>() {
                    @Override
                    public String call() {
                        System.out.println("Calling resourceFactory");
                        return "resourceFactory";
                    }
                },
                new Func1<String, Observable<String>>() {
                    @Override
                    public Observable<String> call(String str) {
                        System.out.println("Calling observableFactory");
                        return Observable.just("observableFactory");
                    }
                },
                new Action1<String>() {
                    @Override
                    public void call(String s) {
                        System.out.println("Calling disposeAction");
                        throw new RuntimeException("disposeAction");
                    }
                }
        ).subscribe(
                new Action1<String>() {
                    @Override
                    public void call(String s) {
                        System.out.println("onNext: " + s);
                    }
                },
                new Action1<Throwable>() {
                    @Override
                    public void call(Throwable e) {
                        System.out.println("onError: " + e.getMessage());
                    }
                },
                new Action0() {
                    @Override
                    public void call() {
                        System.out.println("onComplete");
                    }
                }
        );
    }
}
```

Output:

```
Calling resourceFactory
Calling observableFactory
onNext: observableFactory
onComplete
Calling disposeAction
```

This program finishes without error, and the `RuntimeException` thrown in dispose action is silently ignored.

akarnokd:
Hi. At that point, it can't notify the subscriber because it received a termination event before. The best thing we can do is to route the exception into the plugin system.

laec:
Can it be made so that `onCompleted` will be called after the dispose action returns successfully? So that if the dispose action throws an error, it can be notified with `onError`. This will make its behaviour consistent with the rest of the error notification workflow.

zsxwing:
@laec, according to the [Rx design guide](http://blogs.msdn.com/b/rxteam/archive/2010/10/28/rx-design-guidelines.aspx), you should not throw an error in the dispose action:

> 6.18. Unsubscription should not throw
[...]

PR #3155, opened 2015-08-14
SafeSubscriber - report onCompleted unsubscribe error to RxJavaPlugin 

davidmoten:
Discussed in #2464, when an observable emits `onCompleted` but `unsubscribe` in `SafeSubscriber` throws then the error should be reported to the `RxJavaPlugin` error handler and if that fails a stack trace is written to `System.err`.

### DIFF
--- a/src/test/java/rx/observers/SafeSubscriberTest.java
+++ b/src/test/java/rx/observers/SafeSubscriberTest.java
@@ -15,15 +15,25 @@
  */
 package rx.observers;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 import java.lang.reflect.Method;
+import java.util.concurrent.atomic.AtomicInteger;
 
-import org.junit.*;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
 
-import rx.exceptions.*;
+import rx.exceptions.OnCompletedFailedException;
+import rx.exceptions.OnErrorFailedException;
+import rx.exceptions.OnErrorNotImplementedException;
+import rx.exceptions.TestException;
+import rx.exceptions.UnsubscribeFailedException;
 import rx.functions.Action0;
-import rx.plugins.*;
+import rx.plugins.RxJavaErrorHandler;
+import rx.plugins.RxJavaPlugins;
 import rx.subscriptions.Subscriptions;
 
 public class SafeSubscriberTest {
@@ -51,10 +61,12 @@ public class SafeSubscriberTest {
             }
         };
         SafeSubscriber<Integer> safe = new SafeSubscriber<Integer>(ts);
-        
-        safe.onCompleted();
-        
-        assertTrue(safe.isUnsubscribed());
+        try {
+            safe.onCompleted();
+            Assert.fail();
+        } catch (OnCompletedFailedException e) {
+            assertTrue(safe.isUnsubscribed());
+        }
     }
     
     @Test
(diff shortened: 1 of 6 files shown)


davidmoten:
There are a lot of repeated try catch blocks so I've made a `RxJavaPluginUtils` class. It will also be called from the `using` fix which I will submit once this PR has been accepted.

akarnokd:
One of the tests fails.

davidmoten:
I'm going to ponder the issue of throwing after reporting to the error handler a bit.

davidmoten:
I've been looking for a bit of guidance about what to do when `unsubscribe` throws in `SafeSubscriber`. *Rx Design Guidelines* doesn't really cover it. *reactive-streams-jvm* does touch upon it saying:

>3.15 Calling Subscription.cancel MUST return normally. The only legal way to signal failure to a Subscriber is via the onError method.

I think given that the contract has been breached it is fair to throw an exception and I think a new wrapper `UnsubscribeFailureException` would be appropriate.

I'd also like to address the case when `onCompleted` throws. *reactive-streams-jvm* has this to say:

>2.13 Calling onSubscribe, onNext, onError or onComplete MUST return normally except when any provided parameter is null in which case it MUST throw a java.lang.NullPointerException to the caller, for all other situations the only legal way for a Subscriber to signal failure is by cancelling its Subscription. In the case that this rule is violated, any associated Subscription to the Subscriber MUST be considered as cancelled, and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.

So in the existing code if in `SafeSubscriber` an `onError` call to the downstream subscriber throws then the RxJavaPlugin error handler is called *and* an `OnErrorFailedException` is thrown. Comments in the method say this is ok because the contract has been broken and this is probably consistent with the advice in 3.13:

>and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.

In the existing code if in `SafeSubscriber` an `onCompleted` call to the downstream subscriber throws then the consequent call to `_onError` tries to signal `onError` to the downstream subscriber.  I'm worried that calling `onError` after calling `onCompleted` is a strong enough breach of the contract that the possible side effects could be worse than just a throw. Shouldn't we throw only (and notify the plugin error handler of course)?

davidmoten:
I've submitted code that matches my thinking above. It still fails because I haven't amended the existing unit tests associated with throwing. I'll do that after another round of review if accepted.

akarnokd:
The failing test is a coverage test for a behavior no longer present. I suggest removing it and rerunning the jacoco to see if all current code path is covered.

davidmoten:
@akarnokd Yep that's right. Are you happy that a throwing `onCompleted()` call should not induce an `onError()` call? If so I'll clean up and update the PR.

davidmoten:
I've fixed the unit tests and added more so that `SafeSubscriber` gets 100% coverage.

akarnokd:
:+1: Looks good to me.

Since this adds public API classes, it requires further approval (otherwise it shouldn't affect programs other than those that badly fail, therefore, the merge policy would allow me to merge it).

src/test/java/rx/observers/SafeSubscriberTest.java
```diff
@@ -51,10 +61,12 @@ public void onCompleted() {
             }
         };
         SafeSubscriber<Integer> safe = new SafeSubscriber<Integer>(ts);
-        
+        try {
         safe.onCompleted();
-        
-        assertTrue(safe.isUnsubscribed());
+        Assert.fail();
```
zsxwing: indent

zsxwing:
LGTM. It would be better if you can fix the indent. :+1: for UnsubscribeFailureException. It's a pity that we cannot change `OnErrorFailedException`s that should have been `UnsubscribeFailureException` now since it's a break change.

davidmoten:
Thanks for review @akarnokd  and @zsxwing. I've fixed indent in test and squashed commits.

akarnokd:
Great, thanks!

PR #3160, opened 2015-08-14
Remove redundant variables

vbauer:
Remove redundant variables from the following classes:

* OperatorTimeoutBase
* EventLoopsScheduler

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorTimeoutBase.java
+++ b/src/main/java/rx/internal/operators/OperatorTimeoutBase.java
@@ -157,10 +157,9 @@ class OperatorTimeoutBase<T> implements Operator<T, T> {
         }
 
         public void onTimeout(long seqId) {
-            long expected = seqId;
             boolean timeoutWins = false;
             synchronized (gate) {
-                if (expected == actual && TERMINATED_UPDATER.getAndSet(this, 1) == 0) {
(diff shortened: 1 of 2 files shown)


benjchristensen:
Often times developers use redundant variables for clarity.

vbauer:
I know that it is common practice for some developers, but I'm not sure that clarification via completely unnecessary variables is a good way / code design. All static analysis tools (like PMD, Findbugs, Checkstyle, Inspections from IDEA) agree with this. I think it will be better to have self-documented code which is well-formed for reading and debugging without this variables.

benjchristensen:
This is why we don't use PMD, Checkstyle, etc on this project, because every developer has a different subjective opinion.

vbauer:
Actually, this tools should prevent argues and stop subjective-thinking (to minimize human-related errors, to make a common code style, etc.), but I understand your opinion and you can reject this PR.

akarnokd:
Cosmetic change in internal classes, I don't think it's worth it.

stealthcode:
RxJava project does not have a style guide line for contributions. My initial thoughts are that the original committer of this particular code thought it was important to have a named value (potentially for ease of debugging or renaming in context) so it would be best to leave it as is. That said, if there were changes that you were making to the functionality then you might find it necessary or beneficial to collapse them (i.e. condense the code for readability) and so long as it wasn't too far off from the general style of the surrounding code then it would be just fine. Closing.

PR #3163, opened 2015-08-18
Making a cross between defer() and just() 

abersnaze:
that takes advantage of existing scalar synchronous optimizations.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -1068,6 +1068,10 @@ public class Observable<T> {
     public final static <T> Observable<T> defer(Func0<Observable<T>> observableFactory) {
         return create(new OnSubscribeDefer<T>(observableFactory));
     }
+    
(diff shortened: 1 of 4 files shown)


src/main/java/rx/Observable.java
```diff
@@ -1771,8 +1775,8 @@ public void call(Subscriber<? super Object> subscriber) {
     @Experimental
     @SuppressWarnings({"unchecked", "rawtypes"})
     public final static <T> Observable<T> merge(Observable<? extends Observable<? extends T>> source, int maxConcurrent) {
-        if (source.getClass() == ScalarSynchronousObservable.class) {
-            return ((ScalarSynchronousObservable<T>)source).scalarFlatMap((Func1)UtilityFunctions.identity());
+        if (source instanceof ScalarSynchronousObservable) {
+            return ((ScalarSynchronousObservable<Observable<T>>)source).get();
```
abersnaze: not sure if this is okay for the `ScalarSynchronousLazyObservable` since it makes it not so lazy.

PR #3169, opened 2015-08-20
Merge can now operate in horizontally unbounded mode.

### LINKED ISSUE #3156: Merge Limiting Concurrency - Can Deadlock
Since the `merge` changes in 1.0.13 we now have a bug in `merge` that limits concurrency and can cause an async "deadlock" since not all `Observable`s being merged will be subscribed to. 

These unit tests show the issue:

```java
    @Test
    public void testUnboundedDefaultConcurrency() {
    	List<Observable<Integer>> os = new ArrayList<Observable<Integer>>();
    	for(int i=0; i < 2000; i++) {
    		os.add(Observable.<Integer>never());
    	}
    	os.add(Observable.range(0, 100));    	
    	
    	TestSubscriber<Integer> ts = TestSubscriber.create();
    	Observable.merge(os).take(1).subscribe(ts);
    	ts.awaitTerminalEvent(5000, TimeUnit.MILLISECONDS);
    	ts.assertValue(0);
    	ts.assertCompleted();
    }
    
    @Test
    public void testConcurrencyLimit() {
    	List<Observable<Integer>> os = new ArrayList<Observable<Integer>>();
    	for(int i=0; i < 2000; i++) {
    		os.add(Observable.<Integer>never());
    	}
    	os.add(Observable.range(0, 100));    	
    	
    	TestSubscriber<Integer> ts = TestSubscriber.create();
    	Observable.merge(os, Integer.MAX_VALUE).take(1).subscribe(ts);
    	ts.awaitTerminalEvent(5000, TimeUnit.MILLISECONDS);
    	ts.assertValue(0);
    	ts.assertCompleted();
    }
```

Surprisingly, even when providing the `maxConcurrent` value the issue still happens.

We have bounded **vertical** buffers, but always must default to unbounded **horizontal** buffers, since it is the code that defines how many items are horizontally buffered. This affects both `merge` and `groupBy`. The `maxConcurrent` overload allows a developer to limit the horizontal buffering, and a developer controls the `groupBy` selector.

akarnokd:
Resolves #3156 

Note that since the default merge operation is unbounded, this change could lead to an excessive memory usage when flatMapping fast sources. Note that the pre 1.0.13 version did this albeit on a slighty slower path.

The change also affects the scalar optimization as well. Pre 1.0.13 implicitly limited the concurrency level to RxRingBuffer.SIZE when scalars were received. This version now fills the queue up to the concurrency level.

For 2.0, I suggest having a bounded behavior by default and require the developer to specify Integer.MAX_VALUE to go for the unbounded behavior so he/she knows about the consequences.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -17,13 +17,15 @@ package rx.internal.operators;
 
 import java.util.*;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.*;
+import java.util.concurrent.atomic.AtomicLong;
 
 import rx.*;
-import rx.Observable.Operator;
 import rx.Observable;
+import rx.Observable.Operator;
 import rx.exceptions.*;
 import rx.internal.util.*;
+import rx.internal.util.atomic.*;
+import rx.internal.util.unsafe.*;
 import rx.subscriptions.CompositeSubscription;
 
 /**
@@ -144,7 +146,7 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
         
         MergeProducer<T> producer;
         
-        volatile RxRingBuffer queue;
+        volatile Queue<Object> queue;
         
         /** Tracks the active subscriptions to sources. */
         volatile CompositeSubscription subscriptions;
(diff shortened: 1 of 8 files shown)


benjchristensen:
I'll have to allocate some time to review this later as it is somewhat detailed. Thank you for tackling this.

benjchristensen:
> Pre 1.0.13 implicitly limited the concurrency level to RxRingBuffer.SIZE when scalars were received.

That was done by design, since scalars behave as `onNext`. And we can limit the number of `Observable`s being merged if we have scalars queued up for delivery.

The reason we can't limit the number of async `Observable`s is because we don't know if they will emit or not and thus risk deadlock.

stealthcode:
:+1:

akarnokd:
I'll rebase this.

akarnokd:
Rebased, let's get this finally in.

stealthcode:
Thanks

PR #3171, opened 2015-08-21
Scan backpressure and first emission fix

### LINKED ISSUE #3168: BehaviorSubject.scan not picking up initial value
I would expect this:

```
final Observable<Void> ob = BehaviorSubject.create();
ob.scan(0, (accum, x) -> 123)
  .subscribe(x -> {
    Timber.d("Hit!");
  });
```

to emit a value (and indeed it does in the equivalent RxJS code).

Is this a bug?

akarnokd:
Fixes #3168

If there is no initial value, there is no need to manipulate the request amounts.

If there is an initial value, we must do a full emitter-loop with value queueing because now the downstream request can race with the first onNext or the setting of the upstream producer. The downside is the increased overhead which should be reduced when the spsc unbounded queues get merged.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorScan.java
+++ b/src/main/java/rx/internal/operators/OperatorScan.java
@@ -15,15 +15,14 @@
  */
 package rx.internal.operators;
 
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.Queue;
 
+import rx.*;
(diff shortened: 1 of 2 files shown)


stevegury:
:+1:

stealthcode:
:+1:

Issue #3063, opened 2015-07-03
toBlocking ignore all


davidmoten:
When I want to use `toBlocking` and I want to ignore results but wait for completion I will often do 

```java
observable.count().toBlocking().single();
```
This works but I'd like something concise for this common use case. We have a `forEach` method that takes an `Action1`. What about a `forEach` overload that takes no argument?

```java
observable.toBlocking().forEach();
```

or perhaps `.ignoreElements()`?

akarnokd:
I've been thinking about this too and could use these new blocking methods:

  - `run()` or `subscribe()` run to completion or exception
  - `subscribe(Observer)`, `subscribe(Action1, Action1)` and `subscribe(Action1, Action1, Action0)` blocking callback style
  - `subscribe(Subscriber)` allow breaking out of the iteration and have the unsubscription properly propagate up.

davidmoten:
Sounds good, may as well duplicate all the `subscribe()` overloads across to `BlockingObservable` that make sense. That symmetry would be nice.

pkafel:
+1 for the topic. I have many use cases where I would love to use something like that !

tommack:
I'll throw ``none()`` out there. Would be a void method that threw an Exception if any elements were output.

```
observable.ignoreElements().toBlocking().none();
```

akarnokd:
See for the proposed methods in #3174.

PR #3174, opened 2015-08-21
Blocking subscribe methods for convenience

akarnokd:
Mirroring some `Observable.subscribe` methods on `BlockingObservable`.

Related discussion: #3063.

### DIFF
--- a/src/main/java/rx/observables/BlockingObservable.java
+++ b/src/main/java/rx/observables/BlockingObservable.java
@@ -15,23 +15,19 @@
  */
 package rx.observables;
 
-import java.util.Iterator;
-import java.util.NoSuchElementException;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Future;
+import java.util.*;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 
+import rx.*;
 import rx.Observable;
-import rx.Subscriber;
-import rx.Subscription;
-import rx.functions.Action1;
-import rx.functions.Func1;
-import rx.internal.operators.BlockingOperatorLatest;
-import rx.internal.operators.BlockingOperatorMostRecent;
-import rx.internal.operators.BlockingOperatorNext;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/observables/BlockingObservable.java
```diff
@@ -477,4 +473,227 @@ private void awaitForComplete(CountDownLatch latch, Subscription subscription) {
             throw new RuntimeException("Interrupted while waiting for subscription to complete.", e);
         }
     }
+    
+    /**
+     * Runs the source observable to a terminal event, ignoring any values and rethrowing any exception.
+     */
+    @Experimental
+    public void run() {
```
abersnaze: can this be renamed to `subscribe()`?

akarnokd: There is already a method named `subscribe()`. I've put in this alias for discoverability.

src/main/java/rx/observables/BlockingObservable.java
```diff
@@ -477,4 +473,227 @@ private void awaitForComplete(CountDownLatch latch, Subscription subscription) {
...
+        run();
+    }
+    
+    /**
+     * Subscribes to the source and calls the given action on the current thread and rethrows any exception wrapped
+     * into OnErrorNotImplementedException.
+     * @param onNext the callback action for each source value
+     */
+    @Experimental
+    public void subscribe(final Action1<? super T> onNext) {
```
abersnaze: should this be an alias for `forEach(Action1)`?

akarnokd: They have different behavior. `forEach` executes the action on the thread that emits the `onNext` events and awaits the terminal event in the current thread. This method calls the action on the current thread.

abersnaze: Could you add a comment to the javadoc for both methods highlighting the difference between them?

akarnokd:
Done.

PR #3177, opened 2015-08-22
BackpressureUtils capped add/multiply methods + tests

akarnokd:
Common methods to add and multiply longs and cap the result at Long.MAX_VALUE in case of an overflow.

### DIFF
--- a/src/main/java/rx/internal/operators/BackpressureUtils.java
+++ b/src/main/java/rx/internal/operators/BackpressureUtils.java
@@ -44,11 +44,7 @@ public final class BackpressureUtils {
         // add n to field but check for overflow
         while (true) {
             long current = requested.get(object);
-            long next = current + n;
-            // check for overflow
-            if (next < 0) {
-                next = Long.MAX_VALUE;
-            }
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/BackpressureUtils.java
```diff
@@ -70,14 +66,41 @@ public static long getAndAddRequest(AtomicLong requested, long n) {
...
+    
+    /**
+     * Adds two positive longs and caps the result at Long.MAX_VALUE.
+     * @param a the first value
+     * @param b the second value
+     * @return the capped sum of a and b
+     */
+    public static long addCap(long a, long b) {
+        long u = a + b;
+        if (u < 0L) {
```
wendigo: Is this negative check necessary? long has 2^63 -1 capacity and let's assume that we will accumulate 10^9 every second we would still need 106 752 days to overflow (roughly 292 years).

akarnokd: Yes, because negative requests are not allowed and two large longs can overflow to a negative value.

PR #3183, opened 2015-08-24
Refactored exception reporting of most operators.

akarnokd:
Added two convenient methods to `Exceptions` which either throws a fatal exception or reports it with our without the causing value to an Observer.

### DIFF
--- a/src/main/java/rx/exceptions/Exceptions.java
+++ b/src/main/java/rx/exceptions/Exceptions.java
@@ -15,10 +15,9 @@
  */
 package rx.exceptions;
 
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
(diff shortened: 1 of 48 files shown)


stealthcode:
:+1:

stealthcode:
What if the `Exceptions.throwOrReport(...)` took a `Subscriber` instead? This way we could check if the downstream subscriber is unsubscribed and then if we are reporting we would instead log to the `RxJavaPlugins.getInstance().getErrorHandler().handleError(e)`. This would standardize usage of the error handler plugin. Thoughts?

akarnokd:
@stealthcode Your suggestion can be applied after the merge with another PR if you wish.

stealthcode:
I agree. Just suggesting for your input.

PR #3186, opened 2015-08-25
Disposable interface and a bunch of general resource containers.

akarnokd:
Mostly internal addition plus the public `Disposable` interface.

Naming of packages and classes can be changed at any time if one wishes...

### DIFF
--- a/src/main/java/io/reactivex/internal/disposables/ArrayCompositeResource.java
+++ b/src/main/java/io/reactivex/internal/disposables/ArrayCompositeResource.java
@@ -1,0 +1,102 @@
+/**
+ * Copyright 2015 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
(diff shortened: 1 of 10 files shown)


src/main/java/io/reactivex/internal/disposables/ArrayCompositeResource.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * A composite resource with a fixed number of slots.
+ *
+ * <p>Note that since the implementation leaks the methods of AtomicReferenceArray, one must be
+ * careful to only call setResource, replaceResource and dispose on it. All other methods may lead to undefined behavior
+ * and should be used by internal means only.
+ * 
+ * @param <T> the resource tpye
+ */
+public final class ArrayCompositeResource<T> extends AtomicReferenceArray<Object> implements Disposable {
```
artem-zinnatullin: Just curious: why you use inheritance over composition for such classes? I saw other classes in RxJava that extend `AtomicInteger`, etc. Why? :)

JakeWharton: Saves bytes. Collapse two object headers into one.

artem-zinnatullin: Sure, but exposes API. int vs enums 2.

JakeWharton: This is not public API

artem-zinnatullin: Yes, but 

mug-meme-okay

akarnokd: It saves bytes, allocation and gives access to intrinsified atomic operations such as getAndIncrement and others.

Issue #1353, opened 2014-06-11
Make operator parameter names consistent across operators.


DavidMGross:
I recommend that some of the parameter names used in Observable operators be changed so that similar parameters have similar names from operator to operator.

Specifically:

#### Collections of Observables

An iterable, list, or array of Observables is called "`sources`" in `amb()` and `combineLatest()`; "`sequences`" in `merge()`, and "`ws`" in `zip()`.

I recommend standardizing on "`observables`" as the parameter name for these.

#### Observables that emit Observables

An Observable that emits Observables is called "`observables`" in `concat()`, "`source`" in `merge()` and `mergeDelayError()`; "`sequenceOfSequences`" in `switchOnNext()`; and "`ws`" in `zip()`.

I recommend standardizing on "`observables`" as the parameter name for these

#### Ordinary Observable

Some operators take an ordinary Observable as a parameter.

The `zip()` operator calls the Observable that you pass as a parameter to zip with the source Observable "`other`"; the `groupJoin()` and `join()` operators use the parameter name "`right`" to describe a similar thing.

(The new operators `ambWith()`, `concatWith()`, and `mergeWith()` use "`t1`" for this sort of parameter; `zipWith()` uses "`other`".)

I recommend standardizing on "`other`" for this.

#### Multiple Observables

Some operators take two or more Observables as parameters.
* `amb()`, `combineLatest()`, and `zip()` call these "`o1`," "`o2`," "`o3`," etc.
* `concat()`, `merge()`, and `mergeDelayError()` call these "`t1`," "`t2`," "`t3`," etc.
* `sequenceEqual()` calls these "`first`" and "`second`"

My recommendation: for at least the operators in the first two bullet points, choose a single naming scheme and standardize over it for all of those operators.

#### Special-Function Observables

Some operators are passed Observables that emit indicators that instruct the operators to do certain things at certain times.  The naming schemes for these could be improved.

Currently:
**`buffer()`**
* opening indicator: `bufferOpenings`
* closing indicator: `bufferClosingSelector`
* boundary indicator: `boundary`

**`window()`**
* opening indicator: `windowOpenings`
* closing indicator: `closingSelector`
* boundary indicator: `boundary`

My recommendation for both:
* opening indicator: `openingIndicator`
* closing indicator: `closingIndicator`
* boundary indicator: `boundaryIndicator`

I also recommend changing the names of the following parameters so that they conform to a similar naming scheme:
[...]

PR #3196, opened 2015-08-26
The skips and filter operators added to Observable

akarnokd:
Forgot to add them in the previous PR.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -307,7 +307,7 @@ public class Observable<T> implements Publisher<T> {
         return fromIterable(sources).flatMap(v -> v, true, maxConcurrency, bufferSize);
     }
 
-    public Observable<T> take(long n) {
+    public final Observable<T> take(long n) {
         if (n < 0) {
             throw new IllegalArgumentException("n >= required but it was " + n);
         } else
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -317,15 +317,17 @@ public static int bufferSize() {
         return lift(new OperatorTake<>(n));
     }
     
-    public <U> Observable<T> takeUntil(Publisher<U> other) {
+    public final <U> Observable<T> takeUntil(Publisher<U> other) {
```
DavidMGross: See #1353 for concerns about the parameter name "other" and suggestions for a more consistent parameter naming scheme.

src/main/java/io/reactivex/Observable.java
```diff
@@ -317,15 +317,17 @@ public static int bufferSize() {
...
     }
     
-    public Observable<T> takeUntil(Predicate<? super T> predicate) {
+    public final Observable<T> takeUntil(Predicate<? super T> predicate) {
+        Objects.requireNonNull(predicate);
         return lift(new OperatorTakeUntilPredicate<>(predicate));
     }
     
-    public Observable<T> takeLast(int n) {
+    public final Observable<T> takeLast(int n) {
```
DavidMGross: See #1353 for a suggestion for consistent naming of parameters that indicate a count of items to be emitted across operators with such parameters.

PR #3198, opened 2015-08-26
Operators publish, replay and cache.

akarnokd:
I'm not 100% certain about the the timing of the onSubscribe() calls inside the operators; I'll have a second run over them once the tests are also added.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -22,7 +22,9 @@ import org.reactivestreams.*;
 
 import io.reactivex.internal.operators.*;
 import io.reactivex.internal.subscriptions.EmptySubscription;
+import io.reactivex.observables.ConnectableObservable;
 import io.reactivex.plugins.RxJavaPlugins;
+import io.reactivex.schedulers.Schedulers;
 import io.reactivex.subscribers.SafeSubscriber;
 
 public class Observable<T> implements Publisher<T> {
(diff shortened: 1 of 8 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -420,4 +422,129 @@ public static int bufferSize() {
...
+    }
+
+    public final ConnectableObservable<T> publish(int bufferSize) {
+        if (bufferSize <= 0) {
+            throw new IllegalArgumentException("bufferSize > 0 required but it was " + bufferSize);
+        }
+        return OperatorPublish.create(this, bufferSize);
+    }
+
+    public final <R> Observable<R> publish(Function<? super Observable<T>, ? extends Observable<R>> selector) {
```
DavidMGross: As per #1353 I recommend something other than "selector" for the parameter name (here and elsewhere) "since these functions aren't really 'selecting' items so much as changing them."

src/main/java/io/reactivex/Observable.java
```diff
@@ -420,4 +422,129 @@ public static int bufferSize() {
...
+
+    public final <R> Observable<R> replay(Function<? super Observable<T>, ? extends Observable<R>> selector) {
+        return OperatorReplay.multicastSelector(this::replay, selector);
+    }
+    
+    public final <R> Observable<R> replay(Function<? super Observable<T>, ? extends Observable<R>> selector, final int bufferSize) {
+        return OperatorReplay.multicastSelector(() -> replay(bufferSize), selector);
+    }
+    
+    public final <R> Observable<R> replay(Function<? super Observable<T>, ? extends Observable<R>> selector, int bufferSize, long time, TimeUnit unit) {
```
DavidMGross: See #1353 for concern about the parameter name "time" (similar parameters in other operators are called "timespan", "timeout", "interval", "period", "intervalDuration", etc.; those parameters that have pretty much the same function should have the same name).

akarnokd: Okay, don't forget it. There will be several rename/refactor passes later.

PR #3203, opened 2015-08-27
Implemented the AsyncOnSubscribe

stealthcode:
This includes the first functioning unit tests. This still has a few iterations to go but opening PR for review.

### DIFF
--- a/src/main/java/rx/observables/AsyncOnSubscribe.java
+++ b/src/main/java/rx/observables/AsyncOnSubscribe.java
@@ -1,0 +1,510 @@
+/**
+ * Copyright 2015 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package rx.observables;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicReference;
+
+import rx.Observable;
+import rx.Observable.OnSubscribe;
+import rx.Observer;
+import rx.Producer;
+import rx.Subscriber;
+import rx.Subscription;
+import rx.annotations.Experimental;
+import rx.functions.Action0;
+import rx.functions.Action1;
+import rx.functions.Action2;
+import rx.functions.Action3;
+import rx.functions.Func0;
+import rx.functions.Func3;
+import rx.internal.operators.BufferUntilSubscriber;
+import rx.observers.SerializedObserver;
+import rx.observers.Subscribers;
+import rx.plugins.RxJavaPlugins;
+import rx.subscriptions.BooleanSubscription;
+;
+/**
+ * A utility class to create {@code OnSubscribe<T>} functions that respond correctly to back
+ * pressure requests from subscribers. This is an improvement over
+ * {@link rx.Observable#create(OnSubscribe) Observable.create(OnSubscribe)} which does not provide
+ * any means of managing back pressure requests out-of-the-box. This variant of an OnSubscribe
+ * function allows for the asynchronous processing of requests.
+ *
+ * @param <S>
+ *            the type of the user-define state used in {@link #generateState() generateState(S)} ,
+ *            {@link #next(Object, Long, Subscriber) next(S, Long, Subscriber)}, and
+ *            {@link #onUnsubscribe(Object) onUnsubscribe(S)}.
+ * @param <T>
+ *            the type of {@code Subscribers} that will be compatible with {@code this}.
+ */
+@Experimental
+public abstract class AsyncOnSubscribe<S, T> implements OnSubscribe<T> {
+
+    /**
+     * Executed once when subscribed to by a subscriber (via {@link OnSubscribe#call(Subscriber)})
+     * to produce a state value. This value is passed into {@link #next(Object, long, Observer)
+     * next(S state, Observer <T> observer)} on the first iteration. Subsequent iterations of
+     * {@code next} will receive the state returned by the previous invocation of {@code next}.
+     * 
+     * @return the initial state value
+     */
+    protected abstract S generateState();
+
+    /**
+     * Called to produce data to the downstream subscribers. To emit data to a downstream subscriber
+     * call {@code observer.onNext(t)}. To signal an error condition call
+     * {@code observer.onError(throwable)} or throw an Exception. To signal the end of a data stream
+     * call {@code observer.onCompleted()}. Implementations of this method must follow the following
+     * rules.
+     * 
+     * <ul>
+     * <li>Must not call {@code observer.onNext(t)} more than 1 time per invocation.</li>
+     * <li>Must not call {@code observer.onNext(t)} concurrently.</li>
+     * </ul>
+     * 
+     * The value returned from an invocation of this method will be passed in as the {@code state}
+     * argument of the next invocation of this method.
+     * 
+     * @param state
+     *            the state value (from {@link #generateState()} on the first invocation or the
+     *            previous invocation of this method.
+     * @param requested
+     *            the amount of data requested. An observable emitted to the observer should not
+     *            exceed this amount.
+     * @param observer
+     *            the observer of data emitted by
+     * @return the next iteration's state value
+     */
+    protected abstract S next(S state, long requested, Observer<Observable<? extends T>> observer);
+
+    /**
+     * Clean up behavior that is executed after the downstream subscriber's subscription is
+     * unsubscribed. This method will be invoked exactly once.
+     * 
+     * @param state
+     *            the last state value returned from {@code next(S, Long, Observer)} or
+     *            {@code generateState()} at the time when a terminal event is emitted from
+     *            {@link #next(Object, long, Observer)} or unsubscribing.
+     */
+    protected void onUnsubscribe(S state) {
+
+    }
+
+    /**
+     * Generates a synchronous {@link AsyncOnSubscribe} that calls the provided {@code next}
+     * function to generate data to downstream subscribers.
+     * 
+     * @param generator
+     *            generates the initial state value (see {@link #generateState()})
+     * @param next
+     *            produces data to the downstream subscriber (see
+     *            {@link #next(Object, long, Observer) next(S, long, Observer)})
+     * @return an OnSubscribe that emits data in a protocol compatible with back-pressure.
+     */
+    @Experimental
+    public static <S, T> OnSubscribe<T> createSingleState(Func0<? extends S> generator, 
+            final Action3<? super S, Long, ? super Observer<Observable<? extends T>>> next) {
+        Func3<S, Long, ? super Observer<Observable<? extends T>>, S> nextFunc = 
+                new Func3<S, Long, Observer<Observable<? extends T>>, S>() {
+                    @Override
+                    public S call(S state, Long requested, Observer<Observable<? extends T>> subscriber) {
+                        next.call(state, requested, subscriber);
+                        return state;
+                    }};
+        return new AsyncOnSubscribeImpl<S, T>(generator, nextFunc);
+    }
+
+    /**
+     * Generates a synchronous {@link AsyncOnSubscribe} that calls the provided {@code next}
+     * function to generate data to downstream subscribers.
+     * 
+     * This overload creates a AsyncOnSubscribe without an explicit clean up step.
+     * 
+     * @param generator
+     *            generates the initial state value (see {@link #generateState()})
+     * @param next
+     *            produces data to the downstream subscriber (see
+     *            {@link #next(Object, long, Observer) next(S, long, Observer)})
+     * @param onUnsubscribe
+     *            clean up behavior (see {@link #onUnsubscribe(Object) onUnsubscribe(S)})
+     * @return an OnSubscribe that emits data downstream in a protocol compatible with
+     *         back-pressure.
+     */
+    @Experimental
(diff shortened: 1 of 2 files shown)


akarnokd:
What's the purpose of this class?

stealthcode:
This is the asynchronous complement to the SyncOnSubscribe as described in #3003.

stealthcode:
The remaining issue with this implementation is that the inner observables are subscribed to an unbounded buffer. This could be improved to a back pressure supporting `RxRingBuffer` that requests from it's producer only to fill it's buffer capacity and then when it connects, drains the buffer and forwards all requests to its producer.

akarnokd:
l still don't understand this class. Looks very similar to a SyncOnSubscribe that emits Observables and flatMaps it for you.

stealthcode:
I understand your concern but I'm not sure about your comment. This operator does not call flatMap. This issue comes from use cases at Netflix with large services and streaming work. I think it's pretty clear to see from the unit tests that this gives a user direct access to the request and allows fulfillment by any observable (which may or may not be asynchronous).

stealthcode:
This is very similar to the work that you tried with the AbstractProducer. It allows batching and eager subscription.

stealthcode:
This is supposed to help with creating observables from asynchronous data sources in a non-blocking way. Consider the example below...

```java
    Observable.<Integer>create((Subscriber<? super Integer> s) -> {
        s.setProducer((long requested) -> {
            getData(requested)
                .observe()
                .subscribe(s::onNext, s::onError, s::onCompleted);
        });
    });
}

public static HystrixCommand<Integer> getData(long requested) {
    return null;
}

public static interface HystrixCommand<T> {
    public Observable<T> observe();
}
```

The use case here is that for each request we want to issue an asynchronous service request and when that request is fulfilled then we want to start onNexting the data for that request. The problem with this naive implementation is that it's very easy and certainly possible to have overlapping requests. You could implement the cas addToRequestAndGet and make sure that only one thread at a time is fulfilling the request at a time but you still have the possibility of onNext events interleaving or overlapping.

abersnaze:
it doesn't help with truly hot sources but it'll work great for progressive async sources. :+1:

benjchristensen:
I have reviewed this work with @stealthcode as he's been working on it and am quite interested in where this goes. It can greatly help us solve the most difficult part of using RxJava: creating correct sources that support backpressure. 

I'll open an issue to summarize the needs for 1.x and 2.x so we can all get aligned on this topic.

I'm +1 on experimenting with this and eventually getting `Observable.create` overloads for all these helpers.

benjchristensen:
Actually, there is already an issue: https://github.com/ReactiveX/RxJava/issues/3003

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+            this(null, next, onUnsubscribe);
+        }
+
+        public AsyncOnSubscribeImpl(Func3<S, Long, Observer<Observable<? extends T>>, S> nextFunc) {
+            this(null, nextFunc, null);
+        }
+
+        @Override
+        protected S generateState() {
+            return generator == null ? null : generator.call();
```
akarnokd: It would be great, if this is to be migrated to 2.x, that user-supplied functions are called in try-catch. Perhaps not here but in call().

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+                        }
+                    } catch (Throwable ex) {
+                        handleThrownError(parent, state, ex);
+                        return;
+                    }
+                } while (r != null && !hasTerminated);
+            }
+        }
+
+        private void handleThrownError(final AsyncOnSubscribe<S, T> p, S st, Throwable ex) {
```
akarnokd: Parameters `p` and `st` are not used in the body; I'm not sure if the intention was to use them but was forgotten.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+
+    private static final class UnicastSubject<T> extends Observable<T>implements Observer<T> {
+        public static <T> UnicastSubject<T> create() {
+            return new UnicastSubject<T>(new State<T>());
+        }
+
+        private State<T> state;
+
+        protected UnicastSubject(final State<T> state) {
+            super(new OnSubscribe<T>() {
```
akarnokd: `State<T>` can implement `OnSubscribe<T>` and you save one object allocation.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+            synchronized (this) {
+                size = size();
+                add(n);
+                r = n;
+            }
+            if (size == 0) {
+                do {
+                    // check if unsubscribed before doing any work
+                    if (isUnsubscribed()) {
+                        unsubscribe();
```
akarnokd: If `isUnsubscribed != 0` then the CAS in `unsubscribe` will fail and nothing happens. If you wanted to delay the unsubscription to the time when nextIteration is not running, then you need a separate indicator, i.e., unsubscribeRequested.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+                    public void call() {
+                        if (!holder.compareAndSet(null, SUBSCRIPTION_SENTINEL)) {
+                            Subscription h = holder.get();
+                            subscriptions.remove(h);
+                        }
+                    }})
+                .subscribe(buffer);
+
+            if (holder.compareAndSet(null, innerSubscription)) {
+                subscriptions.add(innerSubscription);
```
akarnokd: The problem with this is that innerSubscription is added to subscriptions and then not removed from it until the whole sequence terminates.

akarnokd: I'm not sure about the thread-safety of accessing `subscriptions`. If doOnTerminate happen to run on a different thread, `subscriptions` may get corrupted and even if backed by CHM, the remove on line 458 may happen just before the add on line 464 and `subscriptions` now leaks the reference. 

What would be better is to make sure `subscriptions` always holds that needs to be removed by adding the innerSubscription before calling t.subscribe() which also means you need to pre-wrap `buffer` so that subscribe returns the same instance. This way, the wrapped buffer instance is added before the call to subscribe and will be guaranteed to be removed when `doOnTerminate` is executed.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+    public final void call(Subscriber<? super T> actualSubscriber) {
+        S state = generateState();
+        UnicastSubject<Observable<T>> subject = UnicastSubject.<Observable<T>> create();
+        AsyncOuterSubscriber<S, T> outerSubscriberProducer = new AsyncOuterSubscriber<S, T>(this, state, subject);
+        actualSubscriber.add(outerSubscriberProducer);
+        Observable.concat(subject).unsafeSubscribe(Subscribers.wrap(actualSubscriber));
+        actualSubscriber.setProducer(outerSubscriberProducer);
+    }
+
+    private static class AsyncOuterSubscriber<S, T> extends ConcurrentLinkedQueue<Long>implements Producer, Subscription, Observer<Observable<? extends T>> {
```
akarnokd: `ConcurrentLinkedQueue` seems to be expensive for the usage pattern; you could get away with `SpscLinkedQueue` or `SpscLinkedArrayQueue` here (no need for `MpscLinkedQueue` because the offer side is inside a synchronized block).

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+                    // check if unsubscribed before doing any work
+                    if (isUnsubscribed()) {
+                        unsubscribe();
+                        return;
+                    }
+                    // otherwise try one iteration for a request of `numRequested` elements
+                    try {
+                        onNextCalled = false;
+                        nextIteration(r);
+                        if (onNextCalled)
```
akarnokd: If terminated, you can skip the poll.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+                            r = poll();
+                        if (hasTerminated || isUnsubscribed()) {
+                            parent.onUnsubscribe(state);
```
akarnokd: If `nextIteration` has terminated, the queue may still hold requested values so it might be worth considering calling `clear()` before quitting.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+     * 
+     * The value returned from an invocation of this method will be passed in as the {@code state}
+     * argument of the next invocation of this method.
+     * 
+     * @param state
+     *            the state value (from {@link #generateState()} on the first invocation or the
+     *            previous invocation of this method.
+     * @param requested
+     *            the amount of data requested. An observable emitted to the observer should not
+     *            exceed this amount.
```
akarnokd: I'm not certain this requirement is enough because if a source doesn't deliver the requested amount, it will keep the client hanging which may not request a new batch. There is also the problem if the first observable delivers less than requested, the missing amount is not replenished/considered by the second Observable and thus the client may starve. This is why `concat` uses the `SubscriptionArbiter` so that if an Observable didn't deliver enough, the next is requested for the missed amount. Similar measures might be required with this `AsyncOnSubscribe`.

src/test/java/rx/observables/AsyncOnSubscribeTest.java
```diff
@@ -0,0 +1,408 @@
...
+                    observer.onNext(o);
+                }
+                else 
+                    observer.onCompleted();
+                return state + 1;
+            }});
+        // initial request emits [[1, 2, 3, 4]] on delay
+        Observable.create(os).subscribe(subscriber);
+        // next request emits [[5, 6, 7, 8]] firing immediately
+        subscriber.requestMore(2);
```
akarnokd: I believe `TestSubscriber` requests `Long.MAX_VALUE` by default so these requests do nothing. In addition, I think if you had a state 3 with yet another observable, a sequence of requests of 1 would overflow the `concat` queue and yield a `MissingBackpressureException` - however, I can't check this scenario right now.

src/test/java/rx/observables/AsyncOnSubscribeTest.java
```diff
@@ -0,0 +1,408 @@
...
+        OnSubscribe<Integer> os = AsyncOnSubscribe.createStateless(
+        new Action2<Long, Observer<Observable<? extends Integer>>>(){
+            @Override
+            public void call(Long requested, Observer<Observable<? extends Integer>> observer) {
+                observer.onNext(Observable.range(1, requested.intValue()));
+            }});
+        Observable.create(os).observeOn(scheduler).subscribe(subscriber);
+        scheduler.advanceTimeBy(10, TimeUnit.DAYS);
+        subscriber.assertNoErrors();
+        subscriber.assertValueCount(RxRingBuffer.SIZE);
```
akarnokd: I don't understand how `subscriber` receives only `RxRingBuffer.SIZE` values and `observeOn` doesn't request more if `subscriber` appears to be requesting Long.MAX_VALUE?

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+        }
+
+        @Override
+        public void unsubscribe() {
+            if (IS_UNSUBSCRIBED.compareAndSet(this, 0, 1)) {
+                // it's safe to process terminal behavior
+                if (isEmpty()) {
+                    parent.onUnsubscribe(state);
+                }
+                for (Subscription s : subscriptions) {
```
akarnokd: I think this may race with the termination of the generated `Observables` if they are async.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: It's merged into 1.x. Feel free to open a pull request to 2.x if you have time or to 1.x if you feel its necessary.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: It was my intention to delete them.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: Feel free to open a pull request.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: Can you write a test that demonstrates a problem?

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: I'm open to alternatives. I used what was already in use in other operators.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: If you could write a test that demonstrates your theory then that would help to debug/refine.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: Checking after `nextIteration(r)` is sufficient. There is not a case where you would be terminated before the first iteration and the only thing that can set `hasTerminated` is the onCompleted or onError which can only be invoked by the next func.

src/main/java/rx/observables/AsyncOnSubscribe.java
stealthcode: Good point. This would affect resubscriptions.

src/main/java/rx/observables/AsyncOnSubscribe.java
```diff
@@ -0,0 +1,510 @@
...
+        protected UnicastSubject(final State<T> state) {
+            super(new OnSubscribe<T>() {
+                @Override
+                public void call(Subscriber<? super T> s) {
+                    if (state.subscriber != null) {
+                        s.onError(new IllegalStateException("There can be only one subscriber"));
+                    } else {
+                        state.subscriber = s;
+                    }
+                }
```
davidmoten: I'm not sure if this class could get used by multiple concurrent subscribers (?) but if it does then I would expect to see an atomic compareAndSet here rather than just !=null.

PR #3208, opened 2015-08-27
Operators interval and intervalRange

akarnokd:
`intervalRange` helps when one needs a fixed amount of timed values and wants to avoid map+take overhead.


benjchristensen:
What is the overhead of take that this saves? I don't see the point. Especially since interval goes through scheduling for each emission which has to cost more than take.

akarnokd:
The operator involves less allocation when a sequence is assembled anda  lover memory footprint when active.

PR #3214, opened 2015-08-27
Fix to Notification equals method.

wrightm:
The Notification method was missing a check to hasValue and hasThrowable. This allowed for a null value onNext Notification to equal an Object notification. A similar bug for onError Notification events also existed.


stealthcode:
:+1: this seems like a valid contribution with tests.

stevegury:
:+1:

wrightm:
Thanks guys, hopefully the first of many contributions.

PR #3258, opened 2015-08-31
ReplaySubject

akarnokd:
Has full backpressure support (replays as many elements as requested).


davidmoten:
@akarnokd Do you also plan to provide  backpressure support for `ReplaySubject` in 1.x?

akarnokd:
There was an attempt before but didn't get through. The problem with 1.x is that it has an expected optimization to goes on a fast path once the subscriber has caught up and no further replay/buffering happens for that particular subscriber. With backpressure, there would be one CAS at minimum for every value delivered which has 6x the overhead compared to a write-through.

PR #3272, opened 2015-09-02
Firehose test fix attempt2x

akarnokd:
Either the test is non-deterministic or I have a bug in the operator(s).


akarnokd:
OOps, wrong branch.

PR #3285, opened 2015-09-02
Added latch to async SyncOnSubscrbeTest

stealthcode:


akarnokd:
#3286 has encountered another failure:

```
rx.observables.SyncOnSubscribeTest > testSubscribeOn FAILED
    java.lang.AssertionError: expected:<4> but was:<3>
        at org.junit.Assert.assertEquals(Assert.java:472)
        at org.junit.Assert.assertEquals(Assert.java:456)
        at rx.observables.SyncOnSubscribeTest.testSubscribeOn(SyncOnSubscribeTest.java:734)
```

stealthcode:
This should also resolve #3287. I have added your example test to the unit tests (3,000 iterations).

src/main/java/rx/observables/SyncOnSubscribe.java
```diff
@@ -337,14 +331,39 @@ private SubscriptionProducer(final Subscriber<? super T> subscriber, SyncOnSubsc
...
-            return isUnsubscribed != 0;
+            return get() < 0L;
         }
         
         @Override
         public void unsubscribe() {
-            IS_UNSUBSCRIBED.compareAndSet(this, 0, 1);
-            if (get() == 0L)
-                parent.onUnsubscribe(state);
+            while(true) {
```
akarnokd: There is a much simpler way of fixing this: undo all these latest changes and request 1!

```java
if (BackpressureUtils.addRequested(this, 1) == 0) {
    parent.onUnsubscribe(state);
}
```

src/test/java/rx/observables/SyncOnSubscribeTest.java
```diff
@@ -969,14 +936,73 @@ public void call() {
...
+        os.call(ts);
+        ts.assertCompleted();
+        ts.assertNoErrors();
+        ts.assertValueCount(1);
+    }
+    
+    @Test
+    public void testConcurrentUnsubscribe3000Iterations() throws InterruptedException, BrokenBarrierException, ExecutionException{
+        for (int i = 0; i < 3000; i++) {
+            testUnsubscribeConcurrently();
```
akarnokd: You could have inlined this loop into the other method because this will create 3000 threads.

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: It's interesting. If there is currently a thread processing over the request count loop (i.e. processing nextIteration()) then won't this have to block the unsubscribing thread? The unsubscribing thread would have to poll over this until it got access. I'm not seeing how this would force a thread currently iterating to terminate. 

If you read through I think you will see that

src/test/java/rx/observables/SyncOnSubscribeTest.java
stealthcode: Inlining the method will still create 3000 threads.

src/main/java/rx/observables/SyncOnSubscribe.java
akarnokd: There is no blocking. My suggestion follows the queue-drain approach where you queue the terminal flag and drain it in the emission loop. Since the loop first checks for the terminal flag there won't be any wrong emissions and the loop quits. If unsubscribe increments from zero, there isn't and won't be any emission loop after that and the resource can be freed.

There is no need for your complicated solution.

src/test/java/rx/observables/SyncOnSubscribeTest.java
akarnokd: Factor out the thread pool creation from the loop!

src/main/java/rx/observables/SyncOnSubscribe.java
stealthcode: I'm not seeing where you are suggesting setting a terminal flag. You are more than welcome to open your own pull request to demonstrate your fix. My pull request does set a flag for the draining thread to signal when it needs to unsubscribe.

src/test/java/rx/observables/SyncOnSubscribeTest.java
stealthcode: Okay, that's not what you said in your comment. I can do this but personally I am not too concerned with the thread or memory efficiency in unit tests.

abersnaze:
:+1:

akarnokd:
Okay, let's merge this so other PRs don't run into the test bug with the current main.

PR #3289, opened 2015-09-02
Added more tests and cleaned up SubscriptionList to program to List

wrightm:
I cleaned up and added more tests to SubscriptionList. 

I also changed the class to implement the List interface instead of hard code LinkedList. 

Changed Data Structure to LinkedList from ArrayList as List is used to iterate over and add exceptions to list. No random access needed and insertions are more efficient in LinkedList.

### DIFF
--- a/src/main/java/rx/internal/util/SubscriptionList.java
+++ b/src/main/java/rx/internal/util/SubscriptionList.java
@@ -27,7 +27,7 @@ import rx.exceptions.Exceptions;
  */
 public final class SubscriptionList implements Subscription {
 
-    private LinkedList<Subscription> subscriptions;
+    private List<Subscription> subscriptions;
     private volatile boolean unsubscribed;
 
     public SubscriptionList() {
@@ -62,7 +62,7 @@ public final class SubscriptionList implements Subscription {
         if (!unsubscribed) {
             synchronized (this) {
                 if (!unsubscribed) {
-                    LinkedList<Subscription> subs = subscriptions;
+                    List<Subscription> subs = subscriptions;
                     if (subs == null) {
                         subs = new LinkedList<Subscription>();
                         subscriptions = subs;
@@ -80,7 +80,7 @@ public final class SubscriptionList implements Subscription {
         if (!unsubscribed) {
             boolean unsubscribe = false;
             synchronized (this) {
-                LinkedList<Subscription> subs = subscriptions;
+                List<Subscription> subs = subscriptions;
                 if (unsubscribed || subs == null) {
                     return;
                 }
@@ -114,7 +114,7 @@ public final class SubscriptionList implements Subscription {
         }
     }
 
-    private static void unsubscribeFromAll(Collection<Subscription> subscriptions) {
+    private void unsubscribeFromAll(Collection<Subscription> subscriptions) {
         if (subscriptions == null) {
             return;
         }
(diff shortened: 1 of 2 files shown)


wrightm:
Looks like test failure not from changed codebase. 

rx.observables.SyncOnSubscribeTest > testUnsubscribeOutsideOfLoop FAILED
    java.lang.AssertionError: Not unsubscribed.

Is someone looking into this?

davidmoten:
I just had a brief look at this. I do notice that white space is screwed up (no tabs please and inconsistent indents).

akarnokd:
:-1:

  - The resource container field is deliberately `LinkedList` so it is not turned into an `ArrayList` and thus making it O(n) removal.
  - When the exceptions are aggregated, it doesn't really matter what kind of list is that since the throwing of any exception has several orders of magnitude more overhead than the list-add overhead
  - the method `unsubscribeFromAll` is static so it doesn't try to access the instance state of the class and cause concurrency issues.

wrightm:
@davidmoten Sorry about that didn't notice it was changed from the initial file. Are we using a style guide that I can use with my IDE?

wrightm:
@akarnokd  Thank you for your comments. 

* On your first I have only change to an interface so that If there is a better data structure in the future we only have to change the initialisation. I understand why ``` LinkedList ```  is used over ``` ArrayList ``` . 

* Changing from the ``` ArrayList ```  to ``` LinkedList```  was motivated by the use of O(n) complexity when adding to the ``` ArrayList```  and also creation of a new ``` ArrayList ``` when capacity*load is met. I should of backed this up with a benchmark to show when and if there is a use case that would benefit the use.

* Could you explain you last point, I can not see any concurrency issues using a non static method. I see there could be a potential fail fast issue with the subscriptions iteration (``` ConcurrentModificationException ``` ), but how the method is currently used in the class it shouldn't be an issue - subscriptions  = null is set in the methods that use ``` unsubscribeFromAll ```

akarnokd:
> On your first I have only change to an interface so that If there is a better data structure in the future we only have to change the initialisation. I understand why LinkedList is used over ArrayList .

I doubt that if this class gets further optimized, that's going to use a `List` implementation.

> Could you explain you last point, I can not see any concurrency issues using a non static method. I see there could be a potential fail fast issue with the subscriptions iteration (ConcurrentModificationException ), but how the method is currently used in the class it shouldn't be an issue - subscriptions = null is set in the methods that use unsubscribeFromAll

This is to avoid programmer mistakes.

wrightm:
> I doubt that if this class gets further optimized, that's going to use a List implementation.
Its nice to have the flexibility that coding to interface brings, rather than an implementation. 

> This is to avoid programmer mistakes.
I believe this would be solved by renaming the input parameter on the method, rather than making it a static method. I can resubmit this pull request to sort the naming issue?

akarnokd:
First of all, this PR, so far, is cosmetic in nature which we rarely accept. 

Second, don't change existing unit tests unless you are fixing a bug in them. This goes also for renaming/refactoring existing stuff or content: if you are improving or fixing an existing operator, you can rename things up to an extent if that gives you more oversight over said change.

wrightm:
Thanks @akarnokd this will focus my contributions in the future. It would be nice to have this type of information in the contributions documentation if not already.

akarnokd:
Yes, many things are not well documented yet in the project.

PR #3304, opened 2015-09-05
Enables assertions chaining

tombujok:
Fluent assertions make the code much more readable.
This change is backward compatible.

### DIFF
--- a/src/main/java/rx/observers/TestObserver.java
+++ b/src/main/java/rx/observers/TestObserver.java
@@ -111,7 +111,7 @@ public class TestObserver<T> implements Observer<T> {
      * @throws AssertionError
      *          if the sequence of items observed does not exactly match {@code items}
      */
-    public void assertReceivedOnNext(List<T> items) {
+    public TestObserver assertReceivedOnNext(List<T> items) {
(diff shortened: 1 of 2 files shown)


JakeWharton:
Changing return types is a binary-incompatible change.

tombujok:
Do you care about binary compatibility of a test util class?

akarnokd:
Since both `TestObserver` and `TestSubscriber` are part of the public API, they are subject to the binary compatibility requirement regarding existing methods.

tombujok:
Do you see a risk here?

akarnokd:
Yes.

I'm sorry, but such change can't be accepted for 1.x, however, you are free to propose this against 2.x.

tombujok:
No problem. 
Hope that users do not release their binary production code with the usage of `TestObserver` and `TestSubscriber`

benjchristensen:
It's not about releasing it to prod. If we break our public API, we break their code bases. Just because it's their test code does not mean we have a lower commitment to not breaking it. 

Please get involved in the 2.x effort which is just getting underway for API related changes.

PR #3305, opened 2015-09-05
2.x: classes Single and NbpObservable

akarnokd:
`Single` is a deferred single-value emitting Observable.

`NbpObservable` is the non-backpressure Observable, named this way until
the main Observable gets all its tests ported so a refactor can deal
with all of them.

I've implemented the most basic construction and operator methods on
them.


benjchristensen:
+1

Ben Christensen
310.782.5511  @benjchristensen

akarnokd:
I'm merging this so the independent work on both can be started.

PR #3339, opened 2015-09-11
2.x: fix delaySubscription with supplied publisher: wrong value to trigger subscription

akarnokd:
The subscription was triggered by the supplied publisher and not one of its value or completion.

Btw, should an `onComplete` from this inner publisher be treated as an indication to not subscribe to the actual source?

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -1405,10 +1405,19 @@ public class Observable<T> implements Publisher<T> {
         return timer(delay, unit, scheduler).flatMap(v -> this);
     }
 
+    
+    static final Object OBJECT = new Object();
+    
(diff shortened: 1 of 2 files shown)


benjchristensen:
What is the purpose of this operator? Why is it added?

benjchristensen:
Huh. That exists in 1.x?! Never seen or heard of it.

benjchristensen:
> Btw, should an onComplete from this inner publisher be treated as an indication to not subscribe to the actual source?

That's a good question ... the 1.x docs say it subscribes if the given Observable emits. `onComplete` is not an "emission" in the normal sense. So probably not.

akarnokd:
I'll add this fix and unit test to a larger set of unit tests + fixes coming separately.

PR #3351, opened 2015-09-16
Make BlockingOperatorToIterator exert backpressure.

vqvu:
The iterator created by `BlockingOperator#getIterator()` doesn't exert backpressure, which causes code like this to never terminate/run out of memory

```java
Observable.from(new Iterable<Integer>() {

    @Override
    public Iterator<Integer> iterator() {
        return new Iterator<Integer>() {

            @Override
            public boolean hasNext() {
                return true;
            }

            @Override
            public Integer next() {
                return 1;
            }
        };
    }
}).toBlocking().getIterator().next();
```

This PR adds the appropriate `request` calls so that this works. I had to combine the implementations of `Subscriber` and `Iterator` into a single class to get access to `request()`.

### DIFF
--- a/src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
+++ b/src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
@@ -24,6 +24,7 @@ import rx.Notification;
 import rx.Observable;
 import rx.Subscriber;
 import rx.exceptions.Exceptions;
+import rx.internal.util.RxRingBuffer;
 
 /**
  * Returns an Iterator that iterates over all items emitted by a specified Observable.
@@ -56,17 +57,19 @@ public final class BlockingOperatorToIterator {
     public static final class SubscriberIterator<T>
         extends Subscriber<Notification<? extends T>> implements Iterator<T> {
 
+        static final int LIMIT = 3 * RxRingBuffer.SIZE / 4;
+
         private final BlockingQueue<Notification<? extends T>> notifications;
         private Notification<? extends T> buf;
+        private int received;
 
         public SubscriberIterator() {
             this.notifications = new LinkedBlockingQueue<Notification<? extends T>>();
-            this.buf = null;
         }
 
         @Override
         public void onStart() {
-            request(0);
+            request(RxRingBuffer.SIZE);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
```diff
@@ -47,68 +46,82 @@ private BlockingOperatorToIterator() {
...
-            }
-        });
+        private final BlockingQueue<Notification<? extends T>> notifications;
+        private Notification<? extends T> buf;
 
-        return new Iterator<T>() {
-            private Notification<? extends T> buf;
+        public SubscriberIterator() {
+            this.notifications = new LinkedBlockingQueue<Notification<? extends T>>();
+            this.buf = null;
```
akarnokd: Setting variables to their default value on construction is a no-op semantically but still mean instructions executed which causes the construction of the object takes longer.

src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
```diff
@@ -47,68 +46,82 @@ private BlockingOperatorToIterator() {
...
+
+        @Override
+        public void onNext(Notification<? extends T> args) {
+            notifications.offer(args);
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (buf == null) {
+                request(1);
```
akarnokd: The problem with this approach is that there can't be any cross-thread pipelining because one thread won't progress unless the other thread does 1 thing. A better approach would be to request in bigger batches and re-request after a certain proportion of the batch has been received.

vqvu: This is equivalent to simply calling `request(batchSize)` in `onStart` right? Also, what batch size would you recommend?

src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
vqvu: I'm surprised the compiler doesn't just optimize that out...but I can remove it.

src/main/java/rx/internal/operators/BlockingOperatorToIterator.java
akarnokd: Right. We use `RxRingBuffer.SIZE` by default and the re-request value is 25% of it. [Here](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorMerge.java#L783) is an example for it.

vqvu:
Ok. I removed the unnecessary variable initialization and added batched requests. The initial batch is `RxRingBuffer.SIZE`, and subsequent ones are `3 * RxRingBuffer.SIZE / 4` each (once we've received that much).

akarnokd:
Excellent! :+1:

abersnaze:
:+1:

Issue #3017, opened 2015-06-10
Eager ConcatMap


benjchristensen:
Is there already a name for something in the ecosystem that is like an eager concatMap, or ordered flatMap?

I often see the pattern of people wanting to eagerly kick off several Observables, but merge them together in order. 

It is absolutely not a good idea for infinite lists, but similar to how `concatMap` and `flatMap` can both be dangerous if used wrong, an `eagerConcatMap` that eagerly subscribes and caches the inner `Observable`s could be useful. 

Of course this can be done without adding anything to RxJava, it's just tedious. 

Bad idea? Does it already have a name that I'm unaware of?

akarnokd:
I can't help you with the name and it is non-trivial to change merge() to support it (#2928): fast-path would require unbounded buffers for the later sources; the scalar queue wouldn't really work if interleaved with regular sources and potentially one would need to wrap each value into an indexed container anyway. 

In theory, after getting rid of the optimizations, the [plain merge's](https://github.com/akarnokd/RxJava/blob/OperatorMergeFullRewrite/src/main/java/rx/internal/operators/OperatorMerge.java#L714) round-robin collector can be modified to stick to the first buffer until it is terminated before moving onto the next.

(In my plans for the semi-auto-parallelization, I thought about the join option to be either ordered or unordered merge.)

benjchristensen:
> non-trivial to change merge() to support it 

Merge does not need to be changed. I can implement this today by composition. For example, I iterate over an array of Observables, .cache().subscribe() them, then concat those together.

abersnaze:
One use case for a eager concat was a bug that I ran into a while ago with using concat and publish with a closure.

`source.publish((pub) -> Observable.eagerConcat(pub.map(...), pub.map(...)));`

stealthcode:
This would also be useful where the `onSubscribe` makes an async call. We will need a way to order the responses in the order that the producer generated them from a request. This would of course require buffering.

davidmoten:
I like the idea. Out of interest I think `concat(Observable<T))` is a little bit eager internally at the moment inasmuch as it requests the next observable before the first has completed. It doesn't subscribe to it before the current observable has completed though which is more what you are referring to.

ypresto:
I was misunderstanding concat operator as it subscribes each observables eagerly.

![image](https://cloud.githubusercontent.com/assets/400558/8453595/161c02de-2031-11e5-9db1-36cfa2ec896a.png)

[...]

PR #3357, opened 2015-09-19
Eager ConcatMap

akarnokd:
Related discussion in #3017.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -4547,6 +4547,412 @@ public class Observable<T> {
     }
 
     /**
+     * Concatenates up to 2 sources eagerly into a single stream of values.
+     * 
+     * <p>
+     * Eager concatenation means that once a Subscriber subscribes, all sources are subscribed to, their values fully buffered and
+     * they are drained in order, one after the previous completes.
+     * 
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>Backpressure is honored towards the downstream, however, due to the eagerness requirement, sources are subscribed to
+     *  in unbounded mode and values queued up in an unbounded buffer.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>This method does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <T> the value type 
+     * @param o1 the first source
+     * @param o2 the second source
+     * @return
+     */
+    @Experimental
+    @SuppressWarnings("unchecked")
+    public static <T> Observable<T> concatEager(Observable<? extends T> o1, Observable<? extends T> o2) {
+        return concatEager(Arrays.asList(o1, o2));
+    }
+    
+    /**
+     * Concatenates up to 3 sources eagerly into a single stream of values.
+     * 
+     * <p>
+     * Eager concatenation means that once a Subscriber subscribes, all sources are subscribed to, their values fully buffered and
+     * they are drained in order, one after the previous completes.
+     * 
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>Backpressure is honored towards the downstream, however, due to the eagerness requirement, sources are subscribed to
+     *  in unbounded mode and values queued up in an unbounded buffer.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>This method does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <T> the value type
+     * @param o1 the first source
+     * @param o2 the second source
+     * @param o3 the third source
+     * @return
+     */
+    @Experimental
+    @SuppressWarnings("unchecked")
+    public static <T> Observable<T> concatEager(
+            Observable<? extends T> o1, Observable<? extends T> o2,
+            Observable<? extends T> o3
+        ) {
+        return concatEager(Arrays.asList(o1, o2, o3));
+    }
+    
+    /**
+     * Concatenates up to 4 sources eagerly into a single stream of values.
+     * 
+     * <p>
+     * Eager concatenation means that once a Subscriber subscribes, all sources are subscribed to, their values fully buffered and
+     * they are drained in order, one after the previous completes.
+     * 
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>Backpressure is honored towards the downstream, however, due to the eagerness requirement, sources are subscribed to
+     *  in unbounded mode and values queued up in an unbounded buffer.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>This method does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <T> the value type
+     * @param o1 the first source
+     * @param o2 the second source
+     * @param o3 the third source
+     * @param o4 the fourth source
+     * @return
+     */
+    @Experimental
+    @SuppressWarnings("unchecked")
+    public static <T> Observable<T> concatEager(
+            Observable<? extends T> o1, Observable<? extends T> o2,
+            Observable<? extends T> o3, Observable<? extends T> o4
+        ) {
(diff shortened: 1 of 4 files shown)


davidmoten:
Thanks a lot for contributing this one @akarnokd. I've had one look through and no problems jump out at me but I will review further. This will be useful.

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+            drain();
+        }
+        
+        @Override
+        public void onCompleted() {
+            done = true;
+            drain();
+        }
+        
+        void drain() {
```
davidmoten: For  readability I'd suggest longer names for variables like d, a, r, e, p.

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+            if (wip.getAndIncrement() != 0) {
+                return;
+            }
+            int missed = 1;
+            
+            final AtomicLong p = sharedProducer;
```
davidmoten: p is not being used in the sense of a producer but rather in its sense of the request count. I'd like to see this renamed to `requested` or something similar in keeping with naming in other operators.

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+                        return;
+                    } else
+                    if (empty) {
+                        a.onCompleted();
+                        return;
+                    }
+                }
+
+                if (!empty) {
+                    long r = p.get();
```
davidmoten: `r` could be `req` say and p gets renamed to `requested`

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+                        }
+                    }
+                    
+                    if (innerDone) {
+                        continue;
+                    }
+                }
+                
+                missed = wip.addAndGet(-missed);
+                if (missed == 0) {
```
davidmoten: A unit test that covers both branches of this statement might be nice

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+        return outer;
+    }
+    
+    static final class EagerOuterProducer extends AtomicLong implements Producer {
+        /** */
+        private static final long serialVersionUID = -657299606803478389L;
+        
+        final EagerOuterSubscriber<?, ?> parent;
+        
+        public EagerOuterProducer(EagerOuterSubscriber<?, ?> parent) {
```
davidmoten: remove `public`? (All the constructors in this file for non-public static classes are marked as public).

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+            for (;;) {
+                
+                if (cancelled) {
+                    cleanup();
+                    return;
+                }
+                
+                EagerInnerSubscriber<R> inner;
+                
+                boolean d = done;
```
davidmoten: rename to `outerDone`?

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
```diff
@@ -0,0 +1,299 @@
...
+                if (!empty) {
+                    long r = p.get();
+                    long e = 0L;
+                    boolean unbounded = r == Long.MAX_VALUE;
+                    Queue<R> q = inner.queue;
+                    boolean innerDone = false;
+                    
+                    
+                    for (;;) {
+                        d = inner.done;
```
davidmoten: There's a bit of reuse of the `d` boolean here for a different meaning (`innerDone`). Could we use a new variable name for this?

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
akarnokd: It doesn't really matter and it helps me visually tell apart the fields and the constructor in general.

akarnokd:
Thanks for the review @davidmoten . Usually I use one letter variables because I find it easier to parse the code and instead of long variable names, I use newlines to separate logical blocks. This way, I don't have to type that many letters and I don't have to wait while Eclipse returns with the content assist (which is blocking by the way and may take hundreds of milliseconds, even on my i7 + SSD).

I took the time and renamed variables as you asked for and added some more unit tests that check the code paths. That being said, I don't plan to do such renames in my contributions very often in the future and encourage anybody to post their PRs with their proposed cleanups/renames.

src/main/java/rx/Observable.java
```diff
@@ -4547,6 +4547,398 @@ public final void onNext(T args) {
...
+     *  <dd>This method does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <T> the value type
+     * @param o1
+     * @param o2
+     * @param o3
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> Observable<T> eagerConcatMap(
```
stealthcode: Should we name these `eagerConcat` instead of `eagerConcatMap`? These overloads do not take a `FuncN` so they aren't really mapping.

davidmoten: Ditto that @stealthcode and for discoverability call it concatEager?

akarnokd:
Thanks for the feedback. I've updated the method names to `concatMapEager` and `concatEager`, added the `@Experimental` tags, some javadoc and a missing test for a 2-parameter `concatEager` overload.

src/main/java/rx/internal/operators/OperatorEagerConcatMap.java
akarnokd: Done. See the reentrancy test.

akarnokd:
Updated with a capacity fix for the `SpscLinkedArrayQueue`.

stevegury:
I reviewed in details in the code and I believe it is correct.

But I wonder if subscribing to the sources in a "unbounded mode" is the right thing to do.
At the expense of complexifying the code, I believe it could be possible to split the `request(n)` between the sources. Maybe requesting `1` to all the `m` sources expect the first one, which we request `n - m + 1` ... and so on.

The devil is in the details, and I believe this would significantly complexify the code, but I would like to know what you are thinking about this?

akarnokd:
@stevegury This operator consumes source observables in order and doesn't make sense to split any request from downstream. If the downstream requests n and the first is requested n / count, that would hang the sequence because only the first is allowed to emit and it won't emit enough to trigger a new request.

It is possible to use a bounded buffer per source so while they are not consumed, they don't grow indefinitely. However, it means that each of them can produce only `RxRingBuffer.SIZE` elements before stopping and thus the operator wouldn't be eager anymore.

stevegury:
@akarnokd sorry I wasn't very clear but what I proposed was roughly what you described (replacing `RxRingBuffer.SIZE` by `1`).

My main concern here is the use of unbounded buffer, which, I think, could be avoided.

akarnokd:
@davidmoten What do you think about the bounded buffering and the fact that such sources would act more like a delayed source?

davidmoten:
I'd be happy to see bounded buffering (`RxRingBuffer.SIZE` sounds a better default size than 1 to me and would be consistent with the request eagerness of other operators like `merge` for instance). Could we add an overload to allow the specification of buffer size as well?

If this is significant rework I'm content to see this unbounded version documented with its buffering characteristics merged and we can defer the bounding work to another PR.

akarnokd:
Done.

davidmoten:
Great, thanks @akarnokd ! Lightning quick as always.

stevegury:
LGTM :+1:

abersnaze:
The three of you seem happy with it so I'm merging it.

PR #3359, opened 2015-09-19
2.x: non-backpressure Subject implementations

akarnokd:
They can be renamed later as necessary. No unit tests but the regular
Subject tests can be ported easily.


akarnokd:

PR #3360, opened 2015-09-19
2.x: non-backpressure Subscribers

akarnokd:
These classes enable the porting of backpressure-aware operators for non-backpressure Observable.

It would be great if this could be merged sooner.

### DIFF
--- a/src/main/java/io/reactivex/NbpObserver.java
+++ b/src/main/java/io/reactivex/NbpObserver.java
@@ -11,27 +11,31 @@
  * the License for the specific language governing permissions and limitations under the License.
  */
 
-package io.reactivex.internal.disposables;
+package io.reactivex;
 
 import io.reactivex.NbpObservable.NbpSubscriber;
 import io.reactivex.disposables.Disposable;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
 
-public enum EmptyDisposable implements Disposable {
-    INSTANCE
-    ;
-    
+public abstract class NbpObserver<T> implements NbpSubscriber<T> {
+    private Disposable s;
     @Override
-    public void dispose() {
-        // no-op
+    public final void onSubscribe(Disposable s) {
+        if (SubscriptionHelper.validateDisposable(this.s, s)) {
+            return;
+        }
+        this.s = s;
+        onStart();
     }
     
-    public static void complete(NbpSubscriber<?> s) {
-        s.onSubscribe(INSTANCE);
-        s.onComplete();
+    protected final void cancel() {
+        s.dispose();
+    }
+    /**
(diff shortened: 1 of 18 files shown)


src/main/java/io/reactivex/NbpObserver.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.reactivex;
+
+import io.reactivex.NbpObservable.NbpSubscriber;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
+
+public abstract class NbpObserver<T> implements NbpSubscriber<T> {
+    private Disposable s;
```
artem-zinnatullin: Shouldn't this be a `volatile`? I guess `cancel()` may be called from another Thread.

// Java is not C, maybe we can use longer names like `disposable`? :)

xpierro: Agreed.

src/main/java/io/reactivex/internal/subscribers/nbp/NbpQueueDrainSubscriber.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+}
+
+// -------------------------------------------------------------------
+// Padding superclasses
+//-------------------------------------------------------------------
+
+/** Pads the header away from other fields. */
+class QueueDrainSubscriberPad0 {
```
artem-zinnatullin: @akarnokd this looks like premature optimization, or you did performance tests?

src/main/java/io/reactivex/NbpObserver.java
akarnokd: No. onSubscribe must be called sequentially, similar to the full-blown RS Subscriber. There is an NbpAsyncObserver where this is indeed volatile because the cancellation is exposed and may happen before a Disposable is set.

If merged, you can run around and rename stuff more easily because of tools. When I type in such classes, it is such a burden to type out everything.

src/main/java/io/reactivex/internal/subscribers/nbp/NbpQueueDrainSubscriber.java
akarnokd: This is due to lessons learned with JCTools and the effect of false sharing.

artem-zinnatullin: ok!

src/main/java/io/reactivex/NbpObserver.java
artem-zinnatullin: ok, np :)

akarnokd:

PR #3361, opened 2015-09-20
Safer error handling in BlockingOperatorToFuture

artem-zinnatullin:
Not very important since `onError` should be called once, but anyway it's `AtomicReference` and we need to store the value returned from `get()` locally.

P.S. I don't use `Future`s, just found this piece of code.


akarnokd:
It is unlikely error changed between the two calls in this operator but regardless, one should read volatile variables once. :+1:

abersnaze:
:+1:

PR #3362, opened 2015-09-20
Fix synchronization on non-final field in BufferUntilSubscriber

artem-zinnatullin:
Compiler, VM, JIT and IDE will be happier :)


akarnokd:
:+1:

zsxwing:
:+1: I was thinking to add `final` but forgot it.

akarnokd:
Great, thanks!

PR #3364, opened 2015-09-20
Possible multithreading bug fix in OnSubscribeAmb

artem-zinnatullin:
Looks like non-volatile `chosen` may cause concurrency problems in this operator, @akarnokd PTAL


davidmoten:
`volatile` keyword not required because of the [Observable contract](http://reactivex.io/documentation/contract.html):

>Observables must issue notifications to observers serially (not in parallel). They may issue these notifications from different threads, but there must be a formal happens-before relationship between the notifications.

The guarantee of a happens-before relationship between emissions means you won't get into synchronization problems with this non-final field.

akarnokd:
No need for volatile because `chosen` is sequence-local and is changed to true after reading a volatile in isSelected.

:-1:

artem-zinnatullin:
ok, agreed, closing :)

PR #3366, opened 2015-09-20
2.x: Added placeholder operators to NbpObservable + special observables

akarnokd:
Nbp prefixed Observable types added. The core operators throw UnsupportedOperationException for now, overloads delegating to them did not change.


akarnokd:

PR #3369, opened 2015-09-21
Lint fixes for unnecessary unboxing

KevinTCoughlin:
Believe I'm removing some unnecessary unboxing calls, but please correct me if I'm wrong.

### DIFF
--- a/src/main/java/rx/schedulers/TestScheduler.java
+++ b/src/main/java/rx/schedulers/TestScheduler.java
@@ -57,9 +57,9 @@ public class TestScheduler extends Scheduler {
         @Override
         public int compare(TimedAction action1, TimedAction action2) {
             if (action1.time == action2.time) {
-                return Long.valueOf(action1.count).compareTo(action2.count);
+                return action1.count < action2.count ? -1 : ((action1.count > action2.count) ? 1 : 0);
             } else {
-                return Long.valueOf(action1.time).compareTo(action2.time);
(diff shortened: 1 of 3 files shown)


src/main/java/rx/schedulers/TestScheduler.java
```diff
@@ -57,9 +57,9 @@ public String toString() {
         @Override
         public int compare(TimedAction action1, TimedAction action2) {
             if (action1.time == action2.time) {
-                return Long.valueOf(action1.count).compareTo(Long.valueOf(action2.count));
+                return Long.valueOf(action1.count).compareTo(action2.count);
```
akarnokd: Could you turn these into `a < b ? -1 : ((a > b) ? 1 : 0)` instead?

akarnokd:
See the comment, otherwise :+1:.

stevegury:
+1 for @akarnokd comment, otherwise :+1:

src/main/java/rx/schedulers/TestScheduler.java
akarnokd: To avoid misunderstanding, don't use `Long.compare()` because we need to be Java 6 compatible and that method is not available in it.

KevinTCoughlin:
@akarnokd @stevegury used ternary in-place of `Long.compareTo` for Java 6 support. Thanks!

stevegury:
:+1:

akarnokd:
:+1:

PR #3371, opened 2015-09-22
2.x: non-backpressure NbpObservable all relevant operators + tests.

akarnokd:
+ non-backpressured subjects.


benjchristensen:
Thanks. Will merge and review as part of the branch over time.

akarnokd:
I have also code for `Single` in #3358. I've rebased it and fixed up NbpObservable with the proper method name (`onError` instead of `onFailure`).

PR #3372, opened 2015-09-24
Observable.empty().collect() should not throw an exception

davidmoten:
`Observable.empty().collect(stateFactory, accumulator)` should return the result of `stateFactory.call()` but throws an exception.

I've added a unit test for this and removed an irrelevant reference to a discussion on `reduce`.


davidmoten:
ah woops, did not throw, closing.

PR #3374, opened 2015-09-25
Optimize Exceptions.throwIfFatal()

artem-zinnatullin:
`Exceptions.throwIfFatal()` used in almost all RxJava operators.

Basically, in this PR I've just divided `throwable` into 3 groups: `RuntimeException`, `Error` and `Exception` (we don't need to check them).

With this PR `Exceptions.throwIfFatal()` becomes ~7% faster than current implementation in case of non-fatal exceptions (because if it's fatal — app will be crashed :smiley_cat:) on the Oracle JRE 1.8.0_45.

I guess, on Android it will be even more faster! (I am not sure in quality of our JITs/VMs).

I've checked the bytecode after compilation in hope that javac will optimize `Exceptions.throwIfFatal()` into same construction, but it does not perform any optimizations over this method at compile time, so the only hope is JIT and even after JIT, optimized version slightly faster (~7%).

Results of the benchmarks:

```
OPTIMIZED VERSION:

Benchmark                                                               Mode   Samples        Score  Score error    Units
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalCheckedException     avgt         5        2.569        0.389    ns/op
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalError                avgt         5        2.332        0.030    ns/op
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalRuntimeException     avgt         5        2.511        0.037    ns/op

Benchmark                                                               Mode   Samples        Score  Score error    Units
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalCheckedException    thrpt         5 395071210.861  1644491.739    ops/s
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalError               thrpt         5 429656328.864  3230560.292    ops/s
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalRuntimeException    thrpt         5 398472020.961  3961447.943    ops/s

OLD VERSION:

Benchmark                                                               Mode   Samples        Score  Score error    Units
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalCheckedException     avgt         5        2.812        0.954    ns/op
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalError                avgt         5        2.338        0.019    ns/op
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalRuntimeException     avgt         5        2.771        0.064    ns/op

Benchmark                                                               Mode   Samples        Score  Score error    Units
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalCheckedException    thrpt         5 370211340.236  9093081.248    ops/s
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalError               thrpt         5 427943417.424  6168313.153    ops/s
r.e.ExceptionsThrowIfFatalPerf.throwIfFatalNonFatalRuntimeException    thrpt         5 368255154.316  4520997.005    ops/s
```

### DIFF
--- a/src/main/java/rx/exceptions/Exceptions.java
+++ b/src/main/java/rx/exceptions/Exceptions.java
@@ -73,25 +73,28 @@ public final class Exceptions {
      * @see <a href="https://github.com/ReactiveX/RxJava/issues/748#issuecomment-32471495">RxJava: StackOverflowError is swallowed (Issue #748)</a>
      */
     public static void throwIfFatal(Throwable t) {
-        if (t instanceof OnErrorNotImplementedException) {
-            throw (OnErrorNotImplementedException) t;
-        } else if (t instanceof OnErrorFailedException) {
-            Throwable cause = t.getCause();
-            if (cause instanceof RuntimeException) {
-                throw (RuntimeException) cause;
-            } else {
-                throw (OnErrorFailedException) t;
+        if (t instanceof RuntimeException) {
+            if (t instanceof OnErrorNotImplementedException) {
+                throw (OnErrorNotImplementedException) t;
+            } else if (t instanceof OnErrorFailedException) {
+                Throwable cause = t.getCause();
+                if (cause instanceof RuntimeException) {
+                    throw (RuntimeException) cause;
+                } else {
+                    throw (OnErrorFailedException) t;
+                }
             }
-        }
-        // values here derived from https://github.com/ReactiveX/RxJava/issues/748#issuecomment-32471495
-        else if (t instanceof StackOverflowError) {
-            throw (StackOverflowError) t;
-        } else if (t instanceof VirtualMachineError) {
-            throw (VirtualMachineError) t;
-        } else if (t instanceof ThreadDeath) {
-            throw (ThreadDeath) t;
-        } else if (t instanceof LinkageError) {
-            throw (LinkageError) t;
+        } else if (t instanceof Error) {
+            // values here derived from https://github.com/ReactiveX/RxJava/issues/748#issuecomment-32471495
+            if (t instanceof StackOverflowError) {
+                throw (StackOverflowError) t;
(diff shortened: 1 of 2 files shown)


davidmoten:
The path for `OnErrorNotImplementedException` or `OnErrorFailedException` will be slower because they are checked for being a `RuntimeException` first. If it's a `RuntimeException` and not one of those then it will be faster because checks against `Error` types will not occur. I think this contribution is probably ok but some estimate of the most commonly used paths through this code should probably be made. In my opinion the path taken by an `Error` will not in general be taken anywhere near as frequently as the path by a `RuntimeException` so the optimization should favour the `RuntimeException` path. The question remains then what `RuntimeException`s are most likely to pass through? I'm assuming `Exception`s won't be travelling by this path very often because I can't think where in our operators we might be throwing checked exceptions (someone can enlighten me of course).

akarnokd:
Your benchmark results are within the error margin to each other so I'd not consider this as a clear-cut improvement. Throwing exceptions is so expensive that a simple `instanceof` check is overshadowed. And besides, optimizing for the exception path is questionable because you are not delivering the exceptions as part of the main business logic.

stevegury:
I agree with @akarnokd here, the gain is not clear enough to justify the logic rewrite.

artem-zinnatullin:
Okay, I agree with all points.

Checked some amount of libraries with RxJava support. Only small amount of them use `Exceptions.throwIfFatal()` in their implementations of `OnSubscribe`, so the case of passing checked `Exception` into the `throwIfFatal()` is pretty rare.

Also, as said in comments, exceptions are not / should not frequent cases of the application lifecycle, so the performance of this method does not really matter.

Thanks for the comments!

PR #3376, opened 2015-09-25
2.x: perf comparing Observable, NbpObservable, Stream and ParallelStream

akarnokd:
This benchmark compares the overhead of backpressured range, non-backpressured range, streamed range and parallel-streamed range. Run on i7 4770K, Windows 7 x64, Java 1.8u60.

image


stevegury:
:+1:

PR #3380, opened 2015-09-27
2.x: CombineLatest removed leftover debug field + better cleanup

akarnokd:
A terminal condition is now clearing the latest array which should help release its contents earlier in case a reference is kept to the class.


abersnaze:
LGTM

PR #3381, opened 2015-09-28
2.x: UnicastSubject fix for the child becoming visible before onSubscribe is called.

akarnokd:
Discovered while the test `NbpOperatorConcatTest.testIssue2890NoStackoverflow` hung on Travis.

What happened is that the reference to the child became visible before the call to its `onSubscribe` method so a concurrent source emitting at the exact same time could already see the "unstarted" child. The PR fixes this in both `NbpUnicastSubject` and `UnicastSubject` by changing the order of calls. The rest of the subjects behave correctly (call `onSubscribe` first, make child visible second).

In addition, `UnicastSubject` now calls `drain()` because when the `onSubscribe` calls `request` and `drain`, the child is not visible and nothing gets replayed. Once both `onSubscribe` call returns and the child is becomes visible, a call to `drain` again will now correctly replay all available contents that were requested.


akarnokd:

PR #3382, opened 2015-09-28
2.x: fix SpscLinkedArrayQueue leaves 1 slot null just before growing

akarnokd:
Discovered while debugging #3381. Confirmed with JCTools in https://github.com/JCTools/JCTools/pull/80


stealthcode:
:+1:

PR #3384, opened 2015-09-29
Fix for take() reentrancy bug.

akarnokd:
Fixes the bug reported in  #3346. (I've done this for 1.x as well since the original poster disappeared).


stevegury:
:+1:

PR #3388, opened 2015-10-01
1.1.0: Subject state methods beta/remove

akarnokd:
Based on votes, this PR removes the common Subject state peeking methods from Subject and removes irrelevant methods from each Subject variant and promotes the rest to `@Beta` state.

Due to the removals, I had to rearrange tests as well.


abersnaze:
For the methods removed can you open a PR that marks them as deprecated? We can merge immediately for 1.0.15. It'll give anyone using them one release to migrate off of them.

akarnokd:
Done: #3407.

Note, however, that it is very likely this PR will need rebasing when #3407 is merged.

akarnokd:
I've rebased this onto the master but the versions are so tangled I don't know how to squash this into a single commit.

artem-zinnatullin:
If you've merged something into this branch — easiest way to squash this without dealing with possible conflicts is to do:

1. `git diff > patch.diff` between this branch and target branch
2. Then drop local branch (don't forget to make a copy via `git checkout -b`) 
3. Create branch with same name from current state of target branch
4. Use `git apply patch.diff` and commit 
5. Use push -f to rewrite branch on GitHub. 

(though squash may pass normally here, I am on mobile and can not check now). Also there is a way to have local branch with different name than on a remote, but I forgot it, with it you can skip steps 2 and 3.

akarnokd:
I'm using Eclipse which has a few git bugs in it right now. I'd better leave this to @abersnaze as he seems to be proficient in squashing PRs.

stealthcode:
I've rebased this in a branch https://github.com/stealthcode/RxJava/tree/SubjectPeekRemoves

@abersnaze and I would like to collapse down all of the 1.1.0 pull requests into a single commit and a single pull request. Any objections? I'll start on the rebasing.

akarnokd:
@stealthcode Great. :+1:

PR #3389, opened 2015-10-01
1.1.0: removal of AbstractOnSubscribe

akarnokd:
Based on votes, the `AbstractOnSubscribe` is to be removed.


Crystark:
Hi, do you have a link to the vote / discussion about this ? I actually didn't see it was experimental and have some code based on it i'd like to adapt.

stealthcode:
The `SyncOnSubscribe` replaced the `AbstractOnSubscribe`. Regarding it's usage, it is similar to it's predecessor with slight differences. There are overloads for creating a stateless, a mutable state, and an immutable/functional style onSubscribe (depending on your use case). Please let us know if the existing javadocs aren't sufficient or could be improved.

PR #3393, opened 2015-10-01
1.1.0: promote TestSubscriber methods to public

akarnokd:
Based on votes, the experimental TestSubscriber methods can now be part
of the public API.

In addition, I've added the missing javadocs for the create() factory
methods and one of the constructors.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -20,7 +20,6 @@ import java.util.concurrent.*;
 
 import rx.*;
 import rx.Observer;
-import rx.annotations.Experimental;
 import rx.exceptions.CompositeException;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/rx/observers/TestSubscriber.java
```diff
@@ -83,39 +81,85 @@ public TestSubscriber(Observer<T> delegate, long initialRequest) {
...
+     * @throws NullPointerException if delegate is null
+     */
     public TestSubscriber(Observer<T> delegate) {
         this(delegate, -1);
     }
 
+    /**
+     * Constructs a TestSubscriber with an initial request of Long.MAX_VALUE and no delegation.
+     */
     public TestSubscriber() {
```
stealthcode: Should this have a `@since` annotation?

stealthcode: Or for that matter, should all of the public methods in this class (which are not overriding implementations) have `@since 1.1.0` ?

akarnokd: This constructor was always part of the class but it missed the Javadoc.

PR #3401, opened 2015-10-01
1.1.0: promote Single + SingleSubscriber to Beta

akarnokd:
Based on votes, Single, SingleSubscriber and Observable.toSingle can now
be promoted to Beta state.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -231,7 +231,7 @@ public class Observable<T> {
      * @see <a href="http://reactivex.io/documentation/single.html">ReactiveX documentation: Single</a>
      * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
      */
-    @Experimental
+    @Beta
     public Single<T> toSingle() {
         return new Single<T>(OnSubscribeSingle.create(this));
     }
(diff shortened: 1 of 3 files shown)


charliek:
I've been a really big fan of Single in our codebase. It helps document the API, reason about our code base, and helps us eliminate cases where people don't properly plan for multiple items when terminating streams.

To date we've not hit any corner cases that I am aware of, but we've not been using it too heavily yet.

+1

beckje01:
I love Single it makes the API very understandable.

dsvoronin:
will be so natural in Retrofit 2.0
but also starting to use it in our codebase as well

passsy:
I'm using it since the first public `@Experimental` release and it makes Interfaces more understandable. My current `Observable`/`Single` ratio is 30/70

I hope all operators known from Observable will promoted to public for Single such as `defer` and `lift`. Doing the following workaround seems unnecessary:

```java
public Single doSomething() {
    return Single.just(..)
        .map(..)
        .toObservable() // to Observable because defer is private for Single
        .defer(...)
        .toSingle(); // back to Single because API says so
}
```

artem-zinnatullin:
@passsy I am planning to use `Single` in production for the same reason of clearer contract of the API (though I [see some cons in this too](https://www.reddit.com/r/androiddev/comments/3nws5l/do_you_use_rxsingle/)), so I'll try to bring most useful operators (`defer` in the list) to the `Single` during this week, and I hope Rx team will merge and release them as part of 1.0.15.

PR #3402, opened 2015-10-01
1.1.0: Promote 2 RxJavaErrorHandler methods to Beta

akarnokd:
Based on votes, RxJavaErrorHandler.render and
RxJavaErrorHandler.handleOnNextValueRendering can now be promoted to
Beta state.


akarnokd:
The test in `BackpressureTests.testMergeAsync` failed. The test is non-deterministic because a small hiccup in one of the threads may produce way more values than the assertion expects.

PR #3405, opened 2015-10-01
1.1.0: promote doOnRequest to be part of the public API

akarnokd:
Based on the majority of votes, doOnRequest can now be part of the
public API.


stealthcode:
This issue is also being discussed in #3386.

PR #3407, opened 2015-10-01
1.0.15. Beta/Deprecation of Subject state peeking methods.

akarnokd:
This should give users one release to prepare for the class structure
changes.


abersnaze:
replaced by #3415

PR #3408, opened 2015-10-02
DoOnEach: report both original exception and callback exception.

akarnokd:
This came up in a [Stackoverflow](http://stackoverflow.com/questions/32889008/do-operators-instead-of-a-whole-subscriber) answer. If the `doOnError`'s callback or the `doOnEach`'s `onError` method throws, any non-fatal exception replaced the original error which got lost. This PR will wrap them both into a `CompositeException`.

2.x note: since Java 8 supports `addSuppressed` all callbacks in this situation either attach to the original exception or the original exception is attached to the callback's exception.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -4362,6 +4362,10 @@ public class Observable<T> {
     /**
      * Modifies the source Observable so that it notifies an Observer for each item it emits.
      * <p>
+     * In case the onError of the supplied observer throws, the downstream will receive a composite exception containing
+     * the original exception and the exception thrown by onError. If the onNext or the onCompleted methods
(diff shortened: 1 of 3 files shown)


abersnaze:
:+1:

frvi:
Is there a way to avoid this wrapping?
I want to throw a specific exception in my `doOnError`, which then gets handled in a certain way.

akarnokd:
No, but you can unwrap via onErrorResumeNext:

```java
source.doOnError(e -> { throw new RuntimeException(); })
.onErrorResumeNext(e -> {
    if (e instanceof CompositeException) {
       return Observable.error(((CompositeException)e).getExceptions().get(1));
    }
    return Observable.error(e);
});
```

frvi:
@akarnokd -- ah, thanks!

PR #3413, opened 2015-10-06
2.x: fix to ReplaySubject not properly accounting for emitted events.

akarnokd:
The subject tries to avoid the request reduction via addAndGet by re-reading the requested amount and continuing the loop as long as possible. The bug was in ignoring the total emission amount and thus a single request(1) from the child ended up replaying all values.


akarnokd:

PR #3417, opened 2015-10-08
Add Single.doOnSuccess()

### LINKED ISSUE #3385: rx.Single - why no `doOnNext()`?
I have a `Single<?>` that I would like to call `.doOnNext()` on. Why is this omitted? I get why `.doOnCompleted()` does not exist, but I want to call a void action once and there seems to be no simple way to do this with `rx.Single`

Technically I could call `.map()`, do my thing, and return the same object - but that seems like too much overhead. Or call `.toObservable().doOnNext().toSingle()`, but I cannot imagine that is too efficient.

artem-zinnatullin:
Closes #3385.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1864,4 +1864,38 @@ public class Single<T> {
 
         return lift(new OperatorDoOnEach<T>(observer));
     }
+    
+    /**
+     * Modifies the source {@link Single} so that it invokes an action when it calls {@code onSuccess}.
+     * <p>
+     * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/doOnNext.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code doOnSuccess} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param onSuccess
+     *            the action to invoke when the source {@link Single} calls {@code onSuccess}
+     * @return the source {@link Single} with the side-effecting behavior applied
+     * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX operators documentation: Do</a>
+     */
+    @Experimental
+    public final Single<T> doOnSuccess(final Action1<? super T> onSuccess) {
+        Observer<T> observer = new Observer<T>() {
+            @Override
+            public void onCompleted() {
+            }
+
+            @Override
+            public void onError(Throwable e) {
+            }
+
+            @Override
+            public void onNext(T t) {
+                onSuccess.call(t);
+            }
+        };
+
+        return lift(new OperatorDoOnEach<T>(observer));
+    }
 }
--- a/src/test/java/rx/SingleTest.java
+++ b/src/test/java/rx/SingleTest.java
@@ -16,6 +16,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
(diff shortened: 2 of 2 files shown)


src/test/java/rx/SingleTest.java
```diff
@@ -461,4 +471,60 @@ public void testToObservable() {
...
+        Single
+                .error(error)
+                .doOnSuccess(action)
+                .subscribe(testSubscriber);
+
+        testSubscriber.assertNoValues();
+        testSubscriber.assertError(error);
+
+        verifyZeroInteractions(action);
+    }
```
akarnokd: Could you also test what happens if the `action` callback throws?

src/main/java/rx/Single.java
```diff
@@ -1789,4 +1790,36 @@ public void onNext(T t) {
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code doOnSuccess} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param onSuccess
+     *            the action to invoke when the source {@link Single} calls {@code onSuccess}
+     * @return the source {@link Single} with the side-effecting behavior applied
+     * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX operators documentation: Do</a>
+     */
+    public final Single<T> doOnSuccess(final Action1<? super T> onSuccess) {
```

artem-zinnatullin: I didn't add `@Experimental` here and in the other PRs because whole class  `Single` is marked with `@Experimental`. Do you think we should mark these basic operators too?

akarnokd: Yep, this is a bit of a conflict because #3401 gets merged first, this would also become beta without any votes.

artem-zinnatullin: Okay, no problem. I'll mark it as @Experimental.

чт, 8 Окт 2015, 12:44, David Karnok <notifications@github.com>:

> In src/main/java/rx/Single.java
> <https://github.com/ReactiveX/RxJava/pull/3417#discussion_r41494676>:
>
> > @@ -1789,4 +1790,36 @@ public void onNext(T t) {
> >          return zip(this, other, zipFunction);
> >      }
> >
> > +    /**
> > +     * Modifies the source {@link Single} so that it invokes an action when it calls {@code onSuccess}.
> > +     * <p>
> > +     * 
> > +     * <dl>
> > +     *  <dt><b>Scheduler:</b></dt>
> > +     *  <dd>{@code doOnSuccess} does not operate by default on a particular {@link Scheduler}.</dd>
> > +     * </dl>
> > +     *
> > +     * @param onSuccess
> > +     *            the action to invoke when the source {@link Single} calls {@code onSuccess}
> > +     * @return the source {@link Single} with the side-effecting behavior applied
> > +     * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX operators documentation: Do</a>
> > +     */
> > +    public final Single<T> doOnSuccess(final Action1<? super T> onSuccess) {
>
> Yep, this is a bit of a conflict because #3401
> <https://github.com/ReactiveX/RxJava/pull/3401> gets merged first, this
> would also become beta without any votes.
>
> —
> <https://github.com/ReactiveX/RxJava/pull/3417/files#r41494676>.
>
-- 

@artem_zin

src/test/java/rx/SingleTest.java
artem-zinnatullin: See `doOnSuccessShouldNotSwallowExceptionThrownByAction()`

src/main/java/rx/Single.java

akarnokd:
:+1:

abersnaze:
@artem-zinnatullin rebase please.

artem-zinnatullin:
@abersnaze rebased!

PR #3418, opened 2015-10-08
Add Single.fromCallable()

artem-zinnatullin:
Second part of #3154.

@akarnokd I'd like to reuse `Observable.fromCallable()` but didn't find apropriate way to do it: `Observable.fromCallable().toSingle()` is not best option, if you have some thoughts on this — please share them.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -12,6 +12,7 @@
  */
 package rx;
 
+import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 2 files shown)


akarnokd:
:+1: I'm not too keen on reusing but the most complicated operators for `Single`.

src/main/java/rx/Single.java
```diff
@@ -605,6 +606,42 @@ public void call(SingleSubscriber<? super T> te) {
...
+     *   <dd>{@code fromCallable} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param func
+     *         function which execution should be deferred, it will be invoked when Observer will subscribe to the {@link Single}.
+     * @param <T>
+     *         the type of the item emitted by the {@link Single}.
+     * @return a {@link Single} whose {@link Observer}s' subscriptions trigger an invocation of the given function.
+     */
+    public static <T> Single<T> fromCallable(final Callable<? extends T> func) {
```

abersnaze:
@artem-zinnatullin rebase please.

artem-zinnatullin:
@abersnaze rebased, I'll rebase one PR at a time, since they're all conflict with each other.

PR #3419, opened 2015-10-08
Add Single.doOnError()

artem-zinnatullin:


src/main/java/rx/Single.java
```diff
@@ -1789,4 +1790,39 @@ public void onNext(T t) {
         return zip(this, other, zipFunction);
     }
 
+    /**
+     * Modifies the source {@link Single} so that it invokes an action if it calls {@code onError}.
+     * <p>
+     * In case the onError action throws, the downstream will receive a composite exception containing
```
akarnokd: Could you add an unit test that verifies this?

src/main/java/rx/Single.java
```diff
@@ -1789,4 +1790,39 @@ public void onNext(T t) {
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code doOnError} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param onError
+     *            the action to invoke if the source {@link Single} calls {@code onError}
+     * @return the source {@link Single} with the side-effecting behavior applied
+     * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX operators documentation: Do</a>
+     */
+    public final Single<T> doOnError(final Action1<Throwable> onError) {
```

src/main/java/rx/Single.java
artem-zinnatullin: Added `doOnErrorShouldThrowCompositeExceptionIfOnErrorActionThrows()`

src/main/java/rx/Single.java

akarnokd:
:+1:

abersnaze:
:+1:

PR #3423, opened 2015-10-08
Renaming Observable#x to Observable#extend

stealthcode:


abersnaze:
:+1:

akarnokd:
Correct me if I'm wrong but the merge policy requires two non-author likes, right?

abersnaze:
I thought it was two committers and if the auth was a committer that opening the PR was an implicit vote.  I'll post some by laws so we can all agree on what the rules are.

stealthcode:
That was my understanding as well. I thought it was the releases that required more votes.

akarnokd:
Oh, I see. Checking my PRs again...

PR #3424, opened 2015-10-08
Added Runnable interface to BlockingObservable

stealthcode:


akarnokd:
Why did you add that interface? The `run()` method was never intended to represent a Runnable BlockingObservable.

abersnaze:
But I thought you added `run()` for discoverability. I felt that this PR was just the logical conclusion to adding that method. Can we pick between both or neither before we release 1.0.15?

akarnokd:
If you have a `Runnable` class, that implies to the developer it has to be submitted to an `ExecutorService`. This isn't the case with `BlockingObservable` because it is meant to block in the current thread, mostly because of the side effects. Submitting it to an `Executor` and awaiting the future is pointless. You want the `BlockingObservable` *to run to completion* so you'll find the method `run()`.

abersnaze:
I believe having the `run()` method implies that `BlockingObservable` should be submittable to an `ExecutorService`.  The implementing of `Runnable` only makes it possible do to so in Java.
[If it looks like a duck, swims like a duck, and quacks like a duck, then it probably is a duck.](https://en.wikipedia.org/wiki/Duck_test)

akarnokd:
In that case, the run() method should be removed.

stealthcode:
I could see that users may wish to submit a blocking observable to an ExecutorService simply because the interface they have available takes a Runnable. I do not think it's pointless. But I would prefer for users to deliberately create a Runnable wrapper for executing subscribe and we delete the erroneous run method.

akarnokd:
Both keeping it and removing it is fine with me. If this PR is to stay, I'd suggest adding `@Override` to runnable as well.

stealthcode:
I'm closing this PR. For 1.0.15 I'd rather delete it and avoid making a decision about how Runnable would interact with `BlockingObservable`

PR #3428, opened 2015-10-09
GroupBy backpressure fix

akarnokd:
This is a backport of the 2.x GroupBy operator which solves #3425.

One unit test in OperatorRetryTest had to be altered a bit. I believe
the original code relied on a GroupBy behavior which caused the bug in
#3425.

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorGroupByTest.java
+++ b/src/test/java/rx/internal/operators/OperatorGroupByTest.java
@@ -15,45 +15,24 @@
  */
 package rx.internal.operators;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyInt;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
+import static org.junit.Assert.*;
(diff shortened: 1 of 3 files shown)


stealthcode:
The following test starts breaking with this branch. It prints out 1 to 128 then stops and does not print anything else (apparently no progress is made). In branch 1.x this continues to print past 128. 

```java
range(1, RxRingBuffer.SIZE *4).repeat()
    .groupBy((d) -> d)
    .flatMap((go) ->
        go.doOnNext(System.out::println)
            .subscribeOn(Schedulers.io())
    )
    .subscribe();
```

akarnokd:
This is because the lack of horizontal unboundedness of flatMap and the bug in groupBy together made it work in 1.0.14. If I run this code with the 2.x fix and add 1024 as the max concurrency value, it prints lots of values indefinitely.

src/test/java/rx/internal/operators/OperatorGroupByTest.java
```diff
@@ -1501,4 +1480,103 @@ public void onNext(Integer t) {
...
+                    public Object call(Integer v) {
+                        return v;
+                    }
+                })
+                .doOnNext(new Action1<GroupedObservable<Object, Integer>>() {
+                    @Override
+                    public void call(GroupedObservable<Object, Integer> g) {
+                        g.subscribe();
+                    }
+                }) // this will request Long.MAX_VALUE
```
stealthcode: Is this comment meant for the inner observable? The outer observable gets a request of 128.

akarnokd:
I've updated the comments.

stealthcode:
:+1:

PR #3433, opened 2015-10-10
Add Single.defer()

artem-zinnatullin:


src/main/java/rx/Single.java
```diff
@@ -1898,4 +1898,48 @@ public void onNext(T t) {
...
+
+                try {
+                    single = singleFactory.call();
+                } catch (Throwable t) {
+                    Exceptions.throwIfFatal(t);
+                    singleSubscriber.onError(t);
+                    return;
+                }
+
+                single.subscribe(singleSubscriber);
```
artem-zinnatullin: Please notice that I'm not wrapping subscriber via `Subscribers.wrap()` because `SingleSubscriber` differs from `Subscriber`, that's why I used `subscribe` instead of `unsafeSubscribe()` here.

If this should be changed, please comment.

akarnokd:
Please add unit test of normal, callback throws and callback returns null cases.

artem-zinnatullin:
@akarnokd added `deferShouldPassNullPointerExceptionToTheSubscriberIfSingleFactoryIsNull` and `deferShouldPassNullPointerExceptionToTheSubscriberIfSingleFactoryReturnsNull`.

Btw, I've used `Callable` instead of `Func0` to save users from dealing with checked exception, hope it's okay :)

akarnokd:
Func0 extends Callable so its okay.

.:+1:

abersnaze:
:+1:

src/main/java/rx/Single.java
akarnokd: `SingleSubscriber` doesn't have `onStart` like `Subscriber` so this is not a problem.

akarnokd:
I'm still :+1: but please rebase.

artem-zinnatullin:
Rebased!

akarnokd:
Thanks! Merged.

passsy:
thanks

Issue #3435, opened 2015-10-10
OperatorFinally calls action twice if action throws exception. Exception thrown by action is swallowed


artem-zinnatullin:
While I was implementing #3434 I've found two problems with `OperatorFinally`.

I'll call `OperatorFinally.action` as `finallyAction` for better readability.

1. If `finallyAction` is `null` — `NullPointerException` will be swallowed by `SafeSubscriber`, this can be solved via `action != null` check in the `OperatorFinally` (I'll make PR).
2. If `finallyAction` throws exception, `lift` calls `onError()` and `OperatorFinally` invokes `finallyAction` again (this may brake user-defined logic)! And second exception is swallowed by the `SafeSubscriber` (see problem 1).

I vote for solving both of these problems before 1.0.15/1.1.10.

artem-zinnatullin:
Here are the tests:

```java
@Test
public void nullFinallyActionShouldBeCheckedASAP() {
    try {
        Observable
                .just("value")
                .finallyDo(null);

        fail();
    } catch (NullPointerException expected) {

    }
}

@Test
public void ifFinallyActionThrowsExceptionShouldNotBeSwallowedAndActionShouldBeCalledOnce() {
    Action0 finallyAction = mock(Action0.class);
    doThrow(new IllegalStateException()).when(finallyAction).call();

    TestSubscriber<String> testSubscriber = new TestSubscriber<String>();

    Observable
            .just("value")
            .finallyDo(finallyAction)
            .subscribe(testSubscriber);

    testSubscriber.assertValue("value");

    verify(finallyAction).call();
    // Actual result:
    // Not only IllegalStateException was swallowed
    // But finallyAction was called twice!
}
```

artem-zinnatullin:
I'll try to send PRs with fixes ASAP, `null` check is easy, but second problem looks tricky.

akarnokd:
If the action throws, you can't do much but call `RxJavaPlugins.getInstance().getErrorHandler().handleError()` for it.

artem-zinnatullin:
We can throw it as `Error`. Leaving the app in an unknown state is very bad idea, especially in something like `finallyDo()`.

akarnokd:
The sequence has already terminated at that point and this operator is there to have side-effects after it. The exception has nowhere to go and there is no guarantee a thrown error reaches its intended target because of a potential async boundary. So it either propagates up to a synchronous caller, gets thrown away by the `ExecutorService` or gets sent to the error handler.

[...]

PR #3436, opened 2015-10-10
1.x: Add action != null check in OperatorFinally

artem-zinnatullin:
Part of #3435.

Personally, I'd also add same test to `Observable.finallyDo()` and `Single.finallyDo()` because there are no guarantees that in future they will use exact same operator as implementation and this contract is more a contract of `Observable.finallyDo()` and `Single.finallyDo()`.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorFinally.java
+++ b/src/main/java/rx/internal/operators/OperatorFinally.java
@@ -33,6 +33,9 @@ public final class OperatorFinally<T> implements Operator<T, T> {
     final Action0 action;
 
     public OperatorFinally(Action0 action) {
+        if (action == null) {
+            throw new NullPointerException("Action can not be null");
+        }
         this.action = action;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorFinally.java
```diff
@@ -33,6 +33,9 @@
     final Action0 action;
 
     public OperatorFinally(Action0 action) {
+        if (action == null) {
```
akarnokd: I'd rather put these checks into `Observable`.

artem-zinnatullin: This operator will be used in `Single.finallyDo()`, I don't want to copy-paste this :)

akarnokd: The benefit of having the null check in the caller is that there is no allocation happening in case the action is null.

artem-zinnatullin: Saving one allocation does not matter in this case because in most of the cases app will crash after this exception :)

akarnokd:
:+1:

zsxwing:
:+1:

PR #3438, opened 2015-10-12
Better null tolerance in rx.exceptions.*Exception classes.

akarnokd:
There seems to be a tendency to send `null` values through `onError` which eventually causes repeated failures. This PR converts these `null` values into `NullPointerException`s so the error origin is can be discovered and prevents the XXXExceptions from throwing NPEs in their constructors as well.

I've also added some missing javadoc.

### DIFF
--- a/src/main/java/rx/exceptions/MissingBackpressureException.java
+++ b/src/main/java/rx/exceptions/MissingBackpressureException.java
@@ -48,9 +48,17 @@ public class MissingBackpressureException extends Exception {
 
     private static final long serialVersionUID = 7250870679677032194L;
 
+    /**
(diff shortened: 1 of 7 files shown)


zsxwing:
:+1:

artem-zinnatullin:
Code is :+1:, but not sure I like null tolerance for exceptions… `subscriber.onError(null)` is like `throw null`.

akarnokd:
The problem is that NPEs will crash the streams further and likely end up hanging somewhere.

artem-zinnatullin:
Yeah, I understand… Okay, I guess.

PR #3444, opened 2015-10-14
1.x: Completable class to support valueless event composition + tests

akarnokd:
This is a backport of #3439.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1,0 +1,2181 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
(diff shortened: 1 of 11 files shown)


stealthcode:
Looks good. Should we annotate `@Experimental` on all the public methods and operators?

akarnokd:
Now, the class annotation should indicate all methods are experimental. If individual methods get promoted, we can add direct annotations then.

stealthcode:
Sounds good :+1:

akarnokd:
Since this adds to the API surface, I'd like to get another like before merging.

stevegury:
:+1:

stealthcode:
Awesome. Thanks!

Issue #3446, opened 2015-10-14
How can I use OperatorConcat<T>?


fougere-mike:
I'm trying to write an operator `then<R,T>` which ignores all emissions from the source (except errors), and continues with a second Observable when completed.

Something like:
```
// Typical Case
Observable.just(1, 2, 3, 4, 5)
   .lift(new then<String, Integer>(Observable.just("Hello")))
   .subscribe(new Action1<String>() {
      @Override
      public void call(String s) {
         // Called once with "Hello"
      });

// Source Errors
Observable.<Integer>error(new RuntimeException())
   .lift(new then<String, Integer>(Observable.just("Hello")))  // <-- the second observable should never be subscribed to since the source error'd
   .subscribe(new Action1<String>() {
      @Override
      public void call(String s) {
         // Not Called

      }, new Action1<Throwable>() {
      @Override
      public void call(Throwable e) {
        System.out.println("Error: "+e); // Should be called with the RuntimeException from above

      }
   });
```

I've come up with an implementation using `ignoreElements` + `map` + `concatWith`:
```
    public static <R, T> Observable<? extends R> then(Observable<T> source, Observable<R> other) {
        return source
                .ignoreElements()
                .map(new Func1<T, R>() {
                    @Override
                    public R call(T integer) {
                        return null;
                    }
                }).concatWith(other);
    }
```

I'm quite new to writing custom operators, and I can't quite figure out how to translate that static function into an operator. I've written a few operators by composing the provided Operator* types, but I'm having trouble with this one.

Any help would be greatly appreciated :)

JakeWharton:
This is discussed in #3113 and implemented awaiting merge in #3443.

fougere-mike:
Ok, well that eliminates the need for my operator, but I'm still interested in how it could be correctly implemented using lift(). I'm sure I'll be writing more operators in the future so I'd like to have a better understanding of how they work. 

This is what I've come up with:
```
/**
 * Ignores all emissions from the source observable. Once the source completes, the provided
 * observable will be subscribed to. If the source errors, the error will terminate the stream and
 * the provided observable will not be subscribed to.
 *
 * @param <T> The type of objects emitted by the source observable.
 * @param <R> The type of objects emitted by the provided `next` observable.
 */
public class then<R, T> implements Observable.Operator<R, T> {
    private final OperatorIgnoreElements<T> mIgnoreElements;
    private final OperatorMap<T, R> mMap;
[...]

Issue #3445, opened 2015-10-14
Add delaySubscription(Observable<U>) overload?


JakeWharton:
Usage of `delaySubscription` has started popping up in our code base, and I've noticed that we always have to define a `Func0` simply to yield an observable that's either cold (in which case the function provides no value) or hot (in which case the function also provides no value). It's not clear to me why the function indirection was added in the first place as it's provided no data and `Observable.defer` could have served in its place.

I would like to propose adding an `Observable`-taking overload and replacing the `Func0` variant to instead wrap its given argument in `Observable.defer` before passing to `OnSubscribeDelaySubscriptionWithSelector`.

Being able to sneak this in before 1.0.15 / 1.1.0 would also be nice :grinning:

akarnokd:
See #3447. I didn't touch the current `Func0` variant to avoid introducing unexpected effects to existing code.

JakeWharton:
:+1: Thanks @akarnokd!

akarnokd:
Fix merged.

PR #3447, opened 2015-10-14
1.x: DelaySubscription with a plain other Observable.

akarnokd:
Requested in #3445.

This is an efficient implementation as it avoids allocating the lifter object, doesn't require a function wrapper and doesn't use producer arbitration.

(Bonus points to those who can recognize the similarities with #3446).


akarnokd:
Any thoughts about this?

abersnaze:
:+1:

stevegury:
:+1:

PR #3448, opened 2015-10-15
Single delay

hyleung:
Adds the `delay(long delay, TimeUnit unit, Scheduler
scheduler)` and 'delay(long delay, TimeUnit unit)` operators to
`rx.Single`.

Includes (in a separate commit) a refactor of `OperatorDelay`. As far as I can tell, the `source` field isn't used at all - by removing it, I was able to reuse `OperatorDelay` in `rx.Single` to implement the delay operator pretty easily.


akarnokd:
:+1:

abersnaze:
LGTM

Issue #1917, opened 2014-12-02
Question: Minor vs Patch Increment?


benjchristensen:
I'd like to get opinions on what justifies incrementing the minor version (1.0.x -> 1.1.x) as opposed to the patch version.

Technically the minor version should increment when new functionality is added. In RxJava though were it's about a long list of operators, that could mean we increment minor versions quite often. Also, should it increment when they are added as `@Experimental` or `@Beta` or only once they become stable?

Examples:

- https://github.com/ReactiveX/RxJava/pull/1905 Beta & Experimental Annotations
- https://github.com/ReactiveX/RxJava/pull/1907 Experimental: onBackpressureBlock
- https://github.com/ReactiveX/RxJava/pull/1916 Experimental onBackpressureBufferWithCapacity

Currently we have the definition of minor and patch as this:

-----

Minor 1.x increments (such as 1.1, 1.2, etc) will occur when non-trivial new functionality is added or significant enhancements or bug fixes occur that may have behavioral changes that may affect some edge cases (such as dependence on behavior resulting from a bug). An example of an enhancement that would classify as this is adding reactive pull backpressure support to an operator that previously did not support it. This should be backwards compatible but does behave differently.

Patch 1.x.y increments (such as 1.0.0 -> 1.0.1, 1.3.1 -> 1.3.2, etc) will occur for bug fixes and trivial functionality (like adding a method overload). New functionality marked with an `@Beta` or `@Experimental` annotation can also be added in patch releases to allow rapid exploration and iteration of unstable new functionality. 

-----

Should we allow additional operators and functionality on patch versions if we deem them trivial (a subjective decision) and only increment the minor version when we subjectively determine significant additions? This would mean we have low minor versions and high patch versions, such as 1.0.17, 1.1.20, etc.

Or should we liberally increment the minor version anytime something is added to the APIs? This would mean high minor versions and low patch versions, such as 1.15.1, 1.27.0, 1.45.3, etc.

I would appreciate guidance on what is expected or preferred.

benjchristensen:
We addressed this by adding `@Experimental` and `@Beta` annotations. All development happens in patch releases. Minor releases are for promoting public APIs from `@Experimental` and `@Beta` to stable.

PR #3451, opened 2015-10-16
1.x: Change the signature of ignoreElements is also an implicit cast to whatever type you want.

abersnaze:
to remove the need for mergeWithEmpty and concatWithEmpty in #3430.  I'm pretty sure that this doesn't count as a breaking change because of type erasure.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -5592,8 +5592,8 @@ public class Observable<T> {
      *         called by the source Observable
      * @see <a href="http://reactivex.io/documentation/operators/ignoreelements.html">ReactiveX operators documentation: IgnoreElements</a>
      */
-    public final Observable<T> ignoreElements() {
-        return lift(OperatorIgnoreElements.<T> instance());
+    public final <R> Observable<R> ignoreElements() {
+        return lift(OperatorIgnoreElements.<R, T>instance());
     }
 
     /**
--- a/src/test/java/rx/internal/operators/OperatorIgnoreElementsTest.java
+++ b/src/test/java/rx/internal/operators/OperatorIgnoreElementsTest.java
@@ -92,7 +92,7 @@ public class OperatorIgnoreElementsTest {
                     }
                 })
                 //
-                .ignoreElements()
+                .<Integer>ignoreElements()
                 //
                 .doOnNext(new Action1<Integer>() {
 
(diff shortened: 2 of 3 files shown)


src/main/java/rx/Observable.java
```diff
@@ -5592,8 +5592,8 @@ public final void forEach(final Action1<? super T> onNext, final Action1<Throwab
      *         called by the source Observable
      * @see <a href="http://reactivex.io/documentation/operators/ignoreelements.html">ReactiveX operators documentation: IgnoreElements</a>
      */
-    public final Observable<T> ignoreElements() {
-        return lift(OperatorIgnoreElements.<T> instance());
+    public final <R> Observable<R> ignoreElements() {
+        return lift(OperatorIgnoreElements.<R, T>instance());
```
akarnokd: Actually, there is no need to add more type parameters but just do an unchecked cast or raw type cast:

```java
//...
return (Observable)lift(OperatorIgnoreElements.<T>instance());
```

Note that `instance()` does this re-cast as well since the operator is stateless.

davidmoten:
ah that's great @abersnaze, so simple!

src/test/java/rx/internal/operators/OperatorIgnoreElementsTest.java
```diff
@@ -92,7 +92,7 @@ public void call(Integer t) {
                     }
                 })
                 //
-                .ignoreElements()
+                .<Integer>ignoreElements()
```
zsxwing: Doesn't it prove this breaks the source compatibility?

abersnaze: Yes, but it doesn't prove if it breaks binary compatibility?

artem-zinnatullin: Would be nice to check binary compatibility as part of CI, there are some checker tools that we can include into the build process. Probably, it deserves a separate issue.

abersnaze: I came across a semver for Java library awhile ago that could take two jars and hint if it was major, minor or patch change.

zsxwing: I just realized we don't have a guidance on source compatibility explicitly. We only discussed backward compatibility in #1917. But I'm supposed backward compatibility should include both binary and source backward compatibility.

davidmoten:
So we need a decision on the importance of source compatibility specifically in the category of generic types, right?

I'm happy to add a generic type to `ignoreElements` where necessary to my codebases when I update to the next release because for a Java only codebase static typing will show me where things went wrong at compile time. Does anyone know if this change would bring about runtime grief for one of the dynamic languages that use RxJava?

stealthcode:
Could this changed be done in a minor release?

abersnaze:
closing in favor of #3567

Issue #2998, opened 2015-06-01
Exception thrown from onError() is swallowed when using groupBy/flatMap


gdela:
This issue is related to #969, where it has been said that following code should throw exception:

    Observable.error(new Exception())
      .subscribe(new Observer<Object>() {
        public void onCompleted() {
          // noop
        }
        public void onError(Throwable e) {
          throw new IllegalStateException("This should crash the app");
        }
        public void onNext(Object o) {
          // noop
        }
      });

And it does throw exception, but when you add `groupBy`/`flatMap` like that:

    Observable.error(new Exception())
      .groupBy(o -> o)
      .flatMap(g -> g)
      .subscribe(new Observer<Object>() {
        public void onCompleted() {
          // noop
        }
        public void onError(Throwable e) {
          throw new IllegalStateException("This should crash the app");
        }
        public void onNext(Object o) {
          // noop
        }
      });

Then the exception is silently swallowed. I tried to debug it a little bit, and seems that it could be because `OnErrorFailedException` is wrapped with `CompositeException`, so `Exceptions.throwIfFatal()` doesn't recognize it as fatal.

akarnokd:
You shouldn't throw from ```onError()``` because that exception, usually, has nowhere to go. The RxJava operators are a bit inconsistent about this; some catch it, some let it through and others wrap it around.

wgd:
I see, so I shouldn't use exception thrown from `onError()` to control flow, because it only seldom bubbles up to the `subscribe()` caller. I think that [this wiki doc](https://github.com/ReactiveX/RxJava/wiki/Error-Handling) is a little bit misleading about that, because it says something like: _"if the onError() call itself fails, the Observable [...] will throw a RuntimeException, an OnErrorFailedException, or an OnErrorNotImplementedException"_.

benjchristensen:
If we are swallowing an error that's a bug.

benjchristensen:
It would be helpful it someone can create a failing unit test for this issue that demonstrates where the exception is being swallowed.

konmik:
@Test(expected = Exception.class)
    public void throws_exception_from_onNext_onError() throws Exception {
        Observable
            .just(1)
            .doOnNext(new Action1<Integer>() {
                @Override
                public void call(Integer integer) {
                    throw new RuntimeException();
                }
            })
            .subscribe(
                new Action1<Integer>() {
                    @Override
                    public void call(Integer integer) {

                    }
[...]

PR #3455, opened 2015-10-16
OnErrorFailedException fix

konmik:
https://github.com/ReactiveX/RxJava/issues/2998

### DIFF
--- a/src/main/java/rx/exceptions/Exceptions.java
+++ b/src/main/java/rx/exceptions/Exceptions.java
@@ -76,12 +76,7 @@ public final class Exceptions {
         if (t instanceof OnErrorNotImplementedException) {
             throw (OnErrorNotImplementedException) t;
         } else if (t instanceof OnErrorFailedException) {
-            Throwable cause = t.getCause();
-            if (cause instanceof RuntimeException) {
-                throw (RuntimeException) cause;
-            } else {
-                throw (OnErrorFailedException) t;
-            }
+            throw (OnErrorFailedException) t;
         }
         // values here derived from https://github.com/ReactiveX/RxJava/issues/748#issuecomment-32471495
         else if (t instanceof StackOverflowError) {
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -164,15 +164,11 @@ public class Observable<T> {
                         // localized capture of errors rather than it skipping all operators 
                         // and ending up in the try/catch of the subscribe method which then
                         // prevents onErrorResumeNext and other similar approaches to error handling
-                        if (e instanceof OnErrorNotImplementedException) {
-                            throw (OnErrorNotImplementedException) e;
-                        }
+                        Exceptions.throwIfFatal(e);
                         st.onError(e);
                     }
                 } catch (Throwable e) {
-                    if (e instanceof OnErrorNotImplementedException) {
-                        throw (OnErrorNotImplementedException) e;
-                    }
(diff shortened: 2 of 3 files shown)


src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -76,12 +76,7 @@ public static void throwIfFatal(Throwable t) {
         if (t instanceof OnErrorNotImplementedException) {
             throw (OnErrorNotImplementedException) t;
         } else if (t instanceof OnErrorFailedException) {
-            Throwable cause = t.getCause();
-            if (cause instanceof RuntimeException) {
-                throw (RuntimeException) cause;
```
akarnokd: I guess by preventing the unwrapping, the error keeps propagating up, therefore, this change is essential. However, there could be code out there that depended on the original behavior (since Exceptions is part of the public API).

akarnokd:
Looks good to me. :+1:

src/main/java/rx/exceptions/Exceptions.java
konmik: I don't think that there is such code. Such unwrapped exceptions were swallowed most of the time anyway.

akarnokd: Sounds reasonable.

zsxwing:
Could you also fix the similar issues in `Observable.subscribe` and `Observable.unsubscribe`? Here are tests to reproduce them:

```
    @Test(expected = OnErrorFailedException.class)
    public void testOnErrorExceptionIsThrownFromSubscribe() {
        Observable.<Integer>create(s1 ->
                        Observable.<Integer>create(s2 -> {
                            throw new IllegalArgumentException("original exception");
                        }).subscribe(s1)
        ).subscribe(System.out::println, e -> {
            throw new RuntimeException();
        });
    }

    @Test(expected = OnErrorFailedException.class)
    public void testOnErrorExceptionIsThrownFromUnsafeSubscribe() {
        Observable.<Integer>create(s1 ->
                        Observable.<Integer>create(s2 -> {
                            throw new IllegalArgumentException("original exception");
                        }).unsafeSubscribe(s1)
        ).subscribe(System.out::println, e -> {
            throw new RuntimeException();
        });
    }

```

konmik:
I would replace error handling in both of these methods with simple

        } catch (Throwable e) {
            Exceptions.throwIfFatal(e);
            try {
                subscriber.onError(hook.onSubscribeError(e));
            } catch (Throwable e2) {
                throw new OnErrorFailedException(e2);
            }
            return Subscriptions.unsubscribed();
        }

However, I think that the problem should be handled more generally. 
The entire error handling policy should be reviewed and enforced.
Otherwise we will always see swallowed exceptions here and there.
Unfortunately, I don't have time right now to investigate the issue, I've just fixed what creates troubles for my current project.

src/main/java/rx/exceptions/Exceptions.java
benjchristensen: Interesting bug. I agree that removing the unwrapping is needed.

benjchristensen:
Thanks for digging in to help with this. I thought we had squashed all of these swallowing of errors! Apparently not :-(

benjchristensen:
:+1: Go ahead with this @akarnokd if you're still good with it.

akarnokd:
It is good as it is. Merging. The thing @zsxwing asked for can be done in a separate PR.

PR #3458, opened 2015-10-17
Clarify contracts of CompositeSubscription in its javadoc

artem-zinnatullin:
Current state of `CompositeSubscription`'s javadoc — totally misleading :smile: 

Every time we need to store some subscriptions in it then clear them and leave `CompositeSubscription` in an operable state we have to read its sources because its actual behavior is not consistent with the javadoc.

`CompositeSubscription.clear()` leaves `CompositeSubscription` in operable state, when on the other hand `CompositeSubscription.unsubscribe()` sets `isUnsubscribed = true` and `CompositeSubscription` is not able to manage new subscriptions.

### DIFF
--- a/src/main/java/rx/subscriptions/CompositeSubscription.java
+++ b/src/main/java/rx/subscriptions/CompositeSubscription.java
@@ -27,6 +27,8 @@ import rx.exceptions.*;
 
 /**
  * Subscription that represents a group of Subscriptions that are unsubscribed together.
+ * <p>
+ * All methods of this class are thread-safe.
  */
 public final class CompositeSubscription implements Subscription {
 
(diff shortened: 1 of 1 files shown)


akarnokd:
:+1:

src/main/java/rx/subscriptions/CompositeSubscription.java
```diff
@@ -116,6 +118,11 @@ public void clear() {
         }
     }
 
+    /**
+     * Unsubscribes itself and all inner subscriptions.
+     * <p>Notice that {@link CompositeSubscription} won't be able
+     * to manage new subscriptions after call of this method.
```
zsxwing: I feel this is unclear. You can still add `Subscription` to a `CompositeSubscription` after call of this method. How about `After call of this method, new {@code Subscription}s added to {@link CompositeSubscription} will be unsubscribed immediately`?

artem-zinnatullin: :+1: fixed!

zsxwing:
:+1:

akarnokd:
Excellent!

Issue #3461, opened 2015-10-19
High CPU usage after time changed on device


tehmur:
On Android using this library, if the user changes the system time, there is a huge spike in the CPU usage emanating from the RXComputationThreadPool, without any expensive calls being made to the RX library.

akarnokd:
Thanks for reporting. It has been already reported in #2943.

tehmur:
Thanks for your swift reply.

Do we have a feel of when we might have a fix for this on v1.x?

artem-zinnatullin:
// one of the possible fixes — Abstraction for Clock #3295.

nachodeh:
I also experienced this issue and it's quite serious. Will this be fixed soon?

akarnokd:
I've proposed a fix for this issue in #3467. I don't know if and when this can get into a release.

Issue #2943, opened 2015-05-13
Scheduler.schedulePeriodically is sensitive to system time changes


psliwowski:
Scheduler.schedulePeriodically can potentially get into busy loop if you set your system time into the future due to using System.currentTimeMillis() to calculate the next delay.

psliwowski:
Confirmed this causes a busy loop and spikes cpu to 100% because of ObjectPool.

akarnokd:
I guess it is due to the way the next delay is calculated from a constant absolute time position and a time set back will result in 0 delay between subsequent executions. Any tips on fixing it?

zsxwing:
If we want to measure elapsed time, we should use `System.nanoTime`, which won't be affected by the user's system time. https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime()

akarnokd:
I wouldn't trust [nanoTime that much](http://shipilev.net/blog/2014/nanotrusting-nanotime/) because the weaker guarantees of its monotonity and consistency across cores.

headinthebox:
This is not an easy thing ;-) http://blogs.msdn.com/b/rxteam/archive/2012/06/20/reactive-extensions-v2-0-release-candidate-available-now.aspx

akarnokd:
@headinthebox  Especially if you *are* platform independent to begin with (i.e., what is the Linux equivalent of ```SERVICE_CONTROL_TIMECHANGE``` callback?)

What the solution boils down:

  1. We need a supervising task scheduled on a small, 5-10s period to detect System.currentTimeMillis() drift or change (i.e., now - last is: > 5s + tolerance / negative).
  2. We need to capture tasks scheduled on any Scheduler that sets a delay/period > 10s and schedule them according to Zeno's paradox.
  3. If the supervising task detects a drastic clock change, it has to catch all those long-delay tasks and re-schedule them based on the new absolute time information, of course, atomically to prevent double-executions.

These all add some non-trivial overhead to scheduling and task-memory-sizes.

My first thought was the same as in the article:

> Observable.Timer/Interval implementation would start its quest to catch up with the new time, draining the battery due to the flood of callbacks it triggered. The fix was more of a band aid, whereby the timer would ignore ticks that occurred in the distant past, resetting the “next” state to be based on the current system time.

But that's still a half-solution at best.

psliwowski:
[...]

PR #3467, opened 2015-10-22
1.x: compensation for significant clock drifts in schedulePeriodically

akarnokd:
There is a problem, reported in #3461 and #2943, in which if the system clock drifts, the periodic calculation inside Scheduler.Worker gets off and either taking a longer time for the next invocation of the task or doing "catching-up" with all the lost invocations.

The solution checks the wall clock difference between the last run and the current run and if it went back or forward significantly, it rebases the timer period and schedules the next execution relative to now.

If the clock goes back, the original code scheduled the next invocation way into the future. This PR will schedule it after the period.

If the clock goes forward, the original code scheduled executions for all the missed time between the last run and the new time immediately, yielding a bunch of 0 delays. This PR will simply schedule the next invocation after the period.

The algorithm for both cases is the same: make sure the next invocation is scheduled relative to now and recalculate the start timestamp as if the whole sequence run under the new drifted clock all along. The subsequent invocations will be scheduled at a fixed rate again.

I've added the system parameter `rx.scheduler.drift-tolerance` (unit: minutes, default: 15 minutes), which is used to determine if the clock drifted too far between invocations of the periodic task.

### DIFF
--- a/src/main/java/rx/Scheduler.java
+++ b/src/main/java/rx/Scheduler.java
@@ -43,6 +43,17 @@ public abstract class Scheduler {
  *    maintenance.
  */
 
+    /** 
+     * The tolerance for a clock drift in nanoseconds where the periodic scheduler will rebase. 
+     * <p>
+     * The associated system parameter, {@code rx.scheduler.drift-tolerance}, expects its value in minutes.
+     */
+    static final long CLOCK_DRIFT_TOLERANCE_NANOS;
+    static {
+        CLOCK_DRIFT_TOLERANCE_NANOS = TimeUnit.MINUTES.toNanos(
+                Long.getLong("rx.scheduler.drift-tolerance", 15));
+    }
+    
     /**
      * Retrieves or creates a new {@link Scheduler.Worker} that represents serial execution of actions.
      * <p>
@@ -109,17 +120,38 @@ public abstract class Scheduler {
          */
         public Subscription schedulePeriodically(final Action0 action, long initialDelay, long period, TimeUnit unit) {
             final long periodInNanos = unit.toNanos(period);
-            final long startInNanos = TimeUnit.MILLISECONDS.toNanos(now()) + unit.toNanos(initialDelay);
+            final long firstNowNanos = TimeUnit.MILLISECONDS.toNanos(now());
+            final long firstStartInNanos = firstNowNanos + unit.toNanos(initialDelay);
 
             final MultipleAssignmentSubscription mas = new MultipleAssignmentSubscription();
             final Action0 recursiveAction = new Action0() {
-                long count = 0;
+                long count;
+                long lastNowNanos = firstNowNanos;
+                long startInNanos = firstStartInNanos;
                 @Override
                 public void call() {
                     if (!mas.isUnsubscribed()) {
                         action.call();
-                        long nextTick = startInNanos + (++count * periodInNanos);
-                        mas.set(schedule(this, nextTick - TimeUnit.MILLISECONDS.toNanos(now()), TimeUnit.NANOSECONDS));
+                        
+                        long nextTick;
+                        
+                        long nowNanos = TimeUnit.MILLISECONDS.toNanos(now());
+                        // If the clock moved in a direction quite a bit, rebase the repetition period
+                        if (nowNanos + CLOCK_DRIFT_TOLERANCE_NANOS < lastNowNanos
+                                || nowNanos >= lastNowNanos + periodInNanos + CLOCK_DRIFT_TOLERANCE_NANOS) {
+                            nextTick = nowNanos + periodInNanos;
+                            /* 
+                             * Shift the start point back by the drift as if the whole thing
+                             * started count periods ago.
+                             */
+                            startInNanos = nextTick - (periodInNanos * (++count));
+                        } else {
+                            nextTick = startInNanos + (++count * periodInNanos);
+                        }
+                        lastNowNanos = nowNanos;
+                        
+                        long delay = nextTick - nowNanos;
+                        mas.set(schedule(this, delay, TimeUnit.NANOSECONDS));
                     }
                 }
             };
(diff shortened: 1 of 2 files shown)


headinthebox:
Time drift is pretty nasty. We spent a lot of time on this in Rx.NET http://blogs.msdn.com/b/rxteam/archive/2012/06/20/reactive-extensions-v2-0-release-candidate-available-now.aspx

src/main/java/rx/Scheduler.java
```diff
@@ -43,6 +43,17 @@
...
  */
 
+    /** 
+     * The tolerance for a clock drift in nanoseconds where the periodic scheduler will rebase. 
+     * <p>
+     * The associated system parameter, {@code rx.scheduler.drift-tolerance}, expects its value in minutes.
+     */
+    static final long CLOCK_DRIFT_TOLERANCE_NANOS;
+    static {
+        CLOCK_DRIFT_TOLERANCE_NANOS = TimeUnit.MINUTES.toNanos(
```
artem-zinnatullin: Any reason to put this into separate `static` block?

akarnokd: Why, where is the other one?

JakeWharton: Implicitly using an initializer.
```java
static final long CLOCK_DRIFT_TOLERANCE_NANOS =
    TimeUnit.MINUTES.toNanos(Long.getLong("rx.scheduler.drift-tolerance", 15));
```

artem-zinnatullin: yep, ^.

BTW, @akarnokd don't you think that minutes is not the best unit of time measurement in programming? I'd prefer millis, `long` seems enough for `CLOCK_DRIFT_TOLERANCE`, and anyway, you're converting it to nanos then so `long` is 100% enough for millis.

Also, `15` minutes seems too much for a default clock drift tolerance…

akarnokd: I didn't inline it in case more parameters are asked for within this PR.

It is a question what is a reasonable tolerance for detecting drifts. One could argue it should be the function of the drift direction and the length of the period instead:

  - if now moves backwards by any number, do the correction,
  - if now moves ahead more than 1.5 times the period, do the correction.

src/main/java/rx/Scheduler.java
```diff
@@ -109,17 +120,39 @@
...
+                        long nowNanos = TimeUnit.MILLISECONDS.toNanos(now());
+                        // If the clock moved in a direction quite a bit, rebase the repetition period
+                        if (nowNanos + CLOCK_DRIFT_TOLERANCE_NANOS < lastNowNanos
+                                || nowNanos >= lastNowNanos + periodInNanos + CLOCK_DRIFT_TOLERANCE_NANOS) {
+                            nextTick = nowNanos + periodInNanos;
+                            /* 
+                             * Shift the start point back by the drift as if the whole thing
+                             * started with the in the current reference time.
+                             */
+                            startInNanos -= (lastNowNanos - nowNanos) + periodInNanos;
```
stealthcode: I think we might have a problem here if `lastNowNanos - nowNanos > 2 * periodInNanos`. Then we schedule the `nextTick` for the right time but the rebased next `nextTick` will result in scheduling an action befor ethe previous action. I suggest replacing lines 147 and 148 with 

```java
startInNanos = nextTick - (periodInNanos * ++count);
```

akarnokd:
I've updated the code with the suggestion of @stealthcode

src/main/java/rx/Scheduler.java
stealthcode: @akarnokd Is it your preference to create a separate factor `rx.scheduler.drift-tolerance-backward` and `rx.scheduler.drift-tolerance-forward`?

I think the tolerance setting would be more easily generalized as a percentage of the period but it would be slightly more difficult to grok (if that is ever actually a concern for users). I'm okay with converting this to a percentage value.

akarnokd: I'd leave it as it is now. I'm not sure about the percentage because what if the task itself is slow instead of a clock drift?

stealthcode:
:+1: LGTM

akarnokd:
Do you need additional changes? If not, feel free to merge this PR.

abersnaze:
:+1:

artem-zinnatullin:
Yay! This was important issue.

So, I did test (similar to #3530):

```java
Observable
  .interval(5, 10, SECONDS)
  .subscribe(aLong -> logger.d("Interval: %d", aLong));
```

And then changed system clock to +2 hours at runtime of the app.

**Before fix**
As expected computation scheduler gone crazy and fired about 700 events during one second.

screen shot 2015-12-15 at 08 44 57
screen shot 2015-12-15 at 08 43 54

**After fix**
Everything was just fine! No incorrect events, no CPU consuming

screen shot 2015-12-15 at 09 01 06
screen shot 2015-12-15 at 09 00 42

---

So, I think #3461 and #3530 can be closed now!

Thanks, @akarnokd!

src/main/java/rx/Scheduler.java
httpdispatch: @akarnokd May you please allow to set `CLOCK_DRIFT_TOLERANCE_NANOS `in seconds/milliseconds? I have a code which controls player position to play short pieces (few seconds) of media and pause them when the position is reached. Such as player may have various playback issues  (fast/slow playback, delay in initialization) i use `Flowable.interval`(50ms) to check its state/position periodically. However when system time synchronization occurs during playback flowable stops fire events until time become the same as it was before synchronization
Here how it looks in my log
```
14:21:43.273 z1 D PlaybackDurationControlCompletable.onTick: player position is unchanged, skipping check
14:21:37.727 z1 D some other event, time syncrhonization occurrs
// ... no onTick events for 6 seconds
14:21:43.372 z1 D PlaybackDurationControlCompletable: Initializing position 41769; elapsedTimerTime=5757 
```

akarnokd: Please post a separate issue about this case.

PR #3476, opened 2015-10-28
1.x: overhead reduction for merge and flatMap

akarnokd:
Changes to the scalar fast-path was inspired by the Project Reactor's `flatMap` which was in turn inspired by RxJava 2.x's implementation of `flatMap`.

**Naturally, this will conflict with #3169 .**

Benchmark for comparison (i7 4770K, Windows 7 x64, Java 8u66):

image

Just by applying the scalar re-batching, the operator gained a massive 45% throughput increase, from **48 MOps/s** to **71 MOps/s**.

When the range optimization is also applied, the improvement is even more impressive:  +60% throughput, from **48 MOps/s** to **79 MOps/s**.

The optimization doesn't really affect `rangeFlatMapRange`, it has a larger run-to-run variance due to GC.

I'm experimenting with the 2.x branch as well and by applying these two optimizations, the throughput increasd from **40 MOps/s** to **58 MOps/s**. I'm investigating if switching to synchronized would help with the remaining overhead gap.

Note also that the perf tests measure the operator overhead only.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -177,6 +177,10 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
         /** An empty array to avoid creating new empty arrays in removeInner. */ 
         static final InnerSubscriber<?>[] EMPTY = new InnerSubscriber<?>[0];
 
+        final int scalarEmissionLimit;
+        
+        int scalarEmissionCount;
+        
         public MergeSubscriber(Subscriber<? super T> child, boolean delayErrors, int maxConcurrent) {
             this.child = child;
             this.delayErrors = delayErrors;
@@ -184,7 +188,13 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
             this.nl = NotificationLite.instance();
             this.innerGuard = new Object();
             this.innerSubscribers = EMPTY;
-            request(maxConcurrent == Integer.MAX_VALUE ? Long.MAX_VALUE : maxConcurrent);
+            if (maxConcurrent == Integer.MAX_VALUE) {
+                scalarEmissionLimit = Integer.MAX_VALUE;
+                request(Long.MAX_VALUE);
+            } else {
+                scalarEmissionLimit = Math.max(1, maxConcurrent >> 1);
+                request(maxConcurrent);
+            }
         }
         
         Queue<Throwable> getOrCreateErrorQueue() {
@@ -488,7 +498,15 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
                 if (r != Long.MAX_VALUE) {
                     producer.produced(1);
                 }
-                this.requestMore(1);
+                
+                int produced = scalarEmissionCount + 1;
+                if (produced == scalarEmissionLimit) {
+                    scalarEmissionCount = 0;
+                    this.requestMore(produced);
+                } else {
+                    scalarEmissionCount = produced;
+                }
+                
                 // check if some state changed while emitting
                 synchronized (this) {
                     skipFinal = true;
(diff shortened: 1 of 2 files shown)


akarnokd:
Rebasing...

akarnokd:
... rebased, sort of.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -488,7 +502,10 @@ protected void emitScalar(T value, long r) {
                 if (r != Long.MAX_VALUE) {
                     producer.produced(1);
                 }
-                this.requestMore(1);
+                if (++scalarEmission == scalarLimit) {
```
artem-zinnatullin: Nice!

artem-zinnatullin:
:+1:/2, I was able to understand what you did for `Merge`, but completely lost in tries of understanding what's going on in `Range`… Those one-lettered variables make code totally unreadable :(

@akarnokd can you please somehow improve readability of `slowPath` in `OnSubscribeRange`?

Performance improvement looks awesome, thank you for doing that!

akarnokd:
For me, it is the opposite; reading long variable names and methods drains my focus. You can always post a cleanup PR.

davidmoten:
@akarnokd there might be some happy medium here. Write your code using one
letter variables. Finish up refactoring some names so more readable. These
terse PRs are harder to evaluate and part of the evaluation is judging the
author's intent based on variable naming.

akarnokd:
Fine, I've renamed the variables. Now it doesn't fit on my screen and I can't tell what is what because all variable names look the same at glance. Distinguishing between `e`, `n`, `r`, `i`, `o` is much easier; a performance optimization for the mind.

akarnokd:
Rebased onto main.

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -184,7 +188,13 @@ public MergeSubscriber(Subscriber<? super T> child, boolean delayErrors, int max
             this.nl = NotificationLite.instance();
             this.innerGuard = new Object();
             this.innerSubscribers = EMPTY;
-            request(maxConcurrent == Integer.MAX_VALUE ? Long.MAX_VALUE : maxConcurrent);
+            if (maxConcurrent == Integer.MAX_VALUE) {
+                scalarEmissionLimit = Integer.MAX_VALUE;
+                request(Long.MAX_VALUE);
+            } else {
+                scalarEmissionLimit = Math.max(1, maxConcurrent >> 1);
```
zsxwing: A small question: any reason for setting `scalarEmissionLimit` to `maxConcurrent / 2`? Or just pick up a magic number `2`?

zsxwing:
:+1:

src/main/java/rx/internal/operators/OperatorMerge.java
zsxwing: Just curious

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -488,7 +498,15 @@ protected void emitScalar(T value, long r) {
                 if (r != Long.MAX_VALUE) {
                     producer.produced(1);
                 }
-                this.requestMore(1);
+                
+                int produced = scalarEmissionCount + 1;
+                if (produced == scalarEmissionLimit) {
+                    scalarEmissionCount = 0;
+                    this.requestMore(produced);
```
zsxwing: nit: use the const `scalarEmissionLimit`?

akarnokd: That equals the producer which is already a local/register loaded value

src/main/java/rx/internal/operators/OperatorMerge.java
akarnokd: I tend to use 50% - 75% of the prefetch as the rerequest amount. It's a heuristic and its optimal value depends on the source emission pattern. The idea is to amortize a 40 cycle atomic increment by doing it less frequently. Basically any value above 1 helps with the overhead reduction but too large and any source to consumer loses pipelining effects.

zsxwing: Thanks for clarifying :+1:

PR #3477, opened 2015-10-28
1.x: add a source OnSubscribe which works from an array directly

akarnokd:
This specialized OnSubscribe performs ~33% better on longer streams.

Benchmark comparison (i7 4770K, Windows 7 x64, Java 8u66):

image

The `slowpath` requests the available amount + 1 to trigger a slow path, `slowpath2` requests in batches of 128 and replenishes only after all were received.


stealthcode:
:+1:

PR #3485, opened 2015-11-02
1.x: fix scan() not accepting a null initial value

akarnokd:
I forgot a NotificationLite conversion in the constructor. Note that
there were no tests verifying null behavior at all.


zsxwing:
:+1:

Issue #3233, opened 2015-08-28
`onBackpressureBuffer(int)` errors on overflow


abersnaze:
Talking with @neerajrj this morning about various Rx issues. One of them is that `onBackpressureBuffer(int)` defaults to sending an error if the limited buffer is full.  He would like to have the option to do different things on overflow.

I'm suggesting adding an `enum OnOverflow { DROP_OLDEST, DROP_LATEST, ERROR }` to the arguments to allow the user to decide what the overflow behavior should be.

Could `onBackpressureDrop()` be rewritten to `onBackpressureBuffer(0, OnOverflow.DROP_OLDEST)` and `onBackpressureLatest()` would be `onBackpressureBuffer(1, OnOverflow.DROP_OLDEST)`?

neerajrj:
+1 Also have an Action0 option to be informed when dropping is occuring.

srvaroa:
I've got this almost ready, I need some polishing but will be pushing a PR today/tomorrow the latest.

PR #3487, opened 2015-11-02
1.x: OnBackpressureBuffer: DROP_LATEST and DROP_OLDEST

srvaroa:
Introduce a new interface BackpressureOverflowStrategy that allows implementing different handlers for an overflow situation.  This patch adds three implementations, reachable via OverflowStrategy:

    static class OverflowStrategy {
        static final BackpressureOverflowStrategy DEFAULT = Error.INSTANCE;
        static final Error ERROR = Error.INSTANCE;
        static final DropOldest DROP_OLDEST = DropOldest.INSTANCE;
        static final DropLatest DROP_LATEST = DropLatest.INSTANCE;
    }

The behavior for each is the following:

- ERROR remains the default as the existing implementation.
- DROP_LATEST will drop newly produced items after the buffer fills up.
- DROP_OLDEST will drop the oldest elements in the buffer, making room for
  newer ones.

In all cases, a drop will result in a notification to the producer by invoking the onOverflow callback.

None of the two new behaviours (DROP_*) will unsubscribe from the source nor onError.

### DIFF
--- a/src/main/java/rx/BackpressureOverflow.java
+++ b/src/main/java/rx/BackpressureOverflow.java
@@ -1,0 +1,90 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx;
+
+import rx.annotations.Experimental;
+import rx.exceptions.MissingBackpressureException;
+
+/**
+ * Generic strategy and default implementations to deal with backpressure buffer overflows.
+ */
+@Experimental
+public final class BackpressureOverflow {
+
+    public interface Strategy {
+
+        /**
+         * Whether the Backpressure manager should attempt to drop the oldest item, or simply
+         * drop the item currently causing backpressure.
+         *
+         * @return true to request drop of the oldest item, false to drop the newest.
+         * @throws MissingBackpressureException
+         */
+        boolean mayAttemptDrop() throws MissingBackpressureException;
+    }
+
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_DEFAULT = Error.INSTANCE;
+    @SuppressWarnings("unused")
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_ERROR = Error.INSTANCE;
+    @SuppressWarnings("unused")
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_DROP_OLDEST = DropOldest.INSTANCE;
+    @SuppressWarnings("unused")
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_DROP_LATEST = DropLatest.INSTANCE;
+
+    /**
+     * Drop oldest items from the buffer making room for newer ones.
+     */
+    static class DropOldest implements BackpressureOverflow.Strategy {
+        static final DropOldest INSTANCE = new DropOldest();
+
+        private DropOldest() {}
+
+        @Override
+        public boolean mayAttemptDrop() {
+            return true;
+        }
+    }
+
+    /**
+     * Drop most recent items, but not {@code onError} nor unsubscribe from source
+     * (as {code OperatorOnBackpressureDrop}).
+     */
+    static class DropLatest implements BackpressureOverflow.Strategy {
+        static final DropLatest INSTANCE = new DropLatest();
+
+        private DropLatest() {}
+
+        @Override
+        public boolean mayAttemptDrop() {
+            return false;
+        }
+    }
+
+    /**
+     * {@code onError} a MissingBackpressureException and unsubscribe from source.
+     */
+    static class Error implements BackpressureOverflow.Strategy {
+
+        static final Error INSTANCE = new Error();
+
+        private Error() {}
+
+        @Override
+        public boolean mayAttemptDrop() throws MissingBackpressureException {
+            throw new MissingBackpressureException("Overflowed buffer");
+        }
+    }
+}
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -6399,7 +6399,8 @@ public class Observable<T> {
      *  <dd>{@code onBackpressureBuffer} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
-     * @return the source Observable modified to buffer items up to the given capacity
+     * @param capacity number of slots available in the buffer.
+     * @return the source {@code Observable} modified to buffer items up to the given capacity.
      * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
      * @since 1.1.0
      */
@@ -6419,12 +6420,49 @@ public class Observable<T> {
      *  <dd>{@code onBackpressureBuffer} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
-     * @return the source Observable modified to buffer items up to the given capacity
+     * @param capacity number of slots available in the buffer.
+     * @param onOverflow action to execute if an item needs to be buffered, but there are no available slots.  Null is allowed.
+     * @return the source {@code Observable} modified to buffer items up to the given capacity
      * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
      * @since 1.1.0
      */
     public final Observable<T> onBackpressureBuffer(long capacity, Action0 onOverflow) {
         return lift(new OperatorOnBackpressureBuffer<T>(capacity, onOverflow));
+    }
+
+    /**
+     * Instructs an Observable that is emitting items faster than its observer can consume them to buffer up to
+     * a given amount of items until they can be emitted. The resulting Observable will behave as determined
+     * by {@code overflowStrategy} if the buffer capacity is exceeded.
+     *
+     * <ul>
+     *     <li>{@code BackpressureOverflow.Strategy.ON_OVERFLOW_ERROR} (default) will {@code onError} dropping all undelivered items,
+     *     unsubscribing from the source, and notifying the producer with {@code onOverflow}. </li>
+     *     <li>{@code BackpressureOverflow.Strategy.ON_OVERFLOW_DROP_LATEST} will drop any new items emitted by the producer while
+     *     the buffer is full, without generating any {@code onError}.  Each drop will however invoke {@code onOverflow}
+     *     to signal the overflow to the producer.</li>j
+     *     <li>{@code BackpressureOverflow.Strategy.ON_OVERFLOW_DROP_OLDEST} will drop the oldest items in the buffer in order to make
+     *     room for newly emitted ones. Overflow will not generate an{@code onError}, but each drop will invoke
+     *     {@code onOverflow} to signal the overflow to the producer.</li>
+     * </ul>
+     *
+     * <p>
+     * <img width="640" height="300" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/bp.obp.buffer.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code onBackpressureBuffer} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param capacity number of slots available in the buffer.
+     * @param onOverflow action to execute if an item needs to be buffered, but there are no available slots.  Null is allowed.
+     * @param overflowStrategy how should the {@code Observable} react to buffer overflows.  Null is not allowed.
+     * @return the source {@code Observable} modified to buffer items up to the given capacity
+     * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
+    public final Observable<T> onBackpressureBuffer(long capacity, Action0 onOverflow, BackpressureOverflow.Strategy overflowStrategy) {
+        return lift(new OperatorOnBackpressureBuffer<T>(capacity, onOverflow, overflowStrategy));
     }
 
     /**
--- a/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
+++ b/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
@@ -19,6 +19,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 
+import rx.BackpressureOverflow;
 import rx.Observable.Operator;
 import rx.Producer;
 import rx.Subscriber;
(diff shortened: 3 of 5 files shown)


src/main/java/rx/Observable.java
```diff
@@ -20,6 +20,7 @@
 import rx.functions.*;
 import rx.internal.operators.*;
 import rx.internal.producers.SingleProducer;
+import rx.internal.operators.OperatorOnBackpressureBuffer.OnOverflow;
```
akarnokd: This enumeration should be part of a more public part of the library instead of the internal part.

src/main/java/rx/Observable.java
```diff
@@ -5925,6 +5926,38 @@ public final Boolean call(T t) {
     }
 
     /**
+     * Instructs an Observable that is emitting items faster than its observer can consume them to buffer up to
+     * a given amount of items until they can be emitted. The resulting * Observable will behave as determined
+     * by {@code overflowBehaviour} if the buffer capacity is exceeded.
```
akarnokd: `overflowBehavior`?

src/main/java/rx/Observable.java
```diff
@@ -5925,6 +5926,38 @@ public final Boolean call(T t) {
...
+     *
+     * <ul>
+     *     <li>{@code OnOverflow.ERROR} (default) will {@code onError} dropping all undelivered items, unsubscribing
+     *     from the source, and notifying the producerwith {@code onOverflow}. </li>
```
akarnokd: ˙producer with`

src/main/java/rx/Observable.java
srvaroa: Makes sense.

Any preference rather than in Observable.java? This seems to be the first case of an enum that's part of the public API (other than Notification.Kind, which is in Notification itself).

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -28,8 +28,11 @@
 
 public class OperatorOnBackpressureBuffer<T> implements Operator<T, T> {
 
+    public enum OnOverflow { DROP_OLDEST, DROP_LATEST, ERROR }
```
akarnokd: This should be moved outside ot the class and the internal package.

src/main/java/rx/Observable.java
akarnokd: could be moved under `rx` and I'd rename it to `BackpressureOverflowStrategy`

akarnokd:
Otherwise, it looks okay. :+1:

srvaroa:
Thanks for the review @akarnokd

src/main/java/rx/BackpressureOverflowStrategy.java
```diff
@@ -0,0 +1,5 @@
+package rx;
```
akarnokd: Please add the copyright header.

src/main/java/rx/BackpressureOverflowStrategy.java
```diff
@@ -0,0 +1,5 @@
...
+
+public enum BackpressureOverflowStrategy {
```
akarnokd: Please add javadoc that explains what this enum is for.

src/main/java/rx/BackpressureOverflowStrategy.java
```diff
@@ -0,0 +1,5 @@
...
+    DROP_OLDEST, DROP_LATEST, ERROR
```
akarnokd: Please add some javadoc to each value.

src/main/java/rx/Observable.java
```diff
@@ -5925,6 +5926,38 @@ public final Boolean call(T t) {
     }
 
     /**
+     * Instructs an Observable that is emitting items faster than its observer can consume them to buffer up to
+     * a given amount of items until they can be emitted. The resulting Observable will behave as determined
+     * by {@code overflowStrategy} if the buffer capacity is exceeded.
+     *
+     * <ul>
+     *     <li>{@code OnOverflow.ERROR} (default) will {@code onError} dropping all undelivered items, unsubscribing
```
akarnokd: The javadoc references the old enum name.

srvaroa:
Rebased - @akarnokd let me know if there are any fixes pending.

akarnokd:
Excellent. :+1:. I don't think there are conflicting PRs right now.

src/main/java/rx/Observable.java
```diff
@@ -5925,6 +5926,38 @@ public final Boolean call(T t) {
...
+     *  <dd>{@code onBackpressureBuffer} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @return the source Observable modified to buffer items up to the given capacity
+     * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
+    public final Observable<T> onBackpressureBuffer(long capacity, Action0 onOverflow, BackpressureOverflowStrategy overflowStrategy) {
+        return lift(new OperatorOnBackpressureBuffer<T>(capacity, onOverflow, overflowStrategy));
```
artem-zinnatullin: Knowing that some RxJava users like to pass `null` :trollface: in places where RxJava developers do not expect that I'd add `null` checks for `onOverflow` and `overflowStrategy` or at least mention it in javadoc.

src/main/java/rx/Observable.java
```diff
@@ -5925,6 +5926,38 @@ public final Boolean call(T t) {
...
+     *     {@code onOverflow} to signal the overflow to the producer.</li>
+     * </ul>
+     *
+     * <p>
+     * <img width="640" height="300" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/bp.obp.buffer.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code onBackpressureBuffer} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
```
artem-zinnatullin: BTW, where are `@param`s?!

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -150,16 +165,27 @@ private boolean assertCapacity() {
...
-                    if (saturated.compareAndSet(false, true)) {
-                        unsubscribe();
-                        child.onError(new MissingBackpressureException(
-                                "Overflowed buffer of "
-                                        + baseCapacity));
-                        if (onOverflow != null) {
-                            onOverflow.call();
-                        }
+                    boolean hasCapacity = false;
+                    switch (overflowStrategy) {
```
artem-zinnatullin: This is awesome

artem-zinnatullin:
Great :+1:, just a few nits.

src/main/java/rx/Observable.java
srvaroa: I copypasted from the old ones (which don't have it either). Adding them.

src/main/java/rx/Observable.java
srvaroa: Done

srvaroa:
Thanks @artem-zinnatullin & @akarnokd. Added fixes to Artem's comments.

akarnokd:
:+1:

stealthcode:
@abersnaze can you take a look and verify that it solves the issue raised in #3233?

srvaroa:
Rebased on 1.x

stealthcode:
The only concern that @abersnaze and I have had with this is that this uses enums to determine the strategy of operator functionality. This pattern hasn't been used much in the public API as of yet. @akarnokd @zsxwing do you guys have any opinion? We could just convert the strategies to instance methods (for consistency) or is this an emerging case that should be encouraged?

akarnokd:
I wouldn't turn a the whole set of `onBackpressureXXX` into enums of Drop|Buffer|Latest because the pattern doesn't allow parametrizing just the Buffer variant. In this case, however, having 3-9 new overload feels unnecessary expansion of the public API. I'm in for the use of enums here, but we should store them in a reasonable subpackage.

srvaroa:
@akarnokd @stealthcode I don't love using enums in this way, but I agree with @akarnokd that this can lead to excessive API expansion.  But happy to move to methods if you guys prefer.  Ditto for moving the enums package from rx to rx.strategies or some similar alternative.

zsxwing:
Agree with @akarnokd 

I prefer to add a new interface `BackpressureOverflowStrategy` instead of `enum`.

stealthcode:
I agree with @zsxwing on the use of an empty interface and impls to compare `instanceof` instead of using an enum. Eventually they might be usable to generalize the strategies.

srvaroa:
Sounds good, I'll update the patch soon. Any suggestion on the package for the strategy?

srvaroa:
Rebased

src/main/java/rx/Observable.java
```diff
@@ -6019,6 +6024,41 @@ public final Boolean call(T t) {
...
+     * @param capacity number of slots available in the buffer.
+     * @param onOverflow action to execute if an item needs to be buffered, but there are no available slots.  Null is allowed.
+     * @param overflowStrategy how should the {@code Observable} react to buffer overflows.  Null is not allowed.
+     * @return the source {@code Observable} modified to buffer items up to the given capacity
+     * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
+    public final Observable<T> onBackpressureBuffer(long capacity, Action0 onOverflow, BackpressureOverflowStrategy overflowStrategy) {
```
akarnokd: This points to an internal interface from the public API, plus the allowed values are also internal classes and the BufferSubscriber is a private class.

stealthcode: Is this still a problem? I'm not sure that I understand this comment.

srvaroa:
@stealthcode I've been busy these days, just returned to this. I had the same doubt re. BufferSubscribe reference, @akarnokd mind to clarify where it happens?

I've fixed the visibility of OverflowStrategy members (new commit on the way)

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -71,6 +175,7 @@ public OperatorOnBackpressureBuffer(long capacity, Action0 onOverflow) {
 
         return parent;
     }
+
     private static final class BufferSubscriber<T> extends Subscriber<T> implements BackpressureDrainManager.BackpressureQueueCallback {
```
akarnokd: `BufferSubscriber` is private here but the public `BackpressureOverflowStrategy` exposes it, which makes it not extendable outside this package, which is not better than having a `BackpressureOverflowStrategy` enum in the first place. Plus, leaving `BufferSubscriber` private generates accessibility bridges.

srvaroa: Thanks. I replaced with the parent interface implemented by BufferSubscriber, which is public.

stevegury:
Thanks for the addition.
Could you also update the PR description?

srvaroa:
@stevegury thanks for the review. PR fixed.

src/main/java/rx/Observable.java
akarnokd: `BackpressureOverflowStrategy` is still in the internals package, it should be moved up to the `rx` package.

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -27,15 +23,93 @@
...
     private static class Holder {
         static final OperatorOnBackpressureBuffer<?> INSTANCE = new OperatorOnBackpressureBuffer<Object>();
     }
-    
+
+    /**
+     * A convenience class to provide singleton instances of available {@code BackpressureOverflowStrategy}.
+     */
+    @SuppressWarnings("unused")
+    public static class OverflowStrategy {
```
akarnokd: Once `BackpressureOverflowStrategy` has been moved to `rx` package, the values of this class can become its constants, but all of them should have the type `BackpressureOverflowStrategy`. The type after the assignment can stay the same as now and can point to the current internal location.

stevegury:
If I'm right the only missing thing is the header in `BackpressureOverflowStrategy.java`.
@srvaroa could you add that?
:+1:

src/main/java/rx/Observable.java
akarnokd: Okay, BackpressureOverflowStrategy is now in rx package.

akarnokd:
Wait, it has still the problem of exposing internal classes and types into the public API:

```java
boolean triggerOn(rx.internal.util.BackpressureDrainManager.BackpressureQueueCallback buffer) throws MissingBackpressureException;
```

In addition, for a proper instance of BackpressureOverflowStrategy, you have to reach into 

```java
rx.internal.operators.OperatorOnBackpressureBuffer.ON_OVERFLOW_DEFAULT
```

for example.

srvaroa:
Thanks @akarnokd @stevegury 

I'm submitting a patch that eliminates the internal API leak. 

While doing it I'm starting to realize that without exposing the buffer or callback the motivation for a strategy vs a plain enum practically dissapears.  The point of using a strategy was to give users ability to implement their own strategies, this led me to expose the internal buffer (which is bad, obviously). However, without it, we're basically doing the same as an enum as the real overflow behaviour is not really in the Strategy.  It actually feels redundant with the onOverflow callback.

To make the strategy worth the extra code we'd need to make BufferSubscriber implement some public API interface defined by BackpressureOverflow through which we allow the user to actually interact with the buffer without leaking the internal API. Something like:

```
BPBuffer {
  boolean dropOldest()
  ...
}
```
And then:

```
BPStrategy {
  boolean triggerOn(BPBuffer buffer)  
}
```
This would become similar to my current patch (https://github.com/srvaroa/RxJava/commit/b09061048f34f539bbc022e846e7361ae6e38606) and enable meaningful strategies without leaking APIs. That said, I think that would be overengineering this patch.

Options I see:

a) revert to enums (problem: we're introducing an enum-based public API).
b) leave the patch I'm submitting next (problem: we're committing to a public API based on a Strategy that doesn't really define any behaviour).
c) define BPBuffer and BPStrategy for the sake of introducing a meaningful strategy-based API. I ca call triggerOn() with a null for now (so it effectively behaves as -b-) and proceed with making BufferSubscriber extend BPBuffer in a later patch.

I'm leaning for (a), but let me know what you prefer.

src/main/java/rx/BackpressureOverflow.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_ERROR = Error.INSTANCE;
+    @SuppressWarnings("unused")
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_DROP_OLDEST = DropOldest.INSTANCE;
+    @SuppressWarnings("unused")
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_DROP_LATEST = DropLatest.INSTANCE;
+
+    /**
+     * Drop oldest items from the buffer making room for newer ones.
+     */
+    public static class DropOldest implements BackpressureOverflow.Strategy {
```
akarnokd: Let's have these package private at least; users should use the `ON_OVERFLOW_*` constants.

src/main/java/rx/BackpressureOverflow.java
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+ */
+package rx;
+
+import rx.exceptions.MissingBackpressureException;
+
+/**
+ * Generic strategy and default implementations to deal with backpressure buffer overflows.
+ */
+public final class BackpressureOverflow {
```
akarnokd: Please add `@Experimental`

akarnokd:
The current API looks good enough to me, only some small visibility and annotation issues.

srvaroa:
Thanks for the reviews, I updated the patch.

akarnokd:
:+1:

stevegury:
Thanks @srvaroa for taking the time of writing a good PR!
:+1:

stealthcode:
:+1: thanks!

PR #3491, opened 2015-11-04
1.x: make scan's delayed Producer independent of event serialization

### LINKED ISSUE #3490: scan infinite loop
The following test fails in 1.x (and 1.0.15):

```java  
    @Test(timeout=1000)
    public void testScanDoesNotHang() {
        Observable.range(0, Integer.MAX_VALUE)
                //
                .scan(1, new Func2<Integer, Integer, Integer>() {

                    @Override
                    public Integer call(Integer t1, Integer t2) {
                        return t1;
                    }
                })
                //
                .subscribe(new Subscriber<Integer>() {

                    int count = 0;

                    @Override
                    public void onCompleted() {

                    }

                    @Override

                    }

                    @Override
                        count++;
                        if (count == 2)
                            unsubscribe();
                    }
                });
    }
```

@akarnokd  I've stepped through in a debugger and in `OperatorScan` L284 initiates a fast path emission in `OnSubscriberRange` which calls `OperatorScan.emit()` L289 but because the method containing L284 has set `emitting` to true the emission goes no further and an a loop occurs that fills the heap.

Though the test might look a bit contrived I just isolated the behaviour to `scan` for the test. I encountered the bug when this hung:

```java
observable.scan(..).elementAt(n); //where n > 1
```
I'd rank this bug as nasty (not an improbable edge case like some concurrency bugs) and possibly deserving of a new release quickly.

akarnokd:
It turns out serializing `request()` calls with regular `onXXX()` calls can be problematic because a `request()` may trigger an emission of events which then end up being queued (since `emitting == true`). If the request is large and the queue otherwise unbounded, this will likely cause OOME.

In case of `scan`, the fix was to make the missing request accounting and arrival of the `Producer` independent of the event's emitter loop; there is no need for them to be serialized in respect to each other.

In case of the `ProducerObserverArbiter` where the request accounting and producer swapping has to be serialized with the value emission, the solution is to call `request()` outside the emitter-loop.

There shouldn't be any issue with 2.x `scan()` because in 2.x, scan receives the `Subscription` before it allows the downstream to request anything so there is no missing requested to be handled.

This should resolve #3490. As far as I can remember, no other operator should have such problems because all others use `ProducerArbiter` which is independent of `onXXX` emission serializations.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorScan.java
+++ b/src/main/java/rx/internal/operators/OperatorScan.java
@@ -16,6 +16,7 @@
 package rx.internal.operators;
 
 import java.util.Queue;
+import java.util.concurrent.atomic.AtomicLong;
 
 import rx.*;
 import rx.Observable.Operator;
@@ -175,12 +176,10 @@ public final class OperatorScan<R, T> implements Operator<R, T> {
         boolean missed;
         /** Missed a request. */
         long missedRequested;
-        /** Missed a producer. */
-        Producer missedProducer;
         /** The current requested amount. */
-        long requested;
+        final AtomicLong requested;
         /** The current producer. */
-        Producer producer;
+        volatile Producer producer;
         
         volatile boolean done;
         Throwable error;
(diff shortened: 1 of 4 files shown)


davidmoten:
Can we get this one reviewed and in a release soon please? This is a blocker for me and precludes me using 1.0.15 and 1.0.16. It's a serious bug with `scan` and I'm nervous for others too that this is out there in another release (1.0.16).

src/main/java/rx/internal/operators/OperatorScan.java
```diff
@@ -270,23 +235,51 @@ public void onCompleted() {
             emit();
         }
         
+        @Override
+        public void request(long n) {
```
stealthcode: @akarnokd in the future if its possible to do so could you not rearrange the methods? It makes it so much easier to to read the changes to `request` side by side instead of seeing one section of code missing entirely only to reappear (modified) later on. Thank you, it's a suggestion to help expedite the PR review process.

akarnokd: I did this move out of way so the interleaved view doesn't confuse the reviewer since the original code was simply wrong. This is rarely happening so other PRs will have changes in place.

stealthcode: Okay thanks for your consideration. As I said it will help a lot for me to be able to see changes interleaved.

stealthcode:
Is `ProducerObserverArbiter` used in the public API? It appears to only be in use in the tests.

akarnokd:
It was added as a tool for building operators but I usually inline the algorithm.

stealthcode:
Okay. So it's more like a pattern that gets pasted into operators. Do you usually inline for performance reasons or for custom functionality?

stealthcode:
:+1:

PR #3497, opened 2015-11-05
2.x: fix takeLast(time) replay behavior causing discontinuity

akarnokd:
The internal queue has to be trimmed while the source is active and the
backpressured-replay shouldn't drop values in case the requests are
delayed relative to each other.


akarnokd:

PR #3498, opened 2015-11-05
1.x: Rename cache(int) to cacheWithInitialCapacity(int)

loganj:
~~The parameter is a capacity hint, but more frequently confused with a
buffer size like replay(int) than it is correctly understood. It also
offers no guarantees, only the weak hope of optimization.~~

The parameter is a capacity hint, but more frequently confused with a buffer size like replay(int) than it is correctly understood. It also offers no guarantees, only the weak hope of optimization.

This change renames the method, deprecating the old name. It also adds javadoc calling out that the parameter is not a bound and referencing replay(int).autoConnect() as a way to achieve that behavior.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -3665,6 +3665,15 @@ public class Observable<T> {
     }
 
     /**
+     * @see #cacheWithInitialCapacity(int)
+     * @deprecated Use {@link #cacheWithInitialCapacity(int)} instead.
+     */
+    @Deprecated
+    public final Observable<T> cache(int initialCapacity) {
+        return cacheWithInitialCapacity(initialCapacity);
+    }
+
+    /**
      * Caches emissions from the source Observable and replays them in order to any subsequent Subscribers.
      * This method has similar behavior to {@link #replay} except that this auto-subscribes to the source
      * Observable rather than returning a {@link ConnectableObservable} for which you must call
@@ -3689,14 +3698,17 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code cache} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
+     * <p>
+     * <em>Note:</em> The capacity hint is not an upper bound on cache size. For that, consider
+     * {@link #replay(int)} in combination with {@link ConnectableObservable#autoConnect()} or similar.
      * 
-     * @param capacityHint hint for number of items to cache (for optimizing underlying data structure)
+     * @param initialCapacity hint for number of items to cache (for optimizing underlying data structure)
      * @return an Observable that, when first subscribed to, caches all of its items and notifications for the
      *         benefit of subsequent subscribers
      * @see <a href="http://reactivex.io/documentation/operators/replay.html">ReactiveX operators documentation: Replay</a>
      */
-    public final Observable<T> cache(int capacityHint) {
-        return CachedObservable.from(this, capacityHint);
+    public final Observable<T> cacheWithInitialCapacity(int initialCapacity) {
+        return CachedObservable.from(this, initialCapacity);
     }
 
     /**
--- a/src/test/java/rx/ObservableTests.java
+++ b/src/test/java/rx/ObservableTests.java
@@ -663,7 +663,7 @@ public class ObservableTests {
                     }
                 }).start();
             }
-        }).cache(1);
+        }).cacheWithInitialCapacity(1);
 
         // we then expect the following 2 subscriptions to get that same value
         final CountDownLatch latch = new CountDownLatch(2);


akarnokd:
Hi. It is part of the public API and can't be just removed. In addition, why did you modify the underlying internal class?

loganj:
How about `@Deprecate`?

I modified the internal class because the only use of that constructor parameter was in the removed operator.  If you'd prefer to deprecate, of course I'll revert that modification.

akarnokd:
We are kind of set for 1.1 in terms of API visibility. How about updating the javadoc and list alternatives?

loganj:
I've pointed developers directly at that javadoc, and they've still misunderstood the parameter.  And that's best case!  Any developer who just looks at the signature, which is most of them, misunderstands the parameter. Everyone assumes it's a bound on the buffer size, and it's so obvious that why would you read the javadoc?

It'd be great to get this resolved in the 1.x series.

artem-zinnatullin:
Huge +1 to `@Deprecated` + simlink like `cacheWithInitialBufferSize`/etc, `cache(int)` is very, very misleading (every member of our team had problems with understanding that `int` is not a size of the cache's buffer).

loganj:
Updated with rename, deprecation, and javadoc.

src/main/java/rx/Observable.java
```diff
@@ -3666,13 +3674,16 @@ public final Boolean call(T first, T second) {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code cache} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
+     * <p>
+     * <em>Note:</em> The capacity hint is not an upper bound on cache size. For that, consider
+     * {@link #replay(int)} in combination with {@link ConnectableObservable#autoConnect()} or similar.
```
artem-zinnatullin: :+1:

src/main/java/rx/Observable.java
```diff
@@ -3641,6 +3641,14 @@ public final Boolean call(T first, T second) {
         return CachedObservable.from(this);
     }
 
+  /**
+   * @see #cacheWithCapacityHint(int)
+   * @deprecated Use {@link #cacheWithCapacityHint(int)} instead.
+   */
+  @Deprecated public final Observable<T> cache(int capacityHint) {
```
artem-zinnatullin: According to "current codestyle" annotation should be placed on the line above.

artem-zinnatullin:
:+1:

src/main/java/rx/Observable.java
```diff
@@ -3666,13 +3675,16 @@ public final Boolean call(T first, T second) {
...
+     * <em>Note:</em> The capacity hint is not an upper bound on cache size. For that, consider
+     * {@link #replay(int)} in combination with {@link ConnectableObservable#autoConnect()} or similar.
      * 
      * @param capacityHint hint for number of items to cache (for optimizing underlying data structure)
      * @return an Observable that, when first subscribed to, caches all of its items and notifications for the
      *         benefit of subsequent subscribers
      * @see <a href="http://reactivex.io/documentation/operators/replay.html">ReactiveX operators documentation: Replay</a>
      */
-    public final Observable<T> cache(int capacityHint) {
+    public final Observable<T> cacheWithCapacityHint(int capacityHint) {
```
akarnokd: Looks good but can't decide if this alias requires Beta or Experimental or could stay plain.

loganj: My inclination would be to leave it plain, since we're deprecating the original name.

It'd be great to see this rename carried into 2.x, or even see this form of the operator dropped altogether.

akarnokd:
You can post a PR against 2.x where you can simply rename and no need for deprecation.

akarnokd:
@stealthcode, @abersnaze what's the situation among Netflix devs regarding this?

artem-zinnatullin:
ping Netflix team

stealthcode:
I don't see the source of the confusion. `java.util.ArrayList` follows a similar convention in the constructor `public ArrayList(int initialCapacity)` and that is sufficient for me to understand it's usage. The array backed data type can be reallocated but it's better to tune it with an appropriate initial value. 

I suggest that we leave it alone in 1.x and rename it to `cacheUnbounded(int initialCapacity)` in 2.x. Also we should improve the javadocs to indicate the intended usage since there still seems to be confusion around this.

JakeWharton:
A list is a data structure designed to grow. When you see `cache(4)` you do not think of a list, but of a fixed size cache, as most will be bounded by some supplied value.

Additionally, an initial size is not required to create an array list and it's a constructor parameter. Not only do you have to opt-in to using this non-default constructor overload, you would have to look at the documentation for any idea what the value is for. This is _very_ different in that the size is required and it's a named method which should describe behavior but is misleading because of prior assumptions around the word "cache".

stealthcode:
@JakeWharton the `cache` operator does not use a fixed size cache. I think that you are saying that the operator name should be enough to to convey the semantic meaning of an unbounded growing cache. I think that by renaming it to `cacheUnbounded(int initialCapacity)` that the usage will be made more clear and it would be discoverable by users who know to look for the `cache` operator. 

If the term "cache" is not sufficient to indicate the contract then we should probably consider changing the name to one that does. What would be a better term for an operator that holds an ordered list of all events emitted from an observable to be replayed to all subscribers after the first?

loganj:
@stealthcode Hence the name `cacheWithInitialCapacity`, introduced in this PR.  That name implies it's unbounded _and_ describes the integer parameter.

loganj:
To clarify: I don't really care what it's called in 2.x, as long as it's reasonably unambiguous (i.e., not `cache(int)`).

abersnaze:
@JakeWharton I think @stealthcode was agreeing that a new method name would make it clearer but not that `cacheWithInitialCapacity` or that it should be done on the 1.x branch.

I'm not very interested in spending a lot of energy bike shedding over the name so can we just have a vote for we want to do: (vote for as many options as you want in the order or preference)

A) do nothing
B) deprecate `cache(int)` and make `cacheWithInitialCapacity(int)` in 1.x
C) deprecate `cache(int)` and make `cacheUnbounded(int)` in 1.x
D) remove `cache(int)` and make `cacheWithInitialCapacity(int)` in 2.x
E) remove `cache(int)` and make `cacheUnbounded(int)` in 2.x

My vote is A, E, C, D, B.

@loganj @stealthcode @akarnokd @JakeWharton @artem-zinnatullin.  We'll [tabulate](https://en.wikipedia.org/wiki/Single_transferable_vote) the votes on Monday.

akarnokd:
My vote is A, E

stealthcode:
I vote E, A, C

iNoles:
I vote B, D, A

artem-zinnatullin:
B.

For 2.x I'd prefer to get rid of cache or replay and leave only one
operator with different settings without autoconnection by default:
cacheWithLimit/replayWithLimit, cacheUnbounded /replayUnbounded, etc.

On Sat, Nov 21, 2015, 00:26 David Karnok <notifications@github.com> wrote:

> My vote is A, E
>
> —
> <https://github.com/ReactiveX/RxJava/pull/3498#issuecomment-158528544>.
>
-- 

@artem_zin

abersnaze:
I would like @JakeWharton votes but so far the A is the winner.

loganj:
B, C, D, E.  presuming a change in 1.x means also making the change in 2.x.

We have empirical evidence that the name and parameter are confusing to users. I can't imagine why we'd ignore that evidence based on our own lack of confusion.

I also think it's funny that this balloting gives a single "do nothing" equal weight to four "do something" options.  By my count, "do something" is winning 3 to 2, before my vote.

dlew:
B, C myself.

I think it's just confusing and have been bitten by `cache(int)` before.

yogurtearl:
I vote B

JakeWharton:
E&C then D&B then A

stealthcode:
I think that E&C equate to to just C... (assuming that all api changes made in 1.x are carried over).

abersnaze:
B won and I'm fine with that. @loganj could do the honors of changing this PR to match. Just to be clear that means that `Observable` will have the methods, `cache()`, `@Deprecated cache(int)` and `cacheWithInitialCapacity(int)` rebased to one commit off of the 1.x branch.

artem-zinnatullin:
Ping @loganj, let's do it!

loganj:
Updated to specified naming. Apologies for the delay, was traveling on vacation.

akarnokd:
:+1:

stealthcode:
:+1:

PR #3500, opened 2015-11-07
Some code clean ups.

dromato:
Nothing that could change logic or application flow, just minor refactors to be consistent with good practices and clean code.
To sum up changes:
- Simplified some conditions
- Changed small l to L in long number, as l tends to look like 1 and might confuse
- Removed "return" statements where those are not necessary (last instruction in function)
- Inlined returns where there were no need for creating new variable (easier to read)
- Deleted unnecessary colons

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeRedo.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeRedo.java
@@ -195,7 +195,7 @@ public final class OnSubscribeRedo<T> implements OnSubscribe<T> {
         final AtomicBoolean resumeBoundary = new AtomicBoolean(true);
         
         // incremented when requests are made, decremented when requests are fulfilled
-        final AtomicLong consumerCapacity = new AtomicLong(0l);
+        final AtomicLong consumerCapacity = new AtomicLong();
 
         final Scheduler.Worker worker = scheduler.createWorker();
         child.add(worker);
--- a/src/test/java/rx/internal/util/IndexedRingBufferTest.java
+++ b/src/test/java/rx/internal/util/IndexedRingBufferTest.java
@@ -191,11 +191,8 @@ public class IndexedRingBufferTest {
             @Override
             public Boolean call(String t1) {
                 list.add(t1);
-                if (i++ == 2) {
-                    return false;
-                } else {
-                    return true;
-                }
+                i++;
+                return i != 3;
             }
 
         }, 0);
--- a/src/main/java/rx/internal/operators/BlockingOperatorNext.java
+++ b/src/main/java/rx/internal/operators/BlockingOperatorNext.java
@@ -79,11 +79,11 @@ public final class BlockingOperatorNext {
             }
             // Since an iterator should not be used in different thread,
             // so we do not need any synchronization.
-            if (hasNext == false) {
+            if (!hasNext) {
                 // the iterator has reached the end.
                 return false;
             }
-            if (isNextConsumed == false) {
+            if (!isNextConsumed) {
                 // next has not been used yet.
                 return true;
             }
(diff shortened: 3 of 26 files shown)


src/main/java/rx/internal/operators/OnSubscribeRedo.java
```diff
@@ -195,7 +195,7 @@ public void call(final Subscriber<? super T> child) {
         final AtomicBoolean resumeBoundary = new AtomicBoolean(true);
         
         // incremented when requests are made, decremented when requests are fulfilled
-        final AtomicLong consumerCapacity = new AtomicLong(0l);
+        final AtomicLong consumerCapacity = new AtomicLong(0L);
```
akarnokd: Unnecessary default value.

dromato: Fixed.

src/main/java/rx/internal/operators/OperatorZip.java
```diff
@@ -112,8 +112,7 @@ public OperatorZip(Func9 f) {
         final Zip<R> zipper = new Zip<R>(child, zipFunction);
         final ZipProducer<R> producer = new ZipProducer<R>(zipper);
         child.setProducer(producer);
-        final ZipSubscriber subscriber = new ZipSubscriber(child, zipper, producer);
-        return subscriber;
+        return new ZipSubscriber(child, zipper, producer);
```
akarnokd: This will conflict with a zip patch. Could you undo this?

dromato: Done :).

akarnokd:
Could you also squash the commits?

dromato:
Squashed :)

akarnokd:
:+1:

src/test/java/rx/internal/util/IndexedRingBufferTest.java
```diff
@@ -191,11 +191,7 @@ public Boolean call(String t1) {
             @Override
             public Boolean call(String t1) {
                 list.add(t1);
-                if (i++ == 2) {
-                    return false;
-                } else {
-                    return true;
-                }
+                return i++ != 2;
```
artem-zinnatullin: Programming on side effects :crying_cat_face:

artem-zinnatullin:
:+1:

src/test/java/rx/internal/util/IndexedRingBufferTest.java
dromato: Well I didn't want to change any existing logic there... 
I don't know code base well enough to feel comfortable enough.
I could take that "i++" one separate line higher.

artem-zinnatullin: >I could take that "i++" one separate line higher.

That's why I don't like such code, it's non-obvious (but short), in fact, "post `++`" will be applied after the condition check, so correct transformation would be:

```java
final boolean result = i != 2;
i++;
return result;
``` 

At least, we don't have undefined behavior here, so, you can leave it as is, but personally, I'd rewrite it.

dromato:
```java
i++;
return i != 3;
```

Should also be fine, right?
We increment both sides by 1.

artem-zinnatullin: Yep.

dromato: Changed.

artem-zinnatullin: Thanks!

dromato: My pleasure :) :smiley:

abersnaze:
looks good.

PR #3507, opened 2015-11-09
1.x: combineLatest can now combine arbitrary number of sources

akarnokd:
This PR is a backport of the 2.x `combineLatest` with `null` support and let's the programmer combine more than 128 sources at once. Note that 2.x supports manual sizing of the prefetch buffer and allows delaying the exceptions to the last moment; I kept these in case such features are required by 1.x some day.

The original had this 128 limit due to using a single and non-parametric RxRingBuffer to store values from sources. In addition, the original algorithm divided downstream requests among the sources with some strange logic in order to keep them running and not overflow the buffer.

This PR uses an unbounded-linked SPSC queue which can now hold as many elements as there are sources times the default ring buffer size (this is not preallocated and grows/shrinks as necessary).

The algorithm stores the current latest array and the source subscriber in the queue in pairs so when that particular row is emitted to downstream, that source subscriber gets the request() replenishment call that generated the row.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeCombineLatest.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeCombineLatest.java
@@ -1,317 +1,409 @@
 /**
- * Copyright 2014 Netflix, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not
- * use this file except in compliance with the License. You may obtain a copy of
- * the License at
- *
+ * Copyright 2015 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * 
  * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
- * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
- * License for the specific language governing permissions and limitations under
- * the License.
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
  */
+
 package rx.internal.operators;
 
-import java.util.BitSet;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.*;
+import java.util.concurrent.atomic.*;
 
+import rx.*;
 import rx.Observable;
 import rx.Observable.OnSubscribe;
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OnSubscribeCombineLatest.java
```diff
@@ -1,320 +1,413 @@
...
-        if (sources.size() == 1) {
-            child.setProducer(new SingleSourceProducer<T, R>(child, sources.get(0), combinator));
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public void call(Subscriber<? super R> s) {
+        Observable<? extends T>[] sources = this.sources;
+        int count = 0;
+        if (sources == null) {
+            if (sourcesIterable instanceof List) {
+                // unchecked & raw: javac type inference problem otherwise
+                sources = (Observable[])(((List)sourcesIterable).toArray(EMPTY));
```
artem-zinnatullin: Why not `toArray(new Observable[list.size()])`?

akarnokd:
@artem-zinnatullin  I didn't bother but now I'm more sensitive to Android reflection issues.

akarnokd:
Rebasing...

akarnokd:
... rebased. (Did I mention rebase doesn't really work with the latest Eclipse; there is no continue/abort menu option and the branch gets stuck and needs to be manually deleted as switching is also disabled).

artem-zinnatullin:
You can use IntelliJ IDEA community edition, it's free and awesome.

On Wed, Nov 11, 2015, 13:30 David Karnok <notifications@github.com> wrote:

> ... rebased. (Did I mention rebase doesn't really work with the latest
> Eclipse; there is no continue/abort menu option and the branch gets stuck
> and needs to be manually deleted as switching is also disabled).
>
> —
> <https://github.com/ReactiveX/RxJava/pull/3507#issuecomment-155729338>.
>
-- 

@artem_zin

akarnokd:
I've tried it a few times and it is too alien for me. I can't find features and its internal compiler plays tricks with me sometimes.

dsvoronin:
we need this one, pls merge

src/test/java/rx/internal/operators/OnSubscribeCombineLatestTest.java
```diff
@@ -884,4 +865,92 @@ public void onNext(Integer t) {
         assertTrue(latch.await(10, TimeUnit.SECONDS));
     }
 
-}
+    @Test
+    public void testCombineMany() {
+        int n = RxRingBuffer.SIZE * 3;
+        
+        Observable<Integer> source = Observable.just(1);
```
artem-zinnatullin: @akarnokd this may play very bad "joke" with us because operator just needs to combine any result (first, second, etc) correct amount of times to pass the test! 

Can you please create unique observables as a source? Probably even better with `String`s to prevent effects from mathematical operations (equality of sum without fixed order in this case)

artem-zinnatullin:
:+1: from me (test needs a fix), @zsxwing can you please take a look?

akarnokd:
Closing out, some unknown merge conflicts appeared, maybe due to age.

akarnokd:
See #3661

PR #3509, opened 2015-11-09
1.x: split Android detection and version detection

akarnokd:
It seems some Samsung devices running 5.x mess with field names so
reflection can't find them.

If the `PlatformDependent.resolveAndroidApiVersion` logic throws due to
the missing field (but not the class), the original code considered it
to be a non-Android platform and went ahead with `Unsafe` stuff which
causes NPEs at many places.

The change splits the class check and the version check. If the version
check fails, it returns a default version (7) value so the `Unsafe`
paths are not triggered.

### DIFF
--- a/src/main/java/rx/internal/util/PlatformDependent.java
+++ b/src/main/java/rx/internal/util/PlatformDependent.java
@@ -34,6 +34,8 @@ public final class PlatformDependent {
 
     private static final boolean IS_ANDROID = ANDROID_API_VERSION != ANDROID_API_VERSION_IS_NOT_ANDROID;
 
+    /** If Android detected but couldn't get the version number. */
+    private static final int ANDROID_API_VERSION_MIN = 7;
     /**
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
If Samsung runtime renames fields in all classes (even `android.os.Build`)  it would break way more libraries that use reflection on fields, especially JSON parsers/serializers and we will have tons of crashes on such devices. 

I mean, this workaround will probably never work.

cc other Android devs, maybe I'm missing something, but I don't think that this PR will fix Samsung problem.

JakeWharton:
Yes. This class and field can never be renamed or the entire phone would fail to work.

akarnokd:
Is there a chance Samsung uses some whitelist that doesn't touch most common libraries but messes with others?

JakeWharton:
Highly unlikely. It seems like a simple reflection bug as people have
confirmed the field is named correctly at runtime.

Issue #3508, opened 2015-11-09
Null pointer using ConcatMapEager on Android 4.2


jjvargas:
I am using a Tablet with Android 4.2  using 'io.reactivex:rxjava:1.0.15'

In Android 5.1 it works fine, but not it 4.2

    public Observable<Artist[]> batchGetArtistsById(ArtistIdList artistIdList) {
            return Observable.from(artistIdList.getArtistIds())
                    .concatMapEager(artist -> getArtistById(artist))
                    .subscribeOn(Schedulers.io())
                    .toList()
                    .observeOn(AndroidSchedulers.mainThread())
                    .doOnError(e ->
                            {
                                System.out.println("Error what to do? " + e);
                            }
                    )
                    .flatMap(artistList -> {
                        Artist[] artistsArray = artistList.toArray(new Artist[artistList.size()]);
                        return Observable.just(artistsArray);
                    });
    }

11-09 16:23:01.485 6762-6782/? E/AndroidRuntime: FATAL EXCEPTION: RxCachedThreadScheduler-1
                                                 java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.Worker thread.
                                                     at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:62)
                                                     at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:390)
                                                     at java.util.concurrent.FutureTask.run(FutureTask.java:234)
                                                     at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:153)
                                                     at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:267)
                                                     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
                                                     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
                                                     at java.lang.Thread.run(Thread.java:856)
                                                  Caused by: java.lang.ExceptionInInitializerError
                                                     at rx.internal.operators.OperatorEagerConcatMap$EagerInnerSubscriber.<init>(OperatorEagerConcatMap.java:281)
                                                     at rx.internal.operators.OperatorEagerConcatMap$EagerOuterSubscriber.onNext(OperatorEagerConcatMap.java:130)
[...]

PR #3510, opened 2015-11-09
1.x: eager concatMap to choose safe or unsafe queue based on platform.

akarnokd:
I forgot to add the choice because 2.x `SpscArrayQueue` doesn't use Unsafe. I copied the `SpscAtomicArrayQueue` from #3169 and I hope it won't conflict. This is the drawback of long trail of unmerged fixes and enhancements...

Reported in #3508 .


stealthcode:
:+1:

Issue #3059, opened 2015-07-01
SafeSubscriber does not call `onCompleted` after `unsubscribe`, allows further calls to `onNext`


dlew:
[The documentation for `SafeSubscriber`](http://reactivex.io/RxJava/javadoc/rx/observers/SafeSubscriber.html) states:

> If `unsubscribe` is called, calls `onCompleted` and forbids any further `onNext` calls.

However, in practice, it appears that `SafeSubscriber` does neither of those.

------

The following code should call `onCompleted`:

```java
Subscription sub = Observable.interval(10, TimeUnit.MILLISECONDS)
  .subscribe(
    l -> System.out.println("Received: " + l),
    e -> System.out.println("Error: " + e),
    () -> System.out.println("Completed")
  );

sleep(50);
sub.unsubscribe();
```

However, the output is as follows:

```
Received: 0
Received: 1
Received: 2
Received: 3
Received: 4
```

------

The following code should not call `onNext` after unsubscription:

```java
Observable<Long> observable = Observable.create(subscriber -> {
  long n = 0;
  while (true) {
    sleep(10);
    subscriber.onNext(n++);
  }
});

Subscription sub = observable
  .subscribeOn(Schedulers.newThread())
  .subscribe(
    l -> System.out.println("Received: " + l),
    e -> System.out.println("Error: " + e),
    () -> System.out.println("Completed")
  );
sleep(50);
sub.unsubscribe();
System.out.println("Unsubscribed!");
```

However, this is the output:

```
Received: 0
Received: 1
Received: 2
Received: 3
Received: 4
Unsubscribed!
Received: 5
Received: 6
Received: 7
Received: 8
...
```

-----

It seems to me that one of three things is going on:

1. `SafeSubscriber` isn't as safe as it should be.
1. The documentation is out-of-date.
1. I'm missing something.

davidmoten:
>If unsubscribe is called, calls onCompleted...

* `onCompleted` should only be called after all the `onNext` notifications have arrived so the behaviour described in the javadoc does not comply with the Observable contract. We should remove this. 

>and forbids any further onNext calls.

[...]

PR #3512, opened 2015-11-09
1.x: fix SafeSubscriber documentation regarding unsubscribe

akarnokd:
The documentation was wrong in two points: unsubscription doesn't call` onCompleted` and unsubscription doesn't directly prevent delivery of `onXXX` events since the implementation doesn't even check isUnsubscribed: (it is the responsibility of the upstream to do that).

Related issue: #3059.


stealthcode:
:+1:

zsxwing:
:+1:

Issue #2835, opened 2015-03-18
Cannot define plugin properties on Android


rosshambrick:
The property key max length on Android devices/emulators is 31 characters thereby preventing the ability to set properties for use by the plug in system.  

The property key template is 29 characters on its own.
`rxjava.plugin.[class-name-here].implementation`

Is it likely that a shorter key could be used to enable this feature on Android?

akarnokd:
I'm not familiar with the constraints on Android. How about we give you the following system property that wraps other system properties:

```
rxjava.props.1.key=rxjava.plugin.[class-name-here].implementation
rxjava.props.1.value=come.somepath.SomeClass
```

i.e., is it possible the use System.setProperty(key, value) on Android with long keys? Does the 31 character limit apply to command-line only?

benjchristensen:
I have to say, that is a ridiculous limitation.

We can't change existing keys, but if someone wants to submit a PR with aliases that set both, that could work.

akarnokd:
I've posted PR #3513 with a shorter scheme proposal.

akarnokd:
Solution merged and delivered in 1.1.0

PR #3513, opened 2015-11-09
1.x: add shorter RxJavaPlugin class lookup approach.

akarnokd:
This adds a new pattern and lookup method that let's the developer specify the custom plugins with shorter system property keys. Android is quite restrictive and allows only 31 characters.

The new pattern splits the target simple class name and its implementation into two separate system properties:

```
rxjava.plugin.1.class=SimpleClassName
rxjava.plugin.1.impl=path.to.impl.Class
```

The index tag (`1`) can be any string of your chosing:

```
rxjava.plugin.mykey.class=SimpleClassName
rxjava.plugin.mykey.impl=path.to.impl.Class
```

but make sure they are paired, otherwise nothing will happen.

If there are multiple `class` entries with the same `SimpleClassName` one of them will be chosen (depending on the walk order in `java.util.Properties`).

Related issue #2835.

### DIFF
--- a/src/main/java/rx/plugins/RxJavaPlugins.java
+++ b/src/main/java/rx/plugins/RxJavaPlugins.java
@@ -15,6 +15,7 @@
  */
 package rx.plugins;
 
+import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -26,7 +27,22 @@ import java.util.concurrent.atomic.AtomicReference;
  * property names)</li>
  * <li>default implementation</li>
  * </ol>
- *
+ * <p>In addition to the {@code rxjava.plugin.[simple classname].implementation} system properties,
+ * you can define two system property:<br>
+ * <pre><code>
+ * rxjava.plugin.[index].class}
+ * rxjava.plugin.[index].impl}
+ * </code></pre>
+ * 
+ * Where the {@code .class} property contains the simple classname from above and the {@code .impl}
+ * contains the fully qualified name of the implementation class. The {@code [index]} can be
+ * any short string or number of your chosing. For example, you can now define a custom 
+ * {@code RxJavaErrorHandler} via two system property:
+ * <pre><code>
+ * rxjava.plugin.1.class=RxJavaErrorHandler
+ * rxjava.plugin.1.impl=some.package.MyRxJavaErrorHandler
+ * </code></pre>
+ * 
  * @see <a href="https://github.com/ReactiveX/RxJava/wiki/Plugins">RxJava Wiki: Plugins</a>
  */
 public class RxJavaPlugins {
@@ -64,13 +80,12 @@ public class RxJavaPlugins {
      * <p>
      * Override the default by calling {@link #registerErrorHandler(RxJavaErrorHandler)} or by setting the
      * property {@code rxjava.plugin.RxJavaErrorHandler.implementation} with the full classname to load.
-     * 
      * @return {@link RxJavaErrorHandler} implementation to use
      */
     public RxJavaErrorHandler getErrorHandler() {
         if (errorHandler.get() == null) {
             // check for an implementation from System.getProperty first
-            Object impl = getPluginImplementationViaProperty(RxJavaErrorHandler.class);
+            Object impl = getPluginImplementationViaProperty(RxJavaErrorHandler.class, System.getProperties());
             if (impl == null) {
                 // nothing set via properties so initialize with default 
                 errorHandler.compareAndSet(null, DEFAULT_ERROR_HANDLER);
@@ -112,7 +127,7 @@ public class RxJavaPlugins {
     public RxJavaObservableExecutionHook getObservableExecutionHook() {
         if (observableExecutionHook.get() == null) {
             // check for an implementation from System.getProperty first
-            Object impl = getPluginImplementationViaProperty(RxJavaObservableExecutionHook.class);
+            Object impl = getPluginImplementationViaProperty(RxJavaObservableExecutionHook.class, System.getProperties());
             if (impl == null) {
                 // nothing set via properties so initialize with default 
                 observableExecutionHook.compareAndSet(null, RxJavaObservableExecutionHookDefault.getInstance());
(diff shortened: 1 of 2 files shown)


stealthcode:
:+1:

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -141,7 +156,7 @@ public void registerObservableExecutionHook(RxJavaObservableExecutionHook impl)
         }
     }
 
-    private static Object getPluginImplementationViaProperty(Class<?> pluginClass) {
+    /* test */ static Object getPluginImplementationViaProperty(Class<?> pluginClass, Properties props) {
```
artem-zinnatullin: Can you change `test` to `VisibleForTesting`? It's pretty common pattern.

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -149,7 +164,27 @@ private static Object getPluginImplementationViaProperty(Class<?> pluginClass) {
          * This will only happen during system startup thus it's okay to use the synchronized
          * System.getProperties as it will never get called in normal operations.
          */
-        String implementingClass = System.getProperty("rxjava.plugin." + classSimpleName + ".implementation");
+        
+        String defaultKey = "rxjava.plugin." + classSimpleName + ".implementation";
```
artem-zinnatullin: Let's declare it as `private static final String`?

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -149,7 +164,27 @@ private static Object getPluginImplementationViaProperty(Class<?> pluginClass) {
...
+        String implementingClass = props.getProperty(defaultKey);
+        
+        for (Map.Entry<Object, Object> e : props.entrySet()) {
+            String key = e.getKey().toString();
+            if (key.startsWith("rxjava.plugin.") && key.endsWith(".class")) {
```
artem-zinnatullin: Here you can use declared string constant

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -149,7 +164,27 @@ private static Object getPluginImplementationViaProperty(Class<?> pluginClass) {
...
+                String value = e.getValue().toString();
+                
+                if (classSimpleName.equals(value)) {
+                    String index = key.substring(0, key.length() - 6).substring(14);
```
artem-zinnatullin: And here you can replace magic numbers with `constant.lenght()` (I hope compiler will replace it with value at compile time).

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -149,7 +164,27 @@ private static Object getPluginImplementationViaProperty(Class<?> pluginClass) {
...
+                    
+                    String implKey = "rxjava.plugin." + index + ".impl";
+                    
+                    implementingClass = props.getProperty(implKey);
```
artem-zinnatullin: What if user had declared the key, but hadn't declared the value or made an error in the value property name? We can be more restrictive and throw an exception here if key property was declared but value wasn't found. It'll allow detect and prevent errors earlier!

src/main/java/rx/plugins/RxJavaPlugins.java
akarnokd: I couldn't find this `VisibleForTesting` in RxJava.

src/main/java/rx/plugins/RxJavaPlugins.java
akarnokd: You mean the constant? They are already in the constant pool of the class file.

artem-zinnatullin: Yes, of course, it's a constant, but I mean explicitly declare it as constant :)

akarnokd:
Updated: made strings local constants and using `length`˙instead of magic numbers. Plus, if the default keying matches, the properties is not traversed for an alternate specification.

src/main/java/rx/plugins/RxJavaPlugins.java
akarnokd: I don't know, crashing while the app is initializing sounds a bit harsh to me. I'm sure the developer who specified the override will notice his override doesn't work.

akarnokd:
An alternative pattern could be this:

```
rxjava.plugin.1=SimpleClassName,org.package.naming.ActualClass
```

src/main/java/rx/plugins/RxJavaPlugins.java
artem-zinnatullin: Initialization is the best time to crash the app because it has almost no state at this point of time and developer could detect problem immediately.

> I'm sure the developer who specified the override will notice his override doesn't work.

But how he will notice if usually developers override schedulers or error handlers? :)

Fail fast, fail early!

src/main/java/rx/plugins/RxJavaPlugins.java
zsxwing: `VisibleForTesting` is in Guava

stealthcode:
I agree that if the plugin cannot be found we should crash the app.

src/main/java/rx/plugins/RxJavaPlugins.java
artem-zinnatullin: I meant not annotation but comment.

On Wed, Nov 11, 2015, 21:32 Shixiong Zhu <notifications@github.com> wrote:

> In src/main/java/rx/plugins/RxJavaPlugins.java
> <https://github.com/ReactiveX/RxJava/pull/3513#discussion_r44567271>:
>
> > @@ -141,7 +156,7 @@ public void registerObservableExecutionHook(RxJavaObservableExecutionHook impl)
> >          }
> >      }
> >
> > -    private static Object getPluginImplementationViaProperty(Class<?> pluginClass) {
> > +    /* test */ static Object getPluginImplementationViaProperty(Class<?> pluginClass, Properties props) {
>
> VisibleForTesting is in Guava
>
> —
> <https://github.com/ReactiveX/RxJava/pull/3513/files#r44567271>.
>
-- 

@artem_zin

akarnokd:
Updated the code to crash if the plugin implementation is missing.

stealthcode:
Thanks!

PR #3514, opened 2015-11-10
1.x: SyncOnSubscribeTest.testConcurrentRequests give more time.

akarnokd:
The test failed on Travis and locally if my machine was under heavy load without interacting with the mock. This change gives more time in the inner await and reports the exception instead of itself throwing.

### DIFF
--- a/src/test/java/rx/observables/SyncOnSubscribeTest.java
+++ b/src/test/java/rx/observables/SyncOnSubscribeTest.java
@@ -16,57 +16,25 @@
 
 package rx.observables;
 
-import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


stealthcode:
Okay :+1: do you think it'd be a good idea to change this to use the test scheduler?

akarnokd:
The request has to happen between while the execution is inside the function. Perhaps if `TestSubscriber` is created before the `Func2`, the body can reference it and issue a `requestMore` synchronously.

artem-zinnatullin:
:+1:

akarnokd:
I'm merging this so PRs won't break on it.

PR #3515, opened 2015-11-11
1.x: BlockingUtils test: clear interrupted flag before/after

akarnokd:
For some strange reason, the interrupted flag is sometimes still set when the next JUnit test method runs and `await` will throw immediately.


artem-zinnatullin:
:+1:

zsxwing:
:+1:

PR #3410, opened 2015-10-04
2.x: null safety checks and fixes

akarnokd:
This PR adds over 900 new null checks. Null checks are essential because RS doesn't allow null values and sometimes a missed null check may lead to operator hangs as well.

PR #3526, opened 2015-11-14
2.x: Reduse verbosity with varargs [WIP]

alexVengrovsk:


akarnokd:
Hi and thanks for the PR, but unfortunately, the Java 6 target makes varargs inconvenient. This is why we have that many overloads for the most common operators so small usages, 2 - 9 data sources, don't force the caller to add `@SuppressWarnings`.

alexVengrovsk:
@akarnokd Thank you for so fast reply. But I saw java8's constructions in the class, I changed.

akarnokd:
Oh sorry, didn't look at the targed branch. Indeed, this shouldn't be much of a problem with Java 7+ if you add the `@SafeVarargs`, however, javac sometimes finds ambiguity with such methods for some reason whereas Eclipse's compiler is fine. This is the main reason I've named the same method you are proposing `concatArray`. So could you do the following?

  - rename the method to `concatArray`,
  - add `@SafeVarargs` to it, and
  - squash the commits.

In addition, note that since this will conflict with #3410, it is likely you'll have to rebase/redo this work.

alexVengrovsk:
Ok. I will do it. For a while this PR will have [WIP] prefix.

akarnokd:
Thanks for your contribution. This PR is quite old, plus, the decision has been made 2.x has to be Java 6 where suppressing varargs is still a problem.

PR #3528, opened 2015-11-16
1.x: Avoid to call next when Iterator is drained

zsxwing:
`delay` delays the `onCompleted` event and `unsubscribe` is called when `onCompleted` finishes. So if we put a `o.zipWith(Iterable, func)` before `delay`, such as `o.zipWith(Iterable, func).delay(...)`, `o` may keep emitting items even if `zipWith` emits an `onCompleted` event. This PR just fixed `OperatorZipIterable` to handle this case.

Fixes https://github.com/ReactiveX/RxScala/issues/180

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorZipIterableTest.java
+++ b/src/test/java/rx/internal/operators/OperatorZipIterableTest.java
@@ -23,6 +23,7 @@ import static org.mockito.Mockito.verify;
 
 import java.util.Arrays;
 import java.util.Iterator;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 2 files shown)


src/test/java/rx/internal/operators/OperatorZipIterableTest.java
```diff
@@ -378,4 +381,22 @@ public String call(Integer t1) {
...
+        Observable<Integer> o = Observable.just(1, 2).zipWith(Arrays.asList(1), new Func2<Integer, Integer, Integer>() {
+            @Override
+            public Integer call(Integer v1, Integer v2) {
+                return v1;
+            }
+        }).delay(500, TimeUnit.MILLISECONDS, scheduler);
+
+        TestSubscriber<Integer> subscriber = new TestSubscriber<Integer>();
+        o.subscribe(subscriber);
+        scheduler.advanceTimeBy(1000, TimeUnit.DAYS.MILLISECONDS);
```
akarnokd: `DAYS` is unnecessary here.

zsxwing: Oops. Fixed.

akarnokd:
:+1:

Issue #3527, opened 2015-11-15
synchronous sample with time


davidmoten:
Just read this 

http://stackoverflow.com/questions/33693958/rxjava-sample-is-cpu-intensive

and it struck me that we should be able to use `sample(long, TimeUnit, scheduler)` with a synchronous scheduler like `Schedulers.immediate`. At the moment this doesn't work (the stream hangs because a worker for a synchronous scheduler has been scheduled periodically and the operator never returns the upstream subscriber). 

I'd like to get this working and in `OperatorSampleWithTime` thought to check the `scheduler` parameter to see if it is an instance of `ImmediateScheduler` or `TrampolineScheduler` and offer a path that just accepts emissions and forwards them if the specified interval has passed since the last emission. We would still hang if provided a custom synchronous `Scheduler`  but I doubt this is a use case to be expected (?).

Before I knock up a PR are people happy with this approach?

akarnokd:
I find `Schedulers.immediate` nothing but trouble and a relic from Rx.NET. Making `sample` synchronous doesn't make sense as it will block its incoming source for the waiting duration.

davidmoten:
>Making sample synchronous doesn't make sense as it will block its incoming source for the waiting duration.

I'm suggesting not using a scheduler at all when the passed scheduler is synchronous.

akarnokd:
You mean by checking `System.currentTimeMillis()` and dropping values that arrived within the timeframe?

davidmoten:
yep

davidmoten:
The main difference between sampling synchronously and asynchronously is this:

* Async sample reports the last emission from each time frame and reports that emission very close to the frame end regardless of when in the frame the emission occurred
* Sync sample reports the first emission from each time frame and reports it on arrival (no wait). This means that it's possible for an emission to be emitted half way through the frame if the source is quietish.

For a fast emitting source this really won't make much difference and I wonder why `sample` is by default asynchronous (with it's associated overhead including relying on a possibly busy scheduler) instead of the simpler synchronous option. 

For my purposes I use a synchronous version (not part of the API).

Can people mention here what use cases they have that are best served by the async version? (I'm wondering if sync version should be default in v1/v2).

akarnokd:
The problem with the synchronous version that it changes the emission pattern from regular to irregular since you can only emit values if the upstream emits a value. There is also the problem what value to emit once it has been determined the time window has elapsed (the last or the current).

abersnaze:
[...]

PR #3529, opened 2015-11-17
1.x: fix time windows in throttleFirst

davidmoten:
As a result of #3527 discussion I noticed that `throttleFirst` does not comply with behaviour described in javadocs. Time windows used for emissions were supposed to be of constant duration yet could be effectively lengthened because the next time window always started at the time of the emission to downstream plus the window duration.

To demonstrate:

Consider  `source.throttleWithFirst(1000, TimeUnit.MILLISECONDS)` where source emits 

```
Time   Value    
0       "a"
1200    "b"
2100    "c"
```

Desired result is:

`a, b, c`

yet previous version produced

`a, b`

A unit test has been added that failed with the previous version.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -8789,6 +8789,9 @@ public class Observable<T> {
      * @see <a href="https://github.com/ReactiveX/RxJava/wiki/Backpressure">RxJava wiki: Backpressure</a>
      */
     public final Observable<T> throttleFirst(long skipDuration, TimeUnit unit, Scheduler scheduler) {
+        if (skipDuration <= 0) {
+            throw new IllegalArgumentException("skipDuration must be greater than zero");
+        }
         return lift(new OperatorThrottleFirst<T>(skipDuration, unit, scheduler));
     }
 
--- a/src/main/java/rx/internal/operators/OperatorThrottleFirst.java
+++ b/src/main/java/rx/internal/operators/OperatorThrottleFirst.java
@@ -25,11 +25,13 @@ import rx.Observable.Operator;
  */
 public final class OperatorThrottleFirst<T> implements Operator<T, T> {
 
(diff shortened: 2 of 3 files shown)


akarnokd:
In 2.x, this operator uses a scheduler, blocks emissions in the given time window and re-enables it once the time runs out. The window is emission relative and not periodic. The javadoc reads "passage of time" which could mean emission-relative blocking. @zsxwing what was the original intention? Is this implemented somewhere else?

davidmoten:
You are right that there is ambiguity in the javadoc:

>Returns an Observable that emits only the first item emitted by the source Observable during sequential time windows of a specified duration.

and 

>@ param windowDuration  time to wait before emitting another item after emitting the last item

zsxwing:
ping @headinthebox @benjchristensen for you comments on the semantics of `throttleFirst`.

zsxwing:
Just found this: http://www.introtorx.com/content/v1.0.10621.0/13_TimeShiftedSequences.html#Throttle

> Unlike Sample though, Throttle's period is a sliding window. Each time Throttle receives a value, the window is reset.

zsxwing:
So I think we only need to fix the javadoc.

davidmoten:
@zsxwing thanks for finding that. That definition seems as good as any and I can accept that one. I'll see what I can do for the javadoc on throttle/sample in another PR to make this difference clear and consistent. 

Given that the operator as I've rewritten it in this PR has the semantics of `sample` perhaps it could be a useful contribution as `sampleFirst`?

akarnokd:
My understanding is that this operator has the purpose of blocking out subsequent onNext events for a specified duration. If an event comes 1.5x the duration after the last, that event can be immediately emitted and a new window starts.

image

Do you want to pursue this further?

akarnokd:
I'm closing this for now. If you want, you can post a PR with the parameter validation.

PR #3536, opened 2015-11-22
1.x: Fix SyncOnSubscribeTest.testConcurrentRequests non-determinism

akarnokd:
The test checks if onUnSubscribe is called but that happens after
onCompleted is sent and as such, may run concurrently with the main
thread where the mock is verified. The change switches to CountDownLatch
to properly await the call to onUnsubscribe.

### DIFF
--- a/src/test/java/rx/observables/SyncOnSubscribeTest.java
+++ b/src/test/java/rx/observables/SyncOnSubscribeTest.java
@@ -476,8 +476,14 @@ public class SyncOnSubscribeTest {
         final CountDownLatch l1 = new CountDownLatch(1);
         final CountDownLatch l2 = new CountDownLatch(1);
         
-        @SuppressWarnings("unchecked")
-        Action1<? super Integer> onUnSubscribe = mock(Action1.class);
+        final CountDownLatch l3 = new CountDownLatch(1);
+
+        final Action1<Object> onUnSubscribe = new Action1<Object>() {
+            @Override
+            public void call(Object t) {
+                l3.countDown();
+            }
+        };
         
         OnSubscribe<Integer> os = SyncOnSubscribe.createStateful(
                 new Func0<Integer>() {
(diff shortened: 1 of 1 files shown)


src/test/java/rx/observables/SyncOnSubscribeTest.java
```diff
@@ -532,7 +538,10 @@ public Integer call(Integer state, Observer<? super Integer> observer) {
         inOrder.verify(o, times(finalCount)).onNext(any());
         inOrder.verify(o, times(1)).onCompleted();
         inOrder.verifyNoMoreInteractions();
-        verify(onUnSubscribe, times(1)).call(any(Integer.class));
+        
+        if (!l3.await(2, TimeUnit.SECONDS)) {
```
artem-zinnatullin: Static import for `SECONDS` will make code little nicer!

src/test/java/rx/observables/SyncOnSubscribeTest.java
```diff
@@ -476,8 +476,14 @@ public void testConcurrentRequests() throws InterruptedException {
...
         final CountDownLatch l2 = new CountDownLatch(1);
         
-        @SuppressWarnings("unchecked")
-        Action1<? super Integer> onUnSubscribe = mock(Action1.class);
+        final CountDownLatch l3 = new CountDownLatch(1);
+
+        final Action1<Object> onUnSubscribe = new Action1<Object>() {
+            @Override
+            public void call(Object t) {
+                l3.countDown();
```
artem-zinnatullin: Previous code checked that `onUnsubscribe` action was called only once, this one guarantees that action will be called at least once. Is this okay or you also want to check that action was called only once?

akarnokd: Detecting a second call is inherently racy with the checking of the first so this async test can't do both.

artem-zinnatullin: okay.

artem-zinnatullin:
:+1:

stealthcode:
:+1:

PR #3539, opened 2015-11-24
Add Single.zip() for Iterable of Singles

artem-zinnatullin:


src/main/java/rx/Single.java
```diff
@@ -1204,6 +1207,34 @@ public void onError(Throwable error) {
...
+     * @param singles
+     *            an Iterable of source Singles
+     * @param zipFunction
+     *            a function that, when applied to an item emitted by each of the source Singles, results in
+     *            an item that will be emitted by the resulting Single
+     * @return a Single that emits the zipped results
+     * @see <a href="http://reactivex.io/documentation/operators/zip.html">ReactiveX operators documentation: Zip</a>
+     */
+    public static <R> Single<R> zip(Iterable<? extends Single<?>> singles, FuncN<? extends R> zipFunction) {
+        List<Observable<?>> listOfSingles = new ArrayList<Observable<?>>();
```
akarnokd: Although not required, I prefer iterating such sources when the child subscribes and not when the sequence is assembled.

akarnokd:
Did you consider adding `zip(Single[], FuncN)` overload as well?

Otherwise, looks good. :+1:

src/main/java/rx/Single.java
artem-zinnatullin: Hm, I am afraid that iterating over sequence when child subscribes may lead to data inconsistency if user will pass the collection of Singles and then change it before `subscribe()`.

artem-zinnatullin:
I'll add `zin(Single[], FuncN)` and `zipWith()` in separate PRs if you don't mind.

Btw, idk if @benjchristensen remembers how I was arguing against adding `Single` to the RxJava. My apologies 😅. I'm starting loving `Single` in our codebase!

akarnokd:
Here is an implementation that avoids conversion to `Observable`:
```java
public static <T, R> Single<R> zip(
        Iterable<? extends Single<? extends T>> singles, 
                FuncN<? extends R> zipper) {
    return Single.create(f -> {
        Single<? extends T>[] sa;
        int count;
        if (singles instanceof List) {
            List<? extends Single<? extends T>> list = 
                    (List<? extends Single<? extends T>>)singles;
            count = list.size();
            sa = list.toArray(new Single[count]);
        } else {
            sa = new Single[8];
            count = 0;
            for (Single<? extends T> s : singles) {
                if (count == sa.length) {
                    Single<? extends T>[] sb = new Single[count + (count >> 2)];
                    System.arraycopy(sa, 0, sb, 0, count);
                    sa = sb;
                }
                sa[count] = s;
                count++;
            }
        }
        
        final AtomicInteger wip = new AtomicInteger(count);
        final AtomicBoolean once = new AtomicBoolean();
        final Object[] values = new Object[count];
        
        CompositeSubscription csub = new CompositeSubscription();
        f.add(csub);
        
        for (int i = 0; i < count; i++) {
            if (csub.isUnsubscribed() || once.get()) {
                break;
            }
            
            final int j = i;
            SingleSubscriber<T> te = new SingleSubscriber<T>() {
                @Override
                    values[j] = value;
                    if (wip.decrementAndGet() == 0) {
                        R r;
                        
                        try {
                            r = zipper.call(values);
                        } catch (Throwable e) {
                            Exceptions.throwIfFatal(e);
                            f.onError(e);
                            return;
                        }
                        
                        f.onSuccess(r);
                    }
                }

                @Override
                        f.onError(error);
                    } else {
                        RxJavaPlugins.getInstance()
                        .getErrorHandler().handleError(error);
                    }
                }
            };
            csub.add(te);
            if (csub.isUnsubscribed() || once.get()) {
                break;
            }
            sa[i].subscribe(te);
        }
    });
}
```

artem-zinnatullin:
@akarnokd I've used your code (rewrote some parts and fixed one bug) and now we don't have to convert `Single`s to `Observable`s! 

Later we can switch other `Single.zip` overloads to `SingleOperatorZip`.

src/main/java/rx/internal/operators/SingleOperatorZip.java
```diff
@@ -0,0 +1,72 @@
...
+                        public void onSuccess(T value) {
+                            values[j] = value;
+                            if (wip.decrementAndGet() == 0) {
+                                R r;
+
+                                try {
+                                    r = zipper.call(values);
+                                } catch (Throwable e) {
+                                    Exceptions.throwIfFatal(e);
+                                    subscriber.onError(e);
```
akarnokd: This should  be just this.onError to avoid error races.

src/main/java/rx/Single.java
```diff
@@ -1997,4 +2025,46 @@ public void call(SingleSubscriber<? super T> singleSubscriber) {
...
+                }
+                tempArray[count] = s;
+                count++;
+            }
+
+            if (tempArray.length == count) {
+                singlesArray = tempArray;
+            } else {
+                singlesArray = new Single[count];
+                System.arraycopy(tempArray, 0, singlesArray, 0, count);
```
akarnokd: This copy can be avoided with my inlined version.

src/main/java/rx/internal/operators/SingleOperatorZip.java
artem-zinnatullin: fixed

src/main/java/rx/Single.java
artem-zinnatullin: Nope! It's a bugfix. Otherwise, you can have an array with `length != count` and have a `null` pointer exception in a loop (for example in  `SingleOperatorZip`). I faced it when I tried to pass `Set` as `Iterable` in one of the tests.

akarnokd: This is why mine run up to count and not array length. If you factor it out, you can't return both the array and the count.

artem-zinnatullin: Oh, got it, so it was my bug when I split your code into two parts, sorry. 

I can return `Map.Entry` from `iterableToArray` where one value will be the array and another — real `count`, but personally I'd prefer current variant.

If you afraid about performance — we can add more general check `instanceof Collection` and initialize the array with correct size even if `Iterable` is not a `List`. For real, I don't even remember when I used something that does not implement `Collection` but implements `Iterable`, I guess will cover a lot of cases with such check: `List`, `Set`, `Queue` and so on.

artem-zinnatullin:
@akarnokd changed `instanceof List` to `instanceof Collection` so we are now really efficient for all collections and fall back to slightly less fast code for other `Iterable`s, rebased. 

PTAL

src/main/java/rx/Single.java
```diff
@@ -1205,6 +1208,31 @@ public void onError(Throwable error) {
...
+     *            an Iterable of source Singles
+     * @param zipFunction
+     *            a function that, when applied to an item emitted by each of the source Singles, results in
+     *            an item that will be emitted by the resulting Single
+     * @return a Single that emits the zipped results
+     * @see <a href="http://reactivex.io/documentation/operators/zip.html">ReactiveX operators documentation: Zip</a>
+     */
+    @SuppressWarnings("unchecked")
+    public static <R> Single<R> zip(Iterable<? extends Single<?>> singles, FuncN<? extends R> zipFunction) {
+        return SingleOperatorZip.zip(iterableToArray(singles), zipFunction);
```
akarnokd: Iterables and Observables are mainly for deferred data streaming. This will iterate at assembly time instead on Subscription time, losing the duality aspects. For example, an Observable source turned into an Iterable via toBlocking().getIterable() is no longer lazy when this zip is applied.

karlicoss: I think it is compatible with the behavior of `Observable.zip`.

stealthcode: [Yep] (https://github.com/ReactiveX/RxJava/blob/89b2e9fc230775e39362248653eaca66a408d2b3/src/main/java/rx/Observable.java#L2835) currently Observable serializes all values in the iterable into an array before zipping them so you can't append after lifting. IMO this is probably intentional since appending to some iterables while iterating results in ConcurrentModificationExceptions. Today the `Observable<Observable<T>>` overload can receive new observables to zip but the zip function won't be called until the outer observable calls `onCompleted()`. This is slightly later than the Iterable overload.

artem-zinnatullin: So, @akarnokd, @stealthcode let's use current implementation from this PR to be consistent with `Observable.zip(Iterable<Observable>)`?

stealthcode: Sounds good to me.

akarnokd: Okay.

artem-zinnatullin:
Rebased!

artem-zinnatullin:
Blocked by #3569.

akarnokd:
This needs to be rebased again.

akarnokd:
:+1:

stealthcode:
:+1:

stevegury:
:+1: (for posterity)

Issue #3542, opened 2015-11-27
Scheduling should serialize notifications.


Dorus:
Currently, `onError` cuts ahead of `onNext` calls when you schedule work, but does not (cannot) when you use a single thread.

	Observable.create(o -> {
		o.onNext(1);
		o.onError(new NotImplementedException());
	})
	.observeOn(Schedulers.newThread())
	.subscribe(i -> System.out.println(i), 
			e -> System.out.println("error:" + e));

This yields `error` instead of `1, error` as expected. Other implementations of Rx like Rx.Net do serialize notifications. This 'cut ahead' behavior is actually described as a [common pitfall](https://realm.io/news/gotocph-mattias-kappler-reactive-architecture-android/) in RxJava by some.

akarnokd:
RxJava 2.0 will allow you to customize this behavior in many operators, including `observeOn`, `delay`, `combineLatest`, `zip`, `groupBy`, `onBackpressureBuffer`, `skipLast` and `takeLast`.

I was planning to post a PR targeting `observeOn` to remove some unnecessary code (#3002) and may well do this as well.

Dorus:
This is really scary nondeterministic behavior, and against Rx contract.I hope the (future) default will be to not lose data. If anything it might be an appropriate setting on the subscriber.

akarnokd:
The `observeOn` operator received an overload that takes a `delayError` parameter you can set.

Closing via #3682.

Issue #3002, opened 2015-06-02
ObserveOn: Unsubscribe of RecursiveScheduler


benjchristensen:
Need to explore the unsubscribe of `recursiveScheduler`: https://github.com/ReactiveX/RxJava/blob/34dce48b4d591a2c574839b4527cbc30cb69e5a2/src/main/java/rx/internal/operators/OperatorObserveOn.java#L122

It seems to wrap with `child.add(scheduledUnsubscribe)` and then defeat that with `child.add(recursiveScheduler)`.

akarnokd:
Yes, looks odd. I can't really remember why the unsubscription of the worker had to be scheduled. Maybe it had something to do with pitfall no. 2: unsubscribing the downstream: something around subscribe and ```SafeSubscriber``` doing this and cutting off the last few events from the wrapped ```Subscriber```.

davidmoten:
Might be another problem here that the call to `child.setProducer` is made before adding all the subscriptions. I might test with #2997.

benjchristensen:
> why the unsubscription of the worker had to be scheduled

Was it was because we could end up not emitting an `onError` that was scheduled but then dropped because the `worker` got unsubscribed?

davidmoten:
I think @akarnokd is right, that `SafeSubscriber` reports `onCompleted` to its wrapped subscriber then unsubscribes which if `child.add(recursiveScheduler)` is present can cut short the schedule in `OperatorObserveOn` so that it doesn't emit its stuff. 

Re `onError` getting dropped this is still a possiblity but still consistent with the Observable contract (?). All the scheduled polls of the queue do check for error first so that once a scheduled task is run the error can shortcut the queue. Perhaps we can do something to aid this scenario without stuffing up the `onComplete` case.

davidmoten:
I've probably got my last comment backwards. I was thinking that SafeSubscriber was upstream of observeOn but should be downstream so wouldn't have the effect I stated.

akarnokd:
```SafeSubscriber``` came into play earlier when the operators used ```subscribe()``` to attach to each other instead of ```unsafeSubscribe()```.

akarnokd:
Closing via #3682.

PR #3544, opened 2015-11-27
1.x: fix observeOn resource handling, add delayError capability

akarnokd:
This PR fixes the "messing around" reported in #3002 and adds an overload to `observeOn` that allows delaying errors without the need for wrapping (see #3542 and maybe there are other reports).

In addition, this PR adds a proper override of the `isEmpty` method to simply compare the two indexes for emptiness directly instead of `size() == 0` to avoid looping, multi-reading and casting.

Benchmark comparison (i7 4790, Windows 7 x64, Java 8u66):

image

Note that the benchmark is generally quite noisy, yielding hectic results (i.e., firing up a thread with newThread may take quite some random microseconds). For example, `observeOnImmediate` shouldn't be affected by any of the changes yet the run-to-run variance is +/- 10%. I'm fine with the results of the benchmark.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -16,22 +16,17 @@
 package rx.internal.operators;
 
 import java.util.Queue;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
+import rx.*;
 import rx.Observable.Operator;
-import rx.Producer;
-import rx.Scheduler;
-import rx.Subscriber;
-import rx.Subscription;
 import rx.exceptions.MissingBackpressureException;
 import rx.functions.Action0;
-import rx.internal.util.RxRingBuffer;
-import rx.internal.util.SynchronizedQueue;
-import rx.internal.util.unsafe.SpscArrayQueue;
-import rx.internal.util.unsafe.UnsafeAccess;
-import rx.schedulers.ImmediateScheduler;
-import rx.schedulers.TrampolineScheduler;
+import rx.internal.util.*;
+import rx.internal.util.atomic.SpscAtomicArrayQueue;
+import rx.internal.util.unsafe.*;
+import rx.plugins.RxJavaPlugins;
+import rx.schedulers.*;
 
 /**
  * Delivers events on the specified {@code Scheduler} asynchronously via an unbounded buffer.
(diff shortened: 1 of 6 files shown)


artem-zinnatullin:
I like the change, but can't we ship this behavior as default? Yes, it's a breaking change, but in a good way. I mean, I'm just not sure that a lot of users will even know about this operator and even less will understand the difference and situations when they'll need to use one instead of another.

src/main/java/rx/internal/operators/OperatorObserveOn.java
```diff
@@ -145,106 +150,126 @@ public void onCompleted() {
         @Override
         public void onError(final Throwable e) {
             if (isUnsubscribed() || finished) {
+                RxJavaPlugins.getInstance().getErrorHandler().handleError(e);
```
artem-zinnatullin: Looks like a candidate for a separate PR, isn't it?

akarnokd: More PRs to the same class ends up being merged much later and usually in the wrong order.

stealthcode:
@akarnokd are you concerned by the performance impacts of this PR?

akarnokd:
No concerns.

src/test/java/rx/internal/operators/OperatorObserveOnTest.java
```diff
@@ -804,5 +783,55 @@ public void onNext(Integer t) {
...
+        ts.assertNotCompleted();
+
+        ts.requestMore(1);
+        s.advanceTimeBy(1, TimeUnit.SECONDS);
+        
+        ts.assertValues(1);
+        ts.assertNoErrors();
+        ts.assertNotCompleted();
+        
+        ts.requestMore(3); // requesting 2 doesn't switch to the error() source for some reason in concat.
```
stevegury: Any idea why requesting 2 didn't work?

akarnokd: The drain loop quits because it runs out of requests but at the same time, the concatenated data source is also empty. I can add a `if (checkTerminated(finished, q.isEmpty(), localChild, q)) return` check again, but this adds overhead for those sources that request again anyway.

stevegury: Ok I see, thanks for the detailed explanation.

stevegury:
:+1:

akarnokd:
Since this adds to the API surface, can I have another +1? (Also feel free to merge this.)

akarnokd:
Closing and reposting again with proper rebase.

PR #3468, opened 2015-10-24
1.x: Fix other places that may swallow OnErrorFailedException

zsxwing:
Reviewed all usages of `OnErrorNotImplementedException` and fixed issues similar to #3455.

PR #3546, opened 2015-11-27
Revert "1.x: Fix other places that may swallow OnErrorFailedException"

grabber24:
Reverts ReactiveX/RxJava#3468


artem-zinnatullin:
Why? It was a bug.

akarnokd:
Troll.

zsxwing:
Is this a new spam bot?

PR #3549, opened 2015-11-27
Remove double whitespace in if conditions

vanniktech:
Just found them while walking through the source code trying to get a better overview of the Rx internals.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorDebounceWithTime.java
+++ b/src/main/java/rx/internal/operators/OperatorDebounceWithTime.java
@@ -163,7 +163,7 @@ public final class OperatorDebounceWithTime<T> implements Operator<T, T> {
                 emitting = true;
             }
 
-            if  (localHasValue) {
+            if (localHasValue) {
                 try {
                     onNextAndComplete.onNext(localValue);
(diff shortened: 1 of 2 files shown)


artem-zinnatullin:
:+1:, you'll find a lot of such issues…

@akarnokd checkstyle one day maybe?

vanniktech:
I actually found only the one in `TestSubscriber`. After that I ran `ack "if  "` over the entire project and found also the other one.

If you guys want, I can add checkstyle in a separate PR.

artem-zinnatullin:
I'd be happy to see checkstyle in RxJava, but my previous attempts to add it failed: #2807, #3147 and other comments.

akarnokd:
It's hard to agree on some styles, i.e. always use blocks with ifs and such.

vanniktech:
Yes I get that, though I don't see what the problem is fixing once they are made.

Maybe we could agree on some very basic checkstyle rules, like don't use double whitespaces (if there's a rule for that). Tab indention etc

vanniktech:
Is there anything against this being merged?

akarnokd:
We usually require two likes from collaborators.

:+1:

PR #3550, opened 2015-12-01
Public API changes for 1.1.0 release

stealthcode:


src/main/java/rx/subjects/SerializedSubject.java
```diff
@@ -69,6 +68,7 @@ public void onNext(T t) {
     public boolean hasObservers() {
         return actual.hasObservers();
     }
+<<<<<<< HEAD
```
akarnokd: Merge conflict not resolved; this causes compilation error.

stealthcode:
This should be merged immediately after the next minor release (1.0.17). The plan is to have a minor release that is identical to the following major release except for changes to the public api (which should not substantially change functionality).

src/main/java/rx/subjects/SerializedSubject.java
stealthcode: My mistake. Thanks for the clue.

akarnokd:
:+1:

stevegury:
:+1:

stealthcode:
Now we need to update the CHANGES.md to reflect the public api changes... That might have been easier in the old format of 1 PR per change. I'll probably do sub-bullets under the PR.

mattbishop:
So Release Notes for 1.1.0 state that Single has been promoted out of Beta, but it looks like it is not actually out of beta.

Issue #3560, opened 2015-12-05
OperatorBufferWithTime error


mariusznet:
Hi,

According to the java docs 
> When the source {@link Observable} completes or produces an error, the current buffer is emitted, and the event is propagated to all subscribed {@link Subscriber}s."

But in the code on error method clears chunks swallowing them without emitting:
```
        @Override
        public void onError(Throwable e) {
            synchronized (this) {
                if (done) {
                    return;
                }
                done = true;
                chunks.clear();
            }
            child.onError(e);
            unsubscribe();
        }
```

According to the documentation http://reactivex.io/documentation/operators/buffer.html
this behaviour is expected, so the java docs need correction.

akarnokd:
The documentation is wrong. It might be worth extending the API with the usual delayError in 2.x though.

akarnokd:
Fix via #3561.

PR #3561, opened 2015-12-05
Corrected buffer operator onError behaviour documentation

mariusznet:
#3560

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -3285,9 +3285,9 @@ public class Observable<T> {
     /**
      * Returns an Observable that emits buffers of items it collects from the source Observable. The resulting
      * Observable emits connected, non-overlapping buffers, each containing {@code count} items. When the source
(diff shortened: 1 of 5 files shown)


src/main/java/rx/Observable.java
```diff
@@ -3285,8 +3285,9 @@ public final Boolean call(T first, T second) {
     /**
      * Returns an Observable that emits buffers of items it collects from the source Observable. The resulting
      * Observable emits connected, non-overlapping buffers, each containing {@code count} items. When the source
-     * Observable completes or encounters an error, the resulting Observable emits the current buffer and
-     * propagates the notification from the source Observable.
+     * Observable completes, the current buffer is emitted, and the event is propagated to all subscribed
```
akarnokd: I don't like "propagate to all subscribed Observers". It suggests this is some hot operation and data is shared (such as in publish()).

mariusznet:
Applied @akarnokd suggestion.

stevegury:
:+1:

akarnokd:
:+1:

PR #3562, opened 2015-12-05
Add Single.doOnUnsubscribe()

artem-zinnatullin:
As promised, adding more operators to `Single`. This one is very useful for releasing resources if task can be cancelled externally via `subscription.unsubscribe()` and subscriber may not receive the result.


akarnokd:
:+1:

stealthcode:
:+1:

PR #3565, opened 2015-12-08
Replace never() calls in BehaviorSubjectTest with verifyNoMoreInteractions

vanniktech:
Advantage is that we verify that there's no interaction at all so also something that we might not expect is not going to happen.


src/test/java/rx/subjects/BehaviorSubjectTest.java
```diff
@@ -43,7 +44,7 @@
 
 public class BehaviorSubjectTest {
 
-    private final Throwable testException = new Throwable();
+    private Throwable testException = new Throwable();
```
akarnokd: Why did this lose `final`?

vanniktech: @akarnokd was a mistake. I added the final back

akarnokd:
:+1:

stevegury:
:+1:

PR #3569, opened 2015-12-09
1.x: fix renamed operator in Single.

akarnokd:
There was a cross dependency between two PRs yielding a broken
compilation in main.


artem-zinnatullin:
Yup, noticed that too, git… :+1:

akarnokd:
I'm merging this to unblock other PRs.

stevegury:
:+1:

PR #3570, opened 2015-12-09
Implemented Completable#andThen(Observable)

stealthcode:
I expect some discussion around the method name.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1081,6 +1081,21 @@ public class Completable {
     }
     
     /**
+     * Returns an Observable which will subscribe to this Completable and once that is completed then 
+     * will subscribe to the {@code next} Observable. An error event from this Completable will be 
+     * propagated to the downstream subscriber and will result in skipping the subscription of the 
+     * Observable.  
+     * 
+     * @param next the Observable to subscribe after this Completable is completed, not null
+     * @return Observable that composes this Completable and next
+     * @throws NullPointerException if next is null
+     */
+    public final <T> Observable<T> andThen(Observable<T> next) {
+        requireNonNull(next);
+        return next.delaySubscription(toObservable());
+    }
+    
+    /**
      * Concatenates this Completable with another Completable.
      * @param other the other Completable, not null
      * @return the new Completable which subscribes to this and then the other Completable
--- a/src/test/java/rx/CompletableTest.java
+++ b/src/test/java/rx/CompletableTest.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.*;
 import org.junit.*;
 
 import rx.Completable.*;
+import rx.Observable.OnSubscribe;
 import rx.exceptions.*;
 import rx.functions.*;
 import rx.observers.TestSubscriber;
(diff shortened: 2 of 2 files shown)


stealthcode:
Adding unit tests and pushing shortly.

src/main/java/rx/Completable.java
```diff
@@ -1081,6 +1081,21 @@ public final Completable compose(CompletableTransformer transformer) {
...
+     * propagated to the downstream subscriber and will result in skipping the subscription of the 
+     * Observable.  
+     * 
+     * @param next the Observable to subscribe after this Completable is completed, not null
+     * @return Observable that composes this Completable and next
+     * @throws NullPointerException if next is null
+     */
+    public final <T> Observable<T> andThen(Observable<T> next) {
+        requireNonNull(next);
+        return next.delaySubscription(toObservable());
```
akarnokd: `Completable` already has an `endWith` overload with the same purpose.

stealthcode: Oh, I missed that.

akarnokd: Is there a verdict on Observable.andThen? Renaming `endWith` to `andThen` may serve some consistency purpose among base classes.

stealthcode: One initial motivation for `andThen` was based on the [Option::and_then method] (https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then) in Rust stdlib. It chains through the computation and transforms it's type/contents to the next Option.

abersnaze: My first thought was `java.util.function.Function a; a.andThen(b)`.  Also `endWith` doesn't end the composition this would look strange `a.endWith(b).endWith(c)` vs `a.andThen(b).andThen(c)`.

stevegury: I think `andThen` is more idiomatic than `endWith`, especially for chaining purposes.

stealthcode: I guess that if we are going to rename or change any of the public API on Completable now is the time before our next release. Should we aim to review the operators before next release?

akarnokd: Yes.

src/test/java/rx/CompletableTest.java
```diff
@@ -357,6 +357,56 @@ public void call(Long v) {
...
+        ts.assertNoValues();
+        ts.assertNoTerminalEvent();
+    }
+    
+    @Test
+    public void andThenError() {
+        TestSubscriber<String> ts = new TestSubscriber<>(0);
+        AtomicBoolean hasRun = new AtomicBoolean(false);
+        Exception e = new Exception();
+        Completable.create(cs -> cs.onError(e))
```
akarnokd: Lambdas are Java 8. Besides, there is already a `Completable.error()` and you should call `cs.onSubscribe(Subscriptions.unsubscribed())` first.

src/test/java/rx/CompletableTest.java
```diff
@@ -357,6 +357,56 @@ public void call(Long v) {
...
+            .andThen(Observable.<String>create(s-> {
```
akarnokd: Lambda.

stealthcode: I'll correct this.

akarnokd:
:+1:

PR #3579, opened 2015-12-17
1.x: No more need to convert Singles to Observables for Single.zip()

artem-zinnatullin:
PTAL

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -956,9 +956,9 @@ public class Single<T> {
      * <dd>{@code zip} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      * 
-     * @param o1
+     * @param s1
      *            the first source Single
-     * @param o2
+     * @param s2
      *            a second source Single
      * @param zipFunction
      *            a function that, when applied to the item emitted by each of the source Singles, results in an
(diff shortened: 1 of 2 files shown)


akarnokd:
:+1:

src/main/java/rx/Single.java
```diff
@@ -966,8 +966,13 @@ public void onError(Throwable error) {
      * @return a Single that emits the zipped results
      * @see <a href="http://reactivex.io/documentation/operators/zip.html">ReactiveX operators documentation: Zip</a>
      */
-    public final static <T1, T2, R> Single<R> zip(Single<? extends T1> o1, Single<? extends T2> o2, final Func2<? super T1, ? super T2, ? extends R> zipFunction) {
-        return just(new Observable<?>[] { asObservable(o1), asObservable(o2) }).lift(new OperatorZip<R>(zipFunction));
+    public static <T1, T2, R> Single<R> zip(Single<? extends T1> s1, Single<? extends T2> s2, final Func2<? super T1, ? super T2, ? extends R> zipFunction) {
```
zsxwing: nit: please add `final` back.

zsxwing:
Could you add `final` back? Otherwise, LGTM

src/main/java/rx/Single.java
JakeWharton: or remove `final` from all static methods in another PR

akarnokd: I'm for the removal as well, it's redundant. (I also remember an argument about it some time back.)

artem-zinnatullin:
Sorry, was busy… @zsxwing added `final` back, will send separate PR to remove them everywhere once this merged.

zsxwing:
:+1:

Issue #3581, opened 2015-12-18
observable.first() not working properly after grouping data


eyeced:
I am calling things like this
```
Observable.just(1, 2, 3, 4, 5, 6)
                .groupBy(i -> i % 3)
                .flatMap(o -> o.first())
                .subscribe(System.out::println);
```
Ideally this should return 1,2,3
But the result is - 
```
1
2
3
4
5
6
```

got this issue in version 1.0.10 and 1.1.0 as well

zsxwing:
This is expected. `groupBy` will emit a new `Observable` if the previous `Observable` of the same group is subscribed.

davidmoten:
This has been discussed in #3411. It's about time I made this clearer in the javadoc so I'll knock up a PR soon.

artem-zinnatullin:
Looks like this issue can be closed?

zsxwing:
#3584 has updated the doc.

Issue #3411, opened 2015-10-05
#groupBy resubscribes to Observable after unsubscribing


NachoSoto:
*Apologies if this is repeated. #283 looked similar but I'm opening a new issue since that doesn't seem to be the current behavior*

`Observable#groupBy` encourages us to unsubscribe from `Observables` that we no longer want values from to avoid leaking that memory, which makes total sense:

Example:
```java
observable
    .groupBy(SomeKey::new)
    .subscribe(groupedObservable -> {
        // This is called again after unsubscribing.
        groupedObservable
            .take(1)
            .subscribe(...)
    })
```

This works fine. However, there's a behavior I was not expecting: after `groupedObservable` has been unsubscribed, the `Observable` returned by `observable.groupBy` emits that observable again!

Is this expected?

*(Currently using version 1.0.10)*

akarnokd:
This is the expected behavior.

```java
Observable.range(1, 10)
.groupBy(v -> 1)
.flatMap(g -> g.take(1))
.subscribe(System.out::println);
```

This will print values from 1-10. 

Whether or not this is reasonable is up for discussion, the alternative would be to drop values once the single subscriber to the group unsubscribes so no groups are opened

NachoSoto:
Seems counter intuitive, and I think it at least deserves a comment in the docs.
How could I implement a similar `groupBy` operator that never emits that same `Observable` again?

davidmoten:
This behaviour is not what I would hope from `groupBy` either. I can imagine both the resubscribing and non-resubscribing behaviour being useful but I would expect the default to be non-resubscribing.

NachoSoto:
Thanks @davidmoten, I agree.

davidmoten:
Anyone out there know what the Microsoft Rx `groupBy` behaviour is in this regard?

akarnokd:
What I can deduce from the [source code](https://github.com/Reactive-Extensions/Rx.NET/blob/a13e3ff05bdded5cef2bf40bface22f8fa4ae316/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/GroupBy.cs) is that groups are created but never removed and once all groups and the main output is disposed, the source is stopped.

Note that Rx.NET doesn't have backpressure and the synchronous cancellation is a mistery to me in it.

> How could I implement a similar groupBy operator that never emits that same Observable again?

GroupBy is perhaps the most complicated operator to think about because of the request/unsubscribe coordination requirement. 

[...]

PR #3584, opened 2015-12-21
clarify behaviour of groupBy in javadoc

davidmoten:
As discussed in #3411, some clarification of the behaviour of `groupBy` looks necessary re its behaviour when the `GroupedObservable` terminates early (for example has `.first()` applied to it). See also  #3581.

I also took the liberty of recommending `ignoreElements` instead of `take(0)` for the section in the javadoc that recommends how to ignore emissions because the `take(0)` approach would have more overhead (every emission from source would prompt the emission of a new `GroupedObservable`) than the `ignoreElements` approach.

Once this is accepted I'll look to submit a PR with `groupByOnce` (mentioned in #3411).

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -5577,14 +5577,17 @@ public class Observable<T> {
     
     /**
      * Groups the items emitted by an {@code Observable} according to a specified criterion, and emits these
-     * grouped items as {@link GroupedObservable}s, one {@code GroupedObservable} per group.
+     * grouped items as {@link GroupedObservable}s. The emitted {@code GroupedObservable} allows only a single 
+     * {@link Subscriber} during its lifetime and if this {@code Subscriber} unsubscribes before the 
+     * source terminates, the next emission by the source having the same key will trigger a new 
+     * {@code GroupedObservable} emission.
      * <p>
      * <img width="640" height="360" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/groupBy.png" alt="">
      * <p>
      * <em>Note:</em> A {@link GroupedObservable} will cache the items it is to emit until such time as it
      * is subscribed to. For this reason, in order to avoid memory leaks, you should not simply ignore those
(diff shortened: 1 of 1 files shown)


davidmoten:
Corrected the documentation that said resubscription occurs when it is rather a new GroupedObservable is emittted for a group.

src/main/java/rx/Observable.java
```diff
@@ -5577,14 +5577,16 @@ public final void forEach(final Action1<? super T> onNext, final Action1<Throwab
     
     /**
      * Groups the items emitted by an {@code Observable} according to a specified criterion, and emits these
-     * grouped items as {@link GroupedObservable}s, one {@code GroupedObservable} per group.
+     * grouped items as {@link GroupedObservable}s. If a {@code GroupedObservable} 
+     * terminates before the source terminates then the next emission by the source with the same key as the
```
akarnokd: GroupedObservable allows only a single Subscriber during its lifetime and if this Subscriber unsubscribes before the main source terminates, the next emission by the main source having the same key will trigger a new GroupedObservable emission.

davidmoten: Thanks @akarnokd, that does add information. I'll use that.

davidmoten: I notice that GroupedObservable itself doesn't have a single Subscriber restriction but certainly the GroupedObservable emitted by groupBy does have this property. I'll rejig the wording to get this across.

davidmoten:
Incorporated @akarnokd suggestion

akarnokd:
:+1:

zsxwing:
:+1: 

> I also took the liberty of recommending ignoreElements instead of take(0) for the section in the javadoc that recommends how to ignore emissions because the take(0) approach would have more overhead (every emission from source would prompt the emission of a new GroupedObservable) than the ignoreElements approach.

By the way, `ignoreElements` has a drawback: It keeps all group `Observable`s in memory even some of them won't emit items.

PR #3585, opened 2015-12-21
1.x: fix Completable.using not disposing the resource if the factory crashes during the subscription phase.

akarnokd:
This PR fixes the cases when the Completable factory throws an exception
or returns null and the resource is not disposed before reporting error
to the subscriber.


artem-zinnatullin:
:+1:

zsxwing:
:+1:

PR #3587, opened 2015-12-25
Remove the need for javac to generate synthetic methods.

JakeWharton:
Outer classes accessing inner class private fields and methods (and vise versa) causes javac to generate package-scoped trampolines. These bloat the class files, adds overhead to the inliner analysis, and for Android create needless method that eat away at our fixed limit of methods in an application. By simply promoting the private interactions to package scope directly, the synthetic methods do not need generated.

2.5% of RxJava's methods were these needless generated trampolines accounting for 1.2% of jar size and 1.4% of dex size.

```
$ dex-method-count before.dex
5005

$ dex-method-count after.dex
4875
```
```
-rw-r--r--   1 jw  jw   699K Dec 25 03:08 after.dex
-rw-r--r--   1 jw  jw   952K Dec 25 03:08 after.jar
-rw-r--r--   1 jw  jw   709K Dec 25 03:06 before.dex
-rw-r--r--   1 jw  jw   964K Dec 25 03:06 before.jar
```

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -60,7 +60,7 @@ public class Observable<T> {
         this.onSubscribe = f;
     }
 
-    private static final RxJavaObservableExecutionHook hook = RxJavaPlugins.getInstance().getObservableExecutionHook();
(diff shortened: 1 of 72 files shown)


akarnokd:
:+1:

zsxwing:
:+1:

PR #3588, opened 2015-12-27
Fix the initialization order in GenericScheduledExecutorService

zsxwing:
The static `GenericScheduledExecutorService.None` should be initialized before creating any GenericScheduledExecutorService instance. Although the previous codes happen to work, it's sill worth to fix it.


akarnokd:
The situation can't happen because class initialization locks out callers.

Otherwise, the cleanup never hurts.

:+1:

PR #3589, opened 2015-12-27
1.x: concat reduce overhead when streaming a source

akarnokd:
This PR reduces the request tracking overhead of `concat` by tracking the produced item count in a plain field and subtracting it from the arbiter and requested values only before the inner source completes. So instead of N decrementAndGet call, we have 1 addAndGet(-N) per source. 

I've added a perf class to measure the difference.

(Intel Celeron 1005M @ 2GHz, Windows 7 x64, Java 8u66)

image

The throughput increased considerably, although I would have expected more, especially in the 1M case where the subscription overhead doesn't matter.

I'll do further investigation on it and post a follow-up PR if this gets merged in the meantime.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorConcat.java
+++ b/src/main/java/rx/internal/operators/OperatorConcat.java
@@ -16,18 +16,14 @@
 package rx.internal.operators;
 
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.*;
 
-import rx.Observable;
+import rx.*;
 import rx.Observable.Operator;
-import rx.Producer;
-import rx.Subscriber;
 import rx.functions.Action0;
 import rx.internal.producers.ProducerArbiter;
 import rx.observers.SerializedSubscriber;
-import rx.subscriptions.SerialSubscription;
-import rx.subscriptions.Subscriptions;
+import rx.subscriptions.*;
 
 /**
  * Returns an Observable that emits the items emitted by two or more Observables, one after the other.
@@ -112,9 +108,19 @@ public final class OperatorConcat<T> implements Operator<T, Observable<? extends
         }
 
         private void requestFromChild(long n) {
-            if (n <=0) return;
+            if (n <= 0) return;
             // we track 'requested' so we know whether we should subscribe the next or not
-            long previous = BackpressureUtils.getAndAddRequest(requested, n);
+            
+            final AtomicLong requestedField = requested;
+            
+            long previous;
+            
+            if (requestedField.get() != Long.MAX_VALUE) {
+                previous = BackpressureUtils.getAndAddRequest(requestedField, n);
+            } else {
+                previous = Long.MAX_VALUE;
+            }
+            
             arbiter.request(n);
             if (previous == 0) {
                 if (currentSubscriber == null && wip.get() > 0) {
@@ -123,10 +129,6 @@ public final class OperatorConcat<T> implements Operator<T, Observable<? extends
                     subscribeNext();
                 }
             } 
-        }
-
-        private void decrementRequested() {
-            requested.decrementAndGet();
         }
 
         @Override
@@ -167,8 +169,10 @@ public final class OperatorConcat<T> implements Operator<T, Observable<? extends
                     child.onCompleted();
                 } else if (o != null) {
                     Observable<? extends T> obs = nl.getValue(o);
+                    
                     currentSubscriber = new ConcatInnerSubscriber<T>(this, child, arbiter);
                     current.set(currentSubscriber);
+
                     obs.unsafeSubscribe(currentSubscriber);
                 }
             } else {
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
:+1:

src/main/java/rx/internal/operators/OperatorConcat.java
```diff
@@ -179,6 +187,14 @@ void subscribeNext() {
                 }
             }
         }
+        
+        void produced(long c) {
+            arbiter.produced(c);
+            AtomicLong requestedField = requested;
+            if (requestedField.get() != Long.MAX_VALUE) {
+                requestedField.addAndGet(-c);
```
zsxwing: Call `getAndAddRequest` here? In case `requested` is changed to `Long.MAX_VALUE` at the same time.

akarnokd: In normal use, such change never happens; consumers either start with MAX_VALUE or some prefetch value, plus you have to be very precise to hit the narrow window between the two lines.

davidmoten: I'd prefer to see correctness. Speculating on the probability of a race
condition seems like dangerous territory to be entering.

akarnokd: This line is correct as it is, what we are talking about is considering the case where one wants to subvert the optimization by racing with get().

davidmoten: I've a very narrow focus on this method so I could well be missing something but normally once `requested` is set to Long.MAX_VALUE we don't reduce its value and @zsxwing has indicated this possibility. I don't think it should be best endeavours as in I would expect to see a CAS loop here. Is this a bit fussy? Do we do this in lots of places already? (I thought a lot had been cleaned up).

akarnokd: The keyword is loop which can lenghten due to failed CAS in bounded mode or safepoint kicking in. AddAndGet is a one step atomic operation.

zsxwing: So no place will update `requested` at the same time. Right?

akarnokd:
There is a non-zero chance there, but it is very small. If it happens, all the effect it will have is to make the operator decrement the requested amount from then on, which adds ~40 cycle overhead after each source completes.

Updated the PR anyway to do a CAS loop on production to remove all worries about the case.

zsxwing:
:+1:

akarnokd:
@zsxwing, @davidmoten are you satisfied with the name `BackpressureUtils.produced`?

src/main/java/rx/internal/operators/OperatorConcat.java
```diff
@@ -112,9 +112,19 @@ public void onStart() {
...
+            if (n <= 0) return;
             // we track 'requested' so we know whether we should subscribe the next or not
-            long previous = BackpressureUtils.getAndAddRequest(requested, n);
+            
+            final AtomicLong requestedField = requested;
+            
+            long previous;
+            
+            if (requestedField.get() != Long.MAX_VALUE) {
+                previous = BackpressureUtils.getAndAddRequest(requestedField, n);
```
davidmoten: This path involves two `AtomicLong.get()`. Can't the whole block just be replaced by 
`long previous = BackpressureUtils.getAndAddRequest(requestedField, n);`?

akarnokd: No, the `getAndAddRequest` doesn't check for max-value and always does the CAS. I'd leave this as is and have a separate PR where the `getAndAddRequest` is changed and all use places are verified/fixed.

src/main/java/rx/internal/operators/OperatorConcat.java
```diff
@@ -187,6 +200,8 @@ void subscribeNext() {
         private final ConcatSubscriber<T> parent;
         private final AtomicInteger once = new AtomicInteger();
```
davidmoten: Why not `AtomicBoolean`?

src/main/java/rx/internal/operators/OperatorConcat.java
```diff
@@ -212,6 +227,11 @@ public void onError(Throwable e) {
         @Override
         public void onCompleted() {
             if (once.compareAndSet(0, 1)) {
+                // signal the production count at once instead of one by one
+                long c = produced;
+                if (c != 0L) {
```
davidmoten: Move the `c!=0` test into the `produced` method? Don't see why you need to assign `produced` to `c` as it can't be changed by another thread due Observable contract.

stevegury: I think @davidmoten is right here, I don't see why you need to capture `produced`.

src/main/java/rx/internal/operators/OperatorConcat.java
stevegury: AtomicBoolean is very similar to AtomicInteger since it uses under the hood an integer.
But I agree that AtomicBoolean makes the intent of the code more explicit.

akarnokd: internally, `AtomicBoolean` is implemented the same as AtomicInteger, but sure, can be changed to `AtomicBoolean`, by a separate PR.

stevegury:
Just a few comments, otherwise :+1:.

src/main/java/rx/internal/operators/OperatorConcat.java
akarnokd: Because of the compareAndSet, the read of parent field would happen all the time, including empty sources. This also affects produced which may or may not be register-cached.

akarnokd:
Changed to AtomicBoolean and moved c != 0 into parent because the parent was used for completing the inner anyway, I overlooked it at first.

stevegury:
:+1:

PR #3592, opened 2015-12-30
1.x: add never test for  PublishSubjectTest

Chaoba:
This PR complete the never assert in PublishSubjectTest.

### DIFF
--- a/src/test/java/rx/subjects/PublishSubjectTest.java
+++ b/src/test/java/rx/subjects/PublishSubjectTest.java
@@ -63,7 +63,7 @@ public class PublishSubjectTest {
         subject.onError(new Throwable());
 
         assertCompletedObserver(observer);
-        // todo bug?            assertNeverObserver(anotherObserver);
+        assertNeverObserver(anotherObserver);
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


src/test/java/rx/subjects/PublishSubjectTest.java
```diff
@@ -150,6 +160,15 @@ private void assertErrorObserver(Observer<String> observer) {
...
     }
 
+    private void assertNeverErrorObserver(Observer<String> observer) {
+        verify(observer, never()).onNext("one");
+        verify(observer, never()).onNext("two");
+        verify(observer, never()).onNext("three");
+        verify(observer, never()).onNext("four");
+        verify(observer, times(1)).onError(any(Throwable.class));
+        verify(observer, never()).onCompleted();
+    }
```
JakeWharton: I think you'll want `verifyNoMoreInteractions(observer)` to end these two methods, otherwise sending an `onNext` of "five" would not fail the test.

akarnokd:
:+1:

vanniktech:
In every test method that is using `verifyNoMoreInteractions` you can remove the `verify(observer, never()).*`

Chaoba:
> In every test method that is using verifyNoMoreInteractions you can remove the verify(observer, never()).*

Your are right, but I think adding these verifies can make tests more clearly.

vanniktech:
Fair enough. I just thought you forgot to delete them and wanted to point it out.
In any case the deletion decision is something the repo owners have to do.

Personally the less test code the preciser it is. Especially since `verifyNoMoreInteractions` clearly states it purpose.

zsxwing:
:+1:

stevegury:
:+1:

PR #3597, opened 2016-01-04
1.x: fix counted buffer and window operators' backpressure behavior

akarnokd:
This PR fixes the backpressure behavior of the counted `buffer` and `window` operators and consists of several changes.

The main issue lies when `count > skip` in the operators, yielding overlapping buffers/windows. 

For `buffer`, when the upstream completed, the logic emitted all remaining partial buffers even if there was no request for new buffers, which can result in `MissingBackpressureException` somewhere. The proper handling of the final buffers required a new backpressure management algorithm which is now part of the `BackpressureUtils` class and consists of two new methods: `postCompleteDone` called from onComplete to take over the emission of queued values and `postCompleteRequest` which manages requests before and after the completed state.

For `window`, the new window opened was emitted regardless of requests which was common due to request-amplification (i.e., requesting n windows results in requesting `count + skip * (n - 1)` elements at first (then `skip * n` later) which opens `ceil(count / skip)` windows upfront. To avoid the overflow, the individual windows have to go through the usual queue/drain logic as well. I've also updated the Javadoc to reflect the backpressure behavior along with parameter validation.

In addition, the window case didn't manage cancellation properly. When the outer observable is unsubscribed, the inner subscribers may be still going and thus cancelling the upstream would stop/hang the inner windows. Instead, the open window count is tracked (also counting the outer as 1 window) and when all get unsubscribed (i.e., count reaches zero), the upstream is unsubscribed. To accomplish this, the `UnicastSubject` had to be retrofitted with a new optional callback `Action0` which gets called at most once whenever either `onError` or `onCompleted` is called or when the single `Subscriber` unsubscribes.

A secondary issue was with the `TestSubscriber`'s initial request; some upstream operators could get triggered with `Long.MAX_VALUE` despite the initial request amount was set. This PR changes it to be set at construction time instead of in `onStart`.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -9654,8 +9654,8 @@ public class Observable<T> {
      * <img width="640" height="400" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/window3.png" alt="">
      * <dl>
      *  <dt><b>Backpressure Support:</b></dt>
-     *  <dd>The operator honors backpressure on its outer subscriber, ignores backpressure in its inner Observables 
-     *  but each of them will emit at most {@code count} elements.</dd>
+     *  <dd>The operator honors backpressure of its inner and outer subscribers, however, the inner Observable uses an
+     *  unbounded buffer that may hold at most {@code count} elements.</dd>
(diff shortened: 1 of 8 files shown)


akarnokd:
Fresh version in #3678.

PR #3609, opened 2016-01-08
delaySubscription(Func0) does not use a scheduler

JakeWharton:
It subscribes to the upstream `Observable` on the emitting thread of the other `Observable` obtained from the `Func0`.


akarnokd:
:+1:

stealthcode:
:+1:

PR #3622, opened 2016-01-18
1.x: Optimizate single just

Chaoba:
This PR add similar optimization as ScalarSynchronousObservable to Single.just method.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -37,6 +37,8 @@ import rx.functions.FuncN;
 import rx.annotations.Beta;
 import rx.internal.operators.*;
 import rx.internal.producers.SingleDelayedProducer;
+import rx.internal.util.ScalarSynchronousSingle;
+import rx.internal.util.UtilityFunctions;
 import rx.singles.BlockingSingle;
 import rx.observers.SafeSubscriber;
 import rx.plugins.*;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/Single.java
```diff
@@ -685,6 +679,9 @@ public void call(SingleSubscriber<? super T> te) {
      * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
      */
     public final static <T> Single<T> merge(final Single<? extends Single<? extends T>> source) {
+        if (source.getClass() == ScalarSynchronousSingle.class) {
```
akarnokd: It appears `instanceof` can be optimized much better by JIT than class equivalence checks. Could you rewrite this and all the other places?

src/main/java/rx/internal/util/ScalarSynchronousSingle.java
```diff
@@ -0,0 +1,156 @@
...
+import rx.functions.Func1;
+import rx.internal.schedulers.EventLoopsScheduler;
+
+public final class ScalarSynchronousSingle<T> extends Single<T> {
+
+    public static final <T> ScalarSynchronousSingle<T> create(T t) {
+        return new ScalarSynchronousSingle<T>(t);
+    }
+
+    final T t;
```
artem-zinnatullin: Let's rename this to `value`? Will improve readability of the code!

artem-zinnatullin:
Nice PR! @Chaoba can you please remove merge commit from the PR (rebase should help)?

Chaoba:
@akarnokd and @artem-zinnatullin  please review on  #3642

Issue #3621, opened 2016-01-17
doOnEmpty() Operator?


thomasnield:
Is it possible we can add an `Observable.doOnEmpty()` operator? Sometimes when I'm debugging or want to handle an empty `Observable` I'd like to raise an error, display a message, or some other action to handle the emptiness...

```
Observable<String> source = ...


source.doOnEmpty(() -> System.out.println("No items were emitted!")).subscribe(System.out::println);
```

thomasnield:
Put in a PR with this operator implemented #3624

zsxwing:
The name of `doOnEmpty` is not consistent with other methods. E.g., doOnNext means triggering a function when `onNext` is called.

thomasnield:
Technically that is true, but I can't really think of a better operator name that expresses the intent as clearly as 'doOnEmpty()'. Trying to conform with 'doOnCompletedIfEmpty()' or some other excessively descriptive name can make it verbose.

You all are welcome to change it if you feel its warranted though.

thomasnield:
Although it does not follow the strict definition of "event" by Rx standards, one could argue that emptiness is an event resulting from the 'onCompleted() event. But the more I think about it I do see what you are saying. I guess the real question is will RxJava users get confused by the name 'doOnEmpty()' implying there is a nonexistant 'onEmpty()' event... or is it more intuitive to call it that?

davidmoten:
I think `doOnEmpty` is a natural name for this one.

thomasnield:
What's the status on this? Is this planned for a layer release?

akarnokd:
You can achieve a similar effect via composition

```java
Observable.empty().switchIfEmpty(Observable.empty().doOnCompleted(() -> { ... })).subscribe();
```

thomasnield:
Oh nice, that's clever. And it doesn't even require any custom Operator implementation. 

I can go ahead and close this then. I've started using Kotlin with extension functions so it's marginalized the need personally for me to put this in a centralized library.

hzsweers:
Sorry to necro this, but I still think this would be useful as a public API, even if it's a small composition under the hood. There's precedent for it a few other areas in the API that just call a few other top-level methods under the hood.

thomasnield:
You can try convincing @akarnokd as I agree this would be a good operator. 

If not, seriously check out Kotlin. You can add your own operators to Observable via extension functions, without any call to lift() or compose(). Kotlin is officially supported on Android as well.

[...]

PR #3624, opened 2016-01-19
1.x: implement OperatorDoOnEmpty, with Observable.doOnEmpty() operator

thomasnield:
See #3621

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -12,31 +12,20 @@
  */
 package rx;
 
-import rx.annotations.Beta;
-import rx.annotations.Experimental;
-import rx.exceptions.Exceptions;
-import rx.exceptions.OnErrorNotImplementedException;
+import java.util.*;
+import java.util.concurrent.*;
+
+import rx.annotations.*;
+import rx.exceptions.*;
 import rx.functions.*;
 import rx.internal.operators.*;
 import rx.internal.producers.SingleProducer;
-import rx.internal.util.RxRingBuffer;
-import rx.internal.util.ScalarSynchronousObservable;
-import rx.internal.util.UtilityFunctions;
-import rx.observables.BlockingObservable;
-import rx.observables.ConnectableObservable;
-import rx.observables.GroupedObservable;
+import rx.internal.util.*;
+import rx.observables.*;
 import rx.observers.SafeSubscriber;
-import rx.plugins.RxJavaObservableExecutionHook;
-import rx.plugins.RxJavaPlugins;
-import rx.schedulers.Schedulers;
-import rx.schedulers.TimeInterval;
-import rx.schedulers.Timestamped;
+import rx.plugins.*;
+import rx.schedulers.*;
 import rx.subscriptions.Subscriptions;
-
-import java.util.*;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
 
 /**
  * The Observable class that implements the Reactive Pattern.
@@ -50,7 +39,7 @@ import java.util.concurrent.TimeUnit;
  * <p>
  * For more information see the <a href="http://reactivex.io/documentation/observable.html">ReactiveX
  * documentation</a>.
- * 
+ *
  * @param <T>
  *            the type of the items emitted by the Observable
  */
@@ -63,7 +52,7 @@ public class Observable<T> {
      * <p>
      * <em>Note:</em> Use {@link #create(OnSubscribe)} to create an Observable, instead of this constructor,
      * unless you specifically have a need for inheritance.
-     * 
+     *
      * @param f
      *            {@link OnSubscribe} to be executed when {@link #subscribe(Subscriber)} is called
      */
@@ -92,7 +81,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code create} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param <T>
      *            the type of the items that this Observable emits
      * @param f
@@ -123,8 +112,8 @@ public class Observable<T> {
     /**
      * Passes all emitted values from this Observable to the provided conversion function to be collected and
      * returned as a single value. Note that it is legal for a conversion function to return an Observable
-     * (enabling chaining). 
-     * 
+     * (enabling chaining).
+     *
      * @param conversion a function that converts from this {@code Observable<T>} to an {@code R}
      * @return an instance of R created by the provided conversion function
      * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
@@ -137,7 +126,7 @@ public class Observable<T> {
                 subscriber.add(Observable.subscribe(subscriber, Observable.this));
             }});
     }
-    
+
     /**
      * Lifts a function to the current Observable and returns a new Observable that when subscribed to will pass
      * the values of the current Observable through the Operator function.
@@ -155,7 +144,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code lift} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param operator the Operator that implements the Observable-operating function to be applied to the source
      *             Observable
      * @return an Observable that is the result of applying the lifted Operator to the source Observable
@@ -172,7 +161,7 @@ public class Observable<T> {
                         st.onStart();
                         onSubscribe.call(st);
                     } catch (Throwable e) {
-                        // localized capture of errors rather than it skipping all operators 
+                        // localized capture of errors rather than it skipping all operators
                         // and ending up in the try/catch of the subscribe method which then
                         // prevents onErrorResumeNext and other similar approaches to error handling
                         Exceptions.throwIfFatal(e);
@@ -187,7 +176,7 @@ public class Observable<T> {
             }
         });
     }
-    
+
     /**
      * Transform an Observable by applying a particular Transformer function to it.
      * <p>
@@ -201,7 +190,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code compose} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param transformer implements the function that transforms the source Observable
      * @return the source Observable, transformed by the transformer function
      * @see <a href="https://github.com/ReactiveX/RxJava/wiki/Implementing-Your-Own-Operators">RxJava wiki: Implementing Your Own Operators</a>
@@ -267,7 +256,7 @@ public class Observable<T> {
     public Completable toCompletable() {
         return Completable.fromObservable(this);
     }
-    
+
 
     /* *********************************************************************************************************
      * Operators Below Here
@@ -283,7 +272,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code amb} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param sources
      *            an Iterable of Observable sources competing to react first
      * @return an Observable that emits the same sequence as whichever of the source Observables first
@@ -303,7 +292,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code amb} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param o1
      *            an Observable competing to react first
      * @param o2
@@ -325,7 +314,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code amb} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param o1
      *            an Observable competing to react first
      * @param o2
@@ -349,7 +338,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code amb} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param o1
      *            an Observable competing to react first
      * @param o2
@@ -403,7 +392,7 @@ public class Observable<T> {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code amb} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * 
+     *
      * @param o1
      *            an Observable competing to react first
      * @param o2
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,66 @@
...
+    public Subscriber<? super T> call(final Subscriber<? super T> child) {
+
+        return new Subscriber<T>() {
+
+            private volatile boolean isEmpty = true;
+
+            @Override
+            public void onCompleted() {
+                if (isEmpty)
+                    action.call();
```
davidmoten: Should this call get a try catch and `Exceptions.throwOrReport` called? If so, I'll put in a PR for `OperatorDoAfterTerminate` as well.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final Action0 action;
+
+    public OperatorDoOnEmpty(Action0 action) {
+        this.action = action;
+    }
+    @Override
+    public Subscriber<? super T> call(final Subscriber<? super T> child) {
+
+        return new Subscriber<T>() {
```
davidmoten: `child` should be passed as a parameter otherwise unsubscription of `child` will not bring about the unsubscription of your anonymous instance of `Subscriber`.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,66 @@
...
+
+            private volatile boolean isEmpty = true;
```
davidmoten: `isEmpty` does not have to be volatile because calls to the event methods are guaranteed by the Observable contract to be ordered

src/test/java/rx/internal/operators/OperatorDoOnEmptyTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+        source.doOnEmpty(new Action0() {
+            @Override
+            public void call() {
+                wasCalled.set(true);
+            }
+        }).subscribe();
+
+        assert(wasCalled.get());
+    }
```
davidmoten: a unit test for unsubscription would be good

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,66 @@
...
+package rx.internal.operators;
+
+import rx.Observable;
+import rx.Subscriber;
+import rx.exceptions.Exceptions;
+import rx.functions.Action0;
+
+
+/**
+ * Perofrms an action of your choosing when the {@code Observable} has emitted no items after onCompleted() has been called
```
davidmoten: typo -> Performs

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,66 @@
...
+ * this transformation as a new {@code Observable}.
```
davidmoten: not making sense with the preceding words, remove?

src/test/java/rx/internal/operators/OperatorDoOnEmptyTest.java
akarnokd: Please add unit test that verifies backpressure composes through.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
thomasnield: It felt like I was missing something... I'll add that.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
thomasnield: Okay I knew this might come up and this topic always confuses me. But different threads can pass through the operator right? How is it guaranteed each thread will see the latest value change?

Not questioning the practice since all the other operators are designed without volatile. Bur just trying to figure out why its not needed...

src/test/java/rx/internal/operators/OperatorDoOnEmptyTest.java
thomasnield: Okay will do.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
akarnokd: Threads can only pass through if they let each other through a method called [serializing acces](http://akarnokd.blogspot.hu/2015/05/operator-concurrency-primitives.html) which always involves memory barriers and thus changes become visible to each other.

thomasnield: Okay cool, so there is some form of synchronization going on elsewhere in other operators.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
thomasnield: I might have misunderstood `Exceptions.throwOrReport`. Does "reporting" pass it up the chain via `onError()` and `throw` simply throws an exception without passing it?

thomasnield:
I saw there was an `OperatorDoOnEach` which was being paired with an `Observer`, so I went ahead and scrapped `OperatorDoOnEmpty` in favor of using that pattern.

thomasnield:
Okay two questions:

1) What exactly am I testing for with unsubscription? That `Subscription.isUnsubscribed()` is true after calling `Subscription.unsubscribe()`? That items are no longer emitting? Do I have to test for hot and cold sources? 

2) Can you point me to another unit test that does the backpressure test you have in mind?

src/main/java/rx/Observable.java
```diff
@@ -12,21 +12,32 @@
...
+import rx.internal.util.RxRingBuffer;
+import rx.internal.util.ScalarSynchronousObservable;
+import rx.internal.util.UtilityFunctions;
+import rx.observables.BlockingObservable;
+import rx.observables.ConnectableObservable;
+import rx.observables.GroupedObservable;
 import rx.observers.SafeSubscriber;
-import rx.plugins.*;
-import rx.schedulers.*;
+import rx.plugins.RxJavaObservableExecutionHook;
```
akarnokd: Please make sure the imports stay with `*` because it is a source of merge conflict otherwise.

akarnokd:
1) unsubscription test:

use a `PublishSubject`, this operator and `take`, push through values up to the limit and check `hasObservers` on the subject

2) backpressure

use a `range`, this operator and subscribe a `TestSubscriber` which has been created with 0 as an initial value. Use `requestMore(1)` and assert the testsubscriber received only a single value, no error and no completion

src/main/java/rx/Observable.java
thomasnield: Will do.

thomasnield:
@akarnokd Done. The unsubscription test made sense to me. Not too familiar with the `TestSubscriber`. Regardless, I implemented what I think you meant and the tests passed.

thomasnield:
@akarnokd Might help if I actually used the operator in question. One second...

thomasnield:
Okay unit tests are complete.

davidmoten:
Hey Thomas, for your convenience what you normally do is make a branch of 1.x say `doOnEmpty`, do your change in that branch and submit your PR from that branch rather than from 1.x. This PR doesn't have to be changed though.

thomasnield:
Okay, thanks @davidmoten. I'll branch the feature separately next time.

src/test/java/rx/internal/operators/OperatorDoOnEmptyTest.java
```diff
@@ -0,0 +1,113 @@
...
+        final AtomicBoolean wasCalled = new AtomicBoolean(false);
+
+        source.doOnEmpty(new Action0() {
+            @Override
+            public void call() {
+                wasCalled.set(true);
+            }
+        }).subscribe();
+
+        assert(!wasCalled.get());
```
davidmoten: Should use `assertTrue` everywhere in the unit tests rather than `assert` to ensure the assertions are run regardless of  runtime flags.

thomasnield:
Dang it IDEA, sorry my habits have caused Observable to have irrelevant changes. One moment.

akarnokd:
There are too many, probably irrelevant, changes in `Observable.java`. Could you start from a clean file?

In addition, your reuse of `doOnEach` may not work because were sharing the empty flag across subscriptions whose source may not be always empty. For example,

```java
Observable<Integer> o = Observable.defer(
    () -> Observable.range(1, counter.getAndIncrement() % 2))
.doOnEmpty(() -> System.out.println("Empty!"));

o.subscribe(System.out::println);
o.subscribe(System.out::println);
o.subscribe(System.out::println);
o.subscribe(System.out::println);
o.subscribe(System.out::println);
```

This should print 3 `Empty!` but only prints 1.

thomasnield:
That seems like a deliberate side effect almost... so what would the expected behavior be? Because I don't really see how it is much different from `onCompleted()` other than it has a condition before performing the action. I'll try to reason through what you're showing...

thomasnield:
Oh shoot, I think I see what you are saying. I was sharing state *between* the subscriptions. Let me figure out how to do this...

thomasnield:
Alright, I started over on the implementation and placed that `Observable.defer()` case as an additional test. All five unit tests passed.

src/test/java/rx/internal/operators/OperatorDoOnEmptyTest.java
```diff
@@ -0,0 +1,148 @@
...
+        assertTrue(source.hasObservers());
+
+        source.onNext(0);
+        source.onNext(1);
+
+        assertTrue(source.hasObservers());
+
+        source.onNext(2);
+
+        assertTrue(!source.hasObservers());
```
akarnokd: Please use `assertFalse`

src/main/java/rx/Observable.java
```diff
@@ -39,7 +39,7 @@
  * <p>
  * For more information see the <a href="http://reactivex.io/documentation/observable.html">ReactiveX
  * documentation</a>.
- * 
+ *
```
akarnokd: Please make sure your IDE doesn't reformat the `Observable` class.

thomasnield: Okay, I'm going to the notepad and then I'm going to figure out how to turn these features off.

src/test/java/rx/internal/operators/OperatorDoOnEmptyTest.java
stevegury: You forgot one `assertTrue`.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,59 @@
...
+            @Override
+            public void onCompleted() {
+                if (done) {
+                    return;
+                }
+                if (isEmpty) {
+                    try {
+                        onEmpty.call();
+                    } catch (Exception e) {
+                        Exceptions.throwIfFatal(e);
```
davidmoten: If a non-fatal error occurs in `onEmpty.call()` then at the moment it would complete normally. Moreover you should catch for `Throwable`s not just `Exception`s.

This should be 
```java 
catch (Throwable e) {
    Exceptions.throwOrReport(e, this);
    return;
}
```

thomasnield: Okay I will put all these on my list today.

thomasnield:
Question. Why is it the policy that `doOnXXX()` operators won't emit an exception that occurs in the supplied side effect function?  Is the idea that it is up to the client to deal with the exception that occurs in _their _function since it is after all a side-effect? And it is not part of the `Observable`?

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,59 @@
...
+            @Override
+            public void onCompleted() {
+                if (done) {
+                    return;
+                }
+                if (isEmpty) {
+                    try {
+                        onEmpty.call();
+                    } catch (Throwable e) {
+                        Exceptions.throwIfFatal(e);
```
thomasnield: Is this the proper way to handle an error occurring in the side effect function `onEmpty`?

davidmoten: No Thomas, I've given you the way to do it in my previous comment (and the method used calls `onError`).

thomasnield: Okay I thought I missed something in my notes. I'll fix it.

thomasnield: `Exceptions.throwOrReport()`, that's what I was looking for (and I think that is what you said). Thanks for your patience, I'm learning a lot.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,59 @@
...
+            public void onCompleted() {
+                if (done) {
+                    return;
+                }
+                if (isEmpty) {
+                    try {
+                        onEmpty.call();
+                    } catch (Throwable e) {
+                        Exceptions.throwOrReport(e,this);
+                    }
```
davidmoten: add `return;` after call to `Exceptions.throwOrReport`. Thomas I suggest you read the comments on the github site because it seems like you are reacting to truncated comments (perhaps email notifications).

thomasnield: You're right I was reading email notifications.

src/main/java/rx/internal/operators/OperatorDoOnEmpty.java
```diff
@@ -0,0 +1,60 @@
...
+                    try {
+                        onEmpty.call();
+                    } catch (Throwable e) {
+                        Exceptions.throwOrReport(e,this);
+                        return;
+                    }
+                }
+                child.onCompleted();
+                done = true;
+            }
```
davidmoten: Because `onEmpty.call` could take a non-trivial time to complete I would normally put in a `isUnsubscribed` check before calling `child.onCompleted()`. I'd suggest this rejig:

```java
            @Override
            public void onCompleted() {
                if (done) {
                    return;
                }
                if (isEmpty) {
                    try {
                        onEmpty.call();
                    } catch (Throwable e) {
                        Exceptions.throwOrReport(e,this);
                        return;
                    }
                    if (!isUnsubscribed()) {
                        child.onCompleted();
                    }
                } else {
                    child.onCompleted();
                }
                done = true;
            }
```

thomasnield:
Okay it's done, @davidmoten. Please let me know if I missed anything else...

thomasnield:
It looks like a test failed. I don't think it has to do with this feature, but I'll confirm tomorrow after I update my fork and see if that rids it.

davidmoten:
I think this operator was about done so I chucked it into [https://github.com/davidmoten/rxjava-extras](rxjava-extras). That ok with you @thomasnield?

thomasnield:
@davidmoten That sounds great! I'm glad you did, thanks!

Issue #3629, opened 2016-01-21
concatMapEager blows up on null values


Petikoch:
While flatMap and concatMap are fine with flattening null values, concatMapEager blows up.

```
import rx.Observable;

public class RxJavaNullValueTolerance {

	public static void main(String[] args) {
		Observable.just("just me")
				.flatMap(nextString -> Observable.just(null))
				.subscribe(nextObject -> System.out.println("Got: " + nextObject));
		System.out.println("flatMap is fine with null values");

		Observable.just("just me")
				.concatMap(nextString -> Observable.just(null))
				.subscribe(nextObject -> System.out.println("Got: " + nextObject));
		System.out.println("concatMap is fine with null values");

		Observable.just("just me")
				.concatMapEager(nextString -> Observable.just(null))
				.subscribe(nextObject -> System.out.println("Got: " + nextObject));
		System.out.println("concatMapEager is fine with null values");
	}
}
```
(RxJava 1.1.0)

Output:
```
Got: null
flatMap is fine with null values
Got: null
concatMap is fine with null values
Exception in thread "main" rx.exceptions.OnErrorNotImplementedException: null elements not allowed
	at rx.Observable$27.onError(Observable.java:7923)
	at rx.observers.SafeSubscriber._onError(SafeSubscriber.java:159)
	at rx.observers.SafeSubscriber.onError(SafeSubscriber.java:120)
	at rx.internal.operators.OperatorEagerConcatMap$EagerOuterSubscriber.drain(OperatorEagerConcatMap.java:216)
	at rx.internal.operators.OperatorEagerConcatMap$EagerInnerSubscriber.onError(OperatorEagerConcatMap.java:302)
	at rx.Observable.unsafeSubscribe(Observable.java:8105)
	at rx.internal.operators.OperatorEagerConcatMap$EagerOuterSubscriber.onNext(OperatorEagerConcatMap.java:144)
	at rx.internal.util.ScalarSynchronousObservable$1.call(ScalarSynchronousObservable.java:46)
	at rx.internal.util.ScalarSynchronousObservable$1.call(ScalarSynchronousObservable.java:35)
	at rx.Observable$2.call(Observable.java:162)
	at rx.Observable$2.call(Observable.java:154)
	at rx.Observable.subscribe(Observable.java:8191)
	at rx.Observable.subscribe(Observable.java:8158)
	at rx.Observable.subscribe(Observable.java:7914)
[...]

PR #3630, opened 2016-01-21
1.x: ConcatMapEager allow nulls from inner Observables.

akarnokd:
The operator was not allowing `null`s in the inner Observables.

Related: #3629.


zsxwing:
:+1:

PR #3635, opened 2016-01-22
1.x: zip performance measure

akarnokd:
Performance measurement of `zip` and upgrade to JMH 1.11.3.

Results (Intel Celeron 1005M, Windows 7 x64, Java 8u72):

Overall throughput values:

image

Comparing fully sync and fully async dual sources:

image

Comparing when either the first or the second is an async source:

image

So the order of sources matter, having the async first allows it to start early and not after the synchronous source completes its prefetch run.

### DIFF
--- a/src/perf/java/rx/operators/ZipPerf.java
+++ b/src/perf/java/rx/operators/ZipPerf.java
@@ -1,0 +1,140 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


src/perf/java/rx/operators/ZipPerf.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+        };
+        
+        bothSync = Observable.zip(o1, o2, plus);
+
+        firstSync = Observable.zip(o1, o2.subscribeOn(Schedulers.computation()), plus);
+
+        secondSync = Observable.zip(o1.subscribeOn(Schedulers.computation()), o2, plus);
+
+        bothAsync = Observable.zip(o1.subscribeOn(Schedulers.computation()), o2.subscribeOn(Schedulers.computation()), plus);
```
artem-zinnatullin: @akarnokd please correct me if I'm wrong, but on machine (VM) with only one core this won't be async at all because `Schedulers.computation()` has `count of threads == Runtime.getAvailableProcessors()`, I'd replace it with `Schedulers.newThread()`

artem-zinnatullin:
One comment, otherwise looks good :+1:

src/perf/java/rx/operators/ZipPerf.java
akarnokd: The thread churn would be inacceptable in that case as well.

artem-zinnatullin: ok

PR #3637, opened 2016-01-23
handle predicate exceptions properly in skipWhile

davidmoten:
A non-fatal exception thrown by `predicate.call()` in `OperatorSkipWhile` could result in the error being reported by an upstream operator. This PR ensures that the error is reported by the operator in which it occurs.

I've added a unit test for this scenario that failed with the original code, and a couple of other tests.

There are a few more of these floating around that I'll submit PRs for as well.


davidmoten:
unrelated unit test failure `BackpressureTests.testMergeAsync`

akarnokd:
:+1:

zsxwing:
:+1:

PR #3638, opened 2016-01-23
fix error handling in OperatorDistinctUntilChanged

davidmoten:
As per #3637, `keySelector.call()` was not wrapped in an appropriate try catch to prevent the error from being reported from an upstream operator.

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorDistinctUntilChangedTest.java
+++ b/src/test/java/rx/internal/operators/OperatorDistinctUntilChangedTest.java
@@ -15,6 +15,7 @@
  */
 package rx.internal.operators;
 
+import static org.junit.Assert.assertFalse;
 import static org.mockito.Matchers.any;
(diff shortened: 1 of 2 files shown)


akarnokd:
:+1:

artem-zinnatullin:
:+1:

src/test/java/rx/internal/operators/OperatorDistinctUntilChangedTest.java
```diff
@@ -50,6 +53,15 @@ public String call(String s) {
             return s.toUpperCase();
         }
     };
+    
+    final Func1<String, String> THROWS_NON_FATAL = new Func1<String, String>() {
```
zsxwing: nit: not used

davidmoten:
Thanks @zsxwing, it was supposed to have been used and now is. Also made stuff private and static that could be. Commits squashed.

zsxwing:
:+1:

PR #3639, opened 2016-01-23
fix error handling in onBackpressureBuffer

davidmoten:
`onOverflow.call()`  was not wrapped in an appropriate try catch to prevent the error from being reported from an upstream operator.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
+++ b/src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
@@ -22,6 +22,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import rx.Observable.Operator;
 import rx.Producer;
 import rx.Subscriber;
+import rx.exceptions.Exceptions;
 import rx.exceptions.MissingBackpressureException;
 import rx.functions.Action0;
 import rx.internal.util.BackpressureDrainManager;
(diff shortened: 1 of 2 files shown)


akarnokd:
:+1:

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -156,7 +157,15 @@ private boolean assertCapacity() {
                                 "Overflowed buffer of "
                                         + baseCapacity));
                         if (onOverflow != null) {
-                            onOverflow.call();
+                            try {
+                                onOverflow.call();
+                            } catch (Throwable e) {
+                                Exceptions.throwIfFatal(e);
+                                manager.terminateAndDrain(e);
+                                // this line not strictly necessary but nice for clarity
```
artem-zinnatullin: :+1:

artem-zinnatullin:
:+1:

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
```diff
@@ -156,7 +157,15 @@ private boolean assertCapacity() {
                                 "Overflowed buffer of "
                                         + baseCapacity));
                         if (onOverflow != null) {
-                            onOverflow.call();
+                            try {
+                                onOverflow.call();
+                            } catch (Throwable e) {
+                                Exceptions.throwIfFatal(e);
```
zsxwing: Need to notify the plugin instead of swallowing the non fatal exceptions.

akarnokd: It does not swallow it but emits via L164 to the child subscriber.

zsxwing:
:+1:

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
zsxwing: yeah. Didn't notice that

PR #3640, opened 2016-01-23
fix error handling in onBackpressureDrop

davidmoten:
`onDrop.call()`  was not wrapped in an appropriate try catch to prevent the error from being reported from an upstream operator.


akarnokd:
:+1:

zsxwing:
:+1:

PR #3641, opened 2016-01-24
fix error handling in combineLatest

davidmoten:
`combinator.call()` was not wrapped in an appropriate try catch to prevent the error from being reported from an upstream operator.


akarnokd:
#3507 already takes care of this as well, but thanks.

davidmoten:
Ah good, thanks @akarnokd. I guess we can leave this open if it's possible that #3507 doesn't make the next release.  By the way, the unit test in here would probably be nice to have in whichever PR makes it in first.

akarnokd:
I've updated #3507 with your test method.

akarnokd:
#3661 merged

PR #3645, opened 2016-01-25
1.x: fix Amb sharing the choice among all subscribers

akarnokd:
The OnSubscribeAmb shared the choice variable among all of its subscribers which prevented reusing the same Observable.

### DIFF
--- a/src/test/java/rx/internal/operators/OnSubscribeAmbTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeAmbTest.java
@@ -288,5 +288,26 @@ public class OnSubscribeAmbTest {
         }).ambWith(Observable.just(2)).toBlocking().single();
         assertEquals(1, result);
     }
-    
+
+    @Test(timeout = 1000)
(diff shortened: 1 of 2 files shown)


src/test/java/rx/internal/operators/OnSubscribeAmbTest.java
```diff
@@ -288,5 +288,26 @@ public void call(Object t) {
         }).ambWith(Observable.just(2)).toBlocking().single();
         assertEquals(1, result);
     }
-    
+
+    @Test(timeout = 1000)
```
zsxwing: Why not use TestScheduler to avoid using timeout?

akarnokd: Doesn't really matter.

zsxwing: okey. not a big deal.

zsxwing:
:+1:

PR #3647, opened 2016-01-26
Adding support for concurrently inserting actions while advancing time

bcorne:
This change is adding a bit of thread safety when some threads are trying to insert some actions while the time is changed in the TestScheduler


akarnokd:
Hi. Why do you need this feature? TestScheduler is for single-threaded testing.

bcorne:
In fact, TestScheduler is also a very nice tool to use in integration tests. I'm using it in some automated integration tests where the time of the day is important and need to be controlled. These tests are moving the time progressively, checking the output of the software is coherent with the simulated time.

PR #3648, opened 2016-01-26
2.x: Add support for concurrently inserting actions while advancing time

bcorne:
This change is adding a bit of thread safety when some threads are trying to insert some actions while the time is changed in the TestScheduler

### DIFF
--- a/src/test/java/io/reactivex/schedulers/TestSchedulerTest.java
+++ b/src/test/java/io/reactivex/schedulers/TestSchedulerTest.java
@@ -18,6 +18,9 @@ import static org.junit.Assert.fail;
 import static org.mockito.Matchers.anyLong;
 import static org.mockito.Mockito.*;
 
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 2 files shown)


src/test/java/io/reactivex/schedulers/TestSchedulerTest.java
```diff
@@ -216,4 +217,40 @@ public void run() {
...
+        }.start();
+        try
+        {
+            for (int i = 0; i < 10; i++) {
+                Thread.sleep(5);
+                scheduler.advanceTimeBy(1, TimeUnit.SECONDS);
+            }
+        } catch (InterruptedException e)
+        {
+            e.printStackTrace();
```
artem-zinnatullin: Just add `throws InterruptedException` to method signature, it should fail the test, I guess

src/test/java/io/reactivex/schedulers/TestSchedulerTest.java
```diff
@@ -216,4 +217,40 @@ public void run() {
...
+        }
+        try
+        {
+        } catch(NullPointerException e) {
```
artem-zinnatullin: Any reasons to catch NPE here? The whole try-catch looks 100% unneeded

src/test/java/io/reactivex/schedulers/TestSchedulerTest.java
```diff
@@ -216,4 +217,40 @@ public void run() {
...
+                        @Override
+                        public void run()
+                        {
+                            System.out.println(scheduler.now(TimeUnit.MILLISECONDS));
+                        }
+
+                    },1, TimeUnit.SECONDS);
+                }
+            }
+        }.start();
```
artem-zinnatullin: Whole test looks very flaky, I don't see any kind of threads joining, it may run differently each time!

akarnokd: And it hung on CI as well.

bcorne: Yes, I need to rework the UT part

akarnokd:
I'm in for this small change. Could you rebase/redo this for the current 2.x branch?

bcorne:
Sure, will do that today

akarnokd:
Thanks.

Issue #3657, opened 2016-01-31
Sample operator: last item before termination is dropped after sampling


kboyarshinov:
Hi. Recently I found that `sample` operator drops last value before `onCompleted` event. This is behaviour is valid according to these discussions https://github.com/ReactiveX/RxJava/pull/566 and https://github.com/ReactiveX/RxJava/pull/571 and this [test](https://github.com/ReactiveX/RxJava/blob/1.x/src/test/java/rx/internal/operators/OperatorSampleTest.java#L190). It originated in Rx.NET and still exists in RxJava to be consistent across all implementations as far as I understood.

But, marble diagram in documentation shows otherwise: item 5 is emitted before termination and sampled by sampler.
![sample_marble](https://cloud.githubusercontent.com/assets/1917615/12701532/b435ef44-c846-11e5-841f-92eb8d4c4418.png)

This behaviour is odd and unexpectable. I stumbled across it trying to resolve backpressure issue with observables where last emitted item is the most important.

So, my question is whether it can be changed? I prepared [a commit](https://github.com/kboyarshinov/RxJava/commit/29ad810a7d134afa0b43703fcd162e1cfee006de) with fixes and corresponding tests. 
Or at least marble diagram should be changed to showcase this behaviour and it would be good to have a note about it in documentation.

Thank you.

akarnokd:
What about when the source completes and the sampler doesn't?

kboyarshinov:
@akarnokd it would emit the last item from completed source and nothing after that, no?

akarnokd:
Right, the diff tricked me. I expected 3 calls to emit, 1 in sampler's onNext, 1 in sampler's onCompleted and 1 in parent's onCompleted.

Besides, I can see that operator still has some issues with unsubscription and backpressure management.

kboyarshinov:
Yep, 2 calls is enough: 1 in sampler's onNext and 1 in parent's onCompleted.

akarnokd:
So if the sampler completes, then the value should be dropped just like now?

kboyarshinov:
It is ok in my usecase. If sampler completes before source completes it means no values are expected before completion. But if source completes before sampler does it should emit last value. 

However, it worth discussion.

akarnokd:
See #3658 that contains other fixes.

kboyarshinov:
Great! Also `OperatorSampleWithTime` will require fixes too.

akarnokd:
Would you like to submit a PR?

[...]

PR #3658, opened 2016-01-31
1.x: fix unsubscription and producer issues in sample(other)

akarnokd:
This PR fixes 2 bugs with `sample`

- Termination of the main or sampler subscriber unsubscribed the child subscriber which is not allowed.
- The sampler wrapped the child subscriber and thus it allowed setting a producer on the child (thus sampling based on request with some sources).

In addition, #3657 wants to emit the very last item on completion to which I marked the required changes in comments (to be uncommented in a separate PR if needed).

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
+++ b/src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
@@ -16,9 +16,9 @@
 package rx.internal.operators;
 
 import java.util.concurrent.atomic.AtomicReference;
-import rx.Observable;
+
+import rx.*;
 import rx.Observable.Operator;
-import rx.Subscriber;
 import rx.observers.SerializedSubscriber;
 
 /**
@@ -44,7 +44,9 @@ public final class OperatorSampleWithObservable<T, U> implements Operator<T, T> 
     
         final AtomicReference<Object> value = new AtomicReference<Object>(EMPTY_TOKEN);
         
-        Subscriber<U> samplerSub = new Subscriber<U>(child) {
+        final AtomicReference<Subscription> main = new AtomicReference<Subscription>();
+        
+        final Subscriber<U> samplerSub = new Subscriber<U>() {
             @Override
             public void onNext(U t) {
                 Object localValue = value.getAndSet(EMPTY_TOKEN);
@@ -58,15 +60,17 @@ public final class OperatorSampleWithObservable<T, U> implements Operator<T, T> 
             @Override
             public void onError(Throwable e) {
                 s.onError(e);
-                unsubscribe();
+                // no need to null check, main is assigned before any of the two gets subscribed
+                main.get().unsubscribe();
             }
 
             @Override
             public void onCompleted() {
+                // onNext(null); // emit the very last value?
                 s.onCompleted();
-                unsubscribe();
+                // no need to null check, main is assigned before any of the two gets subscribed
+                main.get().unsubscribe();
             }
-            
         };
         
         Subscriber<T> result = new Subscriber<T>() {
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
```diff
@@ -57,16 +59,18 @@ public void onNext(U t) {
...
+                // no need to null check, main is assigned before any of the two gets subscribed
+                main.get().unsubscribe();
                 s.onError(e);
-                unsubscribe();
             }
 
             @Override
             public void onCompleted() {
+                // no need to null check, main is assigned before any of the two gets subscribed
+                main.get().unsubscribe();
```
zsxwing: Why change the order of `unsubscribe` and `onCompleted `? Is it possible that `s.onCompleted` uses some resource in `main` and that will be destroyed in `unsubscribe`?

src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
```diff
@@ -77,18 +81,22 @@ public void onNext(T t) {
...
             @Override
             public void onCompleted() {
+                samplerSub.unsubscribe();
+                // samplerSub.onNext(null); // emit the very last value?
                 s.onCompleted();
-                unsubscribe();
             }
         };
         
+        main.lazySet(result);
```
zsxwing: Why `lazySet` here?

akarnokd: child.add() has a full barrier.

src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
akarnokd: The main and child are on different subscription container so child can't use anything from main. But if you think otherwise, please write a unit test.

zsxwing: The child subscriber may store some resource coming from `main` and use it `onCompleted`, but `main` will close it in its subscription. E.g.,

```
   @Test
    public void foo() {
        Observable<InputStream> source = Observable.create(new OnSubscribe<InputStream>() {
            @Override
            public void call(Subscriber<? super InputStream> subscriber) {
                try {
                    final InputStream input = new URL("http://www.google.com").openStream();
                    subscriber.add(Subscriptions.create(new Action0() {
                        @Override
                        public void call() {
                            try {
                                input.close();
                            } catch (IOException e) {
                                e.printStackTrace();
                            }
                        }
                    }));
                    Thread.sleep(1000);
                    subscriber.onNext(input);
                    Thread.sleep(1000);
                    subscriber.onCompleted();
                } catch (Exception e) {
                    subscriber.onError(e);
                }
            }
        });
        source.sample(Observable.interval(1500, 100, TimeUnit.MILLISECONDS).take(2)).subscribe(new Subscriber<InputStream>() {

            private InputStream lastInputStream;

            @Override
            public void onCompleted() {
                if (lastInputStream != null) {
                    BufferedReader reader = new BufferedReader(new InputStreamReader(lastInputStream));
                    String line = null;
                    try {
                        while ((line = reader.readLine()) != null) {
                            System.out.println(line);
                        }
                    } catch (IOException e) {
                        e.printStackTrace();
                    }
                }
            }

            @Override
                e.printStackTrace();
            }

            @Override
            public void onNext(InputStream inputStream) {
                lastInputStream = inputStream;
            }
        });
    }
```

akarnokd:
Restored original completion-unsubscription order + verification

zsxwing:
:+1:

artem-zinnatullin:
:+1:

PR #3501, opened 2015-11-09
1.x: Gradle Wrapper 2.8, RxJava project plugin 3.0.0

artem-zinnatullin:
// Actually, I am slowly preparing RxJava for running tests on Android during CI.

Netflix team: PTAL, hope you're ok with new rxjava-project plugin.

PR #3659, opened 2016-02-01
Add wrapper task and update to gradle 2.10

hzsweers:
Was using a pretty old version of gradle. Tests all run normally for me locally with this

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,3 +33,8 @@ if (project.hasProperty('release.useLastTag')) {
 test{
      maxHeapSize = "2g"
 }
+
+task wrapper(type: Wrapper) {
+    gradleVersion = '2.10'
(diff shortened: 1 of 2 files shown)


JakeWharton:
You can rarely update Gradle when you use Netflix's Nebula plugins. They are violently linked to implementation details because reasons.

hzsweers:
Ah, guess that explains why there haven't been any updates lately. Should I close this then?

JakeWharton:
Or step the version backwards until it works. It's a shame since the speed
improvements in 2.4 and 2.9+ are great. For RxAndroid we purged all the
Nebula plugins because they held the project back.

hzsweers:
eesh, yeah not even 2.3 will work. I was misled locally since the `wrapper` task worked the first time, but that made sense since `2.2.1` worked fine. Closing for now then. Wonder what the status is, came across #3501 after digging a bit more and seems it was closed without explanation.

PR #3660, opened 2016-02-01
Add combineLatest overload for Collection

JakeWharton:
2.x already uses `Iterable`, but that's a very drastic change compared to just `Collection`.

Motivation here is that I'm using set bindings to create observables in a dependency injector and I want to skip the current `new ArrayList<>(sources)` that I have to do on the set.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -786,6 +786,31 @@ public class Observable<T> {
         return combineLatest(Arrays.asList(o1, o2, o3, o4, o5, o6, o7, o8, o9), Functions.fromFunc(combineFunction));
     }
     /**
+     * Combines a collection of source Observables by emitting an item that aggregates the latest values of each of
+     * the source Observables each time an item is received from any of the source Observables, where this
+     * aggregation is defined by a specified function.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code combineLatest} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <T>
+     *            the common base type of source values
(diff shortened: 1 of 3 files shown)


akarnokd:
Can we revisit this once #3507 is merged?

JakeWharton:
Yep. That allows exposing `Iterable` as well which is even better.

src/main/java/rx/Observable.java
```diff
@@ -786,6 +786,31 @@ public Completable toCompletable() {
...
+     *            the result type
+     * @param sources
+     *            the collection of source Observables
+     * @param combineFunction
+     *            the aggregation function used to combine the items emitted by the source Observables
+     * @return an Observable that emits items that are the result of combining the items emitted by the source
+     *         Observables by means of the given aggregation function
+     * @see <a href="http://reactivex.io/documentation/operators/combinelatest.html">ReactiveX operators documentation: CombineLatest</a>
+     */
+    public static <T, R> Observable<R> combineLatest(Collection<? extends Observable<? extends T>> sources, FuncN<? extends R> combineFunction) {
```
artem-zinnatullin: `final`? :trollface:

artem-zinnatullin:
:+1: 

Don't you want to add tests for your new `Observable.combineLatest()`? `OnSubscribeCombineLatest.java` is internal, will be great to cover public API as well for possible changes in future implementation and to test operator from user's point of view.

JakeWharton:
The tests were modified to use it.

artem-zinnatullin:
Tests for `OnSubscribeCombineLatest` were modified, but I'm talking about tests for `Observable.combineLatest(Collection, FuncN)`.

JakeWharton:
They use that API.

PR #3662, opened 2016-02-01
1.x: fix doOnRequest premature requesting.

akarnokd:
When the operator is assembled and it sets the producer on the child, if the child doesn't request until the assembly is complete, `doOnRequest` requested the default Long.MAX_VALUE.

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorDoOnRequestTest.java
+++ b/src/test/java/rx/internal/operators/OperatorDoOnRequestTest.java
@@ -1,19 +1,16 @@
 package rx.internal.operators;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.*;
+import java.util.concurrent.atomic.*;
 
-import org.junit.Test;
+import org.junit.*;
 
+import rx.*;
 import rx.Observable;
-import rx.Subscriber;
-import rx.functions.Action0;
-import rx.functions.Action1;
+import rx.Observable.OnSubscribe;
+import rx.functions.*;
 
 public class OperatorDoOnRequestTest {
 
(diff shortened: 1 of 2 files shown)


davidmoten:
Good find thanks @akarnokd

src/test/java/rx/internal/operators/OperatorDoOnRequestTest.java
```diff
@@ -76,5 +73,51 @@ public void onNext(Integer t) {
...
+            
+            @Override
+            public void setProducer(Producer p) {
+                producer.set(p);
+            }
+        });
+        
+        producer.get().request(1);
+        
+        System.out.println(requested);
```
davidmoten: May as well cull the sysout to reduce test noise

akarnokd:
Done.

src/test/java/rx/internal/operators/OperatorDoOnRequestTest.java
```diff
@@ -76,5 +73,55 @@ public void onNext(Integer t) {
                 });
         assertEquals(Arrays.asList(3L,1L,2L,3L,4L,5L), requests);
     }
+    
+    @Test
+    public void dontRequestIfDownstreamRequestsLate() {
```
artem-zinnatullin: Yay! You did it! `@Test` without `test` prefix, oh, I'm so happy now :smile_cat:

src/test/java/rx/internal/operators/OperatorDoOnRequestTest.java
```diff
@@ -76,5 +73,55 @@ public void onNext(Integer t) {
...
+            @Override
+            public void setProducer(Producer p) {
+                producer.set(p);
+            }
+        });
+        
+        producer.get().request(1);
 
+        int s = requested.size();
+        if (s == 1) {
```
artem-zinnatullin: ? this may work differently from time to time or you want to support changes in future?

akarnokd: This is for a future case where doOnRequest doesn't request anything on its own. Didn't want to mess with that right now.

artem-zinnatullin: well…ok

artem-zinnatullin:
:+1:

zsxwing:
:+1:

PR #3668, opened 2016-02-04
2.x: full JDK 6 compatible backport + including bugfixes up to today

akarnokd:
Due to the issue with AtomicXFieldUpdaters on certain Android devices, I had to manually replace all of them with regular AtomicX classes.


JakeWharton:
This is basically unreviewable, so I vote that it is just merged. People can go through the code base passively and bring up any issues/questions/etc. as they're discovered.

akarnokd:
It's not possible incrementally anyway, i.e., by using our own Function, it affects all operators that take Function.

zsxwing:
:+1: Agree

src/main/java/io/reactivex/Optional.java
```diff
@@ -0,0 +1,80 @@
...
+        Objects.requireNonNull(value, "value is null");
+        return new Optional<T>(value);
+    }
+    
+    public boolean isPresent() {
+        return value != null;
+    }
+    
+    public T get() {
+        return value;
```
artem-zinnatullin: This method throws exception if value is `null` in JDK 8'th impl, is this intended, @akarnokd?

akarnokd: By us, it's `of` that throws before even creating the optional.

JakeWharton: Code can (through abstraction) end up doing `Optional.empty().get()` which leaks a null reference into the system though.

artem-zinnatullin: Yup, exactly, deserves a `throw` in my opinion, to prevent leaking that `null` into user's code. 

Also

stevegury: Agree, `of` and `get` need to check for null.

stevegury:
I'm ok for merging it as is. 
Let's review it incrementally.

hzsweers:
:+1: for all of this

akarnokd:
Merged. Happy reviewing!

PR #3677, opened 2016-02-08
1.x: negative argument check for skip's count and merge's maxConcurrent

zsxwing:
As per #3675, we should also check the `count` argument of `skip`. Also added the missing check for `merge`'s `maxConcurrent`.


artem-zinnatullin:
I'd suggest to check exception message here and in the other tests, otherwise :+1:

zsxwing:
> I'd suggest to check exception message here and in the other tests, otherwise

Done

artem-zinnatullin:
Thanks, @zsxwing! :+1: :+1:

akarnokd:
:+1:

stevegury:
:+1:

Issue #3674, opened 2016-02-05
concat not subscribing to the next source if requested is at 0


akarnokd:
The following test fails because the sequence misses the last value of the source:

```java
    @Test
    public void publishConcatBackpressure() {
        TestSubscriber<Integer> ts = TestSubscriber.create(5);
        
        Observable.range(1, 6)
        .publish(o -> Observable.concat(o.take(5), o.takeLast(5)))
        .subscribe(ts);
        
        ts.requestMore(1);
        
        ts.assertValues(1, 2, 3, 4, 5, 6);
        ts.assertNoErrors();
        ts.assertCompleted();
    }
```

The operator `publish()` has the property that if all subscribers unsubscribe but the upstream hasn't completed yet, it "slowly" consumes the upstream until a subscriber shows up (at which case it consumes the upstream at the rate of the slowest subscriber again). This may be a discussable property (i.e., why not "pause" until more interesting parties show up), but when it is combined with `concat`, unexpected data loss may happen.

`concat` seems to be implemented in a way that if a source completes on a request boundary, that is downstream's requested count is zero, it doesn't subscribe to the next immediately but only when the downstream requests more. This might be a residue from a case where the next source ignores backpressure (`just`?) and would flood the downstream that didn't request anything yet. However, such sources would ignore any request amount anyway so this behavior might be unnecessary.

In the test above, this is simulated by an initial request of 5. After the `subscribe` returns, however, the `publish()` already consumed the rest of the source so the `requestMore(1)` making the `concat` subscribe to the second part, there is no value available anymore. (Besides, it won't complete either, see #3673 .)

I'm not sure if these two behaviors are bugs or just unexpected but legal outcomes, so my question is what should we do about them. I believe, tests exist that verify both `concat` and `publish()` for their current behavior.

The fix to `publish()` would be to remove the `n == 0` case in its drain loop, making it pause when nobody is interested; and the fix to `concat` would be a different completion-subscription trampolining that subscribes immediately to the next but doesn't request anything (or zero). 

Both can individually ensure the continuity of values in the example above.

Thoughts?

hzsweers:
Might want to @ mention people to get this on their radar

stevegury:
It makes sense to pause the publish operator when no one is subscribed, and now `concat` have to  subscribe immediately to the next source without requesting any data (until the subscriber does).

akarnokd:
See #3681 that solves this issue without changing `concat` whose behavior is somewhat expected.

[...]

Issue #3673, opened 2016-02-05
publish(Func1) may leave operators hang


akarnokd:
The following unit test fails because the sequence doesn't complete:

```java
    public void publishConcat() {
        TestSubscriber<Integer> ts = TestSubscriber.create();
        
        Observable.range(1, 3)
        .publish(o -> o.take(5).concatWith(o.takeLast(5)))
        .subscribe(ts);
        
        ts.assertValues(1, 2, 3);
        ts.assertNoErrors();
        ts.assertCompleted();
    }
```
 The problem lies in operator `publish()`. When a published source completes, the operator switches back to a ready state where more subscribers can be subscribed to it and they all receive values from the upstream on the next `connect()`. Otherwise, it wouldn't be possible to observe the sequence again from the beginning by more than one subscriber.

In the test, the source is sort and by the time take(5) completes and makes `concat` subscribe to `o`, the published source, the underlying OperatorPublish is already in the ready state and awaits another connect, which will never arrive, thus leaving the second takeLast hang forever.

This anomaly is there since the rewrite of `publish()` to support backpressure, namely when it doubles as a multicast source. Previously, a PublishSubject was the mediator between the upstream and the use places in the lambda. PublishSubject, when completed will complete any latecommer subscriber such as the one by `takeLast()` thus the unit test passes.

The solution would be to write a request-coordinating internal PublishSubject variant and dispatch upstream values over it in the multicast scenario.

(Note that unbounded `replay` has no such problem because it stays in the completed state until the next `connect()` call, thus latecommers still receive the orignal sequence. Because it buffers all values, there is no penalty for latecommers after a new `connect()` call.)

akarnokd:
Closing via #3681

PR #3681, opened 2016-02-08
1.x: change publish(Func1) to use a dedicated subject-like dispatcher

akarnokd:
Solves the problems #3673 and #3674 by introducing a new subject-like dispatcher class `OnSubscribePublishMulticast` that does support backpressure, pauses when there are no subscribers and signals the terminal event for late subscribers; all ideal for the multicasting `publish(Func1)` overload.

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorPublishFunctionTest.java
+++ b/src/test/java/rx/internal/operators/OperatorPublishFunctionTest.java
@@ -1,0 +1,255 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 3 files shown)


akarnokd:
Updated with more unit tests and behavior clarification/validation (`publish(Func1)` had almost no tests).

src/test/java/rx/internal/operators/OperatorPublishFunctionTest.java
```diff
@@ -0,0 +1,255 @@
...
+            }
+        }, true).unsafeSubscribe(ts);
+        
+        for (int i = 0; i < RxRingBuffer.SIZE * 2; i++) {
+            ps.onNext(i);
+        }
+        
+        ts.requestMore(RxRingBuffer.SIZE);
+        
+        ts.assertValueCount(128);
```
stevegury: Should you use RxRingBuffer.SIZE here?

stevegury:
Other than this minor comment, :+1:

PR #3682, opened 2016-02-09
1.x: fix observeOn resource handling, add delayError capability

akarnokd:
This PR fixes the "messing around" reported in #3002 and adds an overload to `observeOn` that allows delaying errors without the need for wrapping (see #3542 and maybe there are other reports).

In addition, this PR adds a proper override of the `isEmpty` method to simply compare the two indexes for emptiness directly instead of `size() == 0` to avoid looping, multi-reading and casting.

Benchmark comparison (i7 4790, Windows 7 x64, Java 8u66):

image

Note that the benchmark is generally quite noisy, yielding hectic results (i.e., firing up a thread with newThread may take quite some random microseconds). For example, `observeOnImmediate` shouldn't be affected by any of the changes yet the run-to-run variance is +/- 10%. I'm fine with the results of the benchmark.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -5999,7 +5999,9 @@ public class Observable<T> {
     
     /**
      * Modifies an Observable to perform its emissions and notifications on a specified {@link Scheduler},
-     * asynchronously with an unbounded buffer.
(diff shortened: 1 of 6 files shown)


stevegury:
:+1:

Issue #3347, opened 2015-09-16
When is OnErrorThrowable created and received?


mgp:
From its Javadoc:

> Represents a `Throwable` that an `Observable` might notify its subscribers of, but that then can be handled by an operator that is designed to recover from or react appropriately to such an error.

From this description I was thinking that operators like `onErrorResumeNext` might wrap emitted `Throwables` in an `OnErrorThrowable` instance using its `from` static factory method, and then pass that `OnErrorThrowable` instance to the global error handler. But the `onError` method of `OperatorOnErrorResumeNextViaFunction` looks like:

```java
@Override
public void onError(Throwable e) {
    if (done) {
        Exceptions.throwIfFatal(e);
        return;
    }
    done = true;
    try {
        RxJavaPlugins.getInstance().getErrorHandler().handleError(e);

        ...
```

And so the "raw" `Throwable` is passed to the global error handler instead.

The problem we're trying to solve is this: the global error handler in our app forwards errors to Crashlytics. But we want to only forward `Throwable` instances that we're not already recovering from by methods like `onErrorResumeNext`. I was hoping that `onErrorResumeNext` would wrap such errors in `OnErrorThrowable` or something equivalent and our global error handler could only forward errors if they are not of this type.

Any thoughts on how we can achieve this behavior? Thanks!

akarnokd:
Looks like that handler call is at the wrong place and belongs to after throwIfFatal. I'll post a PR for this.

akarnokd:
See #3685

akarnokd:
Closing via #3689

PR #3685, opened 2016-02-09
1.x: onErrorResumeNext(Func1) should not call plugin handler there

akarnokd:
When the operator switches to the other, that counts as a handled error. 

See also #3347.


benkomalo:
There are similar patterns in the other `onError*` operators, namely `OperatorOnErrorResumeNextViaObservable` and `OperatorOnErrorReturn` - do those need changing as well?

akarnokd:
Yes, I'll update this PR.

akarnokd:
Updated, although all of these classes have backpressure bugs in them.

akarnokd:
Replaced by #3689

Issue #3687, opened 2016-02-09
NullPointerException on OperatorZip


jkettmann:
I get crashreports caused by a NullPointerException in OperatorZip's tick() method. You find the stacktrace and my code, causing the problem below.

The responsible line in tick() method is

```java
RxRingBuffer buffer = ((InnerSubscriber) observers[i]).items;
```

where observers[i] seems to be null. 

The app crashes only rarely on one device, which I don't have acces to (Samsung S3, Android 4.3).  The reason seems to be, that  the property `missedProducer` in `ProducerArbiter` must be set, which is never the case on my devices.

We're not in production yet, so I can not tell how many other devices will show this error. 

Stacktrace:
```
0 java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.Worker thread.
1	at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:62)
2	at rx.schedulers.ExecutorScheduler$ExecutorSchedulerWorker.run(ExecutorScheduler.java:98)
3	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1080)
4	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:573)
5	at java.lang.Thread.run(Thread.java:841)
6 Caused by: java.lang.NullPointerException
7	at rx.internal.operators.OperatorZip$Zip.tick(OperatorZip.java:237)
8	at rx.internal.operators.OperatorZip$ZipProducer.request(OperatorZip.java:175)
9	at rx.internal.producers.ProducerArbiter.emitLoop(ProducerArbiter.java:181)
10	at rx.internal.producers.ProducerArbiter.request(ProducerArbiter.java:69)
11	at rx.internal.operators.OperatorConcat$ConcatSubscriber.requestFromChild(OperatorConcat.java:118)
12	at rx.internal.operators.OperatorConcat$ConcatSubscriber.access$100(OperatorConcat.java:79)
13	at rx.internal.operators.OperatorConcat$ConcatProducer.request(OperatorConcat.java:74)
14	at rx.internal.producers.ProducerArbiter.request(ProducerArbiter.java:66)
15	at rx.internal.operators.OperatorSubscribeOn$1$1$1$1$1.call(OperatorSubscribeOn.java:94)
16	at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
17	... 4 more
```

[...]

PR #3688, opened 2016-02-09
1.x: Fix zip() - observer array becoming visible too early and causing NPE

akarnokd:
Reported in #3687.

By the way, I see several other small problems with zip: it is implemented as an `Operator` and even though one can know the number of sources upfront, this indirection hides the fact. In addition, it uses SpMc queue but could just use Spsc because there is only one drain thread at a time.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorZip.java
+++ b/src/main/java/rx/internal/operators/OperatorZip.java
@@ -177,7 +177,10 @@ public final class OperatorZip<R> implements Operator<R, Observable<?>[]> {
 
     }
 
-    private static final class Zip<R> extends AtomicLong {
+    static final class Zip<R> extends AtomicLong {
+        /** */
+        private static final long serialVersionUID = 5995274816189928317L;
+        
         final Observer<? super R> child;
         private final FuncN<? extends R> zipFunction;
         private final CompositeSubscription childSubscription = new CompositeSubscription();
@@ -186,7 +189,7 @@ public final class OperatorZip<R> implements Operator<R, Observable<?>[]> {
         int emitted = 0; // not volatile/synchronized as accessed inside COUNTER_UPDATER block
 
         /* initialized when started in `start` */
-        private Object[] observers;
+        private volatile Object[] subscribers;
         private AtomicLong requested;
 
         public Zip(final Subscriber<? super R> child, FuncN<? extends R> zipFunction) {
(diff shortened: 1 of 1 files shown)


src/main/java/rx/internal/operators/OperatorZip.java
```diff
@@ -177,7 +177,10 @@ public void request(long n) {
 
     }
 
-    private static final class Zip<R> extends AtomicLong {
+    static final class Zip<R> extends AtomicLong {
+        /** */
+        private static final long serialVersionUID = 5995274816189928317L;
```
artem-zinnatullin: We really need it? Also, javadoc comment above does not make sense

src/main/java/rx/internal/operators/OperatorZip.java
```diff
@@ -197,14 +200,16 @@ public Zip(final Subscriber<? super R> child, FuncN<? extends R> zipFunction) {
...
+            Object[] observers = new Object[os.length];
             for (int i = 0; i < os.length; i++) {
                 InnerSubscriber io = new InnerSubscriber();
                 observers[i] = io;
                 childSubscription.add(io);
             }
-
+            
+            this.requested = requested;
+            this.observers = observers; // full memory barrier: release all above
```
artem-zinnatullin: :+1:

src/main/java/rx/internal/operators/OperatorZip.java
```diff
@@ -197,14 +200,16 @@ public Zip(final Subscriber<? super R> child, FuncN<? extends R> zipFunction) {
 
         @SuppressWarnings("unchecked")
         public void start(@SuppressWarnings("rawtypes") Observable[] os, AtomicLong requested) {
-            observers = new Object[os.length];
-            this.requested = requested;
+            Object[] observers = new Object[os.length];
```
artem-zinnatullin: `final` and different name please to avoid misusage by mistake

artem-zinnatullin:
Few small things, otherwise lgtm :+1:

src/main/java/rx/internal/operators/OperatorZip.java
akarnokd: Less javac warnings.

akarnokd:
Updated.

stevegury:
:+1:

PR #3689, opened 2016-02-09
1.x: unified onErrorX and onExceptionResumeNext and fixed backpressure

akarnokd:
The operators didn't honor backpressure when they switched over to the other Observable.

This PR also includes the fix #3685.


akarnokd:
There was a test failure in BackpressureTests.testMergeAsync, run before any of the changes of this PR and shouldn't have any effect. Overloaded servers?

src/test/java/rx/internal/operators/OperatorOnErrorReturnTest.java
```diff
@@ -217,6 +219,33 @@ public void run() {
...
+        
+        ps.onNext(1);
+        ps.onNext(2);
+        ps.onError(new TestException("Forced failure"));
+
+        ts.assertValues(1, 2);
+        ts.assertNoErrors();
+        ts.assertNotCompleted();
+
+        ts.requestMore(2);
```
stevegury: Not a big deal, but it should be `requestMore(1)` if you want to have the same behavior (than the other tests).

stevegury:
:+1:

src/test/java/rx/internal/operators/OperatorOnErrorReturnTest.java
akarnokd: It's deliberately overrequesting here.

PR #3692, opened 2016-02-10
1.x: Fix Completable using JDK 7 suppressed exceptions feature

akarnokd:
The upgraded gradle switched my IDE to use a JDK 6 library I have installed all along and these popped up immediately.


zsxwing:
:+1:

stevegury:
:+1:

zsxwing:
Although nebula gradle plugin requires Java 7, I'm still wondering if we can setup a build to compile codes with Java 6 to avoid such issue. E.g., I can just use `find src/main/java/* -name *.java | xargs javac` to compile them using Java 6 and see if any error.

PR #3696, opened 2016-02-11
1.x: Added Single execution hooks

### LINKED ISSUE #3595: Execution hook for Single
`RxJavaObservableExecutionHook` is not compatible with [`Single`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Single.java) type. For example: [`RxJavaObservableExecutionHook#onCreate`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/plugins/RxJavaObservableExecutionHook.java#L54) method tooks [`Observable#OnSubscribe`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L101) type as a parameter. However, `Single` is using it's own [`OnSubscribe`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Single.java#L149) type. Shall we implement new class to handle `Single`'s hooks let's say `RxJavaSingleExecutionHook` ?

phajduk:
Resolves #3595

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -29,6 +29,7 @@ import rx.observers.SafeSubscriber;
 import rx.observers.SerializedSubscriber;
 import rx.plugins.RxJavaObservableExecutionHook;
 import rx.plugins.RxJavaPlugins;
+import rx.plugins.RxJavaSingleExecutionHook;
 import rx.schedulers.Schedulers;
 import rx.singles.BlockingSingle;
 import rx.subscriptions.Subscriptions;
@@ -101,7 +102,7 @@ public class Single<T> {
         this.onSubscribe = f;
     }
 
-    static final RxJavaObservableExecutionHook hook = RxJavaPlugins.getInstance().getObservableExecutionHook();
+    static RxJavaSingleExecutionHook hook = RxJavaPlugins.getInstance().getSingleExecutionHook();
 
     /**
      * Returns a Single that will execute the specified function when a {@link SingleSubscriber} executes it or
@@ -130,7 +131,7 @@ public class Single<T> {
      * @see <a href="http://reactivex.io/documentation/operators/create.html">ReactiveX operators documentation: Create</a>
      */
     public static <T> Single<T> create(OnSubscribe<T> f) {
-        return new Single<T>(f); // TODO need hook 
+        return new Single<T>(hook.onCreate(f));
     }
 
     /**
@@ -1570,14 +1571,12 @@ public class Single<T> {
      * @param subscriber
      *            the Subscriber that will handle the emission or notification from the Single
      */
-    public final void unsafeSubscribe(Subscriber<? super T> subscriber) {
+    public final Subscription unsafeSubscribe(Subscriber<? super T> subscriber) {
         try {
             // new Subscriber so onStart it
             subscriber.onStart();
-            // TODO add back the hook
-            //            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
-            onSubscribe.call(subscriber);
-            hook.onSubscribeReturn(subscriber);
+            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
+            return hook.onSubscribeReturn(subscriber);
         } catch (Throwable e) {
             // special handling for certain Throwable/Error/Exception types
             Exceptions.throwIfFatal(e);
@@ -1594,6 +1593,7 @@ public class Single<T> {
                 // TODO why aren't we throwing the hook's return value.
                 throw r;
             }
+            return Subscriptions.unsubscribed();
         }
     }
 
@@ -1685,9 +1685,7 @@ public class Single<T> {
         // The code below is exactly the same an unsafeSubscribe but not used because it would add a sigificent depth to alreay huge call stacks.
         try {
             // allow the hook to intercept and/or decorate
-            // TODO add back the hook
-            //            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
-            onSubscribe.call(subscriber);
+            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
             return hook.onSubscribeReturn(subscriber);
         } catch (Throwable e) {
             // special handling for certain Throwable/Error/Exception types
--- a/src/main/java/rx/plugins/RxJavaSingleExecutionHookDefault.java
+++ b/src/main/java/rx/plugins/RxJavaSingleExecutionHookDefault.java
@@ -1,0 +1,28 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.plugins;
+
+/**
+ * Default no-op implementation of {@link RxJavaSingleExecutionHook}
+ */
+class RxJavaSingleExecutionHookDefault extends RxJavaSingleExecutionHook {
+
+    private static final RxJavaSingleExecutionHookDefault INSTANCE = new RxJavaSingleExecutionHookDefault();
+
+    public static RxJavaSingleExecutionHook getInstance() {
+        return INSTANCE;
+    }
+}
--- a/src/test/java/rx/plugins/RxJavaPluginsTest.java
+++ b/src/test/java/rx/plugins/RxJavaPluginsTest.java
@@ -16,6 +16,7 @@
 package rx.plugins;
 
 import static org.junit.Assert.*;
+import static org.mockito.Mockito.mock;
 
 import java.util.*;
 import java.util.concurrent.TimeUnit;
@@ -105,6 +106,15 @@ public class RxJavaPluginsTest {
         p.registerObservableExecutionHook(new RxJavaObservableExecutionHookTestImpl());
         RxJavaObservableExecutionHook impl = p.getObservableExecutionHook();
         assertTrue(impl instanceof RxJavaObservableExecutionHookTestImpl);
+    }
+
+    @Test
+    public void testSingleExecutionHookViaRegisterMethod() {
+        RxJavaPlugins p = new RxJavaPlugins();
+        RxJavaSingleExecutionHook customHook = mock(RxJavaSingleExecutionHook.class);
+        p.registerSingleExecutionHook(customHook);
+        RxJavaSingleExecutionHook impl = p.getSingleExecutionHook();
+        assertSame(impl, customHook);
     }
 
     @Test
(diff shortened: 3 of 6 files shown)


akarnokd:
Would you also apply these plugin methods in Single?

  - onSubscribeStart
  - onSubscribeReturn
  - onSubscribeError
  - onLift

phajduk:
Beloew hooks are already called (I replaced previous calls of `RxJavaObservableExecutionHook` by `RxJavaSingleExecutionHook` calls).
- onSubscribeReturn
- onSubscribeError
- onLift

However there is problem with `onSubscribeStart`. This hook should be called in `onSubscribe` and `unsafeSubscribe`. Unfortunately, in `Single` `onSubscribe` and `unsafeSubscribe` methods get `Subscriber` type as parameter instead of `SingleSubscriber`. I will make a commit to cover it soon.

akarnokd:
There is something wrong, a bunch of Single test failed.

akarnokd:
:+1:

src/main/java/rx/Single.java
```diff
@@ -1569,14 +1569,12 @@ public final void onNext(T args) {
      * @param subscriber
      *            the Subscriber that will handle the emission or notification from the Single
      */
-    public final void unsafeSubscribe(Subscriber<? super T> subscriber) {
+    public final Subscription unsafeSubscribe(Subscriber<? super T> subscriber) {
```
artem-zinnatullin: Looks like this is incompatible binary change https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.15

src/main/java/rx/plugins/RxJavaSingleExecutionHookDefault.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package rx.plugins;
+
+/**
+ * Default no-op implementation of {@link RxJavaSingleExecutionHook}
+ */
+/* package */class RxJavaSingleExecutionHookDefault extends RxJavaSingleExecutionHook {
+
+    private static RxJavaSingleExecutionHookDefault INSTANCE = new RxJavaSingleExecutionHookDefault();
```
artem-zinnatullin: `final`

src/main/java/rx/plugins/RxJavaSingleExecutionHookDefault.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.plugins;
+
+/**
+ * Default no-op implementation of {@link RxJavaSingleExecutionHook}
+ */
+/* package */class RxJavaSingleExecutionHookDefault extends RxJavaSingleExecutionHook {
```
artem-zinnatullin: `/* package */` is kind of weird :) Let's remove it?

src/test/java/rx/plugins/RxJavaPluginsTest.java
```diff
@@ -108,6 +108,14 @@ public void testObservableExecutionHookViaRegisterMethod() {
     }
 
     @Test
+    public void testSingleExecutionHookViaRegisterMethod() {
```
artem-zinnatullin: no need for `test` prefix

src/test/java/rx/plugins/RxJavaPluginsTest.java
```diff
@@ -108,6 +108,14 @@ public void testObservableExecutionHookViaRegisterMethod() {
...
+        RxJavaPlugins p = new RxJavaPlugins();
+        p.registerSingleExecutionHook(new RxJavaSingleExecutionHookTestImpl());
+        RxJavaSingleExecutionHook impl = p.getSingleExecutionHook();
+        assertTrue(impl instanceof RxJavaSingleExecutionHookTestImpl);
```
artem-zinnatullin: Please change to `assertSame()`

src/test/java/rx/plugins/RxJavaPluginsTest.java
```diff
@@ -238,6 +246,11 @@ public Date call(Date date) {
         // just use defaults
     }
 
+    // inside test so it is stripped from Javadocs
+    public static class RxJavaSingleExecutionHookTestImpl extends RxJavaSingleExecutionHook {
```
artem-zinnatullin: You can just use `Mockito.mock(RxJavaSingleExecutionHook.class)` for the test

src/main/java/rx/Single.java
```diff
@@ -129,7 +129,7 @@ private Single(final Observable.OnSubscribe<T> f) {
      * @see <a href="http://reactivex.io/documentation/operators/create.html">ReactiveX operators documentation: Create</a>
      */
     public static <T> Single<T> create(OnSubscribe<T> f) {
-        return new Single<T>(f); // TODO need hook 
+        return new Single<T>(hook.onCreate(f));
```
artem-zinnatullin: Needs test

src/main/java/rx/Single.java
```diff
@@ -1569,14 +1569,12 @@ public final void onNext(T args) {
...
-    public final void unsafeSubscribe(Subscriber<? super T> subscriber) {
+    public final Subscription unsafeSubscribe(Subscriber<? super T> subscriber) {
         try {
             // new Subscriber so onStart it
             subscriber.onStart();
-            // TODO add back the hook
-            //            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
-            onSubscribe.call(subscriber);
-            hook.onSubscribeReturn(subscriber);
+            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
```
artem-zinnatullin: needs test

src/main/java/rx/Single.java
```diff
@@ -1569,14 +1569,12 @@ public final void onNext(T args) {
...
+            return hook.onSubscribeReturn(subscriber);
```
artem-zinnatullin: needs test

src/main/java/rx/Single.java
```diff
@@ -1593,6 +1591,7 @@ public final void unsafeSubscribe(Subscriber<? super T> subscriber) {
                 // TODO why aren't we throwing the hook's return value.
                 throw r;
             }
+            return Subscriptions.unsubscribed();
```
artem-zinnatullin: needs test

src/main/java/rx/Single.java
```diff
@@ -1660,9 +1659,7 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
         // The code below is exactly the same an unsafeSubscribe but not used because it would add a sigificent depth to alreay huge call stacks.
         try {
             // allow the hook to intercept and/or decorate
-            // TODO add back the hook
-            //            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
-            onSubscribe.call(subscriber);
+            hook.onSubscribeStart(this, onSubscribe).call(subscriber);
```
artem-zinnatullin: needs test

src/main/java/rx/Single.java
phajduk: I strive to be as much as possible compatible with `Observable` type. Do you have some proposition how to handle it? If we remove `Subscription` as return type then we won't be able to decorate or replace the `Subscription` instance in hook.

src/main/java/rx/plugins/RxJavaSingleExecutionHookDefault.java
phajduk: I see similar issues in other places in RxJava codebase Shall I fix it in this PR or create separated one?

src/test/java/rx/plugins/RxJavaPluginsTest.java
phajduk: Majority of test methods which we have are prefixed by `test`. [Contribution](https://github.com/ReactiveX/RxJava/blob/1.x/CONTRIBUTING.md) doc states to follow existing conventions. Do we have some separated guidelines for tests convention somewhere?

src/main/java/rx/Single.java
artem-zinnatullin: Right right, I mean that we will have to release it as 1.2 instead of 1.1.x,  I guess. 

Maintainers: what do you think?

src/main/java/rx/plugins/RxJavaSingleExecutionHookDefault.java
artem-zinnatullin: Please fix your case in this PR and feel free to submit separate one for other cases

src/test/java/rx/plugins/RxJavaPluginsTest.java
artem-zinnatullin: Right, but `@Test` clearly shows that it's a test, I hope we'll have some kind of codestyle document, IDE settings and checkstyle rules in future

src/main/java/rx/Single.java
akarnokd: Unsafesubscribe is called by operators and they don't use the returned Subscription. These unsubscribe the Subscriber.

JakeWharton: Single isn't stable API so this is acceptable if it makes the API better.

artem-zinnatullin: k, let's just mention this as kind of warning in changelog

src/test/java/rx/plugins/RxJavaPluginsTest.java
phajduk: If so then I believe it would be better to leave it as is and fix all test methods in bulk when we prepare IDE settings and SCA tools. What do you think?

artem-zinnatullin: k, as you wish, it doesn't really matter of course

src/test/java/rx/plugins/RxJavaPluginsTest.java
phajduk: I get your point. However I think `mock` won't help here as I need default behavior of default hook (otherwise in each test we will need to setup mock behavior). I would like to use `spy` to make both `verify` and default behavior possible. I'll make a commit soon so you'll be able to check my proposition.

phajduk:
As far I see that failing test is not related with my changes. Is `rx.schedulers.CachedThreadSchedulerTest > testUnSubscribeForScheduler` some kind of flaky test?

src/main/java/rx/Single.java
stevegury: Yes this is a breaking binary change, but thanks to the fact that Single is not yet marked as stable, I vote for making this change.

src/main/java/rx/Single.java
```diff
@@ -100,7 +100,7 @@ private Single(final Observable.OnSubscribe<T> f) {
         this.onSubscribe = f;
     }
 
-    static final RxJavaObservableExecutionHook hook = RxJavaPlugins.getInstance().getObservableExecutionHook();
+    static RxJavaSingleExecutionHook hook = RxJavaPlugins.getInstance().getSingleExecutionHook();
```
stevegury: Why did you remove `final`?

stevegury: Nevermind, I just saw your last comment.

stevegury:
More tests would be appreciated, but otherwise LGTM.

akarnokd:
Could you squash the commits and rebase this PR?

stevegury:
:+1: (after you rebase/squash the commits)

akarnokd:
:+1:

PR #3697, opened 2016-02-11
1.x: fix ScalarSynchronousObservable expects EventLoopsScheduler from

akarnokd:
Schedulers.computation()

When one installs a scheduler hook which makes Schedulers.computation() return a different scheduler, the constant function in ScalarSynchronousObservable crashes if that hasn't been initialized.


stevegury:
:+1:

Issue #3698, opened 2016-02-12
OperatorGroupBy doesn't complete when children are not subscribed to


duncani:
This issue affects <= 1.0.17, but seems to be fixed in the 1.1.1 back-port of the 2.x OperatorGroupBy.

The problem:
When `GroupBySubscriber` receives a completed event from upstream, it immediately emits calls onCompleted() on all children, but then waits for all children to unsubscribe before propagating onCompleted() to the immediate downstream subscribers.

It seems to me that the Operator ought to immediately complete the `Observable<GroupedObservable>` since there will be no further `GroupedObservable`s.  It would then be the responsibility of downstream operators to manage the output groups and await termination as appropriate - e.g. as OperatorMerge does.

This can be seen in two primary scenarios:
1. where one or more of the `GroupedObservable`s are never subscribed to.
2. where the `Observable<GroupedObservable>` is collected into a List or similar collection type.

In both cases the chain will never complete.

duncani:
I've pushed a Failing test case on a fork of the 1.0.17 release, along with a simple fix for it that passes the test cases on that branch, should there ever be a 1.0.18 release.
I've also cherry-picked the tests onto 1.x (where they all pass) as a regression test.

akarnokd:
The test has been merged in #3699. I don't think versions of the pattern 1.0.x will be updated any further. We will only support 1.x and 2.x together.

PR #3699, opened 2016-02-12
1.x: GroupBy Regression Test Case

duncani:
This is passing against 1.1.1 with the back-ported OperatorGroupBy, but would be good to have it in as a regression test.

See also: #3698

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorGroupByTest.java
+++ b/src/test/java/rx/internal/operators/OperatorGroupByTest.java
@@ -1581,4 +1581,224 @@ public class OperatorGroupByTest {
         ts2.assertNotCompleted();
     }
 
+    @Test
+    public void testGroupedObservableCollection() {
+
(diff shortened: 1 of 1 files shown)


src/test/java/rx/internal/operators/OperatorGroupByTest.java
```diff
@@ -1581,4 +1581,224 @@ public void call(GroupedObservable<Object, Integer> g) {
...
+
+            @Override
+            public void onNext(Map<Integer, Observable<Integer>> integerObservableMap) {
+                integerObservableMap.get(0).subscribe(inner1);
+                integerObservableMap.get(1).subscribe(inner2);
+            }
+        });
+
+        Observable<Map<Integer, Observable<Integer>>> mapObservable = Observable.range(0, 10)
+                                                                                                .groupBy(new Func1<Integer, Integer>() {
```
akarnokd: Could you fix the indentation?

akarnokd:
Could you please squash the commits?

PR #3707, opened 2016-02-14
1.x: make Completable.subscribe() report isUnsubscribed consistently

akarnokd:
The empty and lambda-based `Completable.subscribe()` returns a `Subscription` whose `isUnsubscribed` should be consistent with the rest of the reactive objects by returning true if the sequence terminated (not just when one truly cancelled it).

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1828,12 +1828,13 @@ public class Completable {
         subscribe(new CompletableSubscriber() {
             @Override
             public void onCompleted() {
-                // nothing to do
+                mad.unsubscribe();
             }
             
             @Override
             public void onError(Throwable e) {
                 ERROR_HANDLER.handleError(e);
+                mad.unsubscribe();
             }
             
             @Override
@@ -1864,11 +1865,13 @@ public class Completable {
                 } catch (Throwable e) {
                     ERROR_HANDLER.handleError(e);
                 }
+                mad.unsubscribe();
             }
             
             @Override
             public void onError(Throwable e) {
                 ERROR_HANDLER.handleError(e);
+                mad.unsubscribe();
             }
             
             @Override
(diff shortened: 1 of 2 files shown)


src/main/java/rx/Completable.java
```diff
@@ -1859,6 +1860,7 @@ public final Subscription subscribe(final Action0 onComplete) {
         subscribe(new CompletableSubscriber() {
             @Override
             public void onCompleted() {
+                mad.unsubscribe();
```
zsxwing: This one should be called after `onComplete.call()`.

src/main/java/rx/Completable.java
```diff
@@ -1896,6 +1899,7 @@ public final Subscription subscribe(final Action1<? super Throwable> onError, fi
...
```
zsxwing: This one should be called after `onComplete.call()`.

src/main/java/rx/Completable.java
```diff
@@ -1905,6 +1909,7 @@ public void onCompleted() {
             
             @Override
             public void onError(Throwable e) {
+                mad.unsubscribe();
```
zsxwing: This one should be called after `onError.call(e)`.

akarnokd:
Updated.

src/main/java/rx/Completable.java
```diff
@@ -1900,7 +1903,9 @@ public void onCompleted() {
                     onComplete.call();
                 } catch (Throwable e) {
                     onError(e);
+                    return;
                 }
+                mad.unsubscribe();
```
zsxwing: Should be in `finally`?

zsxwing: Or just remove the above return?

akarnokd: OnError calls it too, shouldn't call again after onError returns.

zsxwing:
I see. :+1:

Issue #3708, opened 2016-02-14
Single and Completable takeUntil()


hzsweers:
In reading the original PR for `Single`, I saw Ben was OK with adding a `takeUntil(Single/Observable)` operator for `Single` later. I was wondering if this was still the plan, and if you would be open to contributions on this front. Same with `Completable`.

One caveat that I've thought of is that there's a potentially conflicting contact with `Single` and `takeUntil()` in the sense that `takeUntil()` calls `onCompleted()` in observables, but `Single`s will actually go to `onError()` if `onCompleted()` is called before any events are emitted (and by extension `onSuccess()`). Not sure what the clear path would be, but it seems like `Single`s would have to only unsubscribe and not propagate any notifications, differing from `Observable`s and likely `Completable`s in this regard.

akarnokd:
`Single.takeUntil` still has to emit an onSuccess or onError; you need to emit a `NoSuchElementException` from the other `Single.onSuccess`. Unsubscribing the main and not emitting anything is bad because it keeps the subscribers hanging. I don't think it is worth it.

`Completable.takeUntil` is just `completable.ambWith(other)`.

hzsweers:
The completable makes sense. For Single, do you think there's a feasible way to do a sort of takeUntil-like behavior where it completes/unsubscribes on the emission of another observable/single? 

I'm working on some lifecycle handling for android, and a use case that's worked well in the past is to emit lifecycle events via behaviorsubject and then just takeUntil(e -> e == DETACH). The idea is you want it to unsubscribe when the detach happens, and perhaps out of convenience we've always been using the fact that takeUntil would call onComplete and eventually cause it to unsubscribe.

hzsweers:
It seems like forcing an onSuccess or onError defeats the purpose of having a subscription that can be unsubscribed prior to that emission.

akarnokd:
The best you can do with `takeUntil` is to signal an error, like `CancellationException` similar to how `Future` does.

hzsweers:
I see. Would that be something you guys would be open to as PR? Or at least consider making `Single.lift()` public since it's still experimental? Otherwise it doesn't seem like it's possible in the current API. Could catch existing `NoSuchElementException`s but there'd be no way to know if it was due to completion or cancellation. 

Alternatively, would there be a simple way to potentially swallow that error upstream rather than force the subscriber to handle it?

akarnokd:
We are open for PRs.

akarnokd:
[...]

PR #3712, opened 2016-02-15
Add takeUntil support in Single

hzsweers:
As discussed in #3708

This adds `takeUntil(Observable)` and `takeUntil(Single)` support in `Single`. It was mostly just adapting the logic from the existing `OperatorTakeUntil` and adjusting it for accepting a `Single` and sending a `CancelattionException` in the event of a submission from `other` prior to a terminal event in the source `Single`.

Any feedback is appreciated it, this is my first time contributing an implementation to this project. Particularly wondering if it's worth keeping both overloads or if the user should just coerce their `other` to one type or ther other. Also particularly looking for feedback on what information to include in the `CancellationException`.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -12,6 +12,9 @@
  */
 package rx;
 
+import java.util.Collection;
+import java.util.concurrent.*;
+
 import rx.Observable.Operator;
 import rx.annotations.Beta;
 import rx.annotations.Experimental;
@@ -23,14 +26,12 @@ import rx.internal.producers.SingleDelayedProducer;
 import rx.internal.util.ScalarSynchronousSingle;
 import rx.internal.util.UtilityFunctions;
 import rx.observers.SafeSubscriber;
+import rx.observers.SerializedSubscriber;
 import rx.plugins.RxJavaObservableExecutionHook;
 import rx.plugins.RxJavaPlugins;
 import rx.schedulers.Schedulers;
 import rx.singles.BlockingSingle;
 import rx.subscriptions.Subscriptions;
-
-import java.util.Collection;
-import java.util.concurrent.*;
 
 /**
  * The Single class implements the Reactive Pattern for a single value response. See {@link Observable} for the
@@ -1797,6 +1798,156 @@ public class Single<T> {
                         Single.this.subscribe(ssub);
                     }
                 });
+            }
+        });
+    }
+
+    /**
+     * Returns a Single that emits the item emitted by the source Single until an Observable emits an item. Upon
+     * emission of an item from {@code other}, this will emit a {@link CancellationException} rather than go to
+     * {@link SingleSubscriber#onSuccess(Object)}.
+     * <p>
+     * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/takeUntil.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code takeUntil} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param other
+     *            the Observable whose first emitted item will cause {@code takeUntil} to emit the item from the source
+     *            Single
+     * @param <E>
+     *            the type of items emitted by {@code other}
+     * @return a Single that emits the item emitted by the source Single until such time as {@code other} emits
+     * its first item
+     * @see <a href="http://reactivex.io/documentation/operators/takeuntil.html">ReactiveX operators documentation: TakeUntil</a>
+     */
+    public final <E> Single<T> takeUntil(final Observable<? extends E> other) {
+        return lift(new Operator<T, T>() {
+            @Override
+            public Subscriber<? super T> call(Subscriber<? super T> child) {
+                final Subscriber<T> serial = new SerializedSubscriber<T>(child, false);
+
+                final Subscriber<T> main = new Subscriber<T>(serial, false) {
+                    @Override
+                    public void onNext(T t) {
+                        serial.onNext(t);
+                    }
+                    @Override
+                    public void onError(Throwable e) {
+                        try {
+                            serial.onError(e);
+                        } finally {
+                            serial.unsubscribe();
+                        }
+                    }
+                    @Override
+                    public void onCompleted() {
+                        try {
+                            serial.onCompleted();
+                        } finally {
+                            serial.unsubscribe();
+                        }
+                    }
+                };
+
+                final Subscriber<E> so = new Subscriber<E>() {
+
+                    @Override
+                    public void onCompleted() {
+                        onError(new CancellationException("Stream was canceled before emitting a terminal event."));
+                    }
+
+                    @Override
+                    public void onError(Throwable e) {
+                        main.onError(e);
+                    }
+
+                    @Override
+                    public void onNext(E e) {
+                        onError(new CancellationException("Stream was canceled before emitting a terminal event."));
+                    }
+                };
+
+                serial.add(main);
+                serial.add(so);
+
+                child.add(serial);
+
+                other.unsafeSubscribe(so);
+
+                return main;
+            }
+        });
+    }
+
+    /**
+     * Returns a Single that emits the item emitted by the source Single until a second Single emits an item. Upon
+     * emission of an item from {@code other}, this will emit a {@link CancellationException} rather than go to
+     * {@link SingleSubscriber#onSuccess(Object)}.
+     * <p>
+     * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/takeUntil.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code takeUntil} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param other
+     *            the Single whose emitted item will cause {@code takeUntil} to emit the item from the source Single
+     * @param <E>
+     *            the type of item emitted by {@code other}
+     * @return a Single that emits the item emitted by the source Single until such time as {@code other} emits its item
+     * @see <a href="http://reactivex.io/documentation/operators/takeuntil.html">ReactiveX operators documentation: TakeUntil</a>
+     */
+    public final <E> Single<T> takeUntil(final Single<? extends E> other) {
+        return lift(new Operator<T, T>() {
+            @Override
+            public Subscriber<? super T> call(Subscriber<? super T> child) {
+                final Subscriber<T> serial = new SerializedSubscriber<T>(child, false);
+
+                final Subscriber<T> main = new Subscriber<T>(serial, false) {
+                    @Override
+                    public void onNext(T t) {
+                        serial.onNext(t);
+                    }
+                    @Override
+                    public void onError(Throwable e) {
+                        try {
+                            serial.onError(e);
+                        } finally {
+                            serial.unsubscribe();
+                        }
+                    }
+                    @Override
+                    public void onCompleted() {
+                        try {
+                            serial.onCompleted();
+                        } finally {
+                            serial.unsubscribe();
+                        }
+                    }
+                };
+
+                final SingleSubscriber<E> so = new SingleSubscriber<E>() {
+                    @Override
+                    public void onSuccess(E value) {
+                        onError(new CancellationException("Stream was canceled before emitting a terminal event."));
+                    }
+
+                    @Override
+                    public void onError(Throwable e) {
+                        main.onError(e);
+                    }
+                };
+
+                serial.add(main);
+                serial.add(so);
+
+                child.add(serial);
+
+                other.subscribe(so);
+
+                return main;
             }
         });
     }
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1803,6 +1803,79 @@ public class Single<T> {
     }
 
     /**
+     * Returns a Single that emits the item emitted by the source Single until a Completable terminates. Upon
+     * termination of {@code other}, this will emit a {@link CancellationException} rather than go to
+     * {@link SingleSubscriber#onSuccess(Object)}.
+     * <p>
+     * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/takeUntil.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code takeUntil} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param other
+     *            the Completable whose termination will cause {@code takeUntil} to emit the item from the source
+     *            Single
+     * @return a Single that emits the item emitted by the source Single until such time as {@code other} terminates.
+     * @see <a href="http://reactivex.io/documentation/operators/takeuntil.html">ReactiveX operators documentation: TakeUntil</a>
+     */
+    public final Single<T> takeUntil(final Completable other) {
+        return lift(new Operator<T, T>() {
+            @Override
+            public Subscriber<? super T> call(Subscriber<? super T> child) {
+                final Subscriber<T> serial = new SerializedSubscriber<T>(child, false);
+
+                final Subscriber<T> main = new Subscriber<T>(serial, false) {
+                    @Override
+                    public void onNext(T t) {
+                        serial.onNext(t);
+                    }
+                    @Override
+                    public void onError(Throwable e) {
+                        try {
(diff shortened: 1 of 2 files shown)


hzsweers:
Just noticed my IDE swapped the wildcard imports for explicit ones. Let me know if I should revert that.

akarnokd:
Yes please.

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
         int numberOfErrors = retryCounter.getOnErrorEvents().size();
         assertEquals(retryCount, numberOfErrors);
     }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testTakeUntilSuccess() {
```
artem-zinnatullin: No need for `test` prefix here and in tests below

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+        TestSingle source = new TestSingle(sSource);
+        TestSingle other = new TestSingle(sOther);
+
+        TestSubscriber<String> result = new TestSubscriber<String>();
+        Single<String> stringSingle = Single.create(source).takeUntil(Single.create(other));
+        stringSingle.subscribe(result);
+        other.sendOnSuccess("one");
+
+        result.assertError(CancellationException.class);
+        verify(sSource, times(1)).unsubscribe();
```
artem-zinnatullin: No need for `times(1)` here and below, `verify(obj).call()` already verifies `times(1)`

src/test/java/rx/SingleTest.java
artem-zinnatullin: Can you please add `Single` and `Observable` to all test names appropriately? 

Idk, but current tests names are not very obvious, would be easier to see what to expect from tests if they have names like `takeUntil_should..When..` but it's personal thing, so feel free to ignore :)

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+        stringSingle.subscribe(result);
+        other.sendOnCompleted();
+
+        result.assertError(CancellationException.class);
+        verify(sSource, times(1)).unsubscribe();
+        verify(sOther, times(1)).unsubscribe();
+
+    }
+
+    private static class TestObservable implements Observable.OnSubscribe<String> {
```
artem-zinnatullin: Any reasons to use custom classes instead of  `PublishSubject` which already covered by tests and known to other developers?

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+
+        @Override
+        public void call(SingleSubscriber<? super String> observer) {
+            this.subscriber = observer;
+            observer.add(s);
+        }
+    }
+
+    @Test
+    public void testUntilFires() {
```
artem-zinnatullin: Typo in test name I guess

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+
+        source.take(1).toSingle().takeUntil(until.take(1).toSingle()).unsafeSubscribe(ts);
+
+        assertTrue(source.hasObservers());
+        assertTrue(until.hasObservers());
+
+        until.onNext(1);
+
+        ts.assertError(CancellationException.class);
+        ts.assertTerminalEvent();
```
artem-zinnatullin: Not needed since you already checked `ts.assertError` but you can check `ts.assertNotCompleted()` instead (not sure it exists though, reviewing from phone)

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+
+        assertFalse("Source still has observers", source.hasObservers());
```
artem-zinnatullin: Not sure that error message adds something here `assertFalse(source.hasObservers()) ` speaks for itself :) Same for assertions below.

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+        assertFalse("Until still has observers", until.hasObservers());
+        assertFalse("TestSubscriber is unsubscribed", ts.isUnsubscribed());
+    }
+
+    @Test
+    public void testUntilFiresObservable() {
```
artem-zinnatullin: Typo in test name

src/test/java/rx/SingleTest.java
```diff
@@ -1353,4 +1386,364 @@ public String call() throws Exception {
...
+        ts.assertError(CancellationException.class);
+        ts.assertTerminalEvent();
+
+        assertFalse("Source still has observers", source.hasObservers());
+        assertFalse("Until still has observers", until.hasObservers());
+        assertFalse("TestSubscriber is unsubscribed", ts.isUnsubscribed());
+    }
+
+    @Test
+    public void testDownstreamUnsubscribes() {
```
artem-zinnatullin: Since you're adding tests to `SingleTest` you should definitely mention target method (operator) name in the test

artem-zinnatullin:
Just style issues, otherwise LGTM :+1:

src/test/java/rx/SingleTest.java
hzsweers: Most of these tests are adapted from the `OperatorTakeUntilTest`, which used these. I was just matching behavior. Could use subjects, but if we do then I think we should update both places.

src/test/java/rx/SingleTest.java
hzsweers: I don't know if it's a typo, I think it's pointing out that this is just firing until the `until` condition fires. This is another one copied over from `OperatorTakeUntilTest`

artem-zinnatullin: Well, since you placed it in `SingleTest` it should have `takeUntil` prefix
because `SingleTest `contains a lot of tests for different operators and we
should be able to understand what does each test check from its name.

On Mon, 15 Feb 2016, 14:02 Zac Sweers <notifications@github.com> wrote:

> In src/test/java/rx/SingleTest.java
> <https://github.com/ReactiveX/RxJava/pull/3712#discussion_r52933591>:
>
> > +        }
> > +
> > +        /* used to simulate subscription */
> > +        public void sendOnError(Throwable e) {
> > +            subscriber.onError(e);
> > +        }
> > +
> > +        @Override
> > +        public void call(SingleSubscriber<? super String> observer) {
> > +            this.subscriber = observer;
> > +            observer.add(s);
> > +        }
> > +    }
> > +
> > +    @Test
> > +    public void testUntilFires() {
>
> I don't know if it's a typo, I think it's pointing out that this is just
> firing until the until condition fires. This is another one copied over
> from OperatorTakeUntilTest
>
> —
> <https://github.com/ReactiveX/RxJava/pull/3712/files#r52933591>.
>
-- 

@artem_zin

src/test/java/rx/SingleTest.java
hzsweers: For the test prefix: 322ae576172d2c55bbaaf258314460e396cd0d55

I wasn't sure what to do about the names. I figured since it was in `SingleTest` that I could assume any test without an explicit "Observable" mention was using "Single". Maybe it's not that obvious though. @akarnokd what do you think?

src/test/java/rx/SingleTest.java
hzsweers: be9f9ce3f1b0a3bb15c47e2116226b2fd8d2dac0

src/test/java/rx/SingleTest.java
hzsweers: These tests use a mix though... I'm wondering if the `OperatorTakeUntilTest` tests were added at different times, looks like there's two different styles used

src/test/java/rx/SingleTest.java
hzsweers: d961f02ebbb281b4478e24cc6c82481a0073ac3c

src/test/java/rx/SingleTest.java
hzsweers: 02e6325bc90e3c6d43963239859f46d63e89077d

src/test/java/rx/SingleTest.java
hzsweers: Also from the original `OperatorTakeUntilTest`. Going to write a general comment to see what people think

src/test/java/rx/SingleTest.java
hzsweers: d961f02ebbb281b4478e24cc6c82481a0073ac3c

src/test/java/rx/SingleTest.java
hzsweers: d961f02ebbb281b4478e24cc6c82481a0073ac3c

hzsweers:
CC @akarnokd @artem-zinnatullin 

For the tests, most of them are adapted from `OperatorTakeUntilTest`.  It looks like these tests have two different styles (one with `TestObservable` and the other with `Subject`). I reused both of them here, but would we maybe want to stick with one or the other? Especially with Single, I think some now overlap.

Also, if we stick with one style here, should we update the style in `OperatorTakeUntilTest` as well to match?

akarnokd:
I'm not too keen on how the tests are named or what test framework objects you use.

Please squash your commits.

hzsweers:
Squashed

akarnokd:
:+1:

hzsweers:
@zsxwing @stevegury any input? Need another collaborator review

Main points I'm wondering about:
- Should we only support one overload? If we support more than one, should it support all three (including `Completable`?)
- Should `takeUntil` test styles formalized, looks like there are two styles right now judging by `OperatorTakeUntilTest`: subject-based and `TestObservable`-based.

stevegury:
:+1:

src/main/java/rx/Single.java
```diff
@@ -1800,6 +1801,156 @@ public void onError(Throwable error) {
...
+     * {@link SingleSubscriber#onSuccess(Object)}.
+     * <p>
+     * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/takeUntil.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code takeUntil} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param other
+     *            the Observable whose first emitted item will cause {@code takeUntil} to emit the item from the source
```
zsxwing: nit: the Observable whose first emitted item will cause {@code takeUntil} to **stop emitting** items from the source Observable

zsxwing:
> sending a CancelattionException in the event of a submission from other prior to a terminal event in the source Single.

This behavior is really confusing. Why not be same as Observable? Am I missing anything?

hzsweers:
@akarnokd and I discussed it in #3708. `Single` currently emits a `NoSuchElementException` if the stream is terminated prior to any event emission. By signaling a `CancellationException`, we can at least signal to the subscriber if the error was due to `takeUntil` unsubscribing it or normal missing emissions.

zsxwing:
@hzsweers Thanks for clarifying. However, I would expect `Single.takeUntil` returns something that contains 0 or 1 item. How about returning Observable instead? Then the semantics of `Single.takeUntil` is `onNext{0, 1} (onCompleted | onError)`, which is consistent with `Observable.takeUntil` (`onNext{0, N} (onCompleted | onError)`)

akarnokd:
The operators should stay in the same type as long as they can and there are operators that simply can't behave the same as their counterpart in other reactive types. If one wishes the `Observable.takeUntil` behavior, one can use toObservable().takeUntil() and get that behavior

zsxwing:
> The operators should stay in the same type as long as they can and there are operators that simply can't behave the same as their counterpart in other reactive types. If one wishes the Observable.takeUntil behavior, one can use toObservable().takeUntil() and get that behavior

If so, I vote for `NoSuchElementException` as `CancellationException` seems counterintuitive to me. `CancellationException` looks some codes cancel Single (e.g., I call `unsubscribe` in some place). But here the error is that `Single.takeUntil` returns something that is not `Single`. Hence, I would expect `NoSuchElementException`.

hzsweers:
That's fair, but is there something we could do to at least indicate whether the source was just unsubscribed or actually misbehaved? I was of the impression that `onError` was primarily for serious, unexpected errors, whereas this seems like not not unexpected or serious.

Consider the following

With `NoSuchElementException`:
```java
PublishSubject<Integer> source = PublishSubject.create();
PublishSubject<Integer> until = PublishSubject.create();

source.take(1).toSingle()
        .takeUntil(until.take(1).toSingle())
        .subscribe(
                new Action1<Integer>() {
                    @Override
                    public void call(Integer integer) {
                        System.out.println("Success");
                    }
                },
                new Action1<Throwable>() {
                    @Override
                    public void call(Throwable throwable) {
                        System.out.println("I don't know if it was due to unsubscribing or the source is misbehaving");
                    }
                });

until.onNext(1);
```

vs. with `CancellationException`

```java
PublishSubject<Integer> source = PublishSubject.create();
PublishSubject<Integer> until = PublishSubject.create();

source.take(1).toSingle()
        .takeUntil(until.take(1).toSingle())
        .subscribe(
                new Action1<Integer>() {
                    @Override
                    public void call(Integer integer) {
                        System.out.println("Success");
                    }
                },
                new Action1<Throwable>() {
                    @Override
                    public void call(Throwable throwable) {
                        if (throwable instanceof CancellationException) {
                            System.out.println("It was canceled.");
                        } else {
                            System.out.println("Source didn't emit.");
                        }
                    }
                });

until.onNext(1);
```

The example I gave in the issue is probably the best example for me. We use something similar to this for lifecycle binding in android. When the lifecycle ends, it might unsubscribe this in the middle. For a normal error, we might show a generic "an error occurred" message. In the event that it's just the lifecycle ending, we don't want to react that way, and rather likely just want to do nothing at all or clean up resources. I'm fine with not using `CancellationException`, but I think it would be useful to add a means of being able to differentiate between unsubscription reasons.

hzsweers:
I'll add an overload for `Completable` as well, and remove the old test style in the new tests here for now. @stevegury just to be clear, the "new" style would be the `Subject`-based approach right?

stevegury:
@hzsweers I actually typed "it's a big deal" but I was willing to say "it's **not** a big deal" (I corrected my previous comment). I prefer the Subject based approach, but I am not strongly opinionated about that.
I still :+1: on the current implementation.

zsxwing:
> That's fair, but is there something we could do to at least indicate whether the source was just unsubscribed or actually misbehaved?

My point here is if Single.takeUntil(...) returns a Single that emits nothing, it should be misbehaved rather than `unsubscribed`.

akarnokd:
I think "CancellationException" is the clearer reaction here. Remember the problems around the Observable.single() and how it is a source of problem to find out exactly who didn't signal? Here, you know that if takeUntil is tripped and not some upstream machinery ends up being empty.

zsxwing:
> I think "CancellationException" is the clearer reaction here. Remember the problems around the Observable.single() and how it is a source of problem to find out exactly who didn't signal? Here, you know that if takeUntil is tripped and not some upstream machinery ends up being empty.

What if we want to add other operator that may return something doesn't signal? If it also emits `CancellationException`, then we still cannot find out exactly who didn't signal. So why not use `NoSuchElementException` to indicate all cases that didn't signal?

hzsweers:
What about a subclass of `NoSuchElementException`? A `CanceledNoSuchElementException`, so to speak. Wouldn't break the downstream APIs but still allows for the subscriber to differentiate.

zsxwing:
> What about a subclass of NoSuchElementException? A CanceledNoSuchElementException, so to speak. Wouldn't break the downstream APIs but still allows for the subscriber to differentiate.

Maybe a more general question, should we add special exceptions for different operators, or we just use a general exception?

hzsweers:
I think people should be aware of the implications of operators they use. Would `CompositeException` possibly be precedent for this?

zsxwing:
> I think people should be aware of the implications of operators they use. Would CompositeException possibly be precedent for this?

CompositeException is fine and clear. It indicates there are multiple errors thrown. All classes in rx.exceptions are well defined, and it's very easy to connect them with the bad cases.

If we can define clearly that when should throw CancellationException/CanceledNoSuchElementException(or whatever you propose), when should throw NoSuchElementException, I won't be against that. For now, the confusing thing for me is that if a Single doesn't signal, I may receive CancellationException or NoSuchElementException.

hzsweers:
Updated with a separate commit for easier reviewing, let me know if you want me to squash again.

ping @akarnokd @stevegury

akarnokd:
:+1:

stevegury:
:+1:

Issue #3711, opened 2016-02-15
flatMapIterable with maxConcurrent


ginbalin:
As far as I saw flatMapIterable doesn't have maxConcurrent parameter and thus doesn't offer a way to apply backpressure. In order to align it with other flatMap would be good to add an overload with maxConcurrent.

What do you think?

akarnokd:
PR is welcome.

ginbalin:
Although you stated here that this was useless https://github.com/ReactiveX/RxJava/pull/2627 . Is it really the case?
I believe with current implementation this would be usefull to control backpressure.

akarnokd:
I was wrong back then.

akarnokd:
Closing via #3722

PR #3722, opened 2016-02-19
Add maxConcurrent parameter to flatMapIterable

lukaciko:
Refs #3711. Should these new overloaded functions be annotated with `@Beta`?


lukaciko:
The failing test (`testNoMoreRequestsAfterUnsubscribe`) seems unrelated. All tests pass locally.

akarnokd:
If you could add one unit test for each method that verifies maxConcurrent is honored, that would be great.

lukaciko:
While writing the test I figured you might have been right when commenting on issue #2627. If I am not mistaken all the observables created with `OperatorMapPair.convertSelector` use the same scheduler (there is no way to subscribe them to different schedulers) and thus are run sequentially and until completion. Would adding a Scheduler parameter make sense?

akarnokd:
There is no scheduler involved here so no.

akarnokd:
:+1: as it is now

stevegury:
:+1:

PR #3727, opened 2016-02-23
scan should pass upstream a request of Long.MAX_VALUE

davidmoten:
Changes in 1.1.1. for the `scan` operator introduced a bug in requesting where when the initial request was for max value the `scan` operator requested one less than that value of upstream (which affects fast-path producers).

I believe this bug only affects 1.1.1 (haven't looked at 2.x).

This PR includes a test that failed on the original code.


akarnokd:
2.x is not affected because it does a full queue-drain where the initial value is in the queue already and requests are forwarded to the upstream as they are.

:+1:

davidmoten:
>2.x is not affected

Thanks @akarnokd

stevegury:
:+1:

Issue #3726, opened 2016-02-23
Completable.subscribe() swallows errors by default


kotucz:
The default behavior of `Completable.subscribe()` for errors is different from that of `Observable` or `Single`. Errors in `Completable` are just passed to `RxJavaErrorHandler` which does nothing by default. While `Observable` and `Single` cause runtime `OnErrorNotImplementedException` which is expected. 

*Completable.java*

            @Override
            public void onError(Throwable e) {
                ERROR_HANDLER.handleError(e);
            }

*Observable.java*

            @Override
            public final void onError(Throwable e) {
                throw new OnErrorNotImplementedException(e);
            }

akarnokd:
Throwing like `Observable` relies on the expectation that the execution won't switch to other thread. This `OnErrorNotImplemented` is remnant of the imperative style of programming. Using try-catch around a sequence doesn't make sense and throwing upward has no guarantees either.

kotucz:
I do not want to wrap the `subscribe()` call in try-catch. Problem is that I do when I not specify `onError` callback in `Completable.subscribe()` I do not expect the completable to emit and error. But when the `Completable` emits an unexpected error, that is an programming error and shall be processed as such. The `OnErrorNotImplementedException` is then expected to e.g. crash the application.

loganj:
@akarnokd Throwing upward does have a crucial guarantee-- reaching the thread's uncaught exception handler.  

Delivering to the plugin is effectively swallowing the exception, since it receives all exceptions and cannot safely propagate any of them anywhere.

Failing to propagate exceptions was a huge problem (#1682) in scheduled actions before 1.0.

akarnokd:
PRs are welcome.

loganj:
:+1: Working on one now. Just going to deliver to the uncaught exception handler after the plugin, wherever it seems to make sense.

loganj:
@akarnokd Can you explain the usage of `toNpe()` in Completable, e.g. here?: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Completable.java#L1946

akarnokd:
You can replace/precede them with `Exceptions.throwIfFatal`.

PR #3733, opened 2016-02-24
Avoid swallowing errors in Completable

loganj:
Instead, deliver them up to the thread's uncaught exception handler.

Fixes reactivex/rxjava#3726

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1835,6 +1835,7 @@ public class Completable {
             public void onError(Throwable e) {
                 ERROR_HANDLER.handleError(e);
                 mad.unsubscribe();
+                deliverUncaughtException(e);
             }
             
             @Override
@@ -1864,14 +1865,17 @@ public class Completable {
                     onComplete.call();
                 } catch (Throwable e) {
                     ERROR_HANDLER.handleError(e);
+                    deliverUncaughtException(e);
+                } finally {
+                    mad.unsubscribe();
                 }
-                mad.unsubscribe();
             }
             
             @Override
             public void onError(Throwable e) {
                 ERROR_HANDLER.handleError(e);
                 mad.unsubscribe();
+                deliverUncaughtException(e);
             }
             
             @Override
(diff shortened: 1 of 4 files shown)


akarnokd:
What happened to `CompletableTest`?

src/main/java/rx/Completable.java
```diff
@@ -1835,6 +1835,8 @@ public void onCompleted() {
             public void onError(Throwable e) {
                 ERROR_HANDLER.handleError(e);
                 mad.unsubscribe();
+                Thread thread = Thread.currentThread();
+                thread.getUncaughtExceptionHandler().uncaughtException(thread, e);
```
artem-zinnatullin: Let's just inline `Thread.currentThread()` call here and below?

akarnokd: Or create a private method perhaps?

loganj: Nope, two usages.

loganj:
@akarnokd sorry, IJ formatted it.  Will clean up and push again.

src/main/java/rx/Completable.java
loganj: private method is fair.  done.

akarnokd:
:+1:

src/main/java/rx/Completable.java
```diff
@@ -1927,7 +1933,12 @@ public void onSubscribe(Subscription d) {
         
         return mad;
     }
-    
+
+    private static void deliverUncaughtException(Throwable e) {
```
artem-zinnatullin: Now it's a good candidate to live in `Exceptions` :smile_cat:

JakeWharton: That would require it to be public, sadly, and this isn't an API that RxJava needs to provide for users.

artem-zinnatullin: Well, we can create `ExceptionsInternal` under `internal` package, but yeah… k, let's extract it if somewhere else same functionality will be needed.

src/test/java/rx/CompletableTest.java
```diff
@@ -2700,7 +2700,64 @@ public void call(CompletableSubscriber s) {
         
         Assert.assertTrue(name.get().startsWith("RxComputation"));
     }
-    
+
+    @Test(timeout = 1000)
```
artem-zinnatullin: I belive you don't need timeout here and in the added tests below, code is synchronous

artem-zinnatullin:
just a nit about `timeout` in tests, otherwise :+1:

loganj:
removed timeouts

stevegury:
:+1:

Issue #3442, opened 2015-10-13
1.x: Writing backpressure supporting operators


davidmoten:
We know that writing a custom operator is a very tricky business if we are not composing existing operators.

I'm particularly interested in implementing custom Operators that for each arriving emission of type `T` a state machine consumes the arrival and emits 0..\* items of type `R`.

I've raised examples of this in #3020 but didn't get any bites in terms of an assessment of `Transformers.stateMachine`. I've been using it for a few months and it has evolved both internally and in terms of its API signature. 

For  me this utility is the missing link for producing new operators (well, a certain class of operators). It is not very complex under the covers in that it just composes a bunch of existing operators but it is complex enough that its worth making a utility from. 

Using this operator I now have my own versions of operators like [`StringObservable.split`](https://github.com/davidmoten/rxjava-extras/blob/master/src/main/java/com/github/davidmoten/rx/internal/operators/TransformerStringSplit.java) and `StringObservable.decode` that support backpressure safely (unlike current versions in RxJavaString) and reasonably efficiently. What really sells me about the approach is that the code is almost identical so that backpressure support is not something that the developer need consider much for this class of problem.

Submissions by @stealthcode like `AsyncOnSubscribe` and `SyncOnSubscribe`  provide means of creating observables safely that support backpressure and are more efficient than just throwing `.onBackpressureBuffer` on the end. I'm interested in getting review of [`Transformers.stateMachine`](https://github.com/davidmoten/rxjava-extras/blob/master/src/main/java/com/github/davidmoten/rx/internal/operators/TransformerStateMachine.java) with the intent of submitting it into RxJava so if you've got a moment have a gander, thanks.

akarnokd:
The thing I don't see is that what how is the 0..* amount determined; i.e., in the callback, should one call `onCompleted` or not to singal a batch is completed?

davidmoten:
At the moment you don't need to call `onCompleted` because once the transition completes synchronously it is called for you on L92. 

It might be good to make a call to `onCompleted` in the transition terminate the entire stream (I've taken this approach with the ability to unsubscribe from the transition, and of course on error so that would be consistent).  That way the user doesn't have to think about signals on batch streams but think only of the overall stream.

akarnokd:
Any new input on this issue?

akarnokd:
Closing as 1.x is EOL.

PR #3744, opened 2016-03-09
1.x: new operators buffer(While/Until) with predicate-based boundary

akarnokd:
The need for a buffer operator that splits the source sequence into continuous non-overlapping buffers based on the values in the sequence comes up from time to time. For, me, it reached the point where I'd consider adding operator(s) with this functionality into RxJava.

There are two variants of the required functionality: split before the element is added to the buffer: `bufferWhile`; and split after the element has been added to the buffer: `bufferUntil`.  Both cases, the predicate has to return `true` to trigger a split and emission of the current buffer.

I've added 2-2 overloads of these methods to `Observable` where the second overloads allow specifying the expected capacity use of the buffers (for performance reasons). These operators don't support backpressure and go unbounded, just like `buffer(Observable)` and others.

In the operator itself, I've added backpressured versions of `bufferWhile` and `bufferUntil`, which use a prefetch queue and a request-dependent queue-drain approach. It is up for a discussion what to do with these (provided this PR is of interest at all): 1) use them as the basis for the methods instead, 2) expose them through a different name and 3) drop them.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
+++ b/src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
@@ -1,0 +1,408 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package rx.internal.operators;
+
+import java.util.*;
+import java.util.concurrent.atomic.*;
+
+import rx.*;
+import rx.Observable.Operator;
+import rx.exceptions.*;
+import rx.functions.Func1;
+import rx.internal.util.atomic.SpscAtomicArrayQueue;
+import rx.internal.util.unsafe.*;
+
+/**
+ * Buffers values into a continuous, non-overlapping Lists where the boundary is determined
+ * by a predicate returning true.
+ *
+ * @param <T> the source and List element type
+ */
+public final class OperatorBufferPredicateBoundary<T> implements Operator<List<T>, T> {
+
+    final Func1<? super T, Boolean> predicate;
+    
+    final int prefetch;
+    
+    final int capacityHint;
+    
+    final boolean after;
+    
+    public OperatorBufferPredicateBoundary(Func1<? super T, Boolean> predicate, int prefetch, int capacityHint, boolean after) {
+        if (predicate == null) {
+            throw new NullPointerException("predicate");
+        }
+        if (prefetch <= 0) {
+            throw new IllegalArgumentException("prefetch > 0 required but it was " + prefetch);
+        }
+        if (capacityHint <= 0) {
+            throw new IllegalArgumentException("capacityHint > 0 required but it was " + capacityHint);
+        }
+        this.predicate = predicate;
+        this.prefetch = prefetch;
+        this.capacityHint = capacityHint;
+        this.after = after;
+    }
+
+    @Override
+    public Subscriber<? super T> call(Subscriber<? super List<T>> child) {
+        final BoundedSubscriber<T> parent = after 
+                ? new BoundedAfterSubscriber<T>(child, capacityHint, predicate, prefetch)
+                : new BoundedBeforeSubscriber<T>(child, capacityHint, predicate, prefetch);
+                
+        child.add(parent);
+        child.setProducer(new Producer() {
+            @Override
+            public void request(long n) {
+                parent.requestMore(n);
+            }
+        });
+        
+        return parent;
+    }
+    
+    static abstract class BoundedSubscriber<T> extends Subscriber<T> {
+        final Subscriber<? super List<T>> actual;
+        
+        final int capacityHint;
+        
+        final Func1<? super T, Boolean> predicate;
+        
+        final Queue<Object> queue;
+        
+        final AtomicLong requested;
+
+        final AtomicInteger wip;
+        
+        final NotificationLite<T> nl;
+        
+        final int limit;
+
+        List<T> buffer;
+        
+        long upstreamConsumed;
+        
+        volatile boolean done;
+        Throwable error;
+
+        public BoundedSubscriber(Subscriber<? super List<T>> actual, int capacityHint,
+                Func1<? super T, Boolean> predicate, int prefetch) {
+            this.actual = actual;
+            this.capacityHint = capacityHint;
+            this.predicate = predicate;
+            Queue<Object> q;
+            if (UnsafeAccess.isUnsafeAvailable()) {
+                q = new SpscArrayQueue<Object>(prefetch);
+            } else {
+                q = new SpscAtomicArrayQueue<Object>(prefetch);
+            }
+            queue = q;
+            buffer = new ArrayList<T>(capacityHint);
+            requested = new AtomicLong();
+            wip = new AtomicInteger();
+            nl = NotificationLite.instance();
+            limit = prefetch - (prefetch >> 2);
+            if (prefetch == Integer.MAX_VALUE) {
+                request(Long.MAX_VALUE);
+            } else {
+                request(prefetch);
+            }
+        }
+
+        @Override
+        public void onNext(T t) {
+            if (!queue.offer(nl.next(t))) {
+                unsubscribe();
+                onError(new MissingBackpressureException());
+            } else {
+                drain();
+            }
+        }
+        
+        @Override
+        public void onError(Throwable e) {
+            error = e;
+            done = true;
+            drain();
+        }
+        
+        @Override
+        public void onCompleted() {
+            done = true;
+            drain();
+        }
+        
+        void requestMore(long n) {
+            if (n > 0) {
+                BackpressureUtils.getAndAddRequest(requested, n);
+                drain();
+            } else
+            if (n < 0) {
+                throw new IllegalArgumentException("n >= 0 required but it was " + n);
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
```diff
@@ -0,0 +1,580 @@
...
+            }
+            
+            final Subscriber<? super List<T>> a = actual;
+            final Queue<Object> q = queue;
+            int missed = 1;
+            
+            for (;;) {
+
+                long r = requested.get();
+                long e = 0L;
```
stevegury: For readability would you mind giving longer names to those variable?

src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
```diff
@@ -0,0 +1,580 @@
...
+            
+            try {
+                emit = predicate.call(t);
+            } catch (Throwable ex) {
+                unsubscribe();
+                Exceptions.throwOrReport(ex, this, t);
+                return;
+            }
+            
+            if (emit) {
```
stevegury: I think it's less confusing to not emit the buffer when it's null.
I'd add a check for `! b.isEmpty()` in the if.

stevegury:
w.r.t. your question, I would prefer having the backpressure-aware operator be the default implementation.

akarnokd:
Updated:
- Using backpressure and a standard prefetch of RxRingBuffer.SIZE
- BufferWhile won't emit an empty initial buffer in case the first value triggers a new buffer
- Added `@Experimental` tags
- Added backpressure-tests and and explicit tests with empty or error sources.
- renamed local variables

stevegury:
:+1:

artem-zinnatullin:
LGTM except operators names.

* `bufferWhile(func)` sounds like it will buffer *while* you're returning `true` from `func` and it'll stop_buffering/will_split when you'll return `false`.
* `bufferUntil(func)` sounds like it will buffer *until* you're returning `false` and it'll stop_buffering/will_split when you'll return `true`.

Maybe something like:

* `bufferAndSplitWhen(func)`
* `bufferAndSplitAfter(func)`

not a great names too, but at least more obvious (to me).

Ping others: what do you think about operators names?

stevegury:
I prefer the `bufferWhile` `bufferUntil` version, it's idiomatic in functional languages like Scala and other to use those words.

src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
```diff
@@ -0,0 +1,487 @@
...
+        volatile boolean done;
+        Throwable error;
+
+        public BoundedAfterSubscriber(Subscriber<? super List<T>> actual, int capacityHint,
+                Func1<? super T, Boolean> predicate, int prefetch) {
+            this.actual = actual;
+            this.capacityHint = capacityHint;
+            this.predicate = predicate;
+            Queue<Object> q;
+            if (prefetch >= 10 * 1000 * 1000) {
```
vanniktech: What about `TimeUnit.Milliseconds.toNanos(10)`?

artem-zinnatullin:
Well, for me it sounds like `takeWhile()`/`takeUntil()` and `skipWhile()`/`skipUntil()`.

Compare:

```java
Observable
  .range(1, 1000000)
  .bufferWhile(number -> number % 10 == 0)
  .subscribe(numbers -> System.out.println(numbers));

// Here I'll think that it'll buffer numbers by some unknown mechanism 
// and then stop on first number that % 10 == 0. 
// Subscriber will receive not complete set of items.
```

And:

```java
Observable
  .range(1, 1000000)
  .bufferAndSplitWhen(number -> number % 10 == 0)
  .subscribe(numbers -> System.out.println(numbers));

// Here I clearly see that subscriber will receive all items.
// And that lambda specifies split points, not the end of the stream.
```

idk about others but I'll have to read javadoc each time to understand what to expect with current operators names.

src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
artem-zinnatullin: It's not time, it's prefetch amount. Also, your variant will require runtime calculation, David's may be calculated at compile time (nano win!).

src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
```diff
@@ -0,0 +1,487 @@
...
+                        emit = predicate.call(value);
+                    } catch (Throwable ex) {
+                        unsubscribe();
+                        buffer = null;
+                        Exceptions.throwOrReport(ex, localSubscriber, value);
+                        return;
+                    }
+                    
+                    if (emit) {
+                        localSubscriber.onNext(localBuffer);
```
davidmoten: As predicate.call could take a non-trivial time we could check for unsubscription before emitting. It's a trade-off between checking the volatile unsubscribed flag and reducing work after unsubscription but in other operators we normally do the unsubscribed check.

artem-zinnatullin: Not sure it worth it, passing every item to a predicate function already slows down the emission… `SafeSubscriber` will check subscriber state for us anyway.

Subscription state checked in the beginning of each iteration of the cycle ^.

davidmoten: >Not sure worth it
Hard to say but it's a convention throughout the codebase at the moment. @akarnokd your opinion (being such a performance oriented fellow)?

akarnokd: The most likely unsubscription will happen just after emission (take). There will be always a window between calling unsubscribe() and receiving an event. So I don't think there is a need for more eagerness than what's in the drain loop now. As for `SafeSubscribe`, last time I checked it didn't do hard-gating on `isUnsubscribed` but if one is really eager, his/her `onNext` of `Subscriber` can have extra checks before acting on the value.

src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
akarnokd: Doesn't matter, I'll remove it.

davidmoten:
I'll get back to reviewing the code in a bit but in terms of names an alternative is `toListWhile` and `toListUntil` that could also exist as aliases to the `buffer` versions. This is the api method I went with in *rxjava-extras* that implements this functionality ([Transformers.java](https://github.com/davidmoten/rxjava-extras/blob/a4edc924477b063586cee00be96412db2f444f95/src/main/java/com/github/davidmoten/rx/Transformers.java)). I had a dedicated operator for this stuff till I knocked up `Transformers.stateMachine` (discussion in #3442 that hasn't got any interest seemingly) and `toListWhile` and `toListUntil` goes through that (via `collectWhile/Until`).  I'll be interested to compare performance with the dedicated operator by @akarnokd in this PR. 

A natural extension to this PR that could be incorporated now is using `collectWhile` and `collectUntil` which `bufferWhile` and `bufferUntil` would be built on top of so that we aren't just list focused but can use arbitrary collection types.

akarnokd:
I could rename it to `bufferSplit`, change the parameters to include an isBefore boolean and only have 2 overloads.

akarnokd:
Updated the PR:

- removed that magic number, wasn't really relevant in the current setup
- combined the operators into `bufferSplitWhen` and an `checkAfter` flag to express the time the predicate should be checked.

akarnokd:
`testErrorThrownIssue1685` keeps failing for some reason on CI (but never locally).

stevegury:
:+1: 
I also noticed recently that `testErrorThrownIssue1685` started failing more frequently on TravisCI.

src/main/java/rx/internal/operators/OperatorBufferPredicateBoundary.java
```diff
@@ -0,0 +1,408 @@
...
+            @Override
+            public void request(long n) {
+                parent.requestMore(n);
+            }
+        });
+        
+        return parent;
+    }
+    
+    static abstract class BoundedSubscriber<T> extends Subscriber<T> {
```
artem-zinnatullin: nice abstraction :+1:

src/test/java/rx/internal/operators/OperatorBufferPredicateBoundaryTest.java
```diff
@@ -0,0 +1,473 @@
...
+import rx.Observable;
+import rx.exceptions.TestException;
+import rx.functions.Func1;
+import rx.internal.util.*;
+import rx.observers.TestSubscriber;
+
+public class OperatorBufferPredicateBoundaryTest {
+
+    @Test
+    public void bufferWhilePredicateNull() {
```
artem-zinnatullin: Pls fix tests names

artem-zinnatullin:
:+1:

src/test/java/rx/internal/operators/OperatorBufferPredicateBoundaryTest.java
akarnokd: It helps me tell the two modes apart.

artem-zinnatullin: k, since they live in dedicated clas, it's ok

davidmoten:
I see that the predicate is based only on the latest emission from upstream. If the predicate was a `Func2<List<T>,T, Boolean>` then the operator would even more useful and with little change to the existing operator.

davidmoten:
I prefer `bufferUntil/While` over `bufferSplit(predicate, true/false)`. The second version is not as readable as the first.

stevegury:
+1 on `bufferUntil`/`bufferWhile` but I'm not strongly opinionated about that.

akarnokd:
What's the verdict? bufferWile/bufferUntil where until's function has to emit false?

akarnokd:
Restored the original naming. In both operators, the predicate has to return true to trigger the next buffer. Let me know if you want some other rule.

stevegury:
:+1: I'm with @davidmoten about the naming.
Ping @stealthcode @abersnaze for their input.

ZacSweers:
👍 . Could see this having a lot of applications for something like batching and analytics when you want to wait for certain conditions.

Issue #3732, opened 2016-02-24
Request: Single.using


GrahamRogers:
To match `Observable.using`

    public static <T, Resource> Single<T> using(
            Func0<Resource> resourceFactory,
            Func1<? super Resource, ? extends Single<? extends T>> singleFactory,
            Action1<? super Resource> disposeAction)

and

    public static <T, Resource> Single<T> using(
            Func0<Resource> resourceFactory,
            Func1<? super Resource, ? extends Single<? extends T>> singleFactory,
            Action1<? super Resource> disposeAction, boolean disposeEagerly)

akarnokd:
Would you like to submit a PR?

akarnokd:
I'll do this.

akarnokd:
See #3752

artem-zinnatullin:
PR was merged and released as part of v1.1.2, this issue can be closed!

akarnokd:
Closing via #3752.

PR #3752, opened 2016-03-13
1.x: Single.using()

akarnokd:
This PR adds the factory method `using()` +1 overload to `Single` as requested in #3732.

For convenience, I've also added `Single.subscribe(Observer)`.

### DIFF
--- a/src/test/java/rx/internal/operators/SingleOnSubscribeUsingTest.java
+++ b/src/test/java/rx/internal/operators/SingleOnSubscribeUsingTest.java
@@ -1,0 +1,492 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.operators;
+
+import static org.junit.Assert.*;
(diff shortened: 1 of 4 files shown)


src/test/java/rx/internal/operators/SingleOnSubscribeUsingTest.java
```diff
@@ -0,0 +1,492 @@
...
+        inOrder.verify(observer).onCompleted();
+        inOrder.verifyNoMoreInteractions();
+
+        // The resouce should be closed
+        verify(resource).dispose();
+    }
+
+    @Test
+    public void withSubscribingTwice() {
+        performTestUsingWithSubscribingTwice(false);
```
artem-zinnatullin: nit: probably worth to create constants with meaningful name to use them instead of `true/false`, something like:

```java
@Test
public void withSubscribingTwice() {
  performTestUsingWithSubscribingTwice(DONT_DISPOSE_EAGERLY);
}
```

akarnokd: Maybe next time. I'm already removing `test` prefixes for you.

src/test/java/rx/internal/operators/SingleOnSubscribeUsingTest.java
```diff
@@ -0,0 +1,492 @@
...
+        Func1<Subscription, Single<Integer>> observableFactory = new Func1<Subscription, Single<Integer>>() {
+            @Override
+            public Single<Integer> call(Subscription subscription) {
+                throw new TestException();
+            }
+        };
+
+        try {
+            Single.using(resourceFactory, observableFactory, disposeSubscription)
+            .toBlocking().value();
```
artem-zinnatullin: super nit: why not just `subscribe()`?

akarnokd: This file based on the OnSubscribeUsingTest. Besides, I like value() throwing instead of having some OnErrorNotImplementedException bubbling up.

artem-zinnatullin:
:+1:

stevegury:
LGTM
:+1:

Issue #1389, opened 2014-06-26
SerializedObserverTest.testNoficationDelay not deterministic


davidmoten:
This [test](https://github.com/Netflix/RxJava/blob/0.19.1/rxjava-core/src/test/java/rx/observers/SerializedObserverTest.java#L274-284) is failing 50% of the time on my old laptop and if I bump up ```n``` to 10000 on a modern desktop it fails all the time.

My laptop runs ubuntu 12.04, i5 4 virtual processors, java 8 runtime, and failed 5 out of 10 tests with n=10.
My desktop runs ubuntu 12.04, xeon 12 virtual processors, java 8 runtime, and fails every time after i gets between 1000 and 3000

The line failing is line 330
```assertSame(t1, t2);```

akarnokd:
Potential fix posted: #3753

akarnokd:
Closing via #3753.

PR #3753, opened 2016-03-13
1.x: fix SerializedObserverTest.testNotificationDelay

akarnokd:
This PR fixes the `SerializedObserverTest.testNotificationDelay` test to use reentrancy to verify the buffer in `SerializedObserver` is properly drained. For now, it loops 10,000 times to see if the CI brings up the issue again. Takes 400ms on my machine but I'll probably reduce it to 1,000 times.

(The previous version was a bit convoluted with all those countdowns that didn't properly lock out the other thread).

Reported in issue #1389

### DIFF
--- a/src/test/java/rx/observers/SerializedObserverTest.java
+++ b/src/test/java/rx/observers/SerializedObserverTest.java
@@ -19,7 +19,7 @@ import static org.junit.Assert.*;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.*;
 
-import java.util.Arrays;
+import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
:+1: 

// Took me awhile to understand what are you doing in the test

src/test/java/rx/observers/SerializedObserverTest.java
```diff
@@ -255,74 +257,62 @@ public void runConcurrencyTest() {
...
-                TestSubscriber<String> to = new TestSubscriber<String>(new Observer<String>() {
-
+                
+                @SuppressWarnings("unchecked")
+                final Observer<Integer>[] os = new Observer[1];
+                
+                final List<Thread> threads = new ArrayList<Thread>();
+                
+                final Observer<Integer> o = new SerializedObserver<Integer>(new Observer<Integer>() {
+                    boolean first;
```
stevegury: Don't you think that being explicit improve readability?
`boolean first = false;`

stevegury:
:+1:

src/test/java/rx/observers/SerializedObserverTest.java
akarnokd: No, there is no proper guarantee in the JVM that default-valued initializers get erased and I personally don't like it either.

Issue #1653, opened 2014-09-01
Optimisations with Observable.empty()


davidmoten:
I notice that calling ```observable.concat(Observable.empty())``` invokes back-pressure. I'd find it useful if when I used ```Observable.empty()``` that optimisations came into play such as ensuring that ```observable.concat(Observable.empty())``` simply returned ```observable```.

To achieve this I'd make a private constant ```EMPTY = from(new ArrayList())```  in ```Observable``` so that calling ```Observable.empty()``` always returned this object cast into the appropriate generic type. Then I'd use a simple object reference equality test to determine if an optimisation could be made.

Do you think this proposal has legs? If so, I'll knock up a PR. I would seek to optimise more than just the ```concat``` operator , it would include ```merge``` and possibly others which I can think about if the time comes.

akarnokd:
Why do you have so many ```empty()``` to merge?

davidmoten:
Ta for improvement. The reason I have ```empty()```s to merge is that this is a common pattern for me:

```java
o.flatMap(t -> {
   if (cond(t)) 
       return someObservable(t);
  else 
       return Observable.empty(); 
  });
```
I'm experimenting now with substituting the special case where the flatMap function returns either 0 or 1 by doing 

```java
o.map(toOptional).filter(isPresent).map(toValue);
```
I'd hope it would be faster but not sure.

akarnokd:
This PR is quite old. Can you create a reasonable benchmark to see how concatMap and flatMap behaves? In addition, I suggest we tackle this after both #2928 and #2960 are merged.

davidmoten:
Yep I'll do benchmarks after those PRs have been merged.

akarnokd:
This may be part of the operator-fusion optimizations in 2.x but I don't see this in 1.x due to non-technical reasons.

akarnokd:
For a start: here is the baseline perf: #3754. I'm going to merge it so subsequent PRs can be benchmarked with it.

akarnokd:
See #3759 for the optimized `concat`s.

akarnokd:
See #3761 for the optimized `merge`s

akarnokd:
Optimizations have been added and released with 1.1.4. Closing the issue. If you have more ideas where it's worth adding extra checks for empty, don't hesitate to open a new issue.

PR #3754, opened 2016-03-13
1.x: measure flatMap/concatMap performance when used as filter

akarnokd:
This PR adds a perf class to measure the overhead of using `empty()` when the `flatMap`/`concatMap` emulates `filter`.

This will establish the comparison baseline for `flatMap`/`concatMap` optimization as requested in #1653.

The baseline [numbers](https://gist.github.com/akarnokd/243a09e28edfc27aeb0d) (i7 4770K, Windows 7 x64, Java 8u72):

image

Comparing `flatMap` against `concatMap`:

image

Here, mask indicates the and-mask that makes emitting `empty` less frequent: 0 = always, 1 = every other, 3 = every fourth, 7 = every eighth. This also means that `rangeEmpty` starts emitting more and more values with higher mask value hence the different throughput values.

### DIFF
--- a/src/perf/java/rx/operators/FlatMapAsFilterPerf.java
+++ b/src/perf/java/rx/operators/FlatMapAsFilterPerf.java
@@ -1,0 +1,119 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


davidmoten:
Thanks @akarnokd, very good of you. I'll set up a PR to use it.

PR #3757, opened 2016-03-14
1.x: Operator sample emits last sampled value before termination.

### LINKED ISSUE #3657: Sample operator: last item before termination is dropped after sampling

kboyarshinov:
Changes:
- `OperatorSampleWithTime` emits last stored value if it is set before `onCompleted`
- `OperatorSampleWithObservable` emits last sampled value if source or sample complete.

As discussed in https://github.com/ReactiveX/RxJava/issues/3657.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
+++ b/src/main/java/rx/internal/operators/OperatorSampleWithObservable.java
@@ -66,7 +66,7 @@ public final class OperatorSampleWithObservable<T, U> implements Operator<T, T> 
 
             @Override
             public void onCompleted() {
-                // onNext(null); // emit the very last value?
+                onNext(null);
                 s.onCompleted();
                 // no need to null check, main is assigned before any of the two gets subscribed
                 main.get().unsubscribe();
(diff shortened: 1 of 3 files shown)


akarnokd:
:+1:

stevegury:
:+1:

PR #3759, opened 2016-03-14
1.x: concatMap full rewrite + delayError + performance

akarnokd:
This is a full rewrite of the `concat`/`concatMap` operator with the following properties:

  - using `OnSubscribe` to save on the allocation of `lift`
  - `concatDelayError` and `concatMapDelayError` to delay errors to the very end
  - new trampolining logic that also subscribes to the next source even if requested is zero
  - using `concatMap` or `concatMapDelayError` on a scalar source now bypasses the operator completely and uses `ScalarSynchronousObservable.scalarFlatMap` as well.
  - the operator now detects `just` and `empty` sources and bypasses subscribing accordingly
  - added `ExceptionsUtils` to work with composite exceptions and terminal state; useful for other similar delaying operators later on.
  - `concat` is now implemented as `concatMap(v -> v)` which saves on the allocation of an entire `map` operator
  - `OperatorWindowWithSizeTest.testBackpressureOuterInexact` had to be ignored because `window` doesn't honor backpressure properly and it is fixed in #3678; once it has been resolved, the test can be reenabled
  - Added varargs constructor to `CompositeException` for convenience; avoids creating a list when the number of initial exceptions are known upfront.

Performance comparison to baseline (i7 4770K, Windows 7 x64, Java 8u72):

image


akarnokd:
This is the benchmark comparison against Rsc:

image

stevegury:
LGTM :+1:

PR #3760, opened 2016-03-14
1.x: fix ExecutorScheduler and GenericScheduledExecutorService reorder bug

akarnokd:
This PR relates to the failure of `errorThrownIssue1685`.

The underlying problem was with the `GenericScheduledExecutorService`. By being multi-threaded, tasks scheduled from the same thread one after the other may get reordered because different worker threads inside the pool could pick them up at the same time. In this case, there is no guarantee they keep their FIFO order.

*(I currently have no idea how one can use trampolining for this case; subsequent tasks may have any relative delays in respect to each other.)*

The solution creates N single threaded `ScheduledExecutorService`s and getInstance() hands one of them out. In turn `ExecutorService` takes one in its worker upfront.

However, there is still the problem when the programmer uses a multi-threaded `ScheduledExecutorService` with `Schedulers.from()` when the same issue comes back. A solution to that problem would be to always use the new `GenericScheduledExecutorService` for delaying timed tasks.

### DIFF
--- a/src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
+++ b/src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
@@ -31,24 +31,29 @@ import rx.schedulers.*;
  * the work asynchronously on the appropriate {@link Scheduler} implementation. This means for example that you would not use this approach
  * along with {@link TrampolineScheduler} or {@link ImmediateScheduler}.
  */
-public final class GenericScheduledExecutorService implements SchedulerLifecycle{
+public final class GenericScheduledExecutorService implements SchedulerLifecycle {
 
     private static final String THREAD_NAME_PREFIX = "RxScheduledExecutorPool-";
     private static final RxThreadFactory THREAD_FACTORY = new RxThreadFactory(THREAD_NAME_PREFIX);
 
-    private static final ScheduledExecutorService NONE;
+    private static final ScheduledExecutorService[] NONE = new ScheduledExecutorService[0];
+
+    private static final ScheduledExecutorService SHUTDOWN;
     static {
-        NONE = Executors.newScheduledThreadPool(0);
-        NONE.shutdownNow();
+        SHUTDOWN = Executors.newScheduledThreadPool(0);
+        SHUTDOWN.shutdown();
     }
 
     /* Schedulers needs acces to this in order to work with the lifecycle. */
     public final static GenericScheduledExecutorService INSTANCE = new GenericScheduledExecutorService();
     
-    private final AtomicReference<ScheduledExecutorService> executor;
+    private final AtomicReference<ScheduledExecutorService[]> executor;
 
+    /** We don't use atomics with this because thread-assignment is random anyway. */
+    private static int roundRobin;
+    
     private GenericScheduledExecutorService() {
-        executor = new AtomicReference<ScheduledExecutorService>(NONE);
+        executor = new AtomicReference<ScheduledExecutorService[]>(NONE);
         start();
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
```diff
@@ -31,24 +31,29 @@
...
 
-    private static final ScheduledExecutorService NONE;
+    private static final ScheduledExecutorService[] NONE = new ScheduledExecutorService[0];
+
+    private static final ScheduledExecutorService SHUTDOWN;
     static {
-        NONE = Executors.newScheduledThreadPool(0);
-        NONE.shutdownNow();
+        SHUTDOWN = Executors.newScheduledThreadPool(0);
+        SHUTDOWN.shutdown();
```
artem-zinnatullin: Ew

I definitely don't like that library will create executor when this class will be loaded. Even without threads used it looks strange.

Maybe return new each time `getInstance()`?

src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
```diff
@@ -63,39 +68,60 @@ public void start() {
...
-     * See class Javadoc for information on what this is for and how to use.
+     * Returns one of the single-threaded ScheduledExecutorService helper executors.
      * 
      * @return {@link ScheduledExecutorService} for generic use.
      */
     public static ScheduledExecutorService getInstance() {
-        return INSTANCE.executor.get();
+        ScheduledExecutorService[] execs = INSTANCE.executor.get();
+        if (execs == NONE) {
+            return SHUTDOWN;
```
artem-zinnatullin: How frequent is this case that we need `static final SHUTDOWN`?

src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
akarnokd: Executors don't start their thread until the first task is submitted.

src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
akarnokd: Unknown. Since Schedulers now have lifecycle, this is the correct behavior when some latent sequence tries to schedule something.

artem-zinnatullin: Well, ok. Another solution is to make `SHUTDOWN` lazy via hiding it under separate class, but it will increase methods count, so, I guess, k, let's live with it

stevegury:
I have trouble understanding how this solve the problem, could you please elaborate?
The only thing I see that we could do to force the FIFO order is to cap the thread number to 1, right?

akarnokd:
Yes, ExecutorScheduler's worker needs a helper ScheduledExecutorService with a single thread only. But we don't want all ExecutorSchedulers to wait in a single thread for their time to run. This change, similar to how computation scheduler works, hands out single hreaded ScheduledExecutorServices on demand.

zsxwing:
Nice catch.

> However, there is still the problem when the programmer uses a multi-threaded ScheduledExecutorService with Schedulers.from() when the same issue comes back. A solution to that problem would be to always use the new GenericScheduledExecutorService for delaying timed tasks.

I vote for `always use the new GenericScheduledExecutorService for delaying timed tasks.`. Correctness is more important than performance. Moreover, when people use `schedule(action, time)`, they usually don't want the codes to run as fast as possible, so the performance lost doesn't really matter.

stevegury:
Agree with @zsxwing that correctness is more important than performance.

src/main/java/rx/internal/schedulers/GenericScheduledExecutorService.java
```diff
@@ -63,39 +68,60 @@ public void start() {
...
-        if (executor.compareAndSet(NONE, exec)) {
-            if (!NewThreadWorker.tryEnableCancelPolicy(exec)) {
-                if (exec instanceof ScheduledThreadPoolExecutor) {
-                    NewThreadWorker.registerExecutor((ScheduledThreadPoolExecutor)exec);
+        // A multi-threaded executor can reorder tasks, having a set of them
+        // and handing one of those out on getInstance() ensures a proper order
+        
+        ScheduledExecutorService[] execs = new ScheduledExecutorService[count];
+        for (int i = 0; i < count; i++) {
+            execs[i] = Executors.newScheduledThreadPool(count, THREAD_FACTORY);
```
zsxwing: `count` should be replaced with `1` to guarantee FIFO. Right?

akarnokd:
Right, I'll fix them shortly.

akarnokd:
Updated.

zsxwing:
:+1:

stevegury:
:+1:

PR #3762, opened 2016-03-14
1.x: Deprecate CompositeException constructor with message prefix

vanniktech:
Before this PR messagePrefix was never used :dancer: 

If you guys want I can revert the change on CompositeExceptionTest.java:177. I just thought that way it' be nicer.

About the message is there any reason for having a space after the dot?

### DIFF
--- a/src/main/java/rx/exceptions/CompositeException.java
+++ b/src/main/java/rx/exceptions/CompositeException.java
@@ -41,6 +41,8 @@ public final class CompositeException extends RuntimeException {
     private final List<Throwable> exceptions;
     private final String message;
 
+    /** @deprecated please use {@link #CompositeException(Collection)} */
+    @Deprecated
     public CompositeException(String messagePrefix, Collection<? extends Throwable> errors) {
         Set<Throwable> deDupedExceptions = new LinkedHashSet<Throwable>();
         List<Throwable> _exceptions = new ArrayList<Throwable>();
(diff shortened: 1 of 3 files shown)


akarnokd:
We don't really use that constructor or the message, so I'd prefer removing it altogether.

vanniktech:
Yeah I also realized that later since no test broke after that change. So do you want me to remove it?

akarnokd:
Yes, but don't rush it as there is another, much heavier PR #3759 that also changed this class.

vanniktech:
Alright just ping me once #3759 is in

vanniktech:
@akarnokd I updated the PR again and changed the commit message and the title. I hope the changes are okay now. If the removal of whitespace bugs you just let me know.

akarnokd:
Yes, please revert those space changes. (Are you using IntelliJ?)

vanniktech:
Those changes I did in Sublime and I set it up in a way that it should remove trailing whitespace ;)

Also updated the PR and removed the spaces

akarnokd:
Almost, but forgot that its a public API component thus can't just remove the constructors. Could you instead deprecate `CompositeException(String message, List<Throwable> errors)`?

vanniktech:
@akarnokd like this?

akarnokd:
Yes.

:+1:

artem-zinnatullin:
:+1:

stevegury:
:+1:

PR #3763, opened 2016-03-14
1.x: combineLatestDelayError

akarnokd:
This PR exposes the `delayError` option in `combineLatest` as `combineLatestDelayError`.

(Note that we have convenience overloads with 2-9 sources that could also use an overload, but that just expands the API surface drastically.)

### DIFF
--- a/src/test/java/rx/internal/operators/OnSubscribeCombineLatestTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeCombineLatestTest.java
@@ -29,6 +29,7 @@ import org.mockito.*;
 import rx.*;
 import rx.Observable;
 import rx.Observer;
+import rx.exceptions.*;
 import rx.functions.*;
 import rx.internal.util.RxRingBuffer;
 import rx.observers.TestSubscriber;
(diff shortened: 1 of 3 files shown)


src/test/java/rx/internal/operators/OnSubscribeCombineLatestTest.java
```diff
@@ -954,5 +955,107 @@ public Integer call(Object... args) {
...
-    }; 
+    };
+    
+    @SuppressWarnings("unchecked")
+    @Test
+    public void firstJustError() {
+        TestSubscriber<Integer> ts = TestSubscriber.create();
+        
+        Observable.combineLatestDelayError(
+                Arrays.asList(Observable.just(1), Observable.<Integer>error(new TestException())),
```
artem-zinnatullin: Looks like this should be in `secondJustError()` and vice versa

artem-zinnatullin:
:+1: 

(I'm starting feeling bad for guy who asked for Map with indexes, since his request we added so many overloads of other operators…)

abersnaze:
:+1:

stevegury:
:+1:

PR #3765, opened 2016-03-14
1.x: switchOnNextDelayError and switchMapDelayError

akarnokd:
This PR adds a delayError version to the `switchOnNext` / `switchMap` operators.


stevegury:
LGM :+1:

PR #3766, opened 2016-03-14
1.x: Add Single.onErrorResumeNext(Func)

### LINKED ISSUE #3440: Single.onErrorResumeNext
Observable has a onErrorResumeNext to continue with a new Observable if onError is called. I'd like something similar for Single, so:

```java
Single<T> onErrorResumeNext(Single<? extends T> resumeSingle)

Single<T> onErrorResumeNext(Func1<? super Throwable, ? extends Single<? extends T>> resumeFunction)
```

Single already has onErrorReturn, it makes sense to have onErrorResumeNext as well.

### LINKED ISSUE #3472: Flatmap Single into Observable and ignore errors
I have UI observable which should work as long as my UI lives. When my UI emits I create a network request which could be an Observable or a Single. A Single fits best for a network call.

```java
RxView.clicks(myButton)
    .flatMap(v - > networkRequest())
    .subscribe(data -> showSomething(data),
            e -> {
                // should only be called when the RxView.clicks() throws. 
                // should not be called for network errors
            });
```

When using an Observable I would use `onErrorResumeNext` returning `Observable.empty()` to prevent errors coming from the network request going into my UI Observable and calling `onError` because the UI Observable should life forever:

```java
Observable networkRequest() {
    mApiService.fireActionObservable()
        .onErrorResumeNext(throwable -> {
            // somehow handle error here
            return Observable.empty();
     });
}
```

This, in my opinion, elegant way does not work for Single because no `Single.empty()` exists.

```java
Single networkRequest() {
    mApiService.fireActionSingle()
        .onErrorResumeNext(throwable -> {
            // somehow handle error here
            return Single.empty(); // <-- does not exist. I have to call success or error :/
     });
}
```

`Single.onErrorResumeNext` is btw only available in 2.x

Converting my `Single` to an `Observable` seems wrong, because the network request is a `Single`!

### LINKED ISSUE #3731: Request: Single.onErrorResumeNext(Func1<? super Throwable, Single<? extends T>> resumeFunction)
To match `Observable`'s.

    public Single<V> onErrorResumeNext(Func1<? super Throwable, Single<? extends T>> resumeFunction);

artem-zinnatullin:
Closes #3440, closes #3731, closes #3472 (whoa, 3 issues at a time!)


src/main/java/rx/internal/operators/SingleOperatorOnErrorResumeNext.java
```diff
@@ -0,0 +1,68 @@
+package rx.internal.operators;
+
+import rx.Single;
+import rx.SingleSubscriber;
+import rx.exceptions.Exceptions;
+import rx.functions.Func1;
+import rx.plugins.RxJavaPlugins;
+
+public class SingleOperatorOnErrorResumeNext<T> implements Single.OnSubscribe<T> {
```
artem-zinnatullin: Just renamed `SingleOperatorOnErrorResumeNextViaSingle` to something more general and it's now works with function instead of `Single` directly.

src/main/java/rx/internal/operators/SingleOperatorOnErrorResumeNext.java
```diff
@@ -0,0 +1,68 @@
...
+    public void call(final SingleSubscriber<? super T> child) {
+        final SingleSubscriber<? super T> parent = new SingleSubscriber<T>() {
+            @Override
+            public void onSuccess(T value) {
+                child.onSuccess(value);
+            }
+
+            @Override
+            public void onError(Throwable error) {
+                RxJavaPlugins.getInstance().getErrorHandler().handleError(error);
```
akarnokd: I think we shouldn't call the error handler if the error can be delivered or replaced on the normal path.

src/main/java/rx/internal/operators/SingleOperatorOnErrorResumeNext.java
```diff
@@ -0,0 +1,68 @@
...
+
+                try {
+                    unsubscribe();
```
akarnokd: This is unnecessary.

artem-zinnatullin: ah, it's single, right

src/main/java/rx/internal/operators/SingleOperatorOnErrorResumeNext.java
```diff
@@ -0,0 +1,68 @@
...
+                    resumeFunctionInCaseOfError.call(error).subscribe(child);
```
akarnokd: I'd separate catching the error and subscribing to the Single outside the try-catch.

artem-zinnatullin:
@akarnokd fixed your comments, btw, should I add `@Experimental` to these operators?

src/test/java/rx/SingleTest.java
```diff
@@ -1220,13 +1220,59 @@ public void onErrorResumeNextViaSingleShouldPreventNullSingle() {
...
+                        return Single.just("fallback");
+                    }
+                })
+                .subscribe(testSubscriber);
+
+        testSubscriber.assertValue("fallback");
+    }
+
+    @Test
+    public void onErrorResumeNextViaFunctionShouldPreventNullFunction() {
```
akarnokd: Please also check/handle if the function returns a null Single.

akarnokd:
Yes, and copy over the experimental and since tags into the javadoc.

src/test/java/rx/SingleTest.java
artem-zinnatullin: Added a test, but didn't handle this specifically in the operator, test will make sure that it won't be swallowed

akarnokd:
:+1: There is this new like option but do you get a notification for them?

JakeWharton:
No

src/main/java/rx/internal/operators/SingleOperatorOnErrorResumeNext.java
```diff
@@ -0,0 +1,65 @@
...
+        final SingleSubscriber<? super T> parent = new SingleSubscriber<T>() {
+            @Override
+            public void onSuccess(T value) {
+                child.onSuccess(value);
+            }
+
+            @Override
+            public void onError(Throwable error) {
+                try {
+                    resumeFunctionInCaseOfError.call(error).subscribe(child);
```
zsxwing: Why remove `unsubscribe()`?

artem-zinnatullin: It's terminal state, we don't do `unsubscribe()` in `onSuccess()` too

zsxwing: If `resumeFunctionInCaseOfError` returns something like `Observable.never.toSingle`, we should still `unsubscribe` the original one.

zsxwing: Never mind. I just found `SafeSubscriber` will do it.

zsxwing:
:+1:

stevegury:
:+1:

PR #3767, opened 2016-03-15
1.x: Update Gradle wrapper to 2.12

artem-zinnatullin:
Release notes https://docs.gradle.org/2.12/release-notes

>Gradle 2.12 has the fastest build script compilation speeds of any Gradle version to date.


stevegury:
You should also check that there isn't any breaking changes in the [gradle plugin](https://github.com/nebula-plugins/gradle-rxjava-project-plugin). Some of the plugins expect specific version of gradle.

In your case, I think it's fine: :+1:

akarnokd:
:+1:

PR #3768, opened 2016-03-15
1.x: observeOn - fix in-sequence termination/unsubscription

akarnokd:
This fixes `observeOn` not completing immediately if the downstream requested exactly the remaining amount. The original version required a new request from downstream to trigger the delivery of the terminal event.

This also fixes the rare failure in `testNoMoreRequestsAfterUnsubscribe` because the unsubscription is now checked in-sequence and doesn't trigger the unwanted replenishment request.

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorObserveOnTest.java
+++ b/src/test/java/rx/internal/operators/OperatorObserveOnTest.java
@@ -834,4 +834,27 @@ public class OperatorObserveOnTest {
         ts.assertError(TestException.class);
         ts.assertNotCompleted();
     }
+    
+    @Test
+    public void requestExactCompletesImmediately() {
+TestSubscriber<Integer> ts = TestSubscriber.create(0);
+        
+        TestScheduler test = Schedulers.test();
+
+        Observable.range(1, 10).observeOn(test).subscribe(ts);
+
+        test.advanceTimeBy(1, TimeUnit.SECONDS);
+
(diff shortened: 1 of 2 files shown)


src/test/java/rx/internal/operators/OperatorObserveOnTest.java
```diff
@@ -834,4 +834,16 @@ public void testErrorDelayedAsync() {
         ts.assertError(TestException.class);
         ts.assertNotCompleted();
     }
+    
+    @Test
+    public void requestExactCompletesImmediately() {
```
artem-zinnatullin: Not sure if it tests "immediately", this test passes without changes in this PR.

Maybe you can avoid going to another thread and `awaitTerminalEvent()`?

artem-zinnatullin:
:+1: for changes, not sure about test

src/test/java/rx/internal/operators/OperatorObserveOnTest.java
akarnokd: Yeah, should request 0 upfront and wait till the range completes before requesting 10. I'll change this to use the TestScheduler.

akarnokd: Updated the test which now clearly fails on master.

artem-zinnatullin:
:+1:

stevegury:
:+1:

Issue #3751, opened 2016-03-13
Confusion when using Backpressure operators


troinine:
Hey,

We have a use case in which a consumer might not be able to process items fast enough than what is emitted from a source observable. I understood that in this case, a backpressure with either ```onBackPressureBuffer()``` or ```onBackPressureDrop()``` might be useful. In case of overflow / drop, we would like to store items to a local storage and try processing them later when the consumer in this case is again able to handle the input rate. Our consumer is actually a remote REST call which might timeout or not be available in which case we retry.

Anyways, I tried alternative ways to address the problem but I can't find a suitable way to solve it. To illustrate my testings, here is some code:

```
package io.reactivex;

import org.junit.Test;
import rx.Observable;
import rx.Subscriber;
import rx.schedulers.Schedulers;

public class BackPressureTest {
    @Test
    public void testOnBackPressureDrop() throws InterruptedException {
        Observable<Integer> emitter = toObservable()
                .subscribeOn(Schedulers.newThread());

        emitter.onBackpressureDrop(i -> System.out.println("Dropped " + i))
                .observeOn(Schedulers.computation())
                .map(this::doWork)
                .doOnNext(i -> System.out.println("Output " + i))
                .toBlocking()
                .subscribe(new SingleItemSubscriber<>());
    }

    @Test
    public void testOnBackPressureBuffer() throws InterruptedException {
        Observable<Integer> emitter = toObservable()
                .subscribeOn(Schedulers.newThread());

        emitter.onBackpressureBuffer(2, () -> System.out.println("Overflow"))
                .observeOn(Schedulers.computation())
                .map(this::doWork)
                .doOnNext(i -> System.out.println("Output " + i))
                .toBlocking()
                .subscribe(i -> System.out.println("Subscriber received " + i));
    }

    private Observable<Integer> toObservable() {
        return Observable.create(subscriber -> {
            for (int i = 0; i < 10; i++) {
                System.out.println("Emitting " + i);

                subscriber.onNext(i);

                try {
                    Thread.sleep(250);
                } catch (InterruptedException e) {
                    subscriber.onError(e);
                }
            }

            subscriber.onCompleted();
        });
    }

    private int doWork(int integer) {
        System.out.println("Consuming " + integer);
        try {
            Thread.sleep(1000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }

        return integer;
    }

    private static class SingleItemSubscriber<T> extends Subscriber<T> {
        @Override
[...]

PR #3777, opened 2016-03-17
observeOn: allow configurable buffer size

srvaroa:
The observeOn operator is backed by a small queue of 128 slots that may
overflow quickly on slow producers.  This could only be avoided by
adding a backpressure operator before the observeOn (not only
inconvenient, but also taking a perf. hit as it forces hops between two
queues).

This patch allows modifying the default queue size on the observeOn
operator.

Fixes: #3751
Signed-off-by: Galo Navarro <anglorvaroa@gmail.com>

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -6291,7 +6291,8 @@ public class Observable<T> {
     
     /**
      * Modifies an Observable to perform its emissions and notifications on a specified {@link Scheduler},
-     * asynchronously with a bounded buffer.
+     * asynchronously with a bounded buffer of {@link RxRingBuffer.SIZE} slots.
+     *
      * <p>Note that onError notifications will cut ahead of onNext notifications on the emission thread if Scheduler is truly
      * asynchronous. If strict event ordering is required, consider using the {@link #observeOn(Scheduler, boolean)} overload.
      * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/rx/Observable.java
```diff
@@ -6308,13 +6309,42 @@ public final void forEach(final Action1<? super T> onNext, final Action1<Throwab
...
+     * @param bufferSize the size of the buffer.
+     * @return the source Observable modified so that its {@link Observer}s are notified on the specified
+     *         {@link Scheduler}
+     * @see <a href="http://reactivex.io/documentation/operators/observeon.html">ReactiveX operators documentation: ObserveOn</a>
+     * @see <a href="http://www.grahamlea.com/2014/07/rxjava-threading-examples/">RxJava Threading Examples</a>
+     * @see #subscribeOn
+     * @see #observeOn(Scheduler)
+     * @see #observeOn(Scheduler, boolean)
+     */
+    public final Observable<T> observeOn(Scheduler scheduler, int bufferSize) {
```
akarnokd: Please add `observeOn(Scheduler, boolean delayError, int bufferSize)` overload as well.

src/test/java/rx/internal/operators/OperatorObserveOnTest.java
```diff
@@ -523,14 +523,20 @@ public boolean hasNext() {
...
 
+    /**
+     * {@link BackpressureTests.testObserveOnWithSlowConsumer} also covers the case with default buffer size.
+     */
     @Test
     public void testQueueFullEmitsError() {
         final CountDownLatch latch = new CountDownLatch(1);
+        // randomize buffer size, note that underlaying implementations may be tuning the real size to a power of 2
+        // which can lead to unexpected results when adding excess capacity (e.g.: see ConcurrentCircularArrayQueue)
+        final int capacity = (int)Math.pow(2, new Random().nextInt(10));
```
akarnokd: There is no need for randomized capacity. If you want to verify multiple sizes, loop over a power-of-2 set.

artem-zinnatullin: oh yes, please don't use `Random` in tests, they will be non-reproducible

akarnokd:
:+1:

src/test/java/rx/internal/operators/OperatorObserveOnTest.java
```diff
@@ -523,14 +523,21 @@ public boolean hasNext() {
         assertEquals(7, generated.get());
     }
 
+    /**
+     * {@link BackpressureTests.testObserveOnWithSlowConsumer} also covers the case with default buffer size.
```
artem-zinnatullin: nit: this link won't work, I guess

artem-zinnatullin:
:+1: though i'd leave old test for default buffer size and just add a new one

srvaroa:
Rebased, fixed the javadoc link, and a spelling mistake.

@artem-zinnatullin I reused this one to avoid redundancy as the code path is virtually the same, and there are tests (the ones referenced in the @link) which use the default size. Let me know if you still prefer the 2 cases and I'll fix that.

artem-zinnatullin:
Yeah, I understand, just afraid that one may change linked  tests and we won't cover overload. 

// one day we'll setup code coverage, I hope

src/main/java/rx/Observable.java
akarnokd: I still miss the overload `observeOn(Scheduler, boolean delayError, int bufferSize)`

src/test/java/rx/internal/operators/OperatorObserveOnTest.java
```diff
@@ -523,14 +523,21 @@ public boolean hasNext() {
...
+    /**
+     * Tests in {@link BackpressureTests} like testObserveOnWithSlowConsumer already cover the case with default buffer
+     * size, the code path is virtually identical.
+     */
     @Test
     public void testQueueFullEmitsError() {
         final CountDownLatch latch = new CountDownLatch(1);
+        // randomize buffer size, note that underlying implementations may be tuning the real size to a power of 2
+        // which can lead to unexpected results when adding excess capacity (e.g.: see ConcurrentCircularArrayQueue)
+        final int capacity = (int)Math.pow(2, new Random().nextInt(10));
```
akarnokd: I agree with @artem-zinnatullin, leave this test as it was and create a new one with a loop over some queue sizes:

```java
for (int i = 1; i <= 1024; i = i * 2) {
   final int capacity = i;
   // ...
}
```

srvaroa:
Agh thanks @akarnokd, I didn't notice I changed computers so I was back on the 1st patch and lost the overload. Resubmitted based on the one with all overloads, and added the extra test instead of rewriting it. Hopefully fine now.

akarnokd:
:+1:

src/test/java/rx/internal/operators/OperatorObserveOnTest.java
```diff
@@ -581,6 +581,69 @@ public void onNext(Integer t) {
     }
 
     @Test
+    public void testQueueFullEmitsErrorWithVaryingBufferSize() {
```
artem-zinnatullin: nit: `test` prefix is not needed

srvaroa:
I've seen some failures on the test locally and there is also a bug in the test (the loop should start at 2, not 1). I can't look into it right now but will take a look asap.

stevegury:
:+1:

akarnokd:
I'm merging this. The changes are algorithmically solid so I'm not sure what failures you see locally.

srvaroa:
Thanks @akarnokd. The failure was no error notifications emitted at random sizes, but I did notice Travis didn't complain. Ping me if it surfaces.

PR #3781, opened 2016-03-18
1.x: Add an operator to throttle data via controlling the requests going upstream

abersnaze:
Came across an interesting use case where someone needed a way to get an external signal (unknown to the subscriber) to throttle the data going through an observable chain.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -7665,6 +7665,26 @@ public class Observable<T> {
     }
 
     /**
+     * Allow the an external signal control the amount of data being set through this Observable chain.
+     * When the control Observable emits false (closes the valve) requests upstream are stopped and any
+     * requests from downstream for more data are buffered until the control Observable emits a true
+     * (opens the valve). Should the control Observable error or complete while closed (last control
+     * emition was a false) an error is sent down the data stream. The granularity breaks up large requests
+     * from downstream to limit the number of onNexts that are possible after the control valve has closed.
+     * The smaller the number the tighter the control on the flow but the more overhead there will be in
+     * managing the requests.
+     * 
+     * @param control
+     *             an Observable that dictates if request signals propagate upstream
+     * @param granularity
+     *             the maximum number of outstanding requests.
+     * @returns an Observable that mostly stops emiting after the control Observable emits a false.
+     */
+    public final Observable<T> pressureValve(Observable<Boolean> control, long granularity) {
+        return lift(new OperatorValve<T>(control, granularity));
+    }
+
+    /**
      * Returns an Observable that applies a specified accumulator function to the first item emitted by a source
      * Observable, then feeds the result of that function along with the second item emitted by the source
      * Observable into the same function, and so on until all items have been emitted by the source Observable,
--- a/src/main/java/rx/internal/operators/OperatorValve.java
+++ b/src/main/java/rx/internal/operators/OperatorValve.java
@@ -1,0 +1,204 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.operators;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import rx.Observable;
+import rx.Observable.Operator;
+import rx.Producer;
+import rx.Subscriber;
+
+/**
+ * An {@code Observable} that emits the first {@code num} items emitted by the source {@code Observable}.
+ * <p>
+ * <img width="640" height="305" src="https://raw.githubusercontent.com/wiki/ReactiveX/RxJava/images/rx-operators/take.png" alt="" />
+ * <p>
+ * You can choose to pay attention only to the first {@code num} items emitted by an {@code Observable} by using
+ * the {@code take} operator. This operator returns an {@code Observable} that will invoke a subscriber's
+ * {@link Subscriber#onNext onNext} function a maximum of {@code num} times before invoking
+ * {@link Subscriber#onCompleted onCompleted}.
+ */
+public final class OperatorValve<T> implements Operator<T, T> {
+    private final Observable<Boolean> onByDefault;
+    private final long _granularity;
+
+    public OperatorValve(Observable<Boolean> onByDefault, long granularity) {
+        this.onByDefault = onByDefault;
+        this._granularity = granularity;
+    }
+
+    @Override
+    public Subscriber<? super T> call(final Subscriber<? super T> child) {
+        return new Subscriber<T>(child) {
+            private final long granularity = _granularity;
+            private Producer p;
+            private long backlog;// synchronized access on Producer p
+            private long outstanding;// synchronized access on Producer p
+            private boolean isOpen = true;// synchronized access on Producer p
+            private AtomicBoolean terminated = new AtomicBoolean();
+
+            @Override
+            public void onCompleted() {
+                if (terminated.compareAndSet(false, true))
+                    child.onCompleted();
+            }
+
+            @Override
+            public void onError(Throwable e) {
+                if (terminated.compareAndSet(false, true))
+                    child.onError(e);
+            }
+
+            @Override
+            public void onNext(T t) {
+                child.onNext(t);
+                final long requestUp;
+                synchronized (this) {
+                    if (--outstanding == 0 && isOpen) {
+                        // all out and still open; check to see if there is a backlog.
+                        if (backlog > granularity) {
+                            // don't request too much at once
+                            requestUp = granularity;
+                        } else if (backlog > 0) {
+                            // the backlog isn't too big
+                            requestUp = backlog;
+                        } else {
+                            // no backlog
+                            requestUp = 0;
+                        }
+                    } else {
+                        // expecting more or closed
+                        requestUp = 0;
+                    }
+                    if (requestUp > 0) {
+                        // do the last of the accounting inside the synchronized block
+                        backlog -= requestUp;
+                        outstanding += requestUp;
+                    }
+                }
+                // do the request work outside the synchronized block
+                if (requestUp != 0)
+                    p.request(requestUp);
+            }
+
+            @Override
+            public void setProducer(final Producer p) {
+                this.p = p;
+
+                onByDefault.unsafeSubscribe(new Subscriber<Boolean>() {
+                    @Override
+                    public void onCompleted() {
+                        boolean _isOpen;
+                        synchronized (this) {
+                            // make sure to get the latest value of isOpen
+                            _isOpen = isOpen;
+                        }
+                        if (!_isOpen) {
+                            if (terminated.compareAndSet(false, true)) {
+                                child.onError(new IllegalStateException("control signal terminated while valve was closed"));
+                            }
+                        }
+                        unsubscribe();
+                    }
+
+                    @Override
+                    public void onError(Throwable e) {
+                        if (terminated.compareAndSet(false, true))
+                            child.onError(e);
+                        unsubscribe();
+                    }
+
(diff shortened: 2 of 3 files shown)


src/main/java/rx/Observable.java
```diff
@@ -1258,6 +1258,7 @@ public Completable toCompletable() {
         final static Observable<Object> INSTANCE = create(new OnSubscribe<Object>() {
             @Override
             public void call(Subscriber<? super Object> subscriber) {
+                subscriber.setProducer(ProducerHolder.INSTANCE);
```
akarnokd: Setting the producer is optional in 1.x and many other operators don't do it either; one has to work around it. Why add it just here?

src/main/java/rx/internal/operators/OperatorValve.java
```diff
@@ -0,0 +1,196 @@
...
+ * <img width="640" height="305" src="https://raw.githubusercontent.com/wiki/ReactiveX/RxJava/images/rx-operators/take.png" alt="" />
+ * <p>
+ * You can choose to pay attention only to the first {@code num} items emitted by an {@code Observable} by using
+ * the {@code take} operator. This operator returns an {@code Observable} that will invoke a subscriber's
+ * {@link Subscriber#onNext onNext} function a maximum of {@code num} times before invoking
+ * {@link Subscriber#onCompleted onCompleted}.
+ */
+public final class OperatorValve<T> implements Operator<T, T> {
+    private final Observable<Boolean> onByDefault;
+    private final long granularity;
```
akarnokd: Generates bridge method.

src/main/java/rx/internal/operators/OperatorValve.java
```diff
@@ -0,0 +1,196 @@
...
+    public OperatorValve(Observable<Boolean> onByDefault, long granularity) {
+        this.onByDefault = onByDefault;
+        this.granularity = granularity;
+    }
+
+    @Override
+    public Subscriber<? super T> call(final Subscriber<? super T> child) {
+        return new Subscriber<T>(child) {
+            private Producer p;
+            private long backlog = 0;// synchronized access on Producer p
```
akarnokd: Initialize to default?

src/main/java/rx/internal/operators/OperatorValve.java
```diff
@@ -0,0 +1,196 @@
...
+            private long outstanding = 0;// synchronized access on Producer p
+            private boolean isOpen = true;// synchronized access on Producer p
+            private AtomicBoolean terminated = new AtomicBoolean();
+
+            @Override
+            public void onCompleted() {
+                if (terminated.compareAndSet(false, true))
```
akarnokd: If the valve is closed, this will terminate the downstream, dropping all values between the close event and the onCompleted event. Is this intentional?

src/main/java/rx/internal/operators/OperatorValve.java
```diff
@@ -0,0 +1,196 @@
...
+            public void onError(Throwable e) {
+                if (terminated.compareAndSet(false, true))
+                    child.onError(e);
+            }
+
+            @Override
+            public void onNext(T t) {
+                child.onNext(t);
+                final long requestUp;
+                synchronized (p) {
```
akarnokd: Producer is optional, why not synchronize on `this` (or a separate guard object)?

src/main/java/rx/internal/operators/OperatorValve.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            @Override
+            public void setProducer(final Producer p) {
+                this.p = p;
+
+                onByDefault.unsafeSubscribe(new Subscriber<Boolean>() {
+                    @Override
+                    public void onCompleted() {
+                        if (!isOpen) {
```
akarnokd: isOpen accessed outside its guard.

src/main/java/rx/internal/operators/OperatorValve.java
```diff
@@ -0,0 +1,196 @@
...
+                                isOpen = false;
+                            }
+                        }
+                    }
+                });
+
+                super.setProducer(new Producer() {
+                    @Override
+                    public void request(long n) {
+                        final long requestUp;
```
akarnokd, line 170: validate n > 0?

src/main/java/rx/Observable.java
```diff
@@ -6705,6 +6706,26 @@ public final Boolean call(T t) {
     }
 
     /**
+     * Allow the an external signal control the amount of data being set through this Observable chain.
+     * When the control Observable emits false (closes the valve) requests upstream are stopped and any
+     * requests from downstream for more data are buffered until the control Observable emits a true
```
akarnokd: I don't see any buffering. The operator acts as a request limiter where request > granularity are chopped up into one or more request(granularity) or the leftover.

src/test/java/rx/internal/operators/OperatorValveTest.java
```diff
@@ -0,0 +1,163 @@
...
+        tSub.assertValues(1, 2, 3);
+    }
+
+    @Test
+    public void testControlCompleteClosed() {
+        TestSubscriber<Void> tSub = TestSubscriber.create();
+        Observable.<Void> never().pressureValve(Observable.just(false), 10).subscribe(tSub);
+        tSub.assertError(IllegalStateException.class);
+    }
+}
```
akarnokd, line 205: Please write tests with the most common operators:

range.pressureValve.observeOn()
range.pressureValve.flatMap()
range.pressureValve.flatMap(maxConcurrent=1, 2, 4,...,128)
range.flatMap(v -> range.pressureValve)
range.pressureValve.concatMap()
range.concatMap(v -> range.pressureValve)

src/main/java/rx/Observable.java
abersnaze: It buffers the n from requests while closed in the backlog variable.

src/main/java/rx/Observable.java
abersnaze: My unit test didn't work because I don't create a producer until setProducer is called called. If setProducer is optional then I'll probably have to refactor to use the producer arbiter.

src/main/java/rx/Observable.java
akarnokd: Ah, okay. First I thought this operator always requests granularity no matter how many items the downstream requests but its about holding off requests by aggregating them if the valve is closed.

src/main/java/rx/internal/operators/OperatorValve.java
abersnaze: I like being explicit. Is that going to cause a problem?

src/main/java/rx/internal/operators/OperatorValve.java
abersnaze, line 170: Will do.

src/main/java/rx/Observable.java
akarnokd: It's not about the arbiter, but when you trigger the subscription to the other subscriber. You'd probably have to use the frequent pattern:

``` java
MainSubscriber parent = new MainSubscriber(...)
ValveSubscriber valve = new ValveSubscriber(parent, ...);

child.add(parent);
child.add(valve);
child.setProducer(r -> parent.requestFromChild(r));

other.subscribe(valve);

return parent;
```

src/main/java/rx/internal/operators/OperatorValve.java
akarnokd: I don't know if it gets optimized away or not, but when one debugs the class, the single step will go over all of these field initializations. And for me, it's distracting and prompts if the author meant some reasonable initial value or not.

src/main/java/rx/Observable.java
abersnaze: I added the granularity parameter because the valve would have anything to turn off if the subscriber just requested max long once.

src/test/java/rx/internal/operators/OperatorValveTest.java
abersnaze, line 205: I'm having a hard time coming up with consistent unit tests for this. It is designed as a very sloppy back pressure mechanism; reactive and not prescriptive.  If I used the TestScheduler then there isn't much difference between the units tests you want and the contrived tests that I have now. For example

``` java
    @Test
    public void testObserveOn() {
        final AtomicLong counter = new AtomicLong();
        Observable<Integer> range = Observable.range(0, Integer.MAX_VALUE);
        Observable<Boolean> control = Observable.interval(1, TimeUnit.SECONDS).map(new Func1<Long, Boolean>() {
            @Override
            public Boolean call(Long i) {
                System.out.println();
                counter.set(0);
                return i % 2 == 1;
            }
        });
        long granularity = 10;
        TestSubscriber<Integer> tSub = new TestSubscriber<Integer>();
        range.pressureValve(control, granularity).observeOn(Schedulers.computation()).toBlocking().forEach(new Action1<Integer>() {
            @Override
            public void call(Integer t) {
                System.out.print(counter.incrementAndGet()+ "         \r");
            }
        });
    }
```

Produces the output like this where the number of `onNext`s received after the valve is closed can be anywhere from 8 to 127.

```
550343         
56             
598760         
23             
593937         
110            
597200         
47             
587776         
127            
586648         
103            
585984         
127            
579781         
58             
580374         
105            
581879         
8              
587867         
36             
585960         
23             
572516         
27             
581530         
101            
579535         
48             
583840         
95             
579200         
127            
583312         
111            
580183         
40             
578295         
8              
597470         
```

akarnokd, line 205: The tests I suggested were mainly to expose concurrency flaws and lack-of-request type hangs. In addition, it is up to you what request pattern you want to achieve and how precisely you want to pause the stream.

src/main/java/rx/Observable.java
```diff
@@ -6705,6 +6706,26 @@ public final Boolean call(T t) {
     }
 
     /**
+     * Allow the an external signal control the amount of data being set through this Observable chain.
```
simonbasle: there's a typo there: "Allow **the** an"

akarnokd:
FYI: https://gist.github.com/akarnokd/1c54e5a4f64f9b1e46bdcf62b4222f08

akarnokd:
Do you want to pursue this further?

akarnokd:
I'm in support for this functionality but the PR has test issues. Will you have time to fix it before 1.1.7? If not, would you want me to propose [my take on it](https://gist.github.com/akarnokd/1c54e5a4f64f9b1e46bdcf62b4222f08)?

My first month goal is to have 0 open 1.x PRs as we reach 1.1.7.

abersnaze:
Not in its current form. I was thinking that this could be made much simpler if this was composed with the rebatch operator since most of the code was batching the initial MAX_LONG request.

Because this rebatch doesn't have any queuing it makes it difficult to come up with invariants to test.

akarnokd:
Okay then, closing this for now. Don't hesitate to post a follow-up PR once you are ready.

ZacSweers:
I kind of attempted the inverse of this and was going to release it separately. Rather than throttling requests to the provider, it throttles emissions from the stream and buffers them. Been busy with work but here's what it looked like - https://gist.github.com/hzsweers/91de6e2b87fa2542dd38034c4cee198c. Could open a separate PR here if you think it's something you'd want here.

akarnokd:
@hzsweers is it like a "burster", i.e., collect a window of values and then burst them out? Not sure about its general use. In addition, it has race conditions and potentially missed signals.

ZacSweers:
It wouldn't "burst" them out, per se. It just ensures a minimum time between emissions, normalizing a flow. I even attempted a marble diagram for it, hope this makes sense. Note that the first three emissions after normalizing are all equally spaced by a minimum window of time. It's admittedly a niche use case, but figured I'd mention it after seeing this discussion.

image

ZacSweers:
There is a potential for missed signals, but only in the event of backpressure (which I don't really know how to account for) or unsubscribing before the normalizer has been able to finish draining its queue.

akarnokd:
Yeah, this came up on Stackoverflow a couple of times. I can't find it but I answered such questions.

davidmoten:
This is an interesting one @hzsweers. I have a use case somewhat similar in that I have an archived timestamped stream and I'd like to replay it sped up by a factor. 

If you do want to make a _Normalize_  operator like this check out @akarnokd's blog Advanced RxJava and check out operators like `OperatorObserveOn` and `OperatorOnBackpressureBuffer`.  I'd be happy to have a go at this operator but also happy to assist your PR. @akarnokd has always been very helpful when I've attempted this sort of thing and I'm sure he'd chip in (actually he'd knock it up in 5 mins but there'd be no fun for the rest of us).

Actually @hzsweers couldn't the `Normalize` use case be sorted by zipping with `Observable.interval`?

Issue #3572, opened 2015-12-13
2.0: Extending Observable/Single


artem-zinnatullin:
During my work on #3564 I've realized that `Observable` and `Single` are not very friendly for extending, I mean, you can extend them but as soon as you apply any operator from `Observable` or `Single` you won't be able to work with child class anymore.

Example:

```java
TestObservable
  .from(Observable.from(asList(1, 2, 3, 4, 5))
  .map(number -> toJson(number)) // Returns Observable instead of TestObservable.
  .expectError(IllegalArgumentException.class) // We can not apply this operator because it's not TestObservable anymore.
  .subscribe()
```

Maybe we can change it in RxJava 2? It's tricky but doable (and RxJava already has some kind of such thing — `extend` operator). I know several projects with custom `Observable`s but not sure if it worth in general for the library.

What do you think about this? 

If you want an example of a library that does something similar — AssertJ does it with its assertions https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractAssert.html

I guess, for RxJava it would require kind of `AbstractObservable` where all common operators will live and then libraries/users will be able to extend it to add additional functionality with keeping child type after applying operators.

roideuniverse:
+1

akarnokd:
[I don't think](http://akarnokd.blogspot.hu/2016/03/writing-custom-reactive-base-type.html) this is possible with Java's typesystem. The main reason for this is the need to return `<U extends AbstractObservable>` from each method but U requires target typing to work. I.e., if chained, the compiler can only infer `AbstractObservable` and not your subtype.

artem-zinnatullin:
Ok, let's close it for now. We're currently solving it Kotlin's extention functions.

PR #3782, opened 2016-03-19
2.x design decisions

stevegury:
This is a first draft of the design document for the 2.x branch.
We are interested in collecting any feedback from the community (implementers but also pure users).

This present document is the collaborative work of @abersnaze, @benjchristensen, @stealthcode, and @stevegury. But we encourage anyone to propose improvements/clarifications via pull-request.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,8 +1,20 @@
 ## RxJava v2 Design
 
-Terminology, principles, contracts, and other aspects of the design of RxJava v2.
+This document explains the terminology, principles, contracts, and other aspects of the design of RxJava v2.
+Its intended audience is the implementers of the library.
 
 ### Terminology & Definitions
+
+##### Interactive
+
+Producer obeys consumer-driven flow control.
+Consumer manages capacity by requesting data.
+
+
+##### Reactive
+
+Producer is in charge. Consumer has to do whatever it needs to keep up.
+
 
 ##### Hot
 
@@ -12,41 +24,94 @@ For example, an `Observable` of mouse events. Subscribing to that `Observable` d
 
 (Note: Yes, there are *some* side-effects of adding a listener, but they are inconsequential as far as the 'hot' usage is concerned).
 
+
 ##### Cold
 
 When used to refer to a data source (such as an `Observable`), it means it has side-effects when subscribed to.
 
 For example, an `Observable` of data from a remote API (such as an RPC call). Each time that `Observable` is subscribed to causes a new network call to occur.
 
-##### Reactive
+
+##### Reactive/Push
 
 Producer is in charge. Consumer has to do whatever it needs to keep up.
 
-##### Interactive
+Examples:
+
+- `Observable` (RxJS, Rx.Net, RxJava v1.x without backpressure, RxJava v2)
+- Callbacks (the producer calls the function at its convenience)
+- IRQ, mouse events, IO interrupts
+- 2.x `Flowable` (with `request(n)` credit always granted faster or in larger quantity than producer)
+- Reactive Streams `Publisher` (with `request(n)` credit always granted faster or in larger quantity than producer)
+- Java 9 `Flow.Publisher` (with `request(n)` credit always granted faster than or in larger quantity producer)
+
+
+##### Synchronous Interactive/Pull
 
 Consumer is in charge. Producer has to do whatever it needs to keep up.
 
-##### Push
+Examples:
 
-Producer emits when it wishes to. Related to "reactive". Callbacks are an instance of push.
+- `Iterable`
+- 2.x/1.x `Observable` (without concurrency, producer and consumer on the same thread)
+- 2.x `Flowable` (without concurrency, producer and consumer on the same thread)
+- Reactive Streams `Publisher` (without concurrency, producer and consumer on the same thread)
+- Java 9 `Flow.Publisher` (without concurrency, producer and consumer on the same thread)
 
-##### Pull
 
-Consumer requests data when it wishes to. Related to "interactive". An `Iterable` is an instance of pull.
+##### Async Pull (Async Interactive)
 
-##### Async Pull
+Consumer requests data when it wishes, and the data is then pushed when the producer wishes to. 
 
-Consumer requests data when it wishes, and the data is then pushed when the producer wishes to. The Reactive Streams `Publisher` is an instance of "async pull", as is the 'AsyncEnumerable' in .Net.
+Examples:
+
+- `Future` & `Promise`
+- `Single` (lazy `Future`)
+- 2.x `Flowable`
+- Reactive Streams `Publisher`
+- Java 9 `Flow.Publisher`
+- 1.x `Observable` (with backpressure)
+- `AsyncEnumerable`/`AsyncIterable`
+
+There is an overhead (performance and mental) for achieving this, which is why we also have the 2.x `Observable` without backpressure.
+
+
+##### Flow Control
+
+Flow control is any mitigation strategies that a consumer applies to reduce the flow of data.
+
+Examples:
+
+- Controlling the production of data, such as with `Iterator.next` or `Subscription.request(n)`
+- Preventing the delivery of data, such as buffer, drop, sample/throttle, and debounce.
+
+
+##### Eager
+
+Containing object immediately start work when it is created.
+
+Examples: 
+
+- A `Future` once created has work being performed and represents the eventual value of that work. It can not be deferred once created.
+
+
+##### Lazy
+
+Containing object does nothing until it is subscribed to or otherwise started.
+
+Examples:
+
+- `Observable.create` does not start any work until `Observable.subscribe` starts the work.
+
 
 ### RxJava & Related Types
 
 ##### Observable
 
-Stream that supports async and synchronous push. It does not support interactive flow control (`request(n)`).
+Stream that supports async and synchronous push. It does *not* support interactive flow control (`request(n)`).
 
 Usable for:
 
-- hot and cold sources
 - sync or async
 - push
 - 0, 1, many or infinite items
@@ -55,6 +120,26 @@ Flow control support:
 
 - buffering, sampling, throttling, windowing, dropping, etc
 - temporal and count-based strategies
+
+*Type Signature*
+
+```java
+class Observable<T> {
+  void subscribe(Observer<T> observer);
+
+  interface Observer<T> {
+    void onNext(T t);
+    void onError(Throwable t);
+    void onComplete();
+    void onSubscribe(Disposable d);
+  }
+}
+```
+
+The rule for using this type signature is:
+
+> onSubscribe onNext* (onError | onComplete)?
+
 
 ##### Flowable
 
@@ -62,10 +147,9 @@ Stream that supports async and synchronous push and pull. It supports interactiv
 
 Usable for:
 
-- hot and cold sources
+- pull sources
+- push Observables with backpressure strategy (ie. `Observable.toFlowable(onBackpressureStrategy)`)
 - sync or async
-- push
-- pull
 - 0, 1, many or infinite items
 
 Flow control support:
@@ -73,18 +157,114 @@ Flow control support:
 - buffering, sampling, throttling, windowing, dropping, etc
 - temporal and count-based strategies
 - `request(n)` consumer demand signal
-  - for pull-based sources, this allows batched "async pull"
-  - for push-based sources, this allows backpressure signals to conditionally apply strategies (i.e. drop, buffer, sample, fail, etc)
+	- for pull-based sources, this allows batched "async pull"
+	- for push-based sources, this allows backpressure signals to conditionally apply strategies (i.e. drop, first, buffer, sample, fail, etc)
+
+You get a flowable from:
+
+- Converting a Observable with a backpressure strategy
+- Create from sync/async OnSubscribe API (which participate in backpressure semantics)
+
+*Type Signature*
+
+```java
+class Flowable<T> implements Flow.Publisher<T>, io.reactivestreams.Publisher<T> {
+  void subscribe(Subscriber<T> subscriber);
+
+  interface Subscriber<T> implements Flow.Subscriber<T>, io.reactivestreams.Subscriber<T> {
+    void onNext(T t);
+    void onError(Throwable t);
+    void onComplete();
+    void onSubscribe(Flowable.Subscription d);
+  }
+
+  interface Subscription implements Flow.Subscription, io.reactivestreams.Subscription {
+    void cancel();
+    void request(long n);
+  }
+}
+```
+
+*NOTE: To support `Flow.Publisher` in Java 9+ without breaking Java 7+ compatibility, we want to use the [multi-release jar file support](http://openjdk.java.net/jeps/238).*
+
+The rule for using this type signature is:
+
+> onSubscribe onNext* (onError | onComplete)?
+
+
+##### Single
+
+Lazy representation of a single response (lazy equivalent of `Future`/`Promise`).
+
+Usable for:
+
+- pull sources
+- push sources being windowed or flow controlled (such as `window(1)` or `take(1)`)
+- sync or async
+- 1 item
+
+Flow control:
+
+- Not applicable (don't subscribe if the single response is not wanted)
+
+*Type Signature*
+
+```java
+class Single<T> {
+  void subscribe(Single.Subscriber<T> subscriber);
+
+  interface Subscriber<T> {
+    void onSuccess(T t);
+    void onError(Throwable t);
+    void onSubscribe(Disposable d);
+  }
+}
+```
+
+> onSubscribe (onError | onSuccess)?
+
+
+##### Completable
+
+Lazy representation of a unit of work that can complete or fail
+
+- Semantic equivalent of `Observable.empty().doOnSubscribe()`. 
+- Alternative for scenarios often represented with types such as `Single<Void>` or `Observable<Void>`.
+
+Usable for:
+
+- sync or async
+- 0 items
+
+*Type Signature*
+
+```java
+class Completable {
+  void subscribe(Completable.Subscriber subscriber);
+
+  interface Subscriber {
+    void onComplete();
+    void onError(Throwable t);
+    void onSubscribe(Disposable d);
+  }
+}
+```
+
+> onSubscribe (onError | onComplete)?
 
 ##### Observer
 
-Consumer of events without flow control.
+Reactive consumer of events (without consumer-driven flow control). Involved in subscription lifecycle to allow unsubscription.
 
 ##### Publisher
 
-[Reactive Streams producer](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#1-publisher-code) of data
+Interactive producer of events (with flow control). Implemented by `Flowable`.
+
+[Reactive Streams producer](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#1-publisher-code) of data.
 
 ##### Subscriber
+
+Interactive consumer of events (with consumer-driven flow control). Involved in subscription lifecycle to allow unsubscription.
 
 [Reactive Streams consumer](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#2-subscriber-code) of data.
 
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+- Converting a Observable with a backpressure strategy
+- Create from sync/async OnSubscribe API (which participate in backpressure semantics)
+
+*Type Signature*
+
+```java
+class Flowable<T> implements Flow.Publisher<T>, io.reactivestreams.Publisher<T> {
+  void subscribe(Subscriber<T> subscriber);
+
+  interface Subscriber<T> implements Flow.Subscriber<T>, io.reactivestreams.Subscriber<T> {
```
akarnokd: Extending both will prevent direct use of any instance of them directly.

Besides, the signature doesn't compile because it requires

```java
void subscribe(io.reactivestreams.Subscriber)
void subscribe(java.util.concurrent.Flow.Subscriber)
```

to be implemented separately.

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+
+- Not applicable (don't subscribe if the single response is not wanted)
+
+*Type Signature*
+
+```java
+class Single<T> {
+  void subscribe(Single.Subscriber<T> subscriber);
+
+  interface Subscriber<T> {
```
akarnokd: Consider naming this `SingleSubscriber` to avoid name conflict with RS `Subscriber`.

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+- sync or async
+- 0 items
+
+*Type Signature*
+
+```java
+class Completable {
+  void subscribe(Completable.Subscriber subscriber);
+
+  interface Subscriber {
```
akarnokd: Consider naming this `CompletableSubscriber` to avoid name conflicts.

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+
+In this example, `subscriber1` and `subscriber2` can consume at different rates, `request(n)` will propagate to the provided  `onBackpressureStrategy`, not the original `Subject` which can't propagate `request(n)` upstream.
 
-- buffering, sampling, throttling, windowing, dropping, etc
-- temporal and count-based strategies
-- It does not support pull-based consumer-driven flow control.
 
 ##### Disposable
 
 A type representing work that can be cancelled or disposed.
```
akarnokd: or resource

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+### Creation
+
+Creation of the various types should be exposed through factory methods that provide safe construction.
+
+```java
+Flowable.create(SyncGenerator generator)
+
+Flowable.create(AsyncGenerator generator)
+
+Observable<T>.create(OnSubscribe<Observer<T>> onSubscribe)
```
akarnokd: So `OnSubscribe` is just a fancy `java.util.function.Consumer`? I found it more convenient to include the subscriber's type thus having FlowableOnSubscribe, ObservableOnSubscribe, SingleOnSubscribe and CompletableOnSubscribe.

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+
+- It must propagate/negotiate the `request(n)` event.
+
+
+### Creation
+
+Creation of the various types should be exposed through factory methods that provide safe construction.
+
+```java
+Flowable.create(SyncGenerator generator)
```
akarnokd: I'm assuming operators will extend Flowable as I suggested thus there is no need for create(FlowableOnSubscribe) or create(Publisher), right?

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+
+A producer can terminate a stream by emitting `onComplete` or `onError`. A consumer can terminate a stream by calling `cancel`/`dispose`.
+
+Any resource cleanup of the source or operators must account for any of these three termination events. In other words, if an operator needs cleanup, then it should register the cleanup callback with `cancel`/`dispose`, `onError` and `onComplete`.
+
+The final `subscribe` will *not* invoke `cancel`/`dispose` after receiving an `onComplete` or `onError`.
+
+### JVM target and source compatibility
+
+The 2.x version will target JDK6+ but the source will be using some of the Java 8 features.
```
akarnokd: There is no specific need for lambdas in operators and most IDEs lets you write lambda and convert it to anonymous class creation with little effort. Lambda in tests is a large convenience, but all existing tests have been written with inner classes thus not useful for them; maybe for newer tests.

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
- - push-based creation can push without consideration of a backpressure strategy
- - pull-based creation should be discouraged
+The intent is to let Android users consume the new version of RxJava.
+(Not sure if JDK6 is the appropriate requirement, JDK7+ is maybe sufficient).
 
-##### Destruction
+##### Custom Observable, Single, Completable, or Flowable (To be confirmed)
 
-A producer can terminate a stream by emitting `onComplete` or `onError`. A consumer can terminate a stream by calling `cancel`.
+An implementation of an Observable which can be subscribed to with an `Observer`. Custom (`Observable`|`Single`|`Completable`|`Flowable`) classes would implement the interface `Consumable<S>` where the generic type `S` is a `Flowable.Subscriber<T>`, `Observable.Observer<T>`, `Completable.Subscriber<T>`, or a `Single.Subscriber<T>` depending on which semantics the custom class will follow. This choice would also affect the kinds of observables the custom class could interop with. For instance, `Flowable#merge` could operate over the standard `Flowable` or any custom class which implements `Consumable<Flowable.Subscriber<T>>`.
```
akarnokd: Flowable should implement Publisher which is our baseline for input, thus allows flatMapping somebody else's Publishers:

```java
flatMap(Function<? super T, ? extends Publisher<? extends R>>)

range(1, 10).flatMap(v -> Flux.just(1));
```

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+##### Custom Observable, Single, Completable, or Flowable (To be confirmed)
 
-A producer can terminate a stream by emitting `onComplete` or `onError`. A consumer can terminate a stream by calling `cancel`.
+An implementation of an Observable which can be subscribed to with an `Observer`. Custom (`Observable`|`Single`|`Completable`|`Flowable`) classes would implement the interface `Consumable<S>` where the generic type `S` is a `Flowable.Subscriber<T>`, `Observable.Observer<T>`, `Completable.Subscriber<T>`, or a `Single.Subscriber<T>` depending on which semantics the custom class will follow. This choice would also affect the kinds of observables the custom class could interop with. For instance, `Flowable#merge` could operate over the standard `Flowable` or any custom class which implements `Consumable<Flowable.Subscriber<T>>`. 
 
-Any resource cleanup of the source or operators must account for any of these three termination events. In other words, if an operator needs cleanup, then it should register the cleanup callback with `cancel`, `onError` and `onComplete`. 
+For more information see the proof of concept project [Consumable](https://github.com/stealthcode/Consumable).
```
akarnokd: That `Consumable` looks odd.The main aim of XOnSubscribe is to capture the subscriber's type so one doesn't have to type it out. Making it rooted in a common base type doesn't really add anything because the subscriber's type is necessary to work with it properly. 

My guess is that the class is aimed at writing generic operators that work with all sorts of Subscribers, but the fact is you have to do instanceof checks and dispatch in runtime instead of the compiler doing it for you at compile time. Besides, given an operator Z, you have to implement ZFlowableSubscriber, ZObservableSubscriber, ZSingleSubscriber and ZCompletableSubscriber inner classes...  If it is not yet convincing, consider Single vs Flowable. How would you combine the two if Single has to call onSuccess and Flowable has call onNext?

akarnokd:
### General observations

#### Class structure
I didn't see any mention of the architecture of the base objects and how one adds operators to it. As I suggested in my [design retrospect](http://akarnokd.blogspot.hu/2016/03/rxjava-design-retrospect.html) post, we can save a lot of allocations by extending the base class instead of using create/lift internally all the time.

#### Error handling
In addition, I don't see any mention of requiring operators to call `RxJavaPlugins.onError` with exceptions that can't be delivered because of the RS contract. For example, if 2 sources on a regular flatMap signal an error, only the first can be delivered reliably and the other has nowhere to go.

Consequently, since RS subscribers shoudn't really throw in their `onXXX` methods but end-subscribers tend to (as evidenced by many issue list reports), we'll need a `safeSubscribe()` method as well. As a result, we can no longer throw `OnErrorNotImplementedException` with `subscribe()` and `subscribe(Consumer)` because a) we don't know where it will bubble up and b) if it reaches the  scheduled action root, it still will end up in `RxJavaPlugins.onError`.

#### Default Subscriber implementations

Since cancelling and disposing is now potentially racy, we can't really expect end-subscribers to implement logic for them in their `onSubscribe()` calls. We need standard subscriber classes that do any internal exposure (access to `Subscription.request` conveniently) or external exposure (a cancellable `AsyncObserver` and `AsyncSubscriber`)

### Operator fusion

Macro-fusion relies on the property that operator classes can be identified in some way. For example, `just()` optimizations work in 1.x because it was implemented as a class extending `Observable` instead of hidden behind `create()`. However, as long as the Observable returned by `create()` can be unwrapped to look at the real `OnSubscribe` class, the fusion can still work, albeit with more (unnecessary) overhead. Using lambdas or anonymous inner classes defeats it.

Micro-fusion requires a library-internal protocol to work amongst operators. One day, it would be great to have standard fusion protocol library that all RS libraries can implement and do cross-library fusion with it.

### Changes to 2.x master

I see the following changes required in current 2.x master:

1. rename `Observable` to `Flowable`, including operators prefixed
2. rename `NbpObservable` to `Observable`, including operators prefixed
3. change `Observer` to interface, fix the depending sites
4. make `PublishSubject` and `BehaviorSubject` extend `Observable` instead of `Flowable`
5. remove operators from `Flowable` that would signal onError if the downstream can't keep up
6. rewrite lifted operators to class-extension form
7. add missing operators and overloads from 1.x,
8. add missing unit tests from 1.x
9. remove unnecessary operators or overloads
10. add JavaDoc
11. define interfaces and protocol for fusion, start applying operator-fusion to `Flowable`
12. define interfaces and protocol for non-backpressured fusion, start applying operator-fusion to `Observable`

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+
+The 2.x version will target JDK6+ but the source will be using some of the Java 8 features.
+We'll use the retrolambda project to generate JDK6 Byte-code from the Java 8 sources.
+The only Java 8 features we intend to use are: lambda expressions, method references and try-with-resources statements.
+Default methods may be considered if retrolambda support reach an acceptable level of maturity.
+
+It still up to discussion to know how we will generate multiple targets?
+Options are:
+
+- two artifcat ids: rxjava-jdk6, rxjava-jdk8
```
JakeWharton: What advantage is there for producing two artifacts? If you're not using JDK 8 APIs why not just make the JDK 6 artifact the canonical one.

Having two makes me _very_ nervous about binary compatibility with the two. Java projects built against the JDK 6 impl which aim to support Android should be binary compatible with JVM users on Java 8+. Either a mechanism should be put in place to verify binary compatibility (e.g., animal sniffer) or it might just be easier to have only a single output artifact.

DESIGN.md
```diff
@@ -73,77 +157,225 @@ Flow control support:
...
+- classifiers: jdk6, jdk8
 
-- If `request(n)` behavior is supported in the stream implementation, then:
- - pull-based creation must support `request(n)` semantics
- - push-based creation must provide a default *onBackpressure* strategy
-- If `request(n)` behavior is not supported in the stream implementation, then:
- - push-based creation can push without consideration of a backpressure strategy
- - pull-based creation should be discouraged
+The intent is to let Android users consume the new version of RxJava.
+(Not sure if JDK6 is the appropriate requirement, JDK7+ is maybe sufficient).
```
JakeWharton: The Android tools will translate some JDK 7 language features from version 51 bytecode. If the mechanism is going to be Retrolambda there's really no point in targeting JDK 7, might as well go all the way to JDK 6 (plus you want to ensure try-with-resources gets converted).

DESIGN.md
stealthcode: > My guess is that the class is aimed at writing generic operators that work with all sorts of Subscribers

No this is not the goal of a core consumable interface. The different types of subscribers each have very specific semantics and this is in no way attempting to bridge them to write operators that function over all subscriber semantics. The purpose of this change is to enable different types of `Flowable`s that can all use the standard flowable operators. Please link to the Consumable proof of concept for more details on why and how.

DESIGN.md
iNoles: Also, "The Android N bases its implementation of lambda expressions on anonymous classes. This approach allows them to be backwards compatible and executable on earlier versions of Android." I think Retrolambda isn't compatible with Android's Jack tool.

DESIGN.md
akarnokd: I've read the linked material twice. There seems to be an obsession with own reactive types lately, mostly because Java doesn't have extension methods that would hide the inconvenient composte/lift/to method usage. If you think about it, RxJava 1 already does this with OnSubscribe. OnSubscribe can only do one thing via its single method and we had to wrap it into an `Observable` to add meaningful operations. Your proposal adds yet another layer on top of it. I see two issues: 1) more wrapping and possibly more intermediate layers; 2) likely prevents fusion. I similar example I had to deal with is in IxJava where the interface `Iterable` had to be wrapped and unwraped in the class `Ix` to chain up operators but not the wrapper itself. The point is, that I don't think RxJava should do anything extra beyond what it already does. You can introduce your own class, wrap an Observable and expose any operators you like, package it into a library and share it with your colleagues.

DESIGN.md
JakeWharton: No one will be compiling RxJava sources with Jack so I'm not sure how
that's relevant.

DESIGN.md
stealthcode: > There seems to be an obsession with own reactive types lately

This tells me there is a need for such an interop layer. 

> RxJava 1 already does this with OnSubscribe

An OnSubscribe does not ensure that the type supports a consistent way to subscribe. The only way that a function can accept both an Observable and a custom Observable is if the custom observable extends Observable. 

> 1) more wrapping and possibly more intermediate layers

In some case where the custom operator has to implement a special `OnSubscribe` I could see this as being true. Otherwise (what I see as a majority of cases) if both types support the same OnSubscribe function then there is no extraneous object allocation. The [`MyFlowable` class](https://github.com/stealthcode/Consumable/blob/master/src/main/java/MyFlowable.java#L48) is written with a special OnSubscribe. However it it just used the same type as Flowable (which should have a `Consumer<Subscriber<T>>`) then there is only a lift into `MyFlowable` when chaining operators. 

> 2) likely prevents fusion

I don't think this is true. Can you explain how fusion would work? And how this would prevent it? My assumption is that chaining operators would create a state of operators that you can later traverse at subscribe time so that you can "fuse" certain combinations into equivalent fused operators. If your subscriber type in a consumable was a `FusionSubscriber` then this proposal would actually support fusion and enable fusion across different types of fusable observables.

akarnokd: If you look at the current 2.x master, you'll see that most operators accept `Publisher` instead of a subclass. The reason for this is that operators require only one method: subscribe() from their upstream. In addition, since the create() also accept a Publisher, your dream for `extend(Func<Publisher, X>)` comes true because the wrapped Publisher can be directly applied.

[Operator fusion](http://akarnokd.blogspot.hu/2016/03/operator-fusion-part-1.html) can happen in assembly time or subscription time. So far, pairs of subsequent operators are considered. For example, the operator of observeOn will look at its source and if it's a `just()`, the resulting Observable will be a custom operator. Given a range() as the inner observable of flatMap, the inner subscriber of `flatMap` sees that range is sync-fuseable and switches to a queue-based protocol with it instead of creating a queue itself. This still can work if one flatmaps a `range().map().filter()`. However, the moment there is some non-fuseable operator, such as your MyFlowable.map(), two things can happen: 1) everything stays plain and no fusion happens, 2) fusion happens bypassing MyFlowable.map()'s runtime effects completely. To avoid 2) operators should never just forward their Subscriptions.

We worked out a bunch of cases with operator-fusion over in [Reactive Streams Commons](https://github.com/reactor/reactive-streams-commons), which library was designed to be a pick what you like set of common operators.

Certain micro-fusion, such as the queue-fusion and conditional subscriber travel upwards during subscription time . However, conditional subscriber can be ignored by design but the fact that map() can participate in fusion doesn't mean it will. One needs a fuseable source and receiver that asks for a fusion mode.

The bottom line is that we already have the `Publisher` from Reactive-Streams as the common baseline and implement operators that take it.

> This tells me there is a need for such an interop layer.

Could you explain the original requirement that spawned `Consumable`?

stealthcode: > your dream for `extend(Func<Publisher, X>)` comes true

Publisher only accepts a `Subscriber`. 

> original requirement that spawned `Consumable`

 * #3572 Extending Observable/Single
 * The concept of a `TemporalObservable` which held all scheduling based operators.
 * [The RxJavaStrig project](https://github.com/ReactiveX/RxJavaString)
 * [The RxJava Parallel project](https://github.com/ReactiveX/RxJavaParallel)
 * `GroupedObservable` - At one time you had to subscribe to all inner observables. It would have bene useful to constrain the API to operators that would retain all inner observables. 
 * `retryWhen` - It would be useful to have a custom observable that had a subset of operators by which you could trigger a retry. This would make the API more usable. 
 * [RxNetty](https://github.com/ReactiveX/RxNetty) - has many internal classes which extend observable and require stateful operations
 * the kitchen sink proliferation of instance methods on `Observable`, `Completable`, and `Single`. This untenable solution is leading to an unwieldy API with lots of pitfalls and a very high learning curve.

akarnokd: I can see that Parallel shouldn't allow all sorts of operators, but the rest is just restrictive and we'd end up with why can't I retry with operator Z or why I only have the option to flatMap a groupBy. Regardless, all these API restrictions and extensions can be achieved by wrapping an underlying reactive streams implementation and as long as everybody implements and expects Publisher, they should work together.

stealthcode: It is worth a broader conversation to determine what cases we have for
extension and api conversion. Also as I said `Publisher` does not support
converting an `Observable` as it is subscribed to by an `Observer` and not
a `Subscriber` and likewise completable and single. Flowable is the only
one covered by Reactive Streams. Additionally there is a need for custom
types.

stealthcode: [in this pr](https://github.com/stealthcode/Consumable/pull/1) to consumable project we showed that `Flowable` can implement both `Publisher<T>` and `Consumable<Subscriber<T>>` so the two can be used without conflict.

akarnokd: Certainly, but I don't see how `Consumable` helps. For example,
```java
public <R> Flowable<R> flatMap(Function<? super T, ? extends Consumable<Subscriber<? super R>>> f)
```

Why would flatMap need the `extend()` method here? You can nicely flatMap with a Publisher target where the inner sources are Reactor-Core `Flux` or an Akka-Streams Publisher without any wrapping on RxJava's side; only a `subscribe(Subscriber<? super R>>)` method is required which is already in `Publisher`.

Or do you want to build a type hierarchy so, for example, Retrofit would return `Publisher+Extendable` instead of `Flowable` thus allowing fluent conversion to Flux with `extend(Flux::wrap)`?

stealthcode: > Why would flatMap need the extend() method here? 

The benefit is interop and it has very little to do with `extend` (which just gives a path for monadic chaining and fluent conversion). 

An example of what I am looking for can already be found in Flowable in the 2.x branch. The `Flowable` `flatMap` takes a `Func1<T, ? extends Publisher<R>>` so our Flowable allows merging any Publisher (since Publisher necessarily uses `Subscriber`). 

Today `NbpObservable` declares a method `concatWith(NbpObservable<? extends T> other)`. This must accept a thing that extends `NbpObservable`. And there are many other operators on Completable or Single that takes a specific class. These are not generic enough to allow the same flexibility that Flowable has in it's operator api. Accepting Consumable would result in a change to many operators in Observable, Completable, and Single where the operator accepts a (Observable | Completable | Single). This would be changed to a (`Consumable<Observer>` | `Consumable<CompletableObserver>` | `Consumable<SingleObserver>`). 

> Or do you want to build a type hierarchy

A type hierarchy is exactly what I'm trying to avoid. I do want to eventually consider namespacing complicated or rarely used operators into a different class that extends `Consumable<Observer<T>>`. 

For example, namespacing the match operator from PR #2821 into a different class (so that it doesn't pollute the operators on the standard `Observable` / `Flowable`. 

```java
class SelectCaseObservable<T> implements Consumable<Observer<T>> {
    public MatchBuilder<T, R> match(...) {...}
    public static <T> SelectCaseObservable<T> fromObservable(OnSubscribe<T> onSubscribe) {...}
}
```

This class could be seamlessly merged, concatted, etc... so long as the api accepted a `Consumable<Observer<T>>`.

stealthcode: Can you give me some feedback as to if the goals make sense now? The [Pros and Cons](https://github.com/stealthcode/Consumable#cons) are listed at the bottom of the project README and if they are unclear then I would like to know how you suggest that I could revise them (or gladly accept a well-formed PR).

akarnokd: I understand it now. If it stays just a base interface then no objection. One thing though is that this being parametric, javac will put in a class cast that runs every time the `subscribe(O)` is invoked; I'd prefer separate interfaces for all the other major types instead.

stealthcode: So what you are saying is that every implementation of `subscribe(...)` should accept a concrete class and not a parametric class? I think that's what the design document proposes as there is an interface for each of the classes, `Completable.Subscriber`, `Single.Subscriber`, and `Observer`. Am I misunderstanding your comment?

akarnokd: I was taling about `Consumable<Observer<T>>`. I'd prefer concrete base interfaces:

```java
interface ConsumableSingle<T> {
    void subscribe(SingleSubscriber<? super T> subscriber);
}
interface ConsumableCompletable {
    void subscribe(CompletableSubscriber subscriber);
}
interface ConsumableObservable<T> {
    void subscribe(Observer<? super T> observer);
}
```

stealthcode: Also simplifies the generics as well. Seems like a good idea to me.

DESIGN.md
hzsweers: Also, how would you include sources for the jdk6 distribution since retrolambda only operates on the bytecode? It would make the IDE have to decompile the classes to represent them rather than being able to just click in. Would this not make debugging unnecessarily more tedious and also sacrifice javadoc/comment access?

DESIGN.md
stealthcode: Right now we are not targeting `Flow.Subscriber` since this planned for Java 1.9. Our current source code compatibility is 1.6. I believe the rationale was to show that we intend that this would one day extend and interoperate with `Flow`. 

This line should just be changed to ...

```java
interface Subscriber<T> extends io.reactivestreams.Subscriber<T> {
```

DESIGN.md
stealthcode: Would you rather `io.reactivex.SingleSubscriber` or `io.reactivex.Single.SingleSubscriber`? I prefer the non-static inner class `SingleSubscriber`.

DESIGN.md
stealthcode: I prefer `io.reactivex.CompletableSubscriber`

DESIGN.md
stealthcode: I think I read your [blog post](http://akarnokd.blogspot.hu/2016/03/rxjava-design-retrospect.html) where you comment as such. I think it would be best to consider this as one possible optimization that we should consider during implementation and not necessarily part of the necessary design.

DESIGN.md
akarnokd: an `io.reactivex.Subscriber` will just name-conflict with `io.reactivestreams.Subscriber` without any benefit.

DESIGN.md
stealthcode: Yes in this model it seems like `OnSubscribe` is an `Action0` (since `Consumer` is a Java1.8 only functional interface). Now that you've had a chance to look at the `Consumable` proposal a little closer do you still share this opinion? I don't think that I have a strong opinion here.

DESIGN.md
stealthcode: So your proposal is to just remove this interface?

DESIGN.md
akarnokd: If `ConsumableX` remain functional interfaces, then this line should be `Observable<T>.create(ConsumableObservable<T> body)`

DESIGN.md
akarnokd: Yes.

DESIGN.md
stealthcode: I am not sure that the `ConsumableObservable` is the right type here. A class that extends `Consumable` (according to my proposal) or equivalently a class that extends `ConsumableObservable` (per #3797) is not an `OnSubscribe` but is an `Observable`. I think it's still an `Action0<Observer<? super T>`.

DESIGN.md
stealthcode: As we have discussed, Flowable can and should implement `Publisher` and it should also implement `Consumable`. Is there still a concern here? I suggest that the `Consumable` proposal be revised with a higher degree of elaboration (potentially in a separate PR) and merged to the design.md.

DESIGN.md
stealthcode: I see that @akarnokd and I have reached a general acceptance level here with `Consumable` however in order to reach a broader audience I'd like to remove this section from this PR so further elaboration and discussion can be encouraged outside of this PR. Concerns?

DESIGN.md
akarnokd: Okay. To avoid any confusion, let's have `create()` of each base type take `FlowableOnSubscribe`, `ObservableOnSubscribe`, `SingleOnSubscribe` and `CompletableOnSubscribe` respectively; all functional interfaces and no generics other than the value type (i.e., no `Action1<Observer<? super T>>`.

DESIGN.md
akarnokd: I wouldn't mention that "the source will be using some of the Java 8 features".

stealthcode:
@akarnokd 

> we can save a lot of allocations by extending the base class instead of using create/lift internally all the time.

> I don't see any mention of requiring operators to call RxJavaPlugins.onError with exceptions that can't be delivered because of the RS contract. 

> disposing is now potentially racy

> Macro-fusion relies on the property that operator classes can be identified in some way.

These sound like additive changes to the existing design revision. Would it be okay if you make a revision to the DESIGN.md document and open a PR with proposals to address these items? 

> I see the following changes required in current 2.x master:

Can you hold off on opening PRs until the DESIGN.md changes have been merged into 2.x? This should serve as a sign that the decisions have been reviewed and commitment has been reached by the RxJava committers. You are of course welcome to play around with code for your own understanding and demonstration purposes.

DESIGN.md
akarnokd: `merge` should take `Publisher` which also implies that for `ConsumableFlowable` to work, it has to extend `Publisher` as well. I'd rephrase this so that each base type's relevant operator will take the most generic type `Publisher` for `Flowable`, `ConsumableSingle` for `Single`, `ConsumableObservable` for `Observable` and `ConsumableCompletable` for `Completable`.

DESIGN.md
akarnokd: Let's remove this part for now.

DESIGN.md
stealthcode: I think this is a great point. We of course do not want to cause unnecessary friction in any subset of user base. The bytecode compatibility can easily be tested as per the proposals in #3764 ([see comment](https://github.com/ReactiveX/RxJava/pull/3764#issuecomment-200699475)).

akarnokd:
> Would it be okay if you make a revision to the DESIGN.md document and open a PR with proposals to address these items?

Sure.

Can you hold off on opening PRs until the DESIGN.md changes have been merged into 2.x?

Yep.

DESIGN.md
stealthcode: Why?

DESIGN.md
stealthcode: Is there anything here that you would change? Is your comment alluding to something that is implementation driven or can this be worked out post-design? If not then would you be okay with driving this from a separate PR after we merge this work?

akarnokd: Let's work this out in a separate PR.

DESIGN.md
akarnokd: For the clear intention, plus they allow proper javadocs.

DESIGN.md
stealthcode: Okay, I can understand the argument of _its not part of the design_ however there should be a discussion about using lambdas in RxJava source code. I have opened an issue to track this #3805 discussion. For now I think that this line in the design document should explicitly state something to the following effect.

```
* RxJava will be published to `io.reactivex.rxjava:2.x` with bytecode compatible with a Java 1.6 runtime. In order to do so, all references to java library classes and interfaces must be likewise compatible (i.e. no references to Java8 runtime libraries). 
* RxJava source code will be implemented with lambda expressions. 
```

The reason I vote for keeping the bit about lambdas in the design is because it expresses the development environment and constraints that we intend to have on the committers/contributors.

DESIGN.md
stealthcode: This indicates to me that there needs to be broader awareness and a larger discussion on the `Consumable`. This does not need to be in this PR.

DESIGN.md
stealthcode: Sounds good to me. If they are individual types then they should be listed in the type declarations in the design for completeness.

DESIGN.md
benjchristensen: Good point if we don't need a concrete implementation (which we don't).

DESIGN.md
benjchristensen: It seems redundant to have `io.reactivex.Single.SingleSubscriber`.

DESIGN.md
benjchristensen: Agreed. Similar to `Single`, if we don't like `Single.Subscriber` and `Completable.Subscriber`, then I suggest not having them as inner classes and being `io.reactivex.CompletableSubscriber` and `io.reactivex.SingleSubscriber`.

DESIGN.md
benjchristensen: I prefer having a single artifact.

benjchristensen:
I suggest merging this document so we can start modifying portions of it via PRs. There appear to be no general disputes, but valid changes to individual items that should be done.

benjchristensen:
It's been a week. Can we merge?

akarnokd:
Fine, :+1:, let's use it as a starting point.

benjchristensen:
Merged. Let's start making changes via small pull requests on the various details that were discussed above.

hzsweers:
It's unfortunately been a month and nothing new has been contributed to the 2.x branch. Netflix folks,  @akornokd, and others: would it be useful create a slack group for reactivex/RxJava to facilitate discussions amongst each other and the community rather than strictly GitHub correspondence? It seems like the latter is adding unnecessary friction to discussions since it doesn't scale well to many people discussing at once. Could help improve organization and planning amongst contributors as well. Would be happy to help organize his, and I know many other developers have wished there was a resource like it having similar ones for other platforms and frameworks.

akarnokd:
My latest refactor PR is still not merged so I can't really start pushing other things. If you want, you can set up a chat somewhere, but note I'm at GMT+2.

Issue #3775, opened 2016-03-17
concat can't display all the groupedBy observers on version 1.1.1


lvzhitingzhou:
Here are the code I‘m using to test：
```java
        final Observable<GroupedObservable<String, AppInfo>> groupBy = Observable.from(appInfoList)
                .groupBy(appInfo -> {
                    SimpleDateFormat sdf = new SimpleDateFormat("MM/yyyy");
                    String groupedStr = sdf.format(new Date(appInfo.getLastUpdateTime()));
                    return groupedStr;
                });
        Observable.concat(groupBy)
                .subscribe(mSubscriber);
```

and the library is :

```
    compile 'io.reactivex:rxjava:1.1.1'
    compile 'io.reactivex:rxandroid:1.1.0'`
```
 I got this result which apparently is not correct,  I only got three results!:
![device-2016-03-17-151005](https://cloud.githubusercontent.com/assets/10875078/13838994/e1f62658-ec52-11e5-93c7-f3c726af5bc8.png)

but it should be like this as shown on the Rxjava-essential-code.
![image](https://cloud.githubusercontent.com/assets/10875078/13839031/283c1870-ec53-11e5-8da5-57add00c3b35.png)

This problem never occurs on Version 1.1.0.

akarnokd:
Strange. This code works for me on 1.1.1:

```java
public class GroupConcat {
    static final class AppInfo {
        String name;
        LocalDate date;
        @Override
        public String toString() {
            return name + " @ " + date;
        }
    }
    public static void main(String[] args) {
        System.setProperty("rx.ring-buffer.size", "16");
        
        List<AppInfo> list = new ArrayList<>();
        for (int i = 0; i < 10; i++) {
            for (int j = 0; j < 3; j++) {
                AppInfo ai = new AppInfo();
                ai.name = i + " - " + j;
                ai.date = LocalDate.of(2016, 3, i + 1);
                list.add(ai);
            }
        }
        
        Observable<GroupedObservable<String, AppInfo>> o = Observable.from(list)
        .groupBy(v -> v.date.format(DateTimeFormatter.ofPattern("MM/yyyy")));
        
        Observable.concat(o)
        .subscribe(System.out::println);
    }
}
```

yanglw:
```java
public class Concat {
    public static class Student {
        public String classId;
        public String name;

        @Override
        public String toString() {
            return "Student{classId='" + classId + "', name='" + name + "'}";
        }
    }

[...]

PR #3787, opened 2016-03-21
1.x: fix GroupBy delaying group completion till all groups were emitted

akarnokd:
In 1.1.1, `groupBy` was fixed to properly honor backpressure on the outer `Observable`. The change included a drain loop that emitted `onCompleted()` to the groups only when all `GroupedObservable`s were drained from the main queue. This delayed the group's completion unnecessarily causing the `concat` operator to hang in some source-consumer cases such as #3775.

This PR fixes the behavior by signalling `onCompleted()` to the groups the moment the main completes. 

Note, however, that concatenating groups is eventually prone to hangs due to the groups not completing until the source completes, thus `concat` can't switch to the next source. One should use `flatMap` or `concatMapEager` instead.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorGroupBy.java
+++ b/src/main/java/rx/internal/operators/OperatorGroupBy.java
@@ -219,6 +219,12 @@ public final class OperatorGroupBy<T, K, V> implements Operator<GroupedObservabl
             if (done) {
                 return;
             }
+
+            for (GroupedUnicast<K, V> e : groups.values()) {
+                e.onComplete();
+            }
+            groups.clear();
+
             done = true;
             GROUP_COUNT.decrementAndGet(this);
             drain();
@@ -328,13 +334,6 @@ public final class OperatorGroupBy<T, K, V> implements Operator<GroupedObservabl
                     return true;
                 } else
                 if (empty) {
-                    List<GroupedUnicast<K, V>> list = new ArrayList<GroupedUnicast<K, V>>(groups.values());
-                    groups.clear();
-                    
-                    for (GroupedUnicast<K, V> e : list) {
-                        e.onComplete();
-                    }
-                    
                     actual.onCompleted();
                     return true;
                 }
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorGroupBy.java
```diff
@@ -219,6 +219,13 @@ public void onCompleted() {
             if (done) {
                 return;
             }
+            List<GroupedUnicast<K, V>> list = new ArrayList<GroupedUnicast<K, V>>(groups.values());
+            groups.clear();
+            
+            for (GroupedUnicast<K, V> e : list) {
```
artem-zinnatullin: Why not just iterate over `groups.values()` here and avoid `list` allocation? 

Or you want array structure for better performance? If so, not sure you're winning something big because you're iterating over `values()` in the constructor of `ArrayList` and then here.

src/test/java/rx/GroupByTests.java
```diff
@@ -90,4 +91,27 @@ public void call(String v) {
 
         System.out.println("**** finished");
     }
+    
+    @Test
+    public void groupsCompleteAsSoonAsMainCompletes() {
```
artem-zinnatullin: Passed on 1.x without fix from this PR…

src/main/java/rx/internal/operators/OperatorGroupBy.java
akarnokd: The idea was to reduce the window when terminating groups remove themselves from the map; but I don't think such case is likely anymore. I'll update the PR accordingly.

src/test/java/rx/GroupByTests.java
akarnokd: It fails on 1.1.1 and 1.1.2 but works on 1.1.0 and earlier for me.

artem-zinnatullin: Sorry, I pulled `master` branch instead of `1.x` (uh), confirm: it fails on 1.1.2.

artem-zinnatullin:
:+1:

stevegury:
:+1:

PR #3795, opened 2016-03-23
1.x: observeOn now replenishes with constant rate

akarnokd:
This PR makes sure `observeOn` requests replenishments in a fixed and predictable quantity of 75% of the `bufferSize`, that is, if an emission counter reaches `0.75 * bufferSize`, that amount is requested and the emission counter is reset to zero. This requires saving the emission count between drain runs. If the bufferSize is 1 or 2, the replenishment will trigger after every 1 or 2 items.

Note that there is only one sensitive operator-builder, `AsyncOnSubscribe`, which is mostly affected by the request pattern as it facilitates user code to respond with an Observable sequence of the requested amount.

In addition, since `observeOn` now supports setting the buffer size, it can act as a rebatching operator via the help of `Schedulers.immediate()`.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -83,7 +83,8 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
         final NotificationLite<T> on;
         final boolean delayError;
         final Queue<Object> queue;
-        final int bufferSize;
+        /** The emission threshold that should trigger a replenishing request. */
+        final int limit;
         
         // the status of the current stream
         volatile boolean finished;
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
:+1:

src/main/java/rx/internal/operators/OperatorObserveOn.java
```diff
@@ -105,12 +109,15 @@ public ObserveOnSubscriber(Scheduler scheduler, Subscriber<? super T> child, boo
             this.recursiveScheduler = scheduler.createWorker();
             this.delayError = delayError;
             this.on = NotificationLite.instance();
-            this.bufferSize = (bufferSize > 0) ? bufferSize : RxRingBuffer.SIZE;
+            int calculatedSize = (bufferSize > 0) ? bufferSize : RxRingBuffer.SIZE;
+            this.limit = calculatedSize - (calculatedSize >> 2);
```
stevegury: What do you think about a less efficient (but more readable): `(int) (calculatedSize * 0.75)` ?

stevegury:
Other than the previous comment, 👍

akarnokd:
That formula doesn't work if calculatedSize == 1. I've updated the PR with a comment on the calculation.

stevegury:
👍

PR #3797, opened 2016-03-24
2.x: add ConsumableX to the base types, update method signatures

akarnokd:
This PR factors out the `XSubscriber` types, makes sure the base types implement `ConsumableX`
and methods accepting X in some way now accept `ConsumableX` type.

Wrapping into the base types are not yet eliminated.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -37,10 +37,11 @@ import io.reactivex.schedulers.Schedulers;
  * 
  * The class follows a similar event pattern as Reactive-Streams: onSubscribe (onError|onComplete)?
  */
-public class Completable {
+public class Completable implements ConsumableCompletable {
     /**
      * Callback used for building deferred computations that takes a CompletableSubscriber.
      */
+    // TODO use ConsumableCompletable directly?
     public interface CompletableOnSubscribe extends Consumer<CompletableSubscriber> {
         
     }
@@ -51,29 +52,6 @@ public class Completable {
      */
     public interface CompletableOperator extends Function<CompletableSubscriber, CompletableSubscriber> {
         
-    }
-    
-    /**
-     * Represents the subscription API callbacks when subscribing to a Completable instance.
-     */
-    public interface CompletableSubscriber {
-        /**
-         * Called once the deferred computation completes normally.
-         */
-        void onComplete();
-        
-        /**
-         * Called once if the deferred computation 'throws' an exception.
-         * @param e the exception, not null.
-         */
-        void onError(Throwable e);
-        
-        /**
-         * Called once by the Completable to set a Disposable on this instance which
-         * then can be used to cancel the subscription at any time.
-         * @param d the Disposable instance to call dispose on for cancellation, not null
-         */
-        void onSubscribe(Disposable d);
     }
     
     /**
@@ -102,6 +80,27 @@ public class Completable {
     });
     
     /**
+     * Wraps a general ConsumableCompletable, if not already a Completable, into 
+     * a Completable instance.
+     * 
+     * @param consumable the consumable to wrap or cast to Completable
+     * 
+     * @return the Completable instance wrapped or cast
+     */
+    public static Completable wrap(final ConsumableCompletable consumable) {
+        if (consumable instanceof Completable) {
+            return (Completable)consumable;
+        }
+        Objects.requireNonNull(consumable, "consumable is null");
+        return create(new CompletableOnSubscribe() {
+            @Override
+            public void accept(CompletableSubscriber cs) {
+                consumable.subscribe(cs);
+            }
+        });
+    }
+    
+    /**
      * Returns a Completable which terminates as soon as one of the source Completables
      * terminates (normally or with an error) and cancels all other Completables.
      * @param sources the array of source Completables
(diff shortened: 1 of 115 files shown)


src/main/java/io/reactivex/Completable.java
```diff
@@ -559,7 +558,7 @@ public void onSubscribe(Subscription s) {
      * @throws NullPointerException if flowable is null
      */
     @SchedulerSupport(SchedulerKind.NONE)
-    public static <T> Completable fromNbpObservable(final Observable<T> observable) {
+    public static <T> Completable fromObservable(final Observable<T> observable) {
```
stealthcode: Should these kinds of `fromX(X)` methods change to something like `fromX(ConsmuableX)`? I think its sufficient to adopt the more generic api but still call out to `Observable` in the name and in the javadoc comment.

akarnokd: Yes, I missed a few places. Updated the PR.

akarnokd:
PR updated with the few places where I forgot to replace with the nw interfaces. In addition, I'm no longer convinced that they should cross-implement each other's base interface because it causes overload ambiguity.

stealthcode:
> I'm no longer convinced that they should cross-implement

I think I agree with that statement. Do you mean that `Completable` should not implement the `ConsumableObservable`? I did not intend that in my POC.

akarnokd:
`Flowable` looks like it could do both `ConsumableObservable` and `Publisher` or `Singe` could also implement `Publisher`, but some methods become ambiguous.

stealthcode:
Yeah, since `Flowable`'s subscribe takes a `Subscriber` and both `ConsumableSubscriber` and `Publisher` both agree on the subscriber type then we should make Flowable implement both Consumable and Publisher. The others are difficult.

akarnokd:
Added `ConsumableFlowable`. I had trouble with typing extend, using `Function<? super ConsumableFlowable<T>, R>` is too restrictive and gives no access to `Flowable's methods` and having `U extends ConsumableFlowable<T>>, Function<? super U, R>` has a type-bounce of ? super ? extends ConsumableFlowable.

stealthcode:
I got extend to work with this.

```java
public class Flowable<T> implements Consumable<Subscriber<? super T>> {
@Override
    public <S2, X extends Consumable<S2>> X extend(Function<Consumer<Subscriber<? super T>>, X> convertion) {
        return convertion.apply(onSubscribe);
    }
}
```

So you could probably implement it similarly.

stealthcode:
I'll try it out locally and see what works (and what doesn't).

stealthcode:
I got it to work but the `Publisher<T> onSubscribe` is not a `Consumer<Subscriber<? super T>` so it'd require a wrapper.

```java
@Override
public <S2, X extends Consumable<S2>> X extend(Function<Consumer<Subscriber<? super T>>, X> conversion) {
    return conversion.apply(new Consumer<Subscriber<? super T>>() {
        @Override
        public void accept(Subscriber<? super T> subscriber) {
            onSubscribe.subscribe(subscriber);
        }});
}
```

This required a base interface for `ConsumableFlowable`. 

```java
public interface Consumable<S> {
    void subscribe(S subscriber);
    public <S2, X extends Consumable<S2>> X extend(Function<Consumer<S>, X> conversion);
}

public interface ConsumableFlowable<T> extends Consumable<Subscriber<? super T>> {
}

public class Flowable<T> implements Publisher<T>, ConsumableFlowable<T> {
    // ...
}
```

akarnokd:
Let's experiment with that in a separate PR; also, I'd wait until the Flowable isn't a wrapper anymore but a base class that gets extended by all internal operators that were create/lift before.

stealthcode:
Why do we need a separate PR? Extend is part one 1.x and with the
Consumable it's even more necessary to switch between types. Sure you can
create custom instance operators for converting between them or subscribe
but this interface is explicitly for converting between and chaining
together different streams. So it's useful to convert between types that do
not necessarily have a direct relationship (I.e. in 2 unrelated library
jars) So I am very disinclined to leave it out.

akarnokd:
Why do you want this single PR to host every modification? There is a lot to do with 2.x and some of them, like extend, can be incremental.

stealthcode:
I can see that it could be implemented incrementally. This PR didn't say
anything to that effect. I think that as a team the RxJava committers
should come to discuss and agree on the 2.0 Design doc. This process should
or may include additional elaboration or amendments to areas that do not
yet have team commitment. This would include items like extend (which you
seem to be uncertain in your support of). And in that process team members
would commit to and merge their  commitments incrementally in the design
doc on the 2.x branch before PRs of the uncommitted features are opened.

If any of us have concerns, suggestions, or alternatives or require
elaboration on the design doc then we should comment or open an issue/pr to
the design. I think we should wait until the Consumable and/or extend are
committed and merged to 2.x in the design doc so others have had a chance
to review and commit to it before continuing our work on this PR.

src/main/java/io/reactivex/ConsumableFlowable.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+
+package io.reactivex;
+
+import org.reactivestreams.Subscriber;
+
+/**
+ * A ConsumableFlowable represents a backpressured sequence of values optionally followed
+ * by a teraminal signal.
```
runningcode: teraminal -> terminal

hzsweers:
@stealthcode the design doc PR is merged, is there anything preventing this PR from proceeding?

stealthcode:
@hzsweers Nope. Apologies for the delay but I won't be getting back to this until next week. Thanks for the reminder.

ZacSweers:
@stealthcode ping

akarnokd:
I'll redo this PR in multiple phases.

Issue #3808, opened 2016-03-29
NullPointerException caused by null elements in merge() queue


dlew:
Take a look at this (admittedly somewhat wonky) sample stream:

```java
Observable.range(0, 2)
    .flatMap(__ -> Observable.just(null))
    .subscribe(new Subscriber<Object>() {
      @Override public void onNext(Object o) {
        System.out.println("onNext(" + o + ")");
      }

      @Override public void onError(Throwable e) {
        e.printStackTrace();
      }

      @Override public void onStart() {
        request(1);
      }

      @Override public void onCompleted() { }
    });
```

In this sample I'm purposefully only requesting one item so that the second one ends up getting queued in the internal `OperatorMerge`. As a result of passing `null` to the queue, though, I end up getting this in `onError`:

```java
java.lang.NullPointerException
	at rx.internal.util.atomic.SpscUnboundedAtomicArrayQueue.offer(SpscUnboundedAtomicArrayQueue.java:71)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.queueScalar(OperatorMerge.java:465)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.tryEmit(OperatorMerge.java:437)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:228)
	at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:142)
	at rx.internal.operators.OperatorMap$1.onNext(OperatorMap.java:54)
	at rx.internal.operators.OnSubscribeRange$RangeProducer.fastpath(OnSubscribeRange.java:126)
	at rx.internal.operators.OnSubscribeRange$RangeProducer.request(OnSubscribeRange.java:63)
	at rx.Subscriber.setProducer(Subscriber.java:211)
	at rx.Subscriber.setProducer(Subscriber.java:205)
	at rx.internal.operators.OnSubscribeRange.call(OnSubscribeRange.java:38)
	at rx.internal.operators.OnSubscribeRange.call(OnSubscribeRange.java:26)
	at rx.Observable$2.call(Observable.java:162)
	at rx.Observable$2.call(Observable.java:154)
	at rx.Observable$2.call(Observable.java:162)
	at rx.Observable$2.call(Observable.java:154)
	at rx.Observable.subscribe(Observable.java:8191)
	at rx.Observable.subscribe(Observable.java:8158)
[...]

PR #3809, opened 2016-03-29
1.x: fix merge/flatMap crashing on an inner scalar of null

akarnokd:
This crash happens on contention where a scalar null value has to go through the scalar-queue which doesn't support nulls and requires NotificationLite transformations. (The drain side had the proper `getValue` in place already.)

Reported in #3808.


artem-zinnatullin:
:+1:

akarnokd:
Wrong test, didn't trigger the code path that goes into the queue. Updated.

artem-zinnatullin:
:+1: 

// New rule: check all PRs locally…

zsxwing:
:+1:

stevegury:
:+1:

PR #3814, opened 2016-03-31
1.x: Fix prolonged Producer retention in switchOnNext

akarnokd:
There are two cases when the operator `switchOnNext` / `switchMap` retained parts of an earlier source through its `Producer` via the arbiter structure:

  - when a backpressure-supporting source was followed by a unsupporting source which never sets a new Producer, keeping the old one there indefinitely
  - when a backpressure-supporting source terminated but was not followed by any other source for longer period of time

The fix swaps in a null-producer when a new source is encountered and when an old source terminates. The swap is guarded by the current index so outdated events won't interfere with newer sequences.

**Edit**

Fixed a `ConcurrentModificationException` because the `queue` wasn't nulled out after assigning it to `localQueue`. Plus, I've added a test to check the async-source and async-inner doesn't cause backpressure exceptions.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSwitch.java
+++ b/src/main/java/rx/internal/operators/OperatorSwitch.java
@@ -20,7 +20,7 @@ import java.util.*;
 import rx.*;
 import rx.Observable;
 import rx.Observable.Operator;
-import rx.exceptions.CompositeException;
+import rx.exceptions.*;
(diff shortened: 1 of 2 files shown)


akarnokd:
Updated with another fix + test.

akarnokd:
Closing this and will post a fresh PR with a completely rewritten operator as it has backpressure problems as well (overflow/hang).

Issue #2297, opened 2014-12-30
Make RxJavaPlugins.reset() public


TomRK1089:
Let me preface this by saying the RxJava library has made my life easier in just the short time I've been using it. However, there's one jagged edge that bit me today.

I'm attempting to use the `RxJavaPlugins` class to override the default schedulers via `registerSchedulersHook`. However, I want to only override the defaults for a single JUnit test. Ideally, I would call `RxJavaPlugins.getInstance().reset()` in my `@After` method...but I can't, because it is package private.

I could get at it via reflection, but it'd be easier if it was public -- unless there is a) a very good reason it is not or b) there is an obvious alternative I am missing (always a possibility!)

akarnokd:
Why don't you use operators which let you parametrize the scheduler and not rely on the defaults?

For the other hooks, you could set up a level of indirection, i.e., you register your own hook which then does allow setting/removing further hooks at will. You can even create it such a way that multiple hooks can work at the same time, but you have to implement a separate Observable-Observer and make careful decisions when these hooks return some value instead of just consuming.

Axxiss:
As a work around for this what I have done by now is defining a RxJavaTestPlugins class on src/test/java/rx/plugins

    public class RxJavaTestPlugins extends RxJavaPlugins {
        RxJavaTestPlugins() {
            super();
        }

        public static void resetPlugins(){
            getInstance().reset();
        }
    }

And then invoke `resetPlugins()`  on the tear down of my units test.

JakeWharton:
I very much would like something like this, too.

I do something similar to what you have but as a JUnit rule so it can be used as:
```java
@Rule public final RxJavaPluginsResetRule pluginsReset = new RxJavaPluginsResetRule();
```
and it calls `.reset()` both before and after each test.

JakeWharton:
FYI RxAndroid shipped with a `@Beta` `reset()` method public on its `RxAndroidPlugins` class.

pakerfeldt:
@JakeWharton Is there really a point of doing .reset() in your Rule? To me it doesn't seem to have the expected effect. Any subsequent call to Schedulers.* will still return the same original Scheduler even after calling .reset(). Since the hooks are only called once when the Schedulers instance is created.

I can't seem to find a valid use case for using RxJavaPlugins to register test schedulers in unit tests really. Maybe it was never intended for that anyway.

pakerfeldt:
[...]

PR #3820, opened 2016-04-01
Making RxPlugins reset() public

shivangshah:
Discussions found here: https://github.com/ReactiveX/RxJava/issues/2297

### DIFF
--- a/src/main/java/rx/plugins/RxJavaPlugins.java
+++ b/src/main/java/rx/plugins/RxJavaPlugins.java
@@ -17,6 +17,7 @@ package rx.plugins;
 
 import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
+import rx.annotations.Experimental;
 
 /**
  * Registry for plugin implementations that allows global override and handles the retrieval of correct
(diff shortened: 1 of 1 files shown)


hzsweers:
:+1:

stevegury:
Based on the discussion, it's definitely helpful, but also dangerous.
Could you add a comment indicating that it's dangerous?

Otherwise 👍

artem-zinnatullin:
Yeah, same concern about dangerous that this method may add. 

It looks like an easy way to do testing but in multithreaded environment (integrational or functional tests for example) it may break a lot of things in "random" manner.

@stevegury @akarnokd let's also mark it as `@Experimental` so we'll be able to remove it if community will have more issues than benefits?

stevegury:
Yeah, good idea for the `@Experimental` flag

shivangshah:
Thanks for the feedback @stevegury , @artem-zinnatullin. Just added the `@experimental` tag. Comments in the PR + commit message.

stevegury:
Please add a javadoc block, and explicitly mention that resetting the plugins is dangerous/unsafe during application runtime (I suspect most people won't check the link), keep the link at the end.

JakeWharton:
@shivangshah Friendly ping. Please update with Javadoc.

shivangshah:
@JakeWharton yes .. thanks for the reminder this totally slipped my mind .. Let me do it now ..

shivangshah:
@JakeWharton : Done ! Let me know if we need to add anything else as a part of the documentation.

akarnokd:
:+1:

stevegury:
👍

PR #3821, opened 2016-04-02
1.x: Add assertSubscribed() to TestSubscriber analog to assertUnsubscribed()

passsy:
`assertUnsubscribed()` exists, `assertSubscribed()` was missing

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -297,6 +297,18 @@ public class TestSubscriber<T> extends Subscriber<T> {
     }
 
     /**
+     * Asserts that this {@code Subscriber} is subscribed.
+     *
+     * @throws AssertionError
+     *          if this {@code Subscriber} is not subscribed
+     */
+    public void assertSubscribed() {
+        if (isUnsubscribed()) {
+            throw new AssertionError("Not subscribed.");
+        }
+    }
+
+    /**
      * Asserts that this {@code Subscriber} has received no {@code onError} notifications.
      * 
      * @throws AssertionError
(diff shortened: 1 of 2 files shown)


akarnokd:
The name is confusing (a freshly created `TestSubscriber` will return false for `isUnsubscribed()` even if it never was used with `Observable.subscribe()`) and not sure we need a `!isUnsubscribed()` assertion.

passsy:
I had the feeling it is missing when I wrote this test:

```java
        final RxProperty<String> prop = new RxProperty<>("myValue");
        final Observable<String> observable = prop.asObservable();

        final TestSubscriber<String> testSubscriber1 = new TestSubscriber<>();
        final Subscription subscription1 = observable.subscribe(testSubscriber1);

        final TestSubscriber<String> testSubscriber2 = new TestSubscriber<>();
        observable.subscribe(testSubscriber2);

        subscription1.unsubscribe();
        prop.set("otherValue");

        testSubscriber1.assertValue("myValue");
        testSubscriber1.assertUnsubscribed();

        testSubscriber2.assertValues("myValue", "otherValue");
        assertThat(testSubscriber2.isUnsubscribed()).isFalse(); // looks bad
        testSubscriber2.assertSubscribed(); // what I really want
```

zsxwing:
I also feel `assertSubscribed` is confusing. It sounds like verifying if `Observable.subscribe is called`. But the meaning in your PR is `Observable.subscribe is called but unsubscribe is not yet called`.

hzsweers:
I would be ok with it if naming could be agreed upon. Maybe something more to the effect of `assertIsSubscribed()` and `assertIsUnsubscribed()` for clarity? Agreed with @zsxwing that the naming sounds ambiguous in the original suggestion.

akarnokd:
> `assertThat(testSubscriber2.isUnsubscribed()).isFalse(); // looks bad`

I think `assertNoTerminalEvent()` is for this use case. I'm leaning towards closing this PR.

akarnokd:
Thanks for the contribution. I'm closing this due to inactivity, plus I think this change would cause confusion. If you have a different suggestion, please don't hesitate to post another PR.

passsy:
agree

PR #3824, opened 2016-04-03
1.x: javac 9 compatibility fixes

akarnokd:
The type inference of javac in JDK 9 has been changed in an incompatible way, marking many generics-related code invalid.

The most common problems:

  - The inference calculates `Observable<? extends T>` but from its perspective, a call to `subscribe()` with a `Subscriber<T>` is ambiguous as it matches `subscribe(Observer<? super T>)` and `subscribe(Subscriber<? super T>)`. Somehow, the more specific class is not considered as a valid choice. Downcasting to `Observable<T>` fixes the error.
  - `A<?>[]` no longer accepts `A<Ti>` elements, requires the use of raw types to get around

The tests appear to compile fine.

Note that I couldn't find a working IDE for JDK 9 and had to revert to command line trickery:

```
@echo off

dir /s /B src\main\java\rx\*.java > sources.txt
dir /s /B src\test\java\rx\*.java >> sources.txt

"c:\program files\java\jdk-9\bin\javac.exe" -cp C:/temp/rx/junit.jar;C:/temp/rx/mockito.jar @sources.txt 

del sources.txt
```

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -609,7 +609,7 @@ public class Observable<T> {
      */
     @SuppressWarnings("unchecked")
     public static <T1, T2, R> Observable<R> combineLatest(Observable<? extends T1> o1, Observable<? extends T2> o2, Func2<? super T1, ? super T2, ? extends R> combineFunction) {
-        return combineLatest(Arrays.asList(o1, o2), Functions.fromFunc(combineFunction));
+        return (Observable<R>)combineLatest(Arrays.asList(o1, o2), Functions.fromFunc(combineFunction));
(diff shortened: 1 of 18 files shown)


artem-zinnatullin:
:+1: 

// @akarnokd maybe report "incorrect" behavior of JDK9 compiler to OpenJDK issue tracker?

akarnokd:
There were reports such as [this](https://bugs.openjdk.java.net/browse/JDK-8075793?focusedCommentId=13865850) before. Bottom line is that they are fine with their breaking change.

stevegury:
:+1:

PR #3827, opened 2016-04-04
simplify OperatorObserveOn now that error is a non-volatile field

davidmoten:
This is a minor simplification in `OperatorObserveOn` where a temporary variable was created to avoid unnecessary additional reads of what used to be the volatile field `error`.  The field was changed to be non-volatile a while back so the temporary variable is no longer required. 

I assume it would have very minor perf impact (I haven't run them).


akarnokd:
Why do you want to read fields multiple times?

akarnokd:
It's not about performance, but code habits. By reading things into local variables, you establish a pattern that won't bite you with volatile variables later on.

akarnokd:
Plus, if the decision is made to clear the `error` field to avoid retention of the exception for too long, this would have to be reverted.

davidmoten:
Fair enough, I'll close.

PR #3828, opened 2016-04-04
1.x: AsyncSubject now supports backpressure

akarnokd:
`AsyncSubject` can trivially support backpressure when it emits a single item by setting the `SingleProducer` on the child `Subscriber` instead of calling `onNext` immediately.


stevegury:
:+1:

PR #3832, opened 2016-04-05
TestScheduler: Fix race condition in queue

nerd2:


akarnokd:
TestScheduler is a synchronous Scheduler that has to run on a single thread and its main purpose is to force async operators to execute synchronously as well.

Is there a reason you wanted this change?

nerd2:
Humm. This is the crash I was trying to fix, looked like a race but perhaps I'm using the testscheduler incorrectly somehow.

java.lang.NullPointerException
	at rx.schedulers.TestScheduler$TimedAction.access$200(TestScheduler.java:37)
	at rx.schedulers.TestScheduler.triggerActions(TestScheduler.java:111)
	at rx.schedulers.TestScheduler.advanceTimeTo(TestScheduler.java:97)
	at rx.schedulers.TestScheduler.advanceTimeBy(TestScheduler.java:84)

nerd2:
My class under test is subscribing to the scheduler asynchronously, I guess you're saying TestScheduler wasn't designed to operate list this?

  Observable.interval(mPollTimeMinutes, TimeUnit.MINUTES, mScheduler)
                             .subscribeOn(Schedulers.newThread()).subscribe(v -> {
                    getMyUpgrade();
                }, v -> {
                });

akarnokd:
It won't work with regular schedulers, for example:

```java
Observable.interval(1, TimeUnit.MILLISECONDS) // <-- note: it uses the default computation() scheduler
.observeOn(testScheduler)
.subscribe();
```

is prone to concurrency issues.

If you use `TestScheduler`, it has to be applied everywhere.

nerd2:
Thanks for helping - is this how it's intended to be used?

Observable.interval(1, TimeUnit.MILLISECONDS, test ? testScheduler : Schedulers.computation())
.observeOn(test ? testScheduler : Schedulers.newThread())
.subscribe();

akarnokd:
Yes.

nerd2:
Awesome, thanks.

PR #3836, opened 2016-04-06
1.x: fix switchMap/switchOnNext producer retention and backpressure

akarnokd:
This PR fixes the producer retention problem and backpressure problem in `switchMap`/`switchOnNext`.

In the original, when an inner source completed, its producer was still referenced after it. If there was no new inner source or the next inner source didn't set a new producer, the operator leaked the previous inner source's internals. The fix involves nulling out the producer field when the inner source completes (and is still the current one).

The original had an additional bug due to a race condition between the request arbitration and switching inner sources. In some cases, the switch didn't properly forward the remaining request amount to the new inner source, causing hangs. In some other cases, both the old and new inner sources were able to emit, causing overflow. The fix involves a new queue-drain loop with some emitter-loop help to work out the state transitions.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSwitch.java
+++ b/src/main/java/rx/internal/operators/OperatorSwitch.java
@@ -16,14 +16,17 @@
 package rx.internal.operators;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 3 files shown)


stevegury:
👍

PR #3839, opened 2016-04-07
1.x: fix takeLast() backpressure

akarnokd:
The counted and timed versions of `takeLast` had a concurrency bug which allowed concurrent emission from the underlying queue that holds onto the values till the upstream completes. When timed correctly, both an `onCompleted()` and a `request()` that went from 0 to n could enter the emit() method and break the internal state of the non-concurrent queue. Note that this queue doesn't have to be concurrent by nature because the operator can emit only *after* the upstream completes (`offer()` never runs concurrently with `poll()` unlike in `observeOn`).

In addition, the emission didn't immediately complete if the queue was emptied and requested == 0. In this operator, the fact that the queue became empty is the indicator of completion.

Both issues can be fixed with the `postCompleteXXX` mechanics in `BackpressureUtils`.

Since the operator has to store `null` values, the `postCompleteXXX` had to be extended with an exit transform overload to allow converting the `NotificationLite` values back to regular `T`/`null`.


stevegury:
👍

PR #3840, opened 2016-04-07
1.x: fix concatMap scalar/empty source behavior

akarnokd:
There were two tiny problems with the rewritten `concatMap` operator, mainly due to copy-paste error:

   - When a scalar was concatenated, the `active` field was not set and thus the next prefetched source overwrote it.
   - When the scalar was set on the arbiter, its custom producer didn't check for n > 0 zero causing instant emission always. Generally Producer.request(0) is allowed and should be no-op (unlike RS).
   - When an empty() was encountered as a last source, the it didn't trigger the check for completion afterwards.


zsxwing:
:+1:

stevegury:
👍

PR #3848, opened 2016-04-10
1.x: deanonymize Observable inner classes

akarnokd:
This PR factors out many anonymous inner classes into named classes and moves them outside `Observable`.

This cleanup makes the `Observable` more compact and helps debugging by showing more meaningful names in stacktraces.

### DIFF
--- a/src/main/java/rx/internal/operators/EmptyObservableHolder.java
+++ b/src/main/java/rx/internal/operators/EmptyObservableHolder.java
@@ -23,7 +23,8 @@ import rx.Observable.OnSubscribe;
  * Holds a singleton instance of an empty Observable which is stateless and completes
  * the child subscriber immediately.
  */
-public enum EmptyObservableHolder {
+public enum EmptyObservableHolder implements OnSubscribe<Object> {
+    INSTANCE
     ;
     
     /**
@@ -36,13 +37,10 @@ public enum EmptyObservableHolder {
     }
     
     /** The singleton instance. */
-    static final Observable<Object> EMPTY = Observable.create(new OnSubscribeEmpty());
+    static final Observable<Object> EMPTY = Observable.create(INSTANCE);
     
-    /** The OnSubscribe implementing the empty behavior. */
-    static final class OnSubscribeEmpty implements OnSubscribe<Object> {
-        @Override
-        public void call(Subscriber<? super Object> child) {
-            child.onCompleted();
-        }
+    @Override
+    public void call(Subscriber<? super Object> child) {
+        child.onCompleted();
     }
 }
--- a/src/main/java/rx/internal/operators/EmptyObservableHolder.java
+++ b/src/main/java/rx/internal/operators/EmptyObservableHolder.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package rx.internal.operators;
+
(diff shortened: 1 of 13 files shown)


src/main/java/rx/internal/operators/EmptyObservableHolder.java
```diff
@@ -0,0 +1,48 @@
...
+package rx.internal.operators;
+
+import rx.*;
+import rx.Observable.OnSubscribe;
+
+/**
+ * Holds a singleton instance of an empty Observable which is stateless and completes
+ * the child subscriber immediately.
+ */
+public enum EmptyObservableHolder {
```
JakeWharton: Any reason to use enum here? You're not using the enum singleton pattern. You can save a class by making this be the OnSubscribe implementor.

src/main/java/rx/internal/operators/NeverObservableHolder.java
```diff
@@ -0,0 +1,47 @@
...
+package rx.internal.operators;
+
+import rx.*;
+import rx.Observable.OnSubscribe;
+
+/**
+ * Holds a singleton instance of a never Observable which is stateless doesn't
+ * call any of the Subscriber's methods.
+ */
+public enum NeverObservableHolder {
```
JakeWharton: Same thing here

src/main/java/rx/internal/operators/ThrowObservable.java
```diff
@@ -0,0 +1,52 @@
...
+
+import rx.*;
+
+/**
+ * An Observable that invokes {@link Observer#onError onError} when the {@link Observer} subscribes to it.
+ * 
+ * @param <T>
+ *            the type of item (ostensibly) emitted by the Observable
+ */
+public final class ThrowObservable<T> extends Observable<T> {
```
JakeWharton: Is this type needed? Why not just the OnSubscribe class which is passed to create()?

src/main/java/rx/internal/util/LambdaNotificationObserver.java
```diff
@@ -0,0 +1,48 @@
...
+
+import rx.*;
+import rx.functions.*;
+
+/**
+ * An Observer that forwards the onXXX method calls to a notification callback
+ * by transforming each signal type into Notifications.
+ * @param <T> the value type
+ */
+public final class LambdaNotificationObserver<T> implements Observer<T> {
```
JakeWharton: I would think these "Lambda"-prefix classes should be "Action"-prefixed since lambdas are language implementation details for creating Action instances but otherwise have nothing to do with the implementation of these classes.

src/main/java/rx/internal/operators/EmptyObservableHolder.java
akarnokd: Good idea.

src/main/java/rx/internal/operators/NeverObservableHolder.java
akarnokd: Here too.

src/main/java/rx/internal/util/LambdaNotificationObserver.java
```diff
@@ -0,0 +1,48 @@
...
+        @Override
+        public void onError(Throwable e) {
+            onNotification.call(Notification.createOnError(e));
+        }
+        
+        @Override
+        public void onCompleted() {
+            onNotification.call(Notification.createOnCompleted());
+        }
+    }
```
JakeWharton: Indent is off on whole class

src/main/java/rx/internal/operators/ThrowObservable.java
akarnokd: Yeah, originally, this was in `Observable` above as a custom class. I don't see any reason why keep the custom type. I'll change it to OnSubscribe.

src/main/java/rx/internal/util/LambdaSubscriber.java
```diff
@@ -0,0 +1,51 @@
...
+        @Override
+        public void onError(Throwable e) {
+            onError.call(e);
+        }
+        
+        @Override
+        public void onCompleted() {
+            onCompleted.call();
+        }
+    }
```
JakeWharton: Indent off

JakeWharton:
Great change. Observable class is a beast so readability and stacktrace improvements are always a :+1:

akarnokd:
Thanks @JakeWharton ! Updated PR based on your suggestions.

stevegury:
👍

Issue #3724, opened 2016-02-21
APIs for easily changing the thread priority of default schedulers via hook.


JakeWharton:
On Android, the main thread is a sacred holy land at which we must all worship. As such this means other threads (ideally) shouldn't be favorably scheduled in priority with it. For network and filesystem operations this usually means using low priority threads.

One solution is to just use low priority threads on our own schedulers, but this doesn't affect code using the standard `io()` and `computation()` schedulers. Unfortunately there's a whole bunch of ceremony around how various schedulers are initialized which includes a mixed bag of public vs. non-public types and APIs which makes replacing them in a scheduler hook difficult.

I think the most easy way to accomplish this is the following:
 * [x] ~~`RxThreadFactory` constructor overload which takes a thread priority to use.~~ Not needed.
 * [ ] Factory methods on `Schedulers` for creating instances of `io()`, `computation()`, and `newThread()` except using a supplied `ThreadFactory`
   * [ ] `Schedulers.createIoScheduler(ThreadFactory)`
   * [ ] `Schedulers.createComputationScheduler(ThreadFactory)`
   * [ ] `Schedulers.createNewThreadScheduler(ThreadFactory)`

Nice to have:
 * [x] `Schedulers` using these factory methods to create the default implementations. This involves moving the thread factories from the individual types "up" to `Shedulers`.

JakeWharton:
Happy to do some of the work if this sounds acceptable.

stevegury:
@JakeWharton I think that sound acceptable, I'm happy to review your PR.

JakeWharton:
This is a little more tricky than I had initially anticipated. Perhaps it's worth talking about a means of setting the default priority of the internal thread pools through a mechanism like properties as a separate, easier first step? Plus that would be a lot lower-overhead for Android users anyway, and we could even use the existing Android platform detection to lower the thread priority by default.

erDaren:
so, how to set thread priority in android with android values(like`Process.THREAD_PRIORITY_BACKGROUND`)?

JakeWharton:
Pass a ThreadFactory

PR #3856, opened 2016-04-13
Provide factories for creating the default scheduler instances.

JakeWharton:
Unlike other hooks, the `RxJavaSchedulersHook` has no access to the real `Scheduler` instances in order to do wrapping/delegation. With these factory methods, a hook can access what would otherwise be the instance used since there is often no other means of creating these specialized schedulers.

For Android this wrapping/delegation use-case is important for UI testing. We have a means to tell the testing framework when the app is idle and to do that we need to hook into the schedulers to know when they're empty. This is easy to do currently, but you cannot wrap the real instance and instead have to supply alternate implementations which might subtly alter the behavior under test.

These three methods are referenced in #3724, and I think providing the defaults is useful as well as eventually adding overloads which take `ThreadFactory` instances for each.

### DIFF
--- a/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
+++ b/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
@@ -13,18 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package rx.schedulers;
+package rx.internal.schedulers;
 
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 
 import rx.*;
 import rx.functions.Action0;
-import rx.internal.schedulers.*;
 import rx.internal.util.RxThreadFactory;
 import rx.subscriptions.*;
 
-/* package */final class CachedThreadScheduler extends Scheduler implements SchedulerLifecycle {
+public final class CachedThreadScheduler extends Scheduler implements SchedulerLifecycle {
     private static final String WORKER_THREAD_NAME_PREFIX = "RxCachedThreadScheduler-";
     static final RxThreadFactory WORKER_THREAD_FACTORY =
             new RxThreadFactory(WORKER_THREAD_NAME_PREFIX);
--- a/src/main/java/rx/internal/schedulers/NewThreadScheduler.java
+++ b/src/main/java/rx/internal/schedulers/NewThreadScheduler.java
@@ -1,33 +1,36 @@
 /**
  * Copyright 2014 Netflix, Inc.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package rx.schedulers;
+package rx.internal.schedulers;
 
 import rx.Scheduler;
+import rx.internal.util.RxThreadFactory;
 
 /**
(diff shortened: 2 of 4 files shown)


artem-zinnatullin:
Please correct me if I'm wrong but looks like as soon as you access one of the new methods, `static final Schedulers.INSTANCE` will be instantiated which will make impossible to use `RxJavaSchedulersHook` as you described. Have you checked described use case? Maybe add such tests as example if possible?

>We have a means to tell the testing framework when the app is idle and to do that we need to hook into the schedulers to know when they're empty.

Just wanted to say for those who may find this PR later: empty/non-empty state of schedulers is not 100% source of truth for ui tests, you may have some background or time related (periodic/etc) code that does not affect the ui directly and may even freeze tests.

JakeWharton:
You are accessing the methods from the hook which was called from the static initializer of this class on first reference to `Schedulers.io()` or the like. I'll add a test.

artem-zinnatullin:
Yup, but there is still potential problem if somebody will try to create schedulers using these methods outside of the hook, for example in `@BeforeClass`/etc or probably even just for non-test needs -> may result in hours of trying to understand why hook is not overriding schedulers…

Maybe move these methods to `RxJavaSchedulersHook`?

JakeWharton:
That prevents package scoped methods from being used in the factory.

artem-zinnatullin:
>That prevents package scoped methods from being used in the factory.

We can move schedulers to `internal.schedulers` to "defend" from that (otherwise somebody can always use reflection to break package scope) and make constructors public.

Ideal solution would be to return schedulers in `RxJavaSchedulersHook` by default instead of returning `null` like it does now, so calling `super.getSomeScheduler()` will give original instance that you can wrap/etc as you wish. Though, we'll have to handle `null` as default too.

// sorry, I'm very tired and want to sleep, so I may come to wrong conclusions.

src/main/java/rx/schedulers/NewThreadScheduler.java
```diff
@@ -16,28 +16,18 @@
...
 import rx.Scheduler;
-import rx.internal.schedulers.NewThreadWorker;
-import rx.internal.util.RxThreadFactory;
 
 /**
- * Schedules work on a new thread.
+ * @deprecated This type was never publicly instantiable. Use {@link Schedulers#newThread()}.
  */
+@Deprecated
 public final class NewThreadScheduler extends Scheduler {
```
JakeWharton: It would be nice to just delete this type. It was `public` yes, but it was `final`, had a private constructor, `instance()` was package scoped, and `Schedulers.newThread()` was never guaranteed to return instance of it. This basically makes it completely unusable from application code.

artem-zinnatullin: `+1` to that, we can just mention that in changelog.

akarnokd: I agree, but not sure when the delete could happen.

JakeWharton:
The factories were moved to the hook.

akarnokd:
:+1:

src/main/java/rx/schedulers/NewThreadScheduler.java
artem-zinnatullin: Should be fine for even patch release, `NewThreadScheduler` has no public fields nor methods + private constructor. Without reflection or unchecked casting it's unusable for users.

I vote for removing it right in this PR.

artem-zinnatullin:
:+1: now, but let me ask again, what about this:

>Ideal solution would be to return schedulers in `RxJavaSchedulersHook` by default instead of returning `null` like it does now, so calling `super.getSomeScheduler()` will give original instance that you can wrap/etc as you wish. Though, we'll have to handle `null` as default too.

No need for new methods, no need to fight with instantiation order and package scope visibility. Yes, user won't be able to instantiate new scheduler of required type outside of the hook, but it's not possible at the moment, so nobody loses nothing.

src/main/java/rx/schedulers/NewThreadScheduler.java
stevegury: I agree, it's relatively safe to remove now (even with a patch release).

stevegury:
👍

PR #3862, opened 2016-04-17
1.x: fix from(Iterable) error handling of Iterable/Iterator

akarnokd:
The `from(Iterable)` ignored non-fatal exceptions thrown by the `Iterable` and `Iterator`.

Apart from wrapping the calls into `try-catch`, the overal emission algorithm has been changed:

  - It calls `hasNext` only once for each available value. Since the subscribe() checks for an outright empty `Iterable`, the drain loop runs only if it wasn't empty and does another `hasNext` check immediately to complete immediately if possible.
  - It uses what I call a fast-flow algorithm to avoid decrementing the requested amount (this) if there were more requests issued during the emission.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeFromIterable.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeFromIterable.java
@@ -20,6 +20,7 @@ import java.util.concurrent.atomic.AtomicLong;
 
 import rx.*;
 import rx.Observable.OnSubscribe;
+import rx.exceptions.Exceptions;
 
 /**
  * Converts an {@code Iterable} sequence into an {@code Observable}.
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OnSubscribeFromIterable.java
```diff
@@ -42,11 +43,24 @@ public OnSubscribeFromIterable(Iterable<? extends T> iterable) {
...
+            Exceptions.throwOrReport(ex, o);
+            return;
+        }
+            
+            
+        if (!b && !o.isUnsubscribed()) {
             o.onCompleted();
-        else 
+        } else { 
             o.setProducer(new IterableProducer<T>(o, it));
```
JakeWharton: Do we need to allocate and set the producer if unsubscribed? Seems like that check could enclose this whole if/else block.

akarnokd: Sure. updated.

artem-zinnatullin:
👍

davidmoten:
L93 no longer needs parameter `n`:
```java
void slowpath(long n) {
```

davidmoten:
:+1:  once last comment dealt with

akarnokd:
Actually, it should be using `n`. Fixed.

davidmoten:
Yep I see, nice.

stevegury:
👍

PR #3864, opened 2016-04-18
1.x: optimize concatMapIterable/flatMapIterable

akarnokd:
This PR reduces the overhead of `concatMapIterable`/`flatMapIterable` by not wrapping them into `Observable.from` sources but handling the generated `Iterable`s directly.

Since `Iterable`s are inherently synchronous, there is no difference between the two operators that now use the common underlying operator (i.e., flatMap can't chose a different source because it can't tell if an `Iterable` source is not ready or not; `hasNext()` is for indicating completion, not temporary lack of data).

Benchmark (i7 4790, Window 7 x64, Java 8u77):

image

`Iterable-M` is the current master, `Plain` is using `concatMap(Observable::from)` as a reference.

The `count=1` cases are bit slower because there is no good way of detecting if an `Iterable` holds only a single element, unlike with `just()`, and have to instantiate the full infrastructure even for a single element.

The same table using the master as baseline:

image

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeFlattenIterable.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeFlattenIterable.java
@@ -1,0 +1,358 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/main/java/rx/internal/operators/OnSubscribeFlattenIterable.java
```diff
@@ -0,0 +1,358 @@
...
+            done = true;
+            drain();
+        }
+        
+        void requestMore(long n) {
+            if (n > 0) {
+                BackpressureUtils.getAndAddRequest(requested, n);
+                drain();
+            } else
+            if (n < 0) {
```
stevegury: Could you move the `if` on the same line as the `else`, it makes me uncomfortable to see a line ending with `else`

stevegury:
Other that my previous comment, 👍.

PR #3866, opened 2016-04-18
1.x: Add Single.toCompletable()

### LINKED ISSUE #3865: flatMap from Single to Completable
Is there an easy way of doing a `flatMap` from a `Single` to a `Completable`?

Currently I need to do something like this:

```java 
 Completable saveData(Data data) {
      ... 
 }

 Single.just(data)
        .flatMapObservable(new Func1<Data, Observable<?>>() {
            @Override
            public Observable<?> call(Data data) {
                return saveData(data).toObservable();
            }
        })
        .toCompletable();
```

I think It would be good to have an operator like `Single.flatMapCompletable()` so we could do:
```java 
Single.just(data)
        .flatMapCompletable(new Func1<Data, Completable>() {
            @Override
            public Completable call(Data data) {
                return saveData(data);
            }
        });
```

artem-zinnatullin:
Closes #3865.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -2096,6 +2096,31 @@ public class Single<T> {
     }
 
     /**
+     * Returns a {@link Completable} that discards result of the {@link Single} (similar to
+     * {@link Observable#ignoreElements()}) and calls {@code onCompleted} when this source {@link Single} calls
+     * {@code onSuccess}. Error terminal event is propagated.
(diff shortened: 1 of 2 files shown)


akarnokd:
:+1:

zsxwing:
👍

zsxwing:
Seems GitHub doesn't add the PR description to the commit message automatically. I added it manually.

PR #3868, opened 2016-04-20
Fix an unsubscribe race in EventLoopWorker

zsxwing:
There is an unsubscribe race condition similar to #3842 in `CachedThreadScheduler.EventLoopWorker` and `EventLoopsScheduler.EventLoopWorker`. Image the following execution order:

| Execution Order  | thread 1 | thread 2 |
| ------------- | ------------- | ------------- |
| 1 |  | submit task A |
| 2 |  | submit task B |
| 3 | unsubscribe Worker  |  |
| 4 | unsubscribe task A |   |
| 5 | | task A won't run as it's unsubscribed |
| 6 | | run task B |
| 7 | unsubscribe task B |  |

So task B will run but its previous task A will be skipped.

This PR adds a check before running an action and moves `workerUnderConcurrentUnsubscribeShouldNotAllowLaterTasksToRunDueToUnsubscriptionRace` to `AbstractSchedulerConcurrencyTests` to test all concurrent schedulers.

### DIFF
--- a/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
+++ b/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
@@ -204,13 +204,21 @@ public final class CachedThreadScheduler extends Scheduler implements SchedulerL
         }
 
         @Override
-        public Subscription schedule(Action0 action, long delayTime, TimeUnit unit) {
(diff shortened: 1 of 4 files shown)


akarnokd:
:+1:

davidmoten:
:+1:  Thanks @zsxwing

PR #3871, opened 2016-04-20
Deprecate remaining public scheduler types.

JakeWharton:
A continuation of some work done in #3856 of removing uninstantiable types from the public API (via deprecation).


akarnokd:
:+1:

stevegury:
:+1:

PR #3879, opened 2016-04-22
Expose scheduler factories which accept thread factories.

### LINKED ISSUE #3724: APIs for easily changing the thread priority of default schedulers via hook.

JakeWharton:
This allows hooks to create schedulers whose threads have different priorities.

Closes #3724.

### DIFF
--- a/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
+++ b/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
@@ -24,31 +24,25 @@ import rx.internal.util.RxThreadFactory;
 import rx.subscriptions.*;
 
 public final class CachedThreadScheduler extends Scheduler implements SchedulerLifecycle {
-    private static final String WORKER_THREAD_NAME_PREFIX = "RxCachedThreadScheduler-";
-    static final RxThreadFactory WORKER_THREAD_FACTORY =
-            new RxThreadFactory(WORKER_THREAD_NAME_PREFIX);
-
-    private static final String EVICTOR_THREAD_NAME_PREFIX = "RxCachedWorkerPoolEvictor-";
-    static final RxThreadFactory EVICTOR_THREAD_FACTORY =
-            new RxThreadFactory(EVICTOR_THREAD_NAME_PREFIX);
-
     private static final long KEEP_ALIVE_TIME = 60;
     private static final TimeUnit KEEP_ALIVE_UNIT = TimeUnit.SECONDS;
     
     static final ThreadWorker SHUTDOWN_THREADWORKER;
     static {
-        SHUTDOWN_THREADWORKER = new ThreadWorker(new RxThreadFactory("RxCachedThreadSchedulerShutdown-"));
+        SHUTDOWN_THREADWORKER = new ThreadWorker(RxThreadFactory.NONE);
         SHUTDOWN_THREADWORKER.unsubscribe();
     }
     
     private static final class CachedWorkerPool {
+        private final ThreadFactory threadFactory;
         private final long keepAliveTime;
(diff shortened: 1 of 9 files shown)


src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
```diff
@@ -24,9 +24,6 @@
 import rx.subscriptions.*;
 
 public class EventLoopsScheduler extends Scheduler implements SchedulerLifecycle {
```
akarnokd: `final`?

src/test/java/rx/plugins/RxJavaSchedulersHookTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        final CountDownLatch latch = new CountDownLatch(1);
+        worker.schedule(new Action0() {
+            @Override
+            public void call() {
+                assertSame(threadRef.get(), Thread.currentThread());
+                latch.countDown();
+            }
+        });
+        assertTrue(latch.await(10, SECONDS));
```
akarnokd: Needs to shut down the scheduler.

src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
```diff
@@ -24,39 +24,39 @@
 import rx.subscriptions.*;
 
 public final class CachedThreadScheduler extends Scheduler implements SchedulerLifecycle {
```
akarnokd: Since you renamed the thread naming, why not rename this to IOScheduler?

src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
akarnokd: Also `ComputationScheduler` ?

src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
JakeWharton: I didn't rename it because it describes the implementation mechanism of the scheduler instead of where it was being used. Since the ThreadFactory was decoupled from this class, it made sense to switch the naming to instead reflect how the scheduler was being used instead of how it was implemented.

I'd prefer to keep the naming of these classes as-is for that reason, unless you feel strongly otherwise.

src/test/java/rx/plugins/RxJavaSchedulersHookTest.java
JakeWharton: Added

stevegury:
👍

JakeWharton:
Thanks!

PR #3883, opened 2016-04-27
1.x: fix multiple chained Single.doAfterTerminate not calling actions

### LINKED ISSUE #3881: Single not calling doAfterTerminate when used in flatMap
I've noticed that when I return a Single from a flatMap and that Single calls doAfterTerminate(Action0) the Action0's call method is never called.  This behaviour is not consistent with what happens when you do the same thing with an Observable.  I am using 1.1.3

When I run:
```java
Single.just("Test")
    .flatMap(s -> 
        Single.just("Test2")
            .doAfterTerminate(() -> System.out.println("singleFlatMapDoAfterTerminate"))
        )
.doAfterTerminate(() -> System.out.println("singleDoAfterTerminate"))
.subscribe(new TestSubscriber<String>());
```

I would expect the following to be printed to the console:
> singleDoAfterTerminate
singleFlatMapDoAfterTerminate

However only the following is printed to the console
>singleDoAfterTerminate

When I do the same with an Observable
```java
Observable.just("Test")
    .flatMap(s -> Observable.just("Test2")
        .doAfterTerminate(() -> System.out.println("observableFlatMapDoAfterTerminate"))
     )
.doAfterTerminate(() -> System.out.println("observableDoAfterTerminate"))
.subscribe(new TestSubscriber<String>());`
```

I see the following printed to the console:
>observableDoAfterTerminate
observableFlatMapDoAfterTerminate

akarnokd:
Resolves #3881.

The bug is caused by the impedance mismatch between `Observable`'s `Subscriber` and `Single`'s `SingleSubscriber`. The original code called `onNext()` which immediately signalled an `onSuccess` but also caused an unsubscription, preventing a delivery of `onCompleted()` in the inner Single.

The fix keeps `onSuccess`/`onError` rails intact  throughout the chain.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -2484,7 +2484,7 @@ public class Single<T> {
      */
     @Experimental
     public final Single<T> doAfterTerminate(Action0 action) {
-        return lift(new OperatorDoAfterTerminate<T>(action));
+        return create(new SingleDoAfterTerminate<T>(this, action));
     }
 
     /**
--- a/src/main/java/rx/internal/operators/SingleDoAfterTerminate.java
+++ b/src/main/java/rx/internal/operators/SingleDoAfterTerminate.java
@@ -1,0 +1,84 @@
+/**
+ * Copyright 2015 Netflix, Inc.
+ * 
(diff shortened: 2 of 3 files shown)


stevegury:
👍

PR #3886, opened 2016-04-27
throwIfFatal() now throws OnCompletedFailedException

### LINKED ISSUE #3885: onCompleted swallows all exceptions
I was fairly surprised by the behavior of this stream:

```java
Observable.empty()
    .subscribe(
        System.out::println, 
        System.err::println,
        () -> { throw new RuntimeException(); }
    );
```

What I expected to see was the thread crashing due to the exception; instead absolutely nothing happens because `onCompleted` swallows the exception entirely.

dlew:
Otherwise, if there's an error in onCompleted, the exception is
swallowed and unreported.

Fixes #3885


akarnokd:
:+1:

zsxwing:
👍

PR #3887, opened 2016-04-27
Have undeliverable errors on subscribe() sent to plugin error handler.

akarnokd:
This change will make sure if the `Subscriber` has unsubscribed, the non-fatal error is sent to the plugin error handler at least. I.e.,

```java
Observable.create(s -> {
    s.onError(new RuntimeException());
    throw new IllegalArgumentException();
}).subscribe();
```

In addition, if the original code threw `RuntimeException` in case the `subscriber.onError()` threw. I've changed this to `OnErrorFailedException` but not sure the original type was intentional or just left out.


stevegury:
👍

PR #3888, opened 2016-04-27
1.x: proposal: onTerminateDetach - detach upstream/downstream for GC

akarnokd:
By default, operators have final link to their child Subscriber and they never clear the upstream's Producer. If the end subscriber is referenced, for example in a `CompositeSubscription`, that keeps an entire chain of objects alive and can cause memory leaks (a use case common on Android).

This proposed operator detaches the structures and nulls out references if the sequence terminates or the downstream unsubscribes - at the cost of atomic operations and mandatory volatile read for each onNext(). 

I know this issue has been brought up several times, but instead of adding the overhead to every operator, I propose an operator that can be applied when the developer really needs it (i.e., could be part of the usual compose(subscribeOn/observeOn) setup).

On the implementation side, this requires deferred `Producer`/`request()` handling because requests from downstream may appear even before the upstream calls `setProducer` (if at all) and we can't use the base class' behavior for this.

### DIFF
--- a/src/test/java/rx/internal/operators/OnSubscribeDetachTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeDetachTest.java
@@ -1,0 +1,160 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.operators;
+
+import java.lang.ref.WeakReference;
+import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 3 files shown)


src/test/java/rx/internal/operators/OnSubscribeDetachTest.java
```diff
@@ -0,0 +1,160 @@
...
+        Observable.just(o).count().onTerminateDetach().subscribe(ts);
+        
+        ts.assertValue(1);
+        ts.assertCompleted();
+        ts.assertNoErrors();
+        
+        o = null;
+        
+        System.gc();
+        Thread.sleep(200);
```
stevegury: I believe that on most JVM implementations, the sleep is useless.

stevegury:
👍

src/test/java/rx/internal/operators/OnSubscribeDetachTest.java
akarnokd: I don't know, calling gc() is not dependable but I don't know any other way of detecting leaks as small as this.

JakeWharton:
Doesn't proper `Subscription` nulling and/or `CompositionException` clearing solve this? I haven't seen a lot of people complaining about leaks recently. Is something motivating this?

akarnokd:
What I've seen in Android examples, is that you add to a composite on creation and clear() the composite on destroy, but haven't seen the subscriber removed from the composite when it finishes normally but still while the activity is active.

JakeWharton:
Ah, I see. And that leaks the observable itself, not the activity.

PR #3890, opened 2016-04-28
1.x: map() and filter() should unsubscribe on crash eagerly

akarnokd:
In current master, when the function in `map` or `filter` crashes, the error is propagated but the upstream is not immediately unsubscribed but expected to happen eventually due to some other operator or `SafeSubscriber`.

The fix, and also the restructuring of the unsubscription chain, makes sure the crash also unsubscribes immediately (unless a fatal one).

Related: [StackOverflow question](http://stackoverflow.com/questions/35343474/why-does-it-need-onbackpressure-here-for-click-events)

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMap.java
+++ b/src/main/java/rx/internal/operators/OperatorMap.java
@@ -15,16 +15,20 @@
  */
 package rx.internal.operators;
 
+import rx.*;
 import rx.Observable.Operator;
-import rx.Subscriber;
-import rx.exceptions.Exceptions;
+import rx.exceptions.*;
 import rx.functions.Func1;
+import rx.internal.util.RxJavaPluginUtils;
 
 /**
  * Applies a function of your choosing to every item emitted by an {@code Observable}, and emits the results of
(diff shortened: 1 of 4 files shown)


src/main/java/rx/internal/operators/OperatorMap.java
```diff
@@ -36,28 +40,64 @@ public OperatorMap(Func1<? super T, ? extends R> transformer) {
...
+                return;
             }
-
-        };
+            actual.onCompleted();
+        }
+        
+        @Override
+        public void setProducer(Producer p) {
+            actual.setProducer(p);
```
JakeWharton: Missing a call to `super.setProducer`?

akarnokd: Map doesn't request on its own, unlike filter, so it can just pass along the producer to downstream.

JakeWharton: Ah, yeah. Makes sense.

stevegury:
👍

PR #3898, opened 2016-04-29
1.x: Add missing license headers

akarnokd:
This PR simply adds missing license header to main and test files.


stevegury:
👍

artem-zinnatullin:
Was always curious if this really required if repo has `LICENSE` file and it referenced in the `pom.xml`?

JakeWharton:
99% of people copy/pasting code from this project aren't going to look at
either those things to retain the correct license.

PR #3899, opened 2016-04-29
AssertionError should be treated as fatal

brianvanderwal:
Many test frameworks throw an `AssertionError` to indicate a failure.  For example, mock objects created using EasyMock will throw an `AssertionError` immediately if the mock receives an unexpected method call.  When using `onErrorResumeNext()`, this error may be silently swallowed.  And since this type of exception is not a part of normal program flow (it is only used in tests), it should not be up to the application to handle these errors specifically.

According to the Java documentation on throwables, I would actually expect RxJava to treat all `Error` subclasses as fatal, but perhaps there are other reasons not to do so.  My first thought was to register a global error handling plugin which would rethrow `AssertionError`s, but `onErrorResumeNext()` prevents the error from reaching the plugin.

### DIFF
--- a/src/main/java/rx/exceptions/Exceptions.java
+++ b/src/main/java/rx/exceptions/Exceptions.java
@@ -90,6 +90,8 @@ public final class Exceptions {
             throw (ThreadDeath) t;
         } else if (t instanceof LinkageError) {
             throw (LinkageError) t;
+        } else if (t instanceof AssertionError) {
+            throw (AssertionError) t;
         }
     }
 
--- a/src/test/java/rx/exceptions/ExceptionsTest.java
+++ b/src/test/java/rx/exceptions/ExceptionsTest.java
@@ -27,10 +27,16 @@ import rx.SingleSubscriber;
 import rx.Subscriber;
 import rx.Observable;
 import rx.Observer;
+import rx.functions.Action0;
 import rx.functions.Action1;
+import rx.functions.Func0;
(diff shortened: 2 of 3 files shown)


akarnokd:
I'm not sure about this, I certainly never had the need for it but let's hear the others' thoughts on this.

JakeWharton:
I have thrown AE in tests or used fail() which throws AE. While I've never
coupled these with error-resuming operators, I have a hard time seeing when
this would ever be an error you would want to recover from in normal
operation.

hzsweers:
Are there any other errors that skip past onError or any of the catching operators? Seems like it would be kind of unexpected behavior and a little too much of a special case. In the use case described, I would suggest that you should be using RxJava's test utility classes to check the error events as part of your test.

artem-zinnatullin:
Same thoughts as @hzsweers ^.

>When using onErrorResumeNext(), this error may be silently swallowed.

@brianvanderwal It means that your test is not good enough and it probably won't detect other kind of errors from the upstream too.

// I'd suggest to read about mutation testing: [1](https://en.wikipedia.org/wiki/Mutation_testing), [2](http://www.tutorialspoint.com/software_testing_dictionary/mutation_testing.htm) and see how you can improve your tests. TL;TR: any change of the code under the test should break the tests (ideally).

akarnokd:
Thank you for your contribution but it seems the other participants didn't see any compelling case to support its inclusion into the library.

PR #3903, opened 2016-05-01
Spare a few redundant calls in SleepingAction around determining the …

rzsombor:
…delay value

- Remove redundant outer conditional.
- Skip calling into System.currentTimeMillis(), which could potentially result in different values.


akarnokd:
:+1:

zsxwing:
👍

Issue #3860, opened 2016-04-15
Completable swallows OnErrorNotImplementedException


kotucz:
In the following case, the `IllegalStateException` should cause `OnErrorNotImplementedException`, because `subscribe()` has no onError handler specified.

```java
        Completable.fromCallable(() -> {
            throw new IllegalStateException("FOO");
        })
            .toSingleDefault(1)
            .subscribe();
```

In the related issue #3726 and PR #3733, similar problem was solved, that the exception was propagated to the thread exception handler. For example it works correctly in the following snippet and the `OnErrorNotImplementedException` delivered.

```java
        Completable.fromCallable(() -> {
            throw new IllegalStateException("FOO");
        })
            .subscribe();
```

Maybe more of the usages of the `toNpe()` shall be replaced with `deliverUncaughtException()`?

akarnokd:
Relying on `OnErrorNotImplementedException` being thrown is a fragile position but for consistency, PR welcome.

kotucz:
Not relying on it. But it hides programming errors. The app then behaves incorrectly instead of crashing.

PR #3904, opened 2016-05-01
1.x: Fix Completable swallows OnErrorNotImplementedException

prt2121:
Fixed https://github.com/ReactiveX/RxJava/issues/3860

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1972,6 +1972,7 @@ public class Completable {
             throw ex;
         } catch (Throwable ex) {
             ERROR_HANDLER.handleError(ex);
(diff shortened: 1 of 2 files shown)


akarnokd:
Looks good, :+1:

I wonder, shouldn't we switch to `Exceptions.throwIfFatal` as in `Observable`? Of course, the whole `Completable` needs rechecking.

prt2121:
Thanks for your review. I guess we need another clean-up PR for `Exceptions.throwIfFatal`?

prt2121:
Updated to use `Exceptions.throwIfFatal`. Thanks!

zsxwing:
👍

Issue #3892, opened 2016-04-28
java.lang.IllegalStateException: more items arrived than were requested


vadims:
We're seeing this exception for a small percent of our users. We haven't been able to reproduce this ourselves so providing a sample might not be feasible. Any idea what could be causing this based on the stacktrace alone?

```java
Non-fatal Exception: java.lang.IllegalStateException: more items arrived than were requested
       at rx.internal.producers.ProducerArbiter.produced(ProducerArbiter.java:98)
       at rx.internal.operators.OnSubscribeConcatMap$ConcatMapSubscriber.innerCompleted(OnSubscribeConcatMap.java:206)
       at rx.internal.operators.OnSubscribeConcatMap$ConcatMapInnerSubscriber.onCompleted(OnSubscribeConcatMap.java:345)
       at rx.internal.operators.OperatorMap$1.onCompleted(OperatorMap.java:43)
       at rx.internal.operators.OperatorSingle$ParentSubscriber.onCompleted(OperatorSingle.java:129)
       at rx.internal.operators.OperatorTakeLastOne$ParentSubscriber.onCompleted(OperatorTakeLastOne.java:107)
       at rx.internal.operators.OperatorMerge$MergeSubscriber.emitLoop(OperatorMerge.java:635)
       at rx.internal.operators.OperatorMerge$MergeSubscriber.emit(OperatorMerge.java:547)
       at rx.internal.operators.OperatorMerge$MergeSubscriber.onCompleted(OperatorMerge.java:268)
       at rx.internal.operators.OnSubscribeFromIterable.call(OnSubscribeFromIterable.java:47)
       at rx.internal.operators.OnSubscribeFromIterable.call(OnSubscribeFromIterable.java:32)
       at rx.Observable$2.call(Observable.java:233)
       at rx.Observable$2.call(Observable.java:225)
       at rx.Observable$2.call(Observable.java:233)
       at rx.Observable$2.call(Observable.java:225)
       at rx.Observable$2.call(Observable.java:233)
       at rx.Observable$2.call(Observable.java:225)
       at rx.Observable$2.call(Observable.java:233)
       at rx.Observable$2.call(Observable.java:225)
       at rx.Observable.unsafeSubscribe(Observable.java:8741)
       at rx.internal.operators.OnSubscribeConcatMap$ConcatMapSubscriber.drain(OnSubscribeConcatMap.java:286)
[...]

PR #3905, opened 2016-05-02
1.x: fix singleOrDefault() backpressure if source is empty

akarnokd:
Issue #3892 is a goldmine for missing backpressure problems. This PR fixes the case when `singleOrDefault` encounters an empty source and has to emit some default value. Fixed via setting the `SingleProducer` on the child on termination.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorSingle.java
+++ b/src/main/java/rx/internal/operators/OperatorSingle.java
@@ -20,6 +20,7 @@ import java.util.NoSuchElementException;
 import rx.Observable.Operator;
 import rx.Subscriber;
 import rx.internal.producers.SingleProducer;
+import rx.internal.util.RxJavaPluginUtils;
 
 /**
  * If the Observable completes after emitting a single item that matches a
@@ -121,6 +122,7 @@ public final class OperatorSingle<T> implements Operator<T, T> {
         @Override
         public void onError(Throwable e) {
             if (hasTooManyElements) {
+                RxJavaPluginUtils.handleException(e);
                 return;
             }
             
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorSingle.java
```diff
@@ -88,23 +74,23 @@ public void request(long n) {
...
-
-        void requestMore(long n) {
-            request(n);
+            request(2); // could go unbounded, but test expect this
         }
 
         @Override
         public void onNext(T value) {
+            if (hasTooManyElements) {
+                return;
```
artem-zinnatullin: Report error to plugin?

src/main/java/rx/internal/operators/OperatorSingle.java
```diff
@@ -136,6 +120,10 @@ public void onCompleted() {
 
         @Override
         public void onError(Throwable e) {
+            if (hasTooManyElements) {
```
artem-zinnatullin: Report to plugin?

src/main/java/rx/internal/operators/OperatorSingle.java
akarnokd: It gets here in case upstream ignored the unsubscription or is simply ignoring backpressure. There is no error here unless you want to new up and then sink some error.

src/main/java/rx/internal/operators/OperatorSingle.java
akarnokd: Yeah, I left it out for you specifically.

stevegury:
👍

zsxwing:
👍

PR #3912, opened 2016-05-05
1.x: fix filter() default-requesting and thus going unbounded

akarnokd:
Forgot that `Subscriber` defaults to requesting `Long.MAX_VALUE` unless suppressed and apparently there was no verification of filter's backpressure behavior; the two existing tests didn't really check for the delivery pattern.


stevegury:
👍 
I think the bug is serious enough that we should release a new version ASAP.

akarnokd:
Agree. Not sure if you can release with 1.1.4.1 or you have to  roll to 1.1.5.

JakeWharton:
Does `map` need the same change?

akarnokd:
No. Map doesn't set the Producer on `this` so no default-request happens.

JakeWharton:
Ah, that's right. I think I made the same mistake on the first PR, too.

Issue #3917, opened 2016-05-06
replay(bufferSize, time, TimeUnit) not working - truncate should be called with replay?


jam01:
The buffer only gets truncated when a new item is added. So any subscribers will always receive that last emission even if it's older than the time-limit. 

        'Observable<Long> retrofit = Observable.interval(2L, 2L, TimeUnit.SECONDS).take(10);

        ConnectableObservable<Long> connectableRetrofit;
        connectableRetrofit = retrofit.replay(1, 5, TimeUnit.SECONDS);

        System.out.println("Subscribing A");
        connectableRetrofit.subscribe(new Action1<Long>() {
            @Override
            public void call(Long aLong) {
                System.out.println("subscriber! A ========== " + aLong);
            }
        });

        System.out.println("Connecting!");
        connectableRetrofit.connect();

        Thread.sleep(10000L);

        System.out.println("Subscribing B");
        connectableRetrofit.subscribe(new Action1<Long>() {
            @Override
            public void call(Long aLong) {
                System.out.println("subscriber! B ========== " + aLong);
            }
        });

        Thread.sleep(30000L);

        System.out.println("Subscribing C");
        connectableRetrofit.subscribe(new Action1<Long>() {
            @Override
            public void call(Long aLong) {
                System.out.println("subscriber! C ========== " + aLong);
            }
        });`

Here subscriber C is subbing after 30 secs so you'd expect the buffer to have truncated the last value since the time-limit is 5 secs.

This is the output: 

> Subscribing A
Connecting!
subscriber! A ========== 0
subscriber! A ========== 1
subscriber! A ========== 2
subscriber! A ========== 3
subscriber! A ========== 4
Subscribing B
subscriber! B ========== 4
subscriber! A ========== 5
subscriber! B ========== 5
subscriber! A ========== 6
subscriber! B ========== 6
subscriber! A ========== 7
subscriber! B ========== 7
subscriber! A ========== 8
subscriber! B ========== 8
subscriber! A ========== 9
subscriber! B ========== 9
Subscribing C
subscriber! C ========== 9

akarnokd:
True, replay starts from the current head which is only moved forward by `onNext`. I've been thinking about rewriting `ReplaySubject` to support backpressure; in that structure, when the child Subscriber picks up [the head](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/processors/ReplayProcessor.java#L975), we can skip old entries right there.

[...]

PR #3918, opened 2016-05-08
1.x: ReplaySubject now supports backpressure

akarnokd:
In addition, the behavior of time-limited mode has been changed. Late subscribers will now skip stale data.

Related issue: #3917

### DIFF
--- a/src/main/java/rx/subjects/ReplaySubject.java
+++ b/src/main/java/rx/subjects/ReplaySubject.java
@@ -16,22 +16,16 @@
 package rx.subjects;
 
 import java.lang.reflect.Array;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.*;
 
+import rx.*;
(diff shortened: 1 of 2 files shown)


hzsweers:
Would it be possible to split this up into a couple commits? GitHub won't show the diff in its current state and says to view locally

akarnokd:
No, ReplaySubject has simply too many expected features: unbounded, size bounded and size+time bounded modes, each adding 100s of lines. It is fully rewritten so comparison wouldn't do much help.

stealthcode:
This makes it very difficult to review. What do you propose @akarnokd ?

akarnokd:
Check out and have your IDE compare this PR against the master version.

stealthcode:
> Late subscribers will now skip stale data.

Does this mean that subscribers were not skipping stale data? This change sounds fixing a bug?

akarnokd:
In the original, once the source completed, the current elements in the bounded buffers were frozen in time. Late subscribers would get all the data, including those who have become older by the time these subscriptions happen. Same was true for live, but inactive replays where a subscriber would started from an outdated element. These were expected by the unit tests but #3917 expected only fresh data.

stevegury:
👍

PR #3919, opened 2016-05-09
1.x: Subscriber.NOT_SET Long -> long, saves an unboxing per instance

akarnokd:
Somehow, Subscriber.NOT_SET was object `Long` instead of primitive `long`.


JakeWharton:
A tiny win, but a win no less :+1:

akarnokd:
It bothered me for some time now, seeing the Long there. This is not just about the unboxing, but remove the indication that `NOT_SET` can be null for meaningful reasons.

hzsweers:
👍

davidmoten:
Looks like my bad from last year, thanks :+1:

stevegury:
👍

Issue #3921, opened 2016-05-09
Possible resource leak in OnSubscribeUsing


dimas:
Hi.
I am considering using of RxJava to represent REST API responses so was looking into the sources to understand how stuff works. In the OnSubscribeUsing operator there is something that looks like a logic error to me that may lead to leaking resources.

(I am reporting "theoretical" issue, I have not tried actually making it happen. Also, I have almost zero knowledge on how RxJava works so can be really missing something. Sorry about that in advance)

So in https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OnSubscribeUsing.java the call() method allocates resource the first thing and then creates Observable, attaches dispose handler etc. In the end it calls unsafeSubscribe handling for the potential exception.
But what if exception happens earlier? Imagine the resource was allocated successfully but observableFactory.call(resource) throws for example. There seems to be nothing to dispose the resource in that case.

Again. I have no deep understanding of RxJava but it looks to me that resource disposal should be happening in the top level try/finally block if resource was allocated but doOnTerminate has not been installed yet (or failed to install). I would probably just remove the inner try/catch and did all the handling at the top level (possibly setting some boolean to indicate that stuff actually started so no cleanup needed).

Cheers

akarnokd:
Thanks for reporting! This is indeed a bug and I've posted a fix PR #3922 for it.

dimas:
Oh that was quick. Impressive )
Just out of curiosity - is it kind of implied that thing like 
```java
 subscriber.add(disposeOnceOnly);
```
always succeed and cannot throw? As well as doOnTerminate / doAfterTerminate.
Isn't it safer to always do dispose in a big finally block _unless_ unsafeSubscribe was successful? That would probably also simplify exception handling as now you have two identical copy&paste blocks...

Sorry if I am asking nonsense - really unfamiliar with the framework right now.

PS: you also have a semicolon following .doAfterTerminate(disposeOnceOnly) on a separate line - probably a typo.

akarnokd:
The Observable can complete outside the method so the classical imperative try-finally doesn't work.

dimas:
For the completion outside the method you install these doOnTerminate / doAfterTerminate "callbacks". I was only talking on what can happen inside OnSubscribeUsing.call method. You already have something like
```java
    public void call(final Subscriber<? super T> subscriber) {
        try {
            final Resource resource = resourceFactory.call();
            // ... do some stuff ...
            try {
[...]

PR #3922, opened 2016-05-09
1.x: fix using() resource cleanup when factory throws or being non-eager

akarnokd:
The operator `using` didn't call the resource cleanup code if the `observableFactory.call()` code crashed. In addition, a non-eager using didn't call the resource cleanup if one subscribed with `unsafeSubscribe` or the subscription to the generated `Observable` crashed.

Related: #3921

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeUsing.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeUsing.java
@@ -15,7 +15,6 @@
  */
 package rx.internal.operators;
 
-import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import rx.*;
@@ -26,6 +25,9 @@ import rx.observers.Subscribers;
 
 /**
  * Constructs an observable sequence that depends on a resource object.
+ * 
+ * @param <T> the output value type
+ * @param <Resource> the resource type
  */
 public final class OnSubscribeUsing<T, Resource> implements OnSubscribe<T> {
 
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OnSubscribeUsing.java
```diff
@@ -56,22 +59,43 @@ public void call(final Subscriber<? super T> subscriber) {
...
+            try {
+                source = observableFactory
+                // create the observable
+                        .call(resource);
+            } catch (Throwable e) {
+                Throwable disposeError = dispose(disposeOnceOnly);
+                Exceptions.throwIfFatal(e);
+                Exceptions.throwIfFatal(disposeError);
+                if (disposeError != null) {
+                    subscriber.onError(new CompositeException(Arrays.asList(e, disposeError)));
```
hzsweers: CompositeException accepts varargs right?

src/main/java/rx/internal/operators/OnSubscribeUsing.java
```diff
@@ -56,22 +59,43 @@ public void call(final Subscriber<? super T> subscriber) {
...
                 observable = source
                 // dispose on completion or error
                         .doOnTerminate(disposeOnceOnly);
-            else
-                observable = source;
+            } else {
+                // dispose after the terminal signals were sent out
+                observable = source
+                        .doAfterTerminate(disposeOnceOnly)
+                        ;
```
hzsweers: Any reason this is its own line?

akarnokd: I kept most of the original style.

stevegury: I suppose you're in the process of changing this?

akarnokd: Not really. You want me to restyle the whole operator?

stevegury: Just to clarify, I'm talking about the lone semicolon on line 91 that would be a perfect fit for ending the line 90.
I don't see any other lone semicolon in that file.

akarnokd:
Changes applied.

stevegury:
👍

Issue #3923, opened 2016-05-09
Should RxRingBuffer's validator really be on the computation scheduler?


jnlopar:
I just helped debug an issue where some Android robolectric tests timed out after upgrading from 1.1.1 to 1.1.5 due to https://github.com/ReactiveX/RxJava/commit/ef1d418f43e55eafb1cf9f09d13f37afb5b80b84

As it turned out, they were replacing the computation scheduler with immediate via RxJavaPlugins. The new commit caused RxRingBuffer.SIZE to load the RxRingBuffer class in any call to observeOn, which in turn created its ObjectPools, which in turn ran their start() which sets up a periodic scheduling on the computation thread. This caused the tests to hang indefinitely because of the repeated SleepingActions triggered by the validation tasks.

So, since RxRingBuffer can be loaded in various places by Rx internally, that seems to make it a lot more difficult to reliably replace the computation scheduler, particularly with immediate. I recommended against the immediate override in this case, but using a test scheduler wasn't really helping in this scenario either, although I am unfamiliar with that codebase and didn't have time to dive too deep into why. Short term fix for this particular case was to call observeOn with a size constant to sidestep the RxRingBuffer load in these tests, until we can clean up the underlying scheduler plugin overrides. However, I'm wondering if maybe internal scheduling things like this should have their own scheduler? Maybe an internal computation scheduler that could be an additional override in the schedulers hook. That would allow them to run unhindered in most cases despite computation being overridden, but in extreme cases they could be overridden as well.

jnlopar:
Btw, I can write the PR for this once a decision is made, but I thought it deserved a little discussion of what the right fix is before I jump to an implementation.

JakeWharton:
This came up recently in another forum to the same effect but I never followed up here. But :+1: to something that doesn't pass through user code in such a dangerous way.

akarnokd:
Yep, they have a cyclic dependency. I believe the `ObjectPool`s can be switched to `GenericScheduledExecutorService`. Let me write up a PR.

akarnokd:
See #3924.

akarnokd:
Fix merged in #3924.

PR #3924, opened 2016-05-09
1.x: fix RxRingBuffer-pool depending on the computation scheduler

akarnokd:
This PR fixed the `ObjectPool` dependency on the `Schedulers.computation()` which can be replaced via plugin hooks causing unexpected and long sleeps (i.e., when `computation()` is replaced by `immediate()`).

See #3923.

I'd also consider removing the `RxRingBuffer` entirely.


stevegury:
👍

PR #3931, opened 2016-05-12
1.x: add groupBy overload with evictingMapFactory

davidmoten:
I have a long running stream using `groupBy` that over time will accumulate 10s of millions of keys. If I can specify an evicting map to `groupBy` then I'll be able to keep it down to ~10,000 keys. 

This PR supports this use case and could be used with Guava's `CacheBuilder` like this:

```java
Func1<Action1<K>, Map<K, Object>> mapFactory = 
    action -> CacheBuilder.newBuilder()
              .maximumSize(1000)
              .expireAfterAccess(12, TimeUnit.HOUR)
              .removalListener(key -> action.call(key))
              .<K, Object> build().asMap();
observable
    .groupBy(keySelector, elementSelector, mapFactory)
    ...
```
I'll enhance this PR with more unit tests if this looks a good direction.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -6376,6 +6376,65 @@ public class Observable<T> {
      * 
      * @param keySelector
      *            a function that extracts the key for each item
+     * @param elementSelector
+     *            a function that extracts the return element for each item
+     * @param evictingMapFactory
+     *            a function that given an eviction action returns a {@link Map} instance that will be used to assign 
+     *            items to the appropriate {@code GroupedObservable}s. The {@code Map} instance must be thread-safe 
+     *            and any eviction must trigger a call to the supplied action (synchronously or asynchronously). 
+     *            This can be used to limit the size of the map by evicting keys by maximum size or access time for 
+     *            instance. Here's an example using Guava's {@code CacheBuilder} from v19.0:
+     *            <pre>
+     *            {@code
+     *            Func1<Action1<K>, Map<K, Object>> mapFactory 
+     *              = action -> CacheBuilder.newBuilder()
+     *                  .maximumSize(1000)
+     *                  .expireAfterAccess(12, TimeUnit.HOURS)
+     *                  .removalListener(notification -> action.call(notification.getKey()))
+     *                  .<K, Object> build().asMap();
+     *            }
+     *            </pre>
+     *            
+     * @param <K>
+     *            the key type
+     * @param <R>
+     *            the element type
+     * @return an {@code Observable} that emits {@link GroupedObservable}s, each of which corresponds to a
+     *         unique key value and each of which emits those items from the source Observable that share that
+     *         key value
+     * @throws NullPointerException
+     *             if {@code evictingMapFactory} is null
+     * @see <a href="http://reactivex.io/documentation/operators/groupby.html">ReactiveX operators documentation: GroupBy</a>
+     */
+    @Experimental
+    public final <K, R> Observable<GroupedObservable<K, R>> groupBy(final Func1<? super T, ? extends K> keySelector, 
+            final Func1<? super T, ? extends R> elementSelector, final Func1<Action1<K>, Map<K, Object>> evictingMapFactory) {
+        if (evictingMapFactory == null) {
+            throw new NullPointerException("evictingMapFactory cannot be null");
+        }
+        return lift(new OperatorGroupBy<T, K, R>(keySelector, elementSelector, evictingMapFactory));
+    }
+    
+    /**
+     * Groups the items emitted by an {@code Observable} according to a specified criterion, and emits these
+     * grouped items as {@link GroupedObservable}s. The emitted {@code GroupedObservable} allows only a single 
+     * {@link Subscriber} during its lifetime and if this {@code Subscriber} unsubscribes before the 
+     * source terminates, the next emission by the source having the same key will trigger a new 
+     * {@code GroupedObservable} emission.
+     * <p>
+     * <img width="640" height="360" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/groupBy.png" alt="">
+     * <p>
+     * <em>Note:</em> A {@link GroupedObservable} will cache the items it is to emit until such time as it
+     * is subscribed to. For this reason, in order to avoid memory leaks, you should not simply ignore those
+     * {@code GroupedObservable}s that do not concern you. Instead, you can signal to them that they may
+     * discard their buffers by applying an operator like {@link #ignoreElements} to them.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code groupBy} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * 
+     * @param keySelector
+     *            a function that extracts the key for each item
      * @param <K>
      *            the key type
      * @return an {@code Observable} that emits {@link GroupedObservable}s, each of which corresponds to a
--- a/src/main/java/rx/internal/operators/OperatorGroupBy.java
+++ b/src/main/java/rx/internal/operators/OperatorGroupBy.java
@@ -21,11 +21,13 @@ import java.util.concurrent.atomic.*;
 
 import rx.*;
 import rx.Observable.*;
+import rx.exceptions.Exceptions;
 import rx.functions.*;
 import rx.internal.producers.ProducerArbiter;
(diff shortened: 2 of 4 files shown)


davidmoten:
This PR may be useful still but I've realised I can get what I want with more control using `defer`, `PublishSubject` and `takeUntil` on the grouped observables.

akarnokd:
Yes, you can `timeout` or `takeUntil` on each group to get it removed from the internal map. I'm not sure about the map factory because we need something concurrently modifiable.

davidmoten:
Yep you're right but I put that requirement in the javadoc and the example
map from cachebuilder is threadsafe.

davidmoten:
I'm happy using `timeout` with my use case now but I'll leave this PR up for a few days to see if anyone else has a use case.

davidmoten:
@akarnokd  I think there's a catch with `timeout`. I'm using it as below but I believe I could lose an emission due to timeout cutting in on say the processing occurring in `operator1`:

```java
o.groupBy(...)
  .flatMap(
    g -> 
        g.operator1()
         .timeout(10, TimeUnit.SECONDS)
         .onErrorResumeNext(completeOnTimeoutException))
)
```
I guess I'll have to use `timeout` or `timer` in a different way so it doesn't shortcut an in-flight emission.

akarnokd:
Yes, there  is  chance values get dropped when an emission and group unsubscribe happens at the same time. You can't know the value made it or not. If your value has its own lifecycle, that's going to be a problem. Rx is not designed for that and it's the next challenge of the field - especially since reactive-io is full of life-cycle managed buffers.

davidmoten:
>If your value has its own lifecycle, that's going to be a problem. Rx is not designed for that and it's the next challenge of the field - especially since reactive-io is full of life-cycle managed buffers.

Your absolutely right that my use case has some special lifecycle that I'm trying to manage with `groupBy`. It's an event sourcing/cqrs  use case where emissions upstream of `groupBy` are persisted (you knew io was part of the scenario you perceptive fellow) and whenever a `groupBy` emitted `GroupedObservable` is subscribed to (or resubscribed to) the stream starts with the events already recorded for that key (read from disk) before processing the incoming event from upstream. 

It does seem to me now that if I don't want to lose emissions for my use case (and want to keep the `groupBy` keys down in numbers) then I have to record information about groups upstream of `groupBy` and possibly modify the upstream to include some sort of keyed sentinel to tell the `GroupedObservable` to complete based on last access time for a key for instance. **OR** I just use the modified `groupBy` in this PR.

I'll progress the implementation of the use case and see what else turns up.

src/main/java/rx/Observable.java
```diff
@@ -5905,6 +5905,60 @@ public final void forEach(final Action1<? super T> onNext, final Action1<Throwab
...
      * @param keySelector
      *            a function that extracts the key for each item
+     * @param elementSelector
+     *            a function that extracts the return element for each item
+     * @param evictingMapFactory
+     *            a function that given an eviction action returns a {@link Map} instance that will be used to assign 
+     *            items to the appropriate {@code GroupedObservable}s. The {@code Map} instance must be thread-safe 
+     *            and any eviction must trigger a call to the supplied action (synchronously or asynchronously). 
+     *            This can be used to limit the size of the map by evicting keys by maximum size or access time for 
+     *            instance. If {@code evictingMapFactory} is null then no eviction strategy will be applied (and a suitable default thread-safe
```
akarnokd: I wouldn't support `null` at the surface level. Those who don't wish to have a custom eviction strategy should use the other overload.

akarnokd:
If you want to pursue this further, please rebase it.

davidmoten:
Yep I'd like to pursue this one. This PR is not great to put in an external library because it is so coupled to the groupBy implementation and its tests. I'll fix the API and I realize I may also need to put try-catch on `mapFactory.call` and `map.get` and `map.put` calls with some specific error handling (catching fatal errors, calling onError etc). I'll have a look at it.

I'd also like to use the Guava `CacheBuilder` in unit tests. Can I add that test dependency?

akarnokd:
I'm not against adding `testCompile` entries. Hopefully, Guava doesn't drag in too many other libraries.

davidmoten:
Guava only has a few optional dependencies that I wouldn't include and the current version 19.0 still supports java 6.

akarnokd:
Great!

akarnokd:
:+1:

akarnokd:
Oops, merged in a `build.gradle` change and broke this. Could you rebase again?

/cc @artem-zinnatullin @stevegury for review on this

src/main/java/rx/internal/operators/OperatorGroupBy.java
```diff
@@ -46,35 +48,50 @@
...
-        final GroupBySubscriber<T, K, V> parent = new GroupBySubscriber<T, K, V>(t, keySelector, valueSelector, bufferSize, delayError);
+    public Subscriber<? super T> call(Subscriber<? super GroupedObservable<K, V>> child) {
+        final GroupBySubscriber<T, K, V> parent;
+        try {
+            parent = new GroupBySubscriber<T, K, V>(child, keySelector, valueSelector, bufferSize, delayError, mapFactory);
+        } catch (Throwable ex) {
+            //Can reach here because mapFactory.call() may throw in constructor of GroupBySubscriber
+            Exceptions.throwOrReport(ex, child);
+            Subscriber<? super T> parent2 = Subscribers.empty();
+            parent2.unsubscribe();
```
artem-zinnatullin, line 83: Let's add `Subscribers.unsubscribed()`?

src/main/java/rx/internal/operators/OperatorGroupBy.java
```diff
@@ -132,6 +151,34 @@ public GroupBySubscriber(Subscriber<? super GroupedObservable<K, V>> actual, Fun
...
+                this.groups = new ConcurrentHashMap<Object, GroupedUnicast<K, V>>();
+                this.evictedKeys = null;
+            } else {
+                this.evictedKeys = new ConcurrentLinkedQueue<K>();
+                this.groups = createMap(mapFactory, new EvictionAction<K>(evictedKeys));
+            }
+        }
+        
+        //declare a class instead of using anonymous class to 
+        //limit enclosing scope. Is this of value?
```
artem-zinnatullin: yup + makes stactraces better, @akarnokd recently made similar change and coverted a lot of anonymous classes into static inner ones

I'd remove comment to this class

src/test/java/rx/internal/operators/OperatorGroupByTest.java
```diff
@@ -1804,4 +1809,212 @@ public Integer call(Integer pair) {
...
+    @Test
+    public void mapFactoryEvictionWorks() {
+        Func1<Integer, Integer> keySelector = new Func1<Integer, Integer> (){
+            @Override
+            public Integer call(Integer t) {
+                return t /10;
+            }};
+        Func1<Integer, Integer> elementSelector = UtilityFunctions.identity();
+        final List<Integer> evictedKeys = new ArrayList<Integer>();
+        //normally we would use Guava CacheBuilder for instance but to save a test dependency
```
artem-zinnatullin: You've added Guava, so either this test can be removed or at least this comment :)

src/main/java/rx/internal/operators/OperatorGroupBy.java
davidmoten, line 83: No worries, separate PR I'd suggest.

src/main/java/rx/internal/operators/OperatorGroupBy.java
davidmoten: will do

src/test/java/rx/internal/operators/OperatorGroupByTest.java
davidmoten: I'll keep the test because i can test expectations of eviction quite precisely but I'll update the comment

davidmoten:
Thanks for the review @artem-zinnatullin.
I've addressed your comments and there's one little addition which is the rename of one of the test methods to use camel case and a number to differentiate from `testGroupByBackpressure` (`testgroupByBackpressure` -> `testGroupByBackpressure2`).

stevegury:
👍

davidmoten:
I take it that being an API enhancement this PR still requires one more collaborator approval?

davidmoten:
oh and thanks @stevegury for giving it a look!

akarnokd:
Thanks @davidmoten for the contribution!

PR #3934, opened 2016-05-12
1.x: TestSubscriber extra info on assertion failures

akarnokd:
This PR adds extra information to assertion failure messages on `TestSubscriber` and `TestObserver`, indicating:

  - the listener didn't receive any `onCompleted` calls, which is an indication of hung or skipping operation,
  - there were errors received, indicating a failure in the event generation process.

Previously, if there was something wrong with the sequence, the order and type of assertions were mostly unhelpful: if `assertValues` was first, the lack of values failure could hide a revealing onError call. If the `assertNoErrors()` was first, the error is visible but no way of knowing how far the sequence got.

Now, it is generally okay to use `assertValues` first, which along the difference, will print the lack of completion and the number of exceptions received, plus, the `AssertionError` will have its cause initialized to the actual or composited exception. The message format thus changes:

```
original assertion message with details (0 completions) (+1 error)
...
caused by
...
```

This extra information saved me a lot of time in 2.x and Rsc development.

Note that this change doesn't make the `assertXXX`s also assert for completion or error at all. If the values match, but there is an additional error instead of completion, one has to assert that separately, just like now.

### DIFF
--- a/src/main/java/rx/observers/TestObserver.java
+++ b/src/main/java/rx/observers/TestObserver.java
@@ -167,16 +167,17 @@ public class TestObserver<T> implements Observer<T> {
      * @param message the message to use for the error
      */
     final void assertionError(String message) {
-        StringBuilder b = new StringBuilder();
-        
-        if (onCompletedEvents.isEmpty()) {
-            b.append("(active) ");
-        }
+        StringBuilder b = new StringBuilder(message.length() + 32);
         
         b.append(message);
         
+        
+        b.append(" (");
+        b.append(onCompletedEvents.size());
+        b.append(" completions)");
+        
         if (!onErrorEvents.isEmpty()) {
-            b.append(" (+ ")
+            b.append(" (+")
             .append(onErrorEvents.size())
             .append(" errors)");
         }
--- a/src/main/java/rx/observers/TestObserver.java
+++ b/src/main/java/rx/observers/TestObserver.java
@@ -174,13 +174,23 @@ public class TestObserver<T> implements Observer<T> {
         
         
         b.append(" (");
-        b.append(onCompletedEvents.size());
-        b.append(" completions)");
+        int c = onCompletedEvents.size();
+        b.append(c);
+        b.append(" completion");
+        if (c != 1) {
+            b.append("s");
+        }
+        b.append(")");
         
         if (!onErrorEvents.isEmpty()) {
+            int size = onErrorEvents.size();
             b.append(" (+")
-            .append(onErrorEvents.size())
-            .append(" errors)");
+            .append(size)
+            .append(" error");
+            if (size != 1) {
+                b.append("s");
+            }
+            b.append(")");
         }
         
         AssertionError ae = new AssertionError(b.toString());
--- a/src/main/java/rx/observers/TestObserver.java
+++ b/src/main/java/rx/observers/TestObserver.java
@@ -15,12 +15,11 @@
  */
 package rx.observers;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 import rx.Notification;
 import rx.Observer;
+import rx.exceptions.CompositeException;
 
 /**
  * Observer usable for unit testing to perform assertions, inspect received events or wrap a mocked Observer.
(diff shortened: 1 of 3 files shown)


src/main/java/rx/observers/TestObserver.java
```diff
@@ -146,22 +145,53 @@ public void assertReceivedOnNext(List<T> items) {
...
         }
     }
 
+    /**
+     * Combines an assertion error message with the current completion and error state of this
+     * TestSubscriber, giving more information when some assertXXX check fails.
+     * @param message the message to use for the error
+     */
+    final void assertionError(String message) {
+        StringBuilder b = new StringBuilder();
```
artem-zinnatullin: nit: you can set some initial capacity to save allocations

src/main/java/rx/observers/TestObserver.java
```diff
@@ -146,22 +145,53 @@ public void assertReceivedOnNext(List<T> items) {
...
+        
+        if (onCompletedEvents.isEmpty()) {
+            b.append("(active) ");
```
artem-zinnatullin: Maybe better `(not completed)`?

src/main/java/rx/observers/TestObserver.java
akarnokd: On an unit test failure report?

src/main/java/rx/observers/TestObserver.java
artem-zinnatullin: And probably add number of `onCompleted` events like you did with errors

src/main/java/rx/observers/TestObserver.java
```diff
@@ -146,22 +145,53 @@ public void assertReceivedOnNext(List<T> items) {
...
+        }
+        
+        b.append(message);
+        
+        if (!onErrorEvents.isEmpty()) {
+            b.append(" (+ ")
+            .append(onErrorEvents.size())
+            .append(" errors)");
```
artem-zinnatullin: nit: `onErrorEvents.size() == 1 ? "error" : "errors"`

artem-zinnatullin:
This is great!

src/main/java/rx/observers/TestObserver.java
artem-zinnatullin: oh damn lol k u good

artem-zinnatullin: #perfmatters

src/main/java/rx/observers/TestObserver.java
```diff
@@ -146,22 +145,53 @@ public void assertReceivedOnNext(List<T> items) {
...
+            assertionError("No terminal events received.");
         }
     }
 
+    /**
+     * Combines an assertion error message with the current completion and error state of this
+     * TestSubscriber, giving more information when some assertXXX check fails.
+     * @param message the message to use for the error
+     */
+    final void assertionError(String message) {
```
artem-zinnatullin: `TestObserver` needs tests too

akarnokd: `assertValues` go to `testObserver.assertReceivedOnNext()`

src/test/java/rx/observers/TestSubscriberTest.java
```diff
@@ -610,9 +612,105 @@ public void assertValuesShouldThrowIfNumberOfItemsDoesNotMatch() {
...
+        ts.onNext("c");
+        ts.onError(new TestException("forced failure"));
+
+        try {
+            ts.assertValues("1", "2");
+            fail();
+        } catch (AssertionError expected) {
+            assertEquals("Number of items does not match. Provided: 2  Actual: 3.\n" +
+                    "Provided values: [1, 2]\n" +
+                    "Actual values: [a, b, c] (0 completions) (+1 errors)",
```
artem-zinnatullin: In case of long `Actual values: []` number of completions and errors may not be visible enough for the developer, let's move them on new line?

akarnokd: Okay.

src/main/java/rx/observers/TestObserver.java
artem-zinnatullin: Sure… but it's implementation detail…

akarnokd:
Added newline and plural/singular form.

zsxwing:
👍 this is really awesome.

PR #3935, opened 2016-05-12
Clean-up of the 2.x Design.md document

stevegury:
I went through all the pending comments that weren't addressed.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -353,7 +353,7 @@ The 2.x version will target JDK6+ to let Android users consume the new version o
 
 ### Future work
 
-This section contains current design work which needs more discussion and elaboration before it is merged into this document as a stated goal for 2.x. Our goal is
+This section contains current design work which needs more discussion and elaboration before it is merged into this document as a stated goal for 2.x.
 
 #### Custom Observable, Single, Completable, or Flowable
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -61,7 +61,7 @@ Examples:
 
 ##### Async Pull (Async Interactive)
 
-Consumer requests data when it wishes, and the data is then pushed when the producer wishes to. 
+Consumer requests data when it wishes, and the data is then pushed when the producer wishes to.
 
 Examples:
 
@@ -90,7 +90,7 @@ Examples:
 
 Containing object immediately start work when it is created.
 
-Examples: 
+Examples:
 
 - A `Future` once created has work being performed and represents the eventual value of that work. It can not be deferred once created.
 
@@ -171,13 +171,6 @@ You get a flowable from:
 class Flowable<T> implements Flow.Publisher<T>, io.reactivestreams.Publisher<T> {
   void subscribe(Subscriber<T> subscriber);
 
-  interface Subscriber<T> implements Flow.Subscriber<T>, io.reactivestreams.Subscriber<T> {
-    void onNext(T t);
-    void onError(Throwable t);
-    void onComplete();
-    void onSubscribe(Flowable.Subscription d);
-  }
-
   interface Subscription implements Flow.Subscription, io.reactivestreams.Subscription {
     void cancel();
     void request(long n);
@@ -212,12 +205,12 @@ Flow control:
 ```java
 class Single<T> {
   void subscribe(Single.Subscriber<T> subscriber);
+}
 
-  interface Subscriber<T> {
-    void onSuccess(T t);
-    void onError(Throwable t);
-    void onSubscribe(Disposable d);
-  }
+interface SingleSubscriber<T> {
+  void onSuccess(T t);
+  void onError(Throwable t);
+  void onSubscribe(Disposable d);
 }
 ```
 
@@ -228,7 +221,7 @@ class Single<T> {
 
 Lazy representation of a unit of work that can complete or fail
 
-- Semantic equivalent of `Observable.empty().doOnSubscribe()`. 
+- Semantic equivalent of `Observable.empty().doOnSubscribe()`.
 - Alternative for scenarios often represented with types such as `Single<Void>` or `Observable<Void>`.
 
 Usable for:
@@ -241,12 +234,12 @@ Usable for:
 ```java
 class Completable {
   void subscribe(Completable.Subscriber subscriber);
+}
 
-  interface Subscriber {
-    void onComplete();
-    void onError(Throwable t);
-    void onSubscribe(Disposable d);
-  }
+interface CompletableSubscriber {
+  void onComplete();
+  void onError(Throwable t);
+  void onSubscribe(Disposable d);
 }
 ```
 
@@ -272,7 +265,7 @@ Interactive consumer of events (with consumer-driven flow control). Involved in 
 
 [Reactive Streams state](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#3-subscription-code) of subscription supporting flow control and cancellation.
 
-`Disposable` is a similar type used for lifecycle management on the `Observable` type without interactive flow control. 
+`Disposable` is a similar type used for lifecycle management on the `Observable` type without interactive flow control.
 
 ##### Processor
 
@@ -298,7 +291,7 @@ Subject s = PublishSubject.create();
 Publisher p = s.toPublisher(onBackpressureStrategy);
 
 // now the request(n) semantics are handled by default
-p.subscribe(subscriber1); 
+p.subscribe(subscriber1);
 p.subscribe(subscriber2);
 ```
 
@@ -307,11 +300,11 @@ In this example, `subscriber1` and `subscriber2` can consume at different rates,
 
 ##### Disposable
 
-A type representing work that can be cancelled or disposed.
+A type representing work or resource that can be cancelled or disposed.
 
 Examples:
 
-- An `Observable.subscribe` passes a `Disposable` to the `Observable.onSubscribe` to allow the `Observer` to dispose of the subscription. 
+- An `Observable.subscribe` passes a `Disposable` to the `Observable.onSubscribe` to allow the `Observer` to dispose of the subscription.
 - A `Scheduler` returns a `Disposable` that you use for disposing of the `Scheduler`.
 
 `Subscription` is a similar type used for lifecycle management on the `Flowable` type with interactive flow control.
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -356,27 +349,15 @@ The final `subscribe` will *not* invoke `cancel`/`dispose` after receiving an `o
...
-
-The intent is to let Android users consume the new version of RxJava.
-(Not sure if JDK6 is the appropriate requirement, JDK7+ is maybe sufficient).
+The 2.x version will target JDK6+ to let Android users consume the new version of RxJava.
 
-### Future work 
+### Future work
 
-This section contains current design work which needs more discussion and elaboration before it is merged into this document as a stated goal for 2.x. Our goal is 
+This section contains current design work which needs more discussion and elaboration before it is merged into this document as a stated goal for 2.x. Our goal is
```
artem-zinnatullin: Not sure about `Our goal is`…? Next header is continuation and goal or it's not-ended sentence?

artem-zinnatullin:
👍

DESIGN.md
```diff
@@ -61,7 +61,7 @@ Examples:
 
 ##### Async Pull (Async Interactive)
 
-Consumer requests data when it wishes, and the data is then pushed when the producer wishes to. 
+Consumer requests data when it wishes, and the data is then pushed when the producer wishes to.
```
akarnokd: , but never more than the outstanding request amount till that point.

akarnokd:
:+1:

Do you want me to write a distilled version of the operator-fusion in a follow-up PR?

DESIGN.md
```diff
@@ -356,27 +349,15 @@ The final `subscribe` will *not* invoke `cancel`/`dispose` after receiving an `o
...
+We are investigate a base interface (similar to `Publisher`) for the `Observable`, `Single`, and `Completable` (currently referred to as `Consumable` or `ConsumableObservable`). This would empower library owners and api developers to implement their own type of `Observable`, `Single`, or `Completable` without extending the class. This would result in a change the type signatures of `subscribe` as well as any operator that operates over an `Observable`, `Single`, or `Completable` to accept a more generic type (i.e. `ConsumableObservable`). For more information see the proof of concept project [Consumable](https://github.com/stealthcode/Consumable).
```
akarnokd: Note that these require RS-level-of-detail specification. In addition, I'm not convinced this flexibility has to be necessarily in RxJava. They could live in an interface only library (but that means +1 dependency for RxJava).

As a comparison, Reactor-Core defined its `Mono` type on top of `Publisher` with patterns `onNext+onComplete` | `onError` | `onComplete`. Being `Publisher` itself, one can use it as input to any `Publisher`-requiring operator.

In RxJava design, `Single` and `Completable` live separately and convenience overloads are required to, for example, `flatMap` over a sequence of `Completable`s. It is possible to redefine `Single` in terms of the RS protocol but `Completable` can't be `Publisher<Void>` or `Publisher<Object>` either because the Java type system is not powerful enough to consider `Void` as a subtype of all types (i.e., `? super Void` would match `Integer`, `String`, etc.).

stealthcode: I support the idea of having a separate interface only library. Forcing all semantics into the `Publisher` model is not ideal. 

I have a task to write up more elaboration to the consumable ideas in a separate PR. I suggest that we address this question in that PR. If necessary then delete this section for me to rewrite and propose separately.

akarnokd:
I'd also like to change the extension points:

Unlike RxJava 1.x, 2.x base classes are abstract and stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes. 

Since Java still doesn't have extension methods, "adding" more operators can only happen through helper methods such as `lift(C -> C)` and `compose(R -> P)` where `C` is the default consumer type (i.e., `rs.Subscriber`), `R` is the base type (i.e., `Flowable`) and `P` is the base interface (i.e., `rs.Publisher`). As before, the library itself may gain or lose standard operators and/or overloads through the same community process.

In concert, `create(OnSubscribe)` will not be available; standard operators extend the base types directly. The conversion of other RS-based libraries will happen through the `Flowable.wrap()` static method. 

(*The unfortunate effect of `create` in 1.x was the ignorance of the Observable contract and beginner's first choice as an entry point. We can't eliminate this path since `rs.Publisher` is a single method functional interface that can be implemented just as badly.*)

Therefore, new standard factory methods will try to address the common entry point requirements: 
  - `create(SyncOnSubscribe)` to safe, synchronous generation of signals, one-by-one
  - `create(AsyncOnSubscribe)` to batch-create signals based on request patterns (I'm still not convinced)
  - `create(SingleEmitter)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
  -  `create(FlowEmitter)` to relay multiple values or error from multi-valued reactive-sources (i.e., button-clicks) while also give flow control options right there (buffer, drop, error, etc.).

By extending the base classes, operator implementations would loose the tracking/wrapping features of 1.x. To avoid this, the methods `subscribe(C)` will be final and operators have to implement a protected `subscribeActual` (or any other reasonable name).

```java
@Override
public final void subscribe(Subscriber<? super T> s) {
   subscribeActual(hook.onSubscribe(s));
} 

protected abstract void subscribeActual(Subscriber<? super T> s);
```

Assembly-time hooks will be moved into the individual standard methods on the base types:

```java
public final Flowable<R> map(Function<? super T, ? extends R> mapper) {
    return hook.onAssembly(new FlowableMap<T, R>(this, mapper));
```

stealthcode:
> "adding" more operators can only happen through helper methods

I do not necessarily agree with this. Yes they provide a very easy way to lift in custom operators. However as I have shown with the `Consumable` API, the `Consumable<S2> Consumable<S1>#extend` method provides a way to switch context to a different kind of consumable. This can be used to isolate various different special case operators that someone can switch to as necessary for their use case.

DESIGN.md
stevegury: No sure either, I'm removing it.

stevegury:
Unless there are any controversial points, I would like to merge this as is.
@akarnokd please add a little something about fusion.

akarnokd:
#### Fusion

Operator fusion exploits the declarative nature of building flows; the developer specifies the "what", "where" and "when", the library then tries to optimize the "how".

There are two main levels of operator fusion: *macro* and *micro*.

##### Macro-fusion

Macro fusion deals with the higher level view of the operators, their identity and their combination (mostly in the form of subsequence). This is partially an internal affair of the operators, triggered by the downstream operator and may work with several cases. Given an operator application pair `a().b()` where `a` could be a source or an intermediate operator itself, when the application of `b` happens in assembly time, the following can happen:

  - `b` identifies `a` and decides to not apply itself. Example: `empty().flatMap()` is functionally a no-op
  - `b` identifies `a` and decides to apply a different, conventional operator. Example: `just().subscribeOn()` is turned into `just().observeOn()`.
  - `b` decides to apply a new custom operator, combining and inlining existing behavior. Example: `just().subscribeOn()` internally goes to `ScalarScheduledPublisher`.
  - `a` is `b` and the two operator's parameter set can be combined into a single application. Example: `filter(p1).filter(p2)` combined into `filter(p1 && p2)`

Participating in the macro-fusion externally is possible by implementing a marker interface when extending `Flowable`. Two kinds of interfaces are available: 

  - `java.util.Callable`: the Java standard, throwing interface, indicating the single value has to be extracted in subscription time (or later).
  - `ScalarCallable`: to indicate the single value can be safely extracted during assembly time and used/inlined in other operators:

```java
interface ScalarCallable<T> extends java.util.Callable<T> {
    @Override
    T call();
}
```

`ScalarCallable` is also `Callable` and thus its value can be extracted practically anytime. For convenience (and for sense), `ScalarCallable` overrides and hides the superclass' `throws Exception` clause - throwing during assembly time is likely unreasonable for scalars.

Since Reactive-Streams doesn't allow `null`s in the value flow, we have the opportunity to define `ScalarCallable`s and `Callable`s returning `null` should be considered as an empty source - allowing operators to dispatch on the type `Callable` first then branch on the nullness of `call()`.

Interoperating with other libraries, at this level is possible. Reactor-Core uses the same pattern and the two libraries can work with each other's `Publisher+Callable` types. Unfortunately, this means subscription-time only fusion as `ScalarCallable`s live locally in each library.

##### Micro-fusion

Micro-fusion goes a step deeper and tries to reuse internal structures, mostly queues, in operator pairs, saving on allocation and sometimes on atomic operations. It's property is that, in a way, subverts the standard Reactive-Streams protocol between subsequent operators that both support fusion. However, from the outside world's view, they still work according to the RS protocol.

Currently, two main kinds of micro-fusion opportunities are available.

###### 1) Conditional Subscriber

This extends the RS `Subscriber`interface with an extra method: `boolean tryOnNext(T value)` and can help avoiding small request amounts in case an operator didn't forward but dropped the value. The canonical use is for the `filter()` operator where if the predicate returns false, the operator has to request 1 from upstream (since the downstream doesn't know there was a value dropped and thus not request itself). Operators wanting to participate in this fusion have to implement and subscribe with an extended Subscriber interface:

```java
interface ConditionalSubscriber<T> {
   boolean tryOnNext(T value);
}

//...
@Override
protected void subscribeActual(Subscriber<? super T> s) {
    if (s instanceof ConditionalSubscriber) {
        source.subscribe(new FilterConditionalSubscriber<>(s, predicate));
    } else {
        source.subscribe(new FilterRegularSubscriber<>(s, predicate));
    }
}
```

(Note that this may lead to extra case-implementations in operators that have some kind of queue-drain emission model.)

###### 2) Queue-fusion

The second category is when two (or more) operators share the same underlying queue and each append activity at the exit point (i.e., poll()) of the queue. This can work in two modes: synchronous and asynchronous.

In synchronous mode, the elements of the sequence is already available (i.e., a fixed `range()` or `fromArray()`, or can be synchronously calculated in a pull fashion in `fromIterable`. In this mode, the requesting and regular onError-path is bypassed and is forbidden. Sources have to return null from `pull()` and false from `isEmpty()` if they have no more values and throw from these methods if they want to indicate an exceptional case.

In asynchronous mode, elements may become available at any time, therefore, `pull` returning null, as with regular queue-drain, is just the indication of temporary lack of source values. Completion and error still has to go through `onComplete` and `onError` as usual, requesting still happens as usual but when a value is available in the shared queue, it is indicated by an `onNext(null)` call. This can trigger a chain of `drain` calls without moving values in or out of different queues.

In both modes, `cancel` works and behaves as usual.

Since this fusion mode is an optional extension, the mode switch has to be negotiated and the shared queue interface established. Operators already working with internal queues then can, mostly, keep their current `drain()` algorithm. Queue-fusion has its own interface and protocol built on top of the existing `onSubscribe`-`Subscription` rail:

```java
interface QueueSubscription<T> implements Queue<T>, Subscription {
    int NONE = 0;
    int SYNC = 1;
    int ASYNC = 2;
    int ANY = SYNC | ASYNC;
    int BOUNDARY = 4;

    int requestFusion(int mode);
}
```

For performance, the mode is an integer bitflags setup, called early during subscription time, and allows negotiating the fusion mode. Usually, producers can do only one mode and consumers can do both mode. Because fused, intermediate operators attach logic (which is many times user-callback) to the exit point of the queue interface (poll()), it may change the computation location of those callbacks in an unwanted way. The flag `BOUNDARY` is added by consumers indicating that they will consume the queue over an async boundary. Intermediate operators, such as `map` and `filter` then can reject the fusion in such sequences.

Since RxJava 2.x is still JDK 6 compatible, the `QueueSubscription` can't itself default unnecessary methods and implementations are required to throw `UnsupportedOperationException` for `Queue` methods other than the following:

  - `poll()`
  - `isEmpty()`
  - `clear()`
  - `size()`

Even though other modern libraries also define this interface, they live in local packages and thus non-reusable without dragging in the whole library. Therefore, until externalized and standardized, cross-library micro-fusion won't happen.

A consequence of the extension of the `onSubscribe`-`Subscription` rail is that intermediate operators are no longer allowed to pass an upstream `Subscription` directly to its downstream `Subscriber.onSubscribe`. Doing so is likely to have the fused sequence skip the operator completely, losing behavior or causing runtime exceptions. Since RS `Subscriber` is an interface, operators can simply implement both `Subscriber` and `Subscription` on themselves, delegating the `request` and `cancel` calls to the upstream and calling `child.onSubscribe(this)`.

smaldini:
It is to be noted that since Reactor 2.5 uses the fusion described above, it would be interesting to envision a cross library fusion contract , as part of RS directly or separately. We already benefit from small cross-library optimization thanks to @akarnokd already (the first of this kind,  scalar single to scalar reactor mono).

akarnokd:
@stevegury do you want to include my text in this PR or merge it as is?

akarnokd:
:+1:

DESIGN.md
```diff
@@ -171,13 +171,6 @@ You get a flowable from:
 class Flowable<T> implements Flow.Publisher<T>, io.reactivestreams.Publisher<T> {
   void subscribe(Subscriber<T> subscriber);
 
-  interface Subscriber<T> implements Flow.Subscriber<T>, io.reactivestreams.Subscriber<T> {
-    void onNext(T t);
-    void onError(Throwable t);
-    void onComplete();
-    void onSubscribe(Flowable.Subscription d);
-  }
-
```
benjchristensen, line 180: The document no longer defines `Flowable.Subscriber` anywhere. Is that because it is intended to just be the ReactiveStreams `Subscriber` so isn't being specifically called out? If so, perhaps the fully qualified class name should be used to be explicit about that?

benjchristensen, line 180: I noticed `Observable.Observer` is also not specified anywhere.

stevegury:
Thanks for merging this in @akarnokd.
Sorry for the lack of activity, I've been pretty busy lately on other topics.

PR #3936, opened 2016-05-13
1.x: promote UnicastSubject to be a standard+experimental Subject

akarnokd:
Plus, the unsubscribe indirection has been inlined, making `State` implement `Subscription` directly (instead of `Action0` + `Subscriptions.create`).

### DIFF
--- a/src/main/java/rx/subjects/UnicastSubject.java
+++ b/src/main/java/rx/subjects/UnicastSubject.java
@@ -13,18 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package rx.internal.operators;
+package rx.subjects;
 
 import java.util.Queue;
 import java.util.concurrent.atomic.*;
 
 import rx.*;
+import rx.annotations.Experimental;
 import rx.exceptions.*;
-import rx.functions.*;
+import rx.functions.Action0;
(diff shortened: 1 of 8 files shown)


src/main/java/rx/subjects/UnicastSubject.java
```diff
@@ -250,7 +251,7 @@ public void request(long n) {
         @Override
         public void call(Subscriber<? super T> subscriber) {
             if (this.subscriber.compareAndSet(null, subscriber)) {
-                subscriber.add(Subscriptions.create(this));
+                subscriber.add(this);
```
JakeWharton: Nice 👍

src/main/java/rx/subjects/UnicastSubject.java
```diff
@@ -35,6 +35,7 @@
  * 
  * @param <T> the input and output value type
  */
+@Experimental
```
artem-zinnatullin: We can't promote it for public usage without tests!

akarnokd: It is tested through `window` operators and there is `BufferUntilSubscriberTest`.

zsxwing:
:+1:

stevegury:
👍

PR #3937, opened 2016-05-13
1.x: Change the workers to capture the stack trace

abersnaze:
Watched @dlew presentation of the pit falls of debugging RxJava https://www.youtube.com/watch?v=QdmkXL7XikQ&feature=youtu.be&t=38m12s

This change is to create an exception when a thread based `Scheduler.Worker` is constructed and reused it for all subsequent scheduled actions to increase the readability of uncaught and fatal exceptions that bubble up to the schedulers but spread the cost out across many scheduled actions.

A future improvement to spread the cost out even more, that I didn't want to do in the first attempt, would to create the exception when `Scheduler.io()` is called and reuse the exception across multiple `Worker`s.

The example from the presentation
```
import rx.Observable;
import rx.schedulers.Schedulers;

public class Main {
    public static void main(String[] args) throws InterruptedException {
        Observable.empty()
                .first()
                .subscribeOn(Schedulers.io())
                .subscribe();

        Thread.sleep(100);
    }
}
```

Used to print this just this exception with no mention the Main class.
```
Exception in thread "RxIoScheduler-2" java.lang.IllegalStateException: Exception thrown on Scheduler.Worker thread. Add `onError` handling.
	at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:65)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: rx.exceptions.OnErrorNotImplementedException: Sequence contains no elements
	at rx.internal.util.InternalObservableUtils$ErrorNotImplementedAction.call(InternalObservableUtils.java:374)
	at rx.internal.util.InternalObservableUtils$ErrorNotImplementedAction.call(InternalObservableUtils.java:1)
	at rx.internal.util.ActionSubscriber.onError(ActionSubscriber.java:44)
	at rx.observers.SafeSubscriber._onError(SafeSubscriber.java:157)
	at rx.observers.SafeSubscriber.onError(SafeSubscriber.java:120)
	at rx.internal.operators.OperatorSubscribeOn$1$1.onError(OperatorSubscribeOn.java:59)
	at rx.internal.operators.OperatorSingle$ParentSubscriber.onCompleted(OperatorSingle.java:116)
	at rx.internal.operators.OperatorTake$1.onCompleted(OperatorTake.java:56)
	at rx.internal.operators.EmptyObservableHolder.call(EmptyObservableHolder.java:44)
	at rx.internal.operators.EmptyObservableHolder.call(EmptyObservableHolder.java:1)
	at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:50)
	at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:1)
	at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:50)
	at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:1)
	at rx.Observable.unsafeSubscribe(Observable.java:8460)
	at rx.internal.operators.OperatorSubscribeOn$1.call(OperatorSubscribeOn.java:94)
	at rx.internal.schedulers.CachedThreadScheduler$EventLoopWorker$1.call(CachedThreadScheduler.java:222)
	at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:60)
	... 7 more
Caused by: java.util.NoSuchElementException: Sequence contains no elements
	... 19 more
```

But will now print the exception above but with an additional caused by
```
Caused by: rx.internal.schedulers.SchedulerContextException: Asynchronous work scheduled at
	at rx.internal.schedulers.NewThreadWorker.<init>(NewThreadWorker.java:36)
	at rx.internal.schedulers.CachedThreadScheduler$ThreadWorker.<init>(CachedThreadScheduler.java:235)
	at rx.internal.schedulers.CachedThreadScheduler$CachedWorkerPool.get(CachedThreadScheduler.java:86)
	at rx.internal.schedulers.CachedThreadScheduler$EventLoopWorker.<init>(CachedThreadScheduler.java:187)
	at rx.internal.schedulers.CachedThreadScheduler.createWorker(CachedThreadScheduler.java:173)
	at rx.internal.operators.OperatorSubscribeOn.call(OperatorSubscribeOn.java:42)
	at rx.internal.operators.OperatorSubscribeOn.call(OperatorSubscribeOn.java:1)
	at rx.Observable.subscribe(Observable.java:8553)
	at rx.Observable.subscribe(Observable.java:8520)
	at rx.Observable.subscribe(Observable.java:8316)
	at rx.schedulers.Main.main(Main.java:10)
```

### DIFF
--- a/src/main/java/rx/internal/schedulers/SchedulerContextException.java
+++ b/src/main/java/rx/internal/schedulers/SchedulerContextException.java
@@ -1,0 +1,45 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.schedulers;
+
+/**
+ * Used only for providing context around where work was scheduled should an error occur in a different thread.
+ */
+public class SchedulerContextException extends Exception {
+    /**
+     * Constant to use when disabled
+     */
+    private static final Throwable CONTEXT_MISSING = new SchedulerContextException("Missing context. Enable by setting the system property \"rxjava.captureSchedulerContext=true\"");
+
+    static {
+        CONTEXT_MISSING.setStackTrace(new StackTraceElement[0]);
+    }
+
+    /**
+     * @return a {@link Throwable} that captures the stack trace or a {@link Throwable} that documents how to enable the feature if needed.
+     */
+    public static Throwable create() {
+        String def = "false";
+        String setTo = System.getProperty("rxjava.captureSchedulerContext", def);
+        return setTo != def && "true".equals(setTo) ? new SchedulerContextException("Asynchronous work scheduled at") : CONTEXT_MISSING;
+    }
+
+    private SchedulerContextException(String message) {
+        super(message);
+    }
+
+    private static final long serialVersionUID = 1L;
+}
--- a/src/test/java/rx/schedulers/AbstractSchedulerTests.java
+++ b/src/test/java/rx/schedulers/AbstractSchedulerTests.java
@@ -43,6 +43,8 @@ import rx.Subscriber;
 import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.functions.Func1;
+import rx.plugins.RxJavaErrorHandler;
+import rx.plugins.RxJavaPlugins;
 
 /**
  * Base tests for all schedulers including Immediate/Current.
@@ -502,4 +504,54 @@ public abstract class AbstractSchedulerTests {
 
     }
 
+    @Test
+    public final void testStackTraceAcrossThreads() throws Throwable {
+        final AtomicReference<Throwable> exceptionRef = new AtomicReference<Throwable>();
+        final CountDownLatch done = new CountDownLatch(1);
+        System.setProperty("rxjava.captureSchedulerContext", "true");
+
+        try {
+
+            RxJavaPlugins.getInstance().reset();
+            RxJavaPlugins.getInstance().registerErrorHandler(new RxJavaErrorHandler() {
+                @Override
+                public void handleError(Throwable e) {
+                    exceptionRef.set(e);
+                    done.countDown();
+                }
+            });
+
+            try {
+                getScheduler().createWorker().schedule(new Action0() {
+                    @Override
+                    public void call() {
+                        throw new RuntimeException();
+                    }
+                });
+            } catch (Exception e) {
+                exceptionRef.set(e);
+                done.countDown();
+            }
+
+            done.await();
+
+            Throwable exception = exceptionRef.get();
+            Throwable e = exception;
+            while (e.getCause() != null) {
+                e = e.getCause();
+            }
+
+            StackTraceElement[] st = e.getStackTrace();
+            for (StackTraceElement stackTraceElement : st) {
+                if (stackTraceElement.getMethodName().equals("testStackTraceAcrossThreads")) {
+                    // pass we found this class in the stack trace.
+                    return;
+                }
+            }
+
+            throw exception;
+        } finally {
+            System.setProperty("rxjava.captureSchedulerContext", "false");
+        }
+    }
 }
--- a/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
+++ b/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
@@ -78,6 +78,7 @@ public final class CachedThreadScheduler extends Scheduler implements SchedulerL
             while (!expiringWorkerQueue.isEmpty()) {
                 ThreadWorker threadWorker = expiringWorkerQueue.poll();
                 if (threadWorker != null) {
+                    threadWorker.resetContext();
                     return threadWorker;
                 }
             }
--- a/src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
+++ b/src/main/java/rx/internal/schedulers/EventLoopsScheduler.java
@@ -142,7 +142,7 @@ public final class EventLoopsScheduler extends Scheduler implements SchedulerLif
 
         EventLoopWorker(PoolWorker poolWorker) {
             this.poolWorker = poolWorker;
-            
+            poolWorker.resetContext();
         }
 
         @Override
--- a/src/main/java/rx/internal/schedulers/ExecutorScheduler.java
+++ b/src/main/java/rx/internal/schedulers/ExecutorScheduler.java
@@ -42,6 +42,7 @@ public final class ExecutorScheduler extends Scheduler {
 
     /** Worker that schedules tasks on the executor indirectly through a trampoline mechanism. */
     static final class ExecutorSchedulerWorker extends Scheduler.Worker implements Runnable {
+        private final Throwable creationContext = SchedulerContextException.create();
         final Executor executor;
         // TODO: use a better performing structure for task tracking
         final CompositeSubscription tasks;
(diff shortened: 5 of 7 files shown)


akarnokd:
No. At least make it conditional to a system parameter / global boolean variable.

stealthcode:
👍  as is. The cost is very low and the benefit is very high.

abersnaze:
@akarnokd added a system property to make if off by default.

stealthcode:
I would like this to be default behavior. Learning Rx is hard and we are making it harder by default. Let users opt into over-optimizations.

src/main/java/rx/internal/schedulers/SchedulerContextException.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package rx.internal.schedulers;
+
+/**
+ * Used only for providing context around where work was scheduled should an error occur in a different thread.
+ */
+public class SchedulerContextException extends Exception {
+    public static Throwable create() {
+        if (System.getProperty("rxjava.captureSchedulerContext", "false").equals("true")) {
```
akarnokd: Better. Do you want this to be changeable at runtime?

abersnaze: Yes. When things go wrong it can be helpful to be able to enable more logging without having to restart the VM.

akarnokd: Okay, in that case, I'd consider a public static AtomicBoolean instead of the hash-lookup and string comparison. From security perspective, if you can set System properties, you could just as set an AtomicBoolean, right?

JakeWharton: Probably just a static volatile boolean, right? Atomicity isn't needed
since it's only read by RxJava and only written by application code.

abersnaze: anything else @JakeWharton @akarnokd?

akarnokd:
You have a lot of test failures.

src/main/java/rx/internal/schedulers/SchedulerContextException.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (Exception e) {
+            captureSchedulerContext = false;
+        }
+    }
+
+    /**
+     * Immediately change the behavior to of the {@link SchedulerContextException#create()} method.
+     * @param captureSchedulerContext
+     */
+    public static void setCaptureSchedulerContext(boolean captureSchedulerContext) {
```
JakeWharton: Is this meant to be public API for applications to use? Because it's currently in an internal package.

src/main/java/rx/internal/schedulers/SchedulerContextException.java
```diff
@@ -0,0 +1,63 @@
...
+        try {
+            captureSchedulerContext = System.getProperty("rxjava.captureSchedulerContext", "false").equals("true");
+        } catch (Exception e) {
+            captureSchedulerContext = false;
+        }
+    }
+
+    /**
+     * Immediately change the behavior to of the {@link SchedulerContextException#create()} method.
+     * @param captureSchedulerContext
```
JakeWharton: Delete `@param`

src/main/java/rx/internal/schedulers/SchedulerContextException.java
```diff
@@ -0,0 +1,63 @@
...
+
+        try {
+            captureSchedulerContext = System.getProperty("rxjava.captureSchedulerContext", "false").equals("true");
+        } catch (Exception e) {
+            captureSchedulerContext = false;
+        }
+    }
+
+    /**
+     * Immediately change the behavior to of the {@link SchedulerContextException#create()} method.
```
JakeWharton: remove "to"

robertroeser:
RXJava is hard to debug - lots of people agree on this. I know a of large non-trivial projects that avoided using RXJava because it hard to debug. Non-trivial projects do I/O, and serialization - those are the performance problems, not micro-optimizations. Users will care more about it being easier to debug than being a micro-optimized. This should be default to make it less hard.

src/main/java/rx/internal/schedulers/SchedulerContextException.java
abersnaze: It wasn't an API until it was changed to `volatile boolean` and was just always reading from `System.getProperty()`.

stealthcode: This seems unnecessary to me. `System#getProperty` checks the security manager. If the security manager throws an exception then the boolean is set to false (making it impossible to set to true). 

I would rather we just read from a property always. Why wouldn't we do this?

abersnaze:
I've changed it back to always getting the value from the system properties because makes a fine public API and it makes the tests work again.  I've added some code to mitigate the overhead of doing the check multiple times.

akarnokd:
Still failing...

abersnaze:
Found it the test worked if run in isolation but failed if they were run en masse because of caching. Added pooled work resets in two places seems to fix the issue.

src/test/java/rx/schedulers/AbstractSchedulerTests.java
```diff
@@ -502,4 +504,54 @@ public void onNext(T args) {
...
+            RxJavaPlugins.getInstance().registerErrorHandler(new RxJavaErrorHandler() {
+                @Override
+                public void handleError(Throwable e) {
+                    exceptionRef.set(e);
+                    done.countDown();
+                }
+            });
+
+            try {
+                getScheduler().createWorker().schedule(new Action0() {
```
akarnokd, line 525: This leaks the worker.

src/main/java/rx/internal/schedulers/SchedulerContextException.java
```diff
@@ -0,0 +1,45 @@
...
+    static {
+        CONTEXT_MISSING.setStackTrace(new StackTraceElement[0]);
+    }
+
+    /**
+     * @return a {@link Throwable} that captures the stack trace or a {@link Throwable} that documents how to enable the feature if needed.
+     */
+    public static Throwable create() {
+        String def = "false";
+        String setTo = System.getProperty("rxjava.captureSchedulerContext", def);
```
akarnokd, line 36: I'm not experienced with it, but `AccessController.doPrivileged` could be used for defending the setting of the volatile boolean field.

abersnaze, line 36: Is this about setting up a public API to toggle the features?  I am considering adding the method to the plugin class or the execution hooks.

akarnokd, line 36: `doPreveleged` only penalizes the mode change and not this `create()` method.

akarnokd:
I want to increase debug support in RxJava, but I'd think, based on the feedback on #3965, that such capture is not enough.

I suggest closing this PR, unless you can spend some time with it before 1.1.7. Also I'd like your feedback on #4007. It's tracking grabs the stacktrace for `observeOn` as well (the triggerer of the Scheduled actions) and should help with non-fatal crashes over async boundaries.

A more elaborate solution would be that in case of a crash, the operator graph could be walked and dumped in a consumable format. Unfortunately, to support this, the `Subscriber` class has to be extended with methods to peek into the upstream and all operators have to set a `Producer` on their child subscribers to have the bidirectional link for the graph. Since `Subscriber` is public API, adding methods to it may not be a compatible change. For 2.x, there is no problem.

akarnokd:
I'm closing this for now due to test failures and merge conflicts.

In addition, I suggest extending #4007, once merged, by adding `Scheduler` wrappers - similar to the OnAssembly wrappers, via the hooks to capture and signal the stacks.

If you have further input on the issue, don't hesitate to reopen this issue or post a new one.

PR #3942, opened 2016-05-14
1.x: add Completable.safeSubscribe option + RxJavaPlugins hook support

akarnokd:
Add option to safely subscribe a `CompletableSubscriber` / regular `Subscriber` and handle `onXXX` failures.

See also: #3938

Naming and whether or not the safe wrapping should be the default is open to discussion.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1932,7 +1932,7 @@ public class Completable {
                     try {
                         onComplete.call();
                     } catch (Throwable e) {
-                        onError(e);
+                        callOnError(e);
                         return;
                     }
                     mad.unsubscribe();
@@ -1943,18 +1943,22 @@ public class Completable {
             public void onError(Throwable e) {
                 if (!done) {
                     done = true;
-                    try {
-                        onError.call(e);
-                    } catch (Throwable ex) {
-                        e = new CompositeException(Arrays.asList(e, ex));
-                        ERROR_HANDLER.handleError(e);
-                        deliverUncaughtException(e);
-                    } finally {
-                        mad.unsubscribe();
-                    }
+                    callOnError(e);
                 } else {
                     ERROR_HANDLER.handleError(e);
                     deliverUncaughtException(e);
+                }
+            }
+            
+            void callOnError(Throwable e) {
+                try {
+                    onError.call(e);
+                } catch (Throwable ex) {
+                    e = new CompositeException(Arrays.asList(e, ex));
+                    ERROR_HANDLER.handleError(e);
+                    deliverUncaughtException(e);
+                } finally {
+                    mad.unsubscribe();
                 }
             }
             
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -173,7 +173,7 @@ public class Completable {
                     }
                     
                     // no need to have separate subscribers because inner is stateless
-                    c.subscribe(inner);
+                    c.unsafeSubscribe(inner);
                 }
             }
         });
@@ -301,7 +301,7 @@ public class Completable {
                     }
                     
                     // no need to have separate subscribers because inner is stateless
-                    c.subscribe(inner);
+                    c.unsafeSubscribe(inner);
                 }
             }
         });
@@ -416,7 +416,7 @@ public class Completable {
                     return;
                 }
                 
-                c.subscribe(s);
+                c.unsafeSubscribe(s);
             }
         });
     }
@@ -899,7 +899,7 @@ public class Completable {
                 
                 final AtomicBoolean once = new AtomicBoolean();
                 
-                cs.subscribe(new CompletableSubscriber() {
+                cs.unsafeSubscribe(new CompletableSubscriber() {
                     Subscription d;
                     void dispose() {
                         d.unsubscribe();
@@ -999,7 +999,7 @@ public class Completable {
         final CountDownLatch cdl = new CountDownLatch(1);
         final Throwable[] err = new Throwable[1];
         
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
 
             @Override
             public void onCompleted() {
@@ -1050,7 +1050,7 @@ public class Completable {
         final CountDownLatch cdl = new CountDownLatch(1);
         final Throwable[] err = new Throwable[1];
         
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
 
             @Override
             public void onCompleted() {
@@ -1190,7 +1190,7 @@ public class Completable {
                 final Scheduler.Worker w = scheduler.createWorker();
                 set.add(w);
                 
-                subscribe(new CompletableSubscriber() {
+                unsafeSubscribe(new CompletableSubscriber() {
 
                     
                     @Override
@@ -1302,7 +1302,7 @@ public class Completable {
         return create(new CompletableOnSubscribe() {
             @Override
             public void call(final CompletableSubscriber s) {
-                subscribe(new CompletableSubscriber() {
+                unsafeSubscribe(new CompletableSubscriber() {
 
                     @Override
                     public void onCompleted() {
@@ -1434,7 +1434,7 @@ public class Completable {
         final CountDownLatch cdl = new CountDownLatch(1);
         final Throwable[] err = new Throwable[1];
         
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
 
             @Override
             public void onCompleted() {
@@ -1478,7 +1478,7 @@ public class Completable {
         final CountDownLatch cdl = new CountDownLatch(1);
         final Throwable[] err = new Throwable[1];
         
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
 
             @Override
             public void onCompleted() {
@@ -1530,7 +1530,7 @@ public class Completable {
 
                     CompletableSubscriber sw = onLift.call(s);
                     
-                    subscribe(sw);
+                    unsafeSubscribe(sw);
                 } catch (NullPointerException ex) {
                     throw ex;
                 } catch (Throwable ex) {
@@ -1571,7 +1571,7 @@ public class Completable {
                 
                 s.onSubscribe(ad);
                 
-                subscribe(new CompletableSubscriber() {
+                unsafeSubscribe(new CompletableSubscriber() {
 
                     @Override
                     public void onCompleted() {
@@ -1633,7 +1633,7 @@ public class Completable {
         return create(new CompletableOnSubscribe() {
             @Override
             public void call(final CompletableSubscriber s) {
-                subscribe(new CompletableSubscriber() {
+                unsafeSubscribe(new CompletableSubscriber() {
 
                     @Override
                     public void onCompleted() {
@@ -1682,7 +1682,7 @@ public class Completable {
             @Override
             public void call(final CompletableSubscriber s) {
                 final SerialSubscription sd = new SerialSubscription();
-                subscribe(new CompletableSubscriber() {
+                unsafeSubscribe(new CompletableSubscriber() {
 
                     @Override
                     public void onCompleted() {
@@ -1708,7 +1708,7 @@ public class Completable {
                             return;
                         }
                         
-                        c.subscribe(new CompletableSubscriber() {
+                        c.unsafeSubscribe(new CompletableSubscriber() {
 
                             @Override
                             public void onCompleted() {
@@ -1844,7 +1844,7 @@ public class Completable {
      */
     public final Subscription subscribe() {
         final MultipleAssignmentSubscription mad = new MultipleAssignmentSubscription();
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
             @Override
             public void onCompleted() {
                 mad.unsubscribe();
@@ -1877,7 +1877,7 @@ public class Completable {
         requireNonNull(onComplete);
         
         final MultipleAssignmentSubscription mad = new MultipleAssignmentSubscription();
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
             @Override
             public void onCompleted() {
                 try {
@@ -1919,7 +1919,7 @@ public class Completable {
         requireNonNull(onComplete);
         
         final MultipleAssignmentSubscription mad = new MultipleAssignmentSubscription();
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
             @Override
             public void onCompleted() {
                 try {
@@ -1963,7 +1963,7 @@ public class Completable {
      * @param s the CompletableSubscriber, not null
      * @throws NullPointerException if s is null
      */
-    public final void subscribe(CompletableSubscriber s) {
+    public final void unsafeSubscribe(CompletableSubscriber s) {
         requireNonNull(s);
         try {
             // TODO plugin wrapping the subscriber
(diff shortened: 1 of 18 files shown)


bryant1410:
In my opinion safe wrapping should be the default as it's the default in `Observable` and `Single`. At least the 3 should behave the same way. When using them, one should expect them to have a difference in their essential semantics, and not in these kind of things.

src/main/java/rx/Completable.java
```diff
@@ -1971,8 +1972,29 @@ public final void subscribe(CompletableSubscriber s) {
...
         } catch (Throwable ex) {
+            Exceptions.throwIfFatal(ex);
             ERROR_HANDLER.handleError(ex);
+            throw toNpe(ex);
+        }
+    }
+
+    /**
+     * Subscribes the given CompletableSubscriber to this Completable instance
+     * and handles exceptions throw by its onXXX methods.
```
bryant1410: throw -> thrown

src/main/java/rx/Completable.java
```diff
@@ -1971,8 +1972,29 @@ public final void subscribe(CompletableSubscriber s) {
...
+     * @param s the CompletableSubscriber, not null
+     * @throws NullPointerException if s is null
+     */
+    public final void safeSubscribe(CompletableSubscriber s) {
+        requireNonNull(s);
+        try {
+            // TODO plugin wrapping the subscriber
```
bryant1410: What's the meaning of this TODO? What is missing here?

bryant1410:
Should the tests about error handling plugin in #3938 be here? Or we leave it in that PR?

src/main/java/rx/Completable.java
akarnokd: Your PR has test failures so I didn't look at them in case you fix things in the meantime.

akarnokd: The plugin system is half mistery to me right. I'll figure that out and add the hooks in a separate PR.

akarnokd:
@bryant1410 Your PR has a test failure:

```
rx.plugins.RxJavaPluginsTest > testOnErrorWhenUsingCompletable FAILED
    java.lang.AssertionError: expected:<java.lang.RuntimeException: test onError> but was:<null>
        at org.junit.Assert.assertEquals(Assert.java:147)
        at rx.plugins.RxJavaPluginsTest.testOnErrorWhenUsingCompletable(RxJavaPluginsTest.java:314)
```

akarnokd:
Fixed typo, renamed methods

bryant1410:
@akarnokd They fail on purpose because I made the test but didn't fix the problem (Completable is not calling the error handling plugin).

src/main/java/rx/Completable.java
```diff
@@ -1992,7 +2014,7 @@ public final void subscribe(CompletableSubscriber s) {
                 throw new NullPointerException("The RxJavaPlugins.onSubscribe returned a null Subscriber");
```
bryant1410: I think this can only happen if `s` is `null`, which is checked in `requiredNonNull(s)`. Same in `subscribe(Subscriber)`

bryant1410: Also, why using `sw`? Is just an alias of `s`.

src/main/java/rx/Completable.java
```diff
@@ -1971,19 +1972,40 @@ public final void subscribe(CompletableSubscriber s) {
         } catch (NullPointerException ex) {
             throw ex;
         } catch (Throwable ex) {
+            Exceptions.throwIfFatal(ex);
             ERROR_HANDLER.handleError(ex);
```
bryant1410, line 1995: `unsafeSubscribe` shouldn't call the `ERROR_HANDLER`.

src/main/java/rx/Completable.java
```diff
@@ -2012,6 +2034,49 @@ public void onSubscribe(Subscription d) {
         } catch (NullPointerException ex) {
             throw ex;
         } catch (Throwable ex) {
+            Exceptions.throwIfFatal(ex);
+            ERROR_HANDLER.handleError(ex);
```
bryant1410, line 2059: `unsafeSubscribe` shouldn't call the `ERROR_HANDLER`.

src/main/java/rx/Completable.java
akarnokd: It's there to capture plugin errors once the plugin API is extended.

bryant1410:
The methods should be the other way around. Following `SIngle` and `Observable`, there should be a single `unsafeSubscribe` method, which accepts a parameter of type `CompletableSubscriber`. And `subscribe` should be present in the following forms: `subscribe()`, `subscribe(Action0)`, `subscribe(Action1<? super Throwable>, Action0)`, `subscribe(Subscriber)` and `subscribe(CompletableSubscriber)`.

src/main/java/rx/Completable.java
akarnokd, line 1995: Why not? That exception has to be reported otherwise it could get lost.

bryant1410, line 1995: I see from the javadocs of `Observable#unsafeSubscribe` (and from the code) that it should subscribe to an Observable and call `OnSubscribe` without error handling.

akarnokd:
The `subscribe()` methods wrap with `SafeComletableSubscriber` similar to how Observable.subscribe wraps with `SafeSubscriber`. subscribe(Subscriber) is there to allow conversion and cross-type APIs to bypass the safety overhead.

src/main/java/rx/Completable.java
bryant1410, line 1995: It won't get lost as it's supposed to be called in nested subscriptions, like when implementing operators.

akarnokd, line 1995: In observable, if the subscription fails, the `subscribe()` tries to call onNext on the subscriber. If the subscriber is unsubscribed, the original exception is sent to the plugin handler. If not but crashes further, a custom fatal exception is thrown. In Completable, we can't know if the completable subscriber is ready to receive the exception or not, thus the safe bet is to call the plugin handler and rethrow as NullPointerException. If nested, NPE is thrown as is further up without calling the handler again.

src/main/java/rx/Completable.java
bryant1410: Do you mean once there is an execution hook for `Completable`? Or what do you mean?

src/main/java/rx/Completable.java
bryant1410, line 1995: But this should get compared to `unsafeSubscribe`, not `subscribe`, doesn't it?

akarnokd:
Updated with plugin support.

src/main/java/rx/Completable.java
akarnokd, line 1995: Observable is kind of in an inconsistent state in this regard, really could use a cleanup. The only difference should be that the incoming subscriber is or isn't wrapped. Crash management should be the same.

bryant1410, line 1995: Oh, ok. I also see that it's supposed to not call the hook, but it does. However, couldn't be the case that multiple calls to the error handler are made for the same error (e.g., due to operators)?

akarnokd, line 1995: No, if the exception is fatal, it skips the hook. Once the hook has been called, the exception is wrapped into a NullPointerException and thrown up, but then there is a catch (NullPointerException) to prevent it from calling the hook again.

bryant1410:
Can you cherry-pick the [commit that adds tests for the error handling plugn in `Completable` and `Single`](https://github.com/bryant1410/RxJava/commit/4cc57dcce8603496136d1e24173ce5ff36b7715b) from #3938?

akarnokd:
Sure.

akarnokd:
Done.

src/main/java/rx/observers/SafeCompletableSubscriber.java
```diff
@@ -52,8 +52,8 @@ public void onCompleted() {
 
     @Override
     public void onError(Throwable e) {
+        RxJavaPluginUtils.handleException(e);
```
bryant1410, line 59: This should be done if `!done`, as in `SafeSubscriber`.

akarnokd, line 59: Done or not, the exception has to be reported to the handler. This also makes your test pass.

bryant1410, line 59: But why `SafeSubscriber` doesn't do this? Either this class or the other one are not behaving right.

akarnokd, line 59: Can you demonstrate that with an unit test?

bryant1410, line 59: Do you want me to make a unit test for the case that `onError` gets called twice in a `SafeSubcriber` and `SafeCompletableSubscriber` to see if the error handler gets called twice? I don't know if it's worth it. What it happens is that it just doesn't look good to me when behavior is different in `Completable` from `Single` and `Observable`.

bryant1410:
The comment `// TODO plugin wrapping onSubscribe` in `Completable#create` can be deleted, as the constructor is already calling the hook now.

src/main/java/rx/Completable.java
```diff
@@ -1843,7 +1846,7 @@ public final Completable startWith(Completable other) {
      */
     public final Subscription subscribe() {
         final MultipleAssignmentSubscription mad = new MultipleAssignmentSubscription();
-        subscribe(new CompletableSubscriber() {
+        unsafeSubscribe(new CompletableSubscriber() {
```
bryant1410, line 1847: `SafeCompletableSubscriber` is not used here nor the others `subscribe` methods, except for `subscribe(Subscriber)`. I think `subscribe` methods should call between themselves (as in `Observable#subscribe` methods) to refactor code and avoid forgetting this kind of things.

bryant1410, line 1847: `unsafeSubscribe` should call between themselves too.

src/main/java/rx/Completable.java
```diff
@@ -1962,56 +1965,112 @@ private static void deliverUncaughtException(Throwable e) {
...
                 }
                 
                 @Override
                 public void onSubscribe(Subscription d) {
-                    sw.add(d);
+                    s.add(d);
                 }
             });
-            
+            RxJavaPlugins.getInstance().getObservableExecutionHook().onSubscribeReturn(s);
```
bryant1410, line 2053: Shouldn't the `Completable` hook be used here?

akarnokd, line 2053: The method deals with `Subscriber`.

akarnokd:
Removed comment, changed to reuse methods, added onStart call.

src/main/java/rx/Completable.java
bryant1410, line 1847: This should call `subscribe` now instead, doesn't it? And the others `subscribe` methods.

akarnokd, line 1847: No, the rest are intermediate operators or already perform custom logic in addition to reporting the error to a plugin. Plus, this particulare use case, the subscriber is not expected to crash.

bryant1410, line 1847: But the idea of the safe subscribe is to ensure the caller that the contract will be met. Why in this particular case it won't?

akarnokd, line 1847: No, the purpose of the safe subscriber is to make sure user exceptions from onXXX methods are handled.

bryant1410, line 1847: But `SafeSubscriber` in the first line of the javadocs says:

> SafeSubscriber is a wrapper around Subscriber that ensures that the Subscriber complies with the Observable contract.

akarnokd, line 1847: Right. However, this method doesn't run any user supplied code and thus no need to wrap it into a safe completable subscriber.

bryant1410, line 1847: You're right. However, the other 2 `subscribe` methods that don't use `SafeCompletableSubscriber` should do it. To assert that the `Action` instances are called at most once, and not both, complying with the contract.

akarnokd, line 1847: Added the safeguard to those methods.

bryant1410, line 1847: I see. Why not reusing `SafeCompletableSubscriber`?

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -211,6 +212,48 @@ public void registerSingleExecutionHook(RxJavaSingleExecutionHook impl) {
...
+     * Retrieves the instance of {@link RxJavaCompletableExecutionHook} to use based on order of precedence as
+     * defined in {@link RxJavaPlugins} class header.
+     * <p>
+     * Override the default by calling {@link #registerCompletableExecutionHook(RxJavaCompletableExecutionHook)}
+     * or by setting the property {@code rxjava.plugin.RxJavaCompletableExecutionHook.implementation} with the
+     * full classname to load.
+     *
+     * @return {@link RxJavaCompletableExecutionHook} implementation to use
+     */
+    public RxJavaCompletableExecutionHook getCompletableExecutionHook() {
```
zsxwing, line 227: `@Experimental`

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -211,6 +212,48 @@ public void registerSingleExecutionHook(RxJavaSingleExecutionHook impl) {
...
+     * Register an {@link RxJavaCompletableExecutionHook} implementation as a global override of any injected or
+     * default implementations.
+     *
+     * @param impl
+     *            {@link RxJavaCompletableExecutionHook} implementation
+     * @throws IllegalStateException
+     *             if called more than once or after the default was initialized (if usage occurs before trying
+     *             to register)
+     */
+    public void registerCompletableExecutionHook(RxJavaCompletableExecutionHook impl) {
```
zsxwing, line 255: `@Experimental`

src/main/java/rx/plugins/RxJavaCompletableExecutionHook.java
```diff
@@ -0,0 +1,100 @@
...
+ * <p>
+ * A single implementation of this class will be used globally so methods on this class will be invoked
+ * concurrently from multiple threads so all functionality must be thread-safe.
+ * <p>
+ * Methods are also invoked synchronously and will add to execution time of the completable so all behavior
+ * should be fast. If anything time-consuming is to be done it should be spawned asynchronously onto separate
+ * worker threads.
+ *
+ */
+public abstract class RxJavaCompletableExecutionHook {
```
zsxwing, line 41: `@Experimental`

src/main/java/rx/observers/SafeCompletableSubscriber.java
```diff
@@ -0,0 +1,90 @@
...
+import rx.Completable.CompletableSubscriber;
+import rx.exceptions.*;
+import rx.Subscription;
+import rx.internal.util.RxJavaPluginUtils;
+
+/**
+ * Wraps another CompletableSubscriber and handles exceptions thrown
+ * from onError and onCompleted.
+ */
+public final class SafeCompletableSubscriber implements CompletableSubscriber, Subscription {
```
zsxwing, line 31: `@Experimental`

zsxwing:
Please add the missing `@Experimental`. Otherwise 👍

akarnokd:
Added missing annotations

src/main/java/rx/plugins/RxJavaCompletableExecutionHook.java
```diff
@@ -43,11 +46,11 @@
      * logging, metrics and other such things and pass through the function.
      *
      * @param f
-     *            original {@link Completable.CompletableOnSubscribe}<{@code T}> to be executed
-     * @return {@link Completable.CompletableOnSubscribe} function that can be modified, decorated, replaced or just
+     *            original {@link rx.Completable.CompletableOnSubscribe}<{@code T}> to be executed
+     * @return {@link rx.Completable.CompletableOnSubscribe} function that can be modified, decorated, replaced or just
      *         returned as a pass through
      */
-    public <T> Completable.CompletableOnSubscribe onCreate(Completable.CompletableOnSubscribe f) {
```
zsxwing: Good catch!

stevegury:
👍

PR #3948, opened 2016-05-18
Add Completable.andThen(Completable), deprecate endWith()

### LINKED ISSUE #3947: Request: Completable.andThen(Completable)
To match `Completable#andThen(Observable<T> next)` and `Completable#andThen(Single<T> next)` it would be nice if there were an operator like this:

`Completable#andThen(Completable next)`.

Unless there is some other way to achieve the same functionality right now.

vanniktech:
See #3947

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -1132,6 +1132,19 @@ public class Completable {
         requireNonNull(next);
         return next.delaySubscription(toObservable());
     }
+
+    /**
+     * Returns a completable that first runs this Completable
+     * and then the other completable.
+     * <p>
+     * This is an alias for {@link #concatWith(Completable)}.
+     * @param next the other Completable, not null
+     * @return the new Completable instance
+     * @throws NullPointerException if other is null
+     */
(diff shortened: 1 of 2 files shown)


akarnokd:
Please update unit tests that use these deprecated methods.

vanniktech:
Done.

akarnokd:
:+1:

artem-zinnatullin:
👍 but I'd either duplicate tests or make universal test functions and test both `endWith` and `andThen` to make sure that we don't break old behavior of public api.

vanniktech:
The question is also since Completable is still `@Experimental` we could if this is wanted go the hard way and remove the deprecated methods and make everyone force to use the standard (new) methods since there are already quite a few deprecated / duplicated methods.

zsxwing:
👍 

> but I'd either duplicate tests or make universal test functions and test both endWith and andThen to make sure that we don't break old behavior of public api.

@artem-zinnatullin I'm not against it but this is really minor as `andThen` just calls `endWith`. I'm going to merge this. Feel free to add tests in a separate PR if you wish.

PR #3949, opened 2016-05-18
1.x: javac- and javadoc-related cleanup in components, part 1

akarnokd:
My Eclipse warned me about several hundred javac and javadoc errors which could hide other significant warnings. This PR is the first part to fix those warnings. Note that `Observable` and `Single` itself is full of those as well (several hundred in each).


DavidMGross:
:+1:
Thanks for getting this done.

PR #3950, opened 2016-05-18
1.x: javac- and javadoc-related cleanup in components, part 2 final

akarnokd:
This PR clears the cast warnings introduced by the need to be JDK 9 compilable.


akarnokd:
I missed a bunch of javadoc problems.

PR #3951, opened 2016-05-18
1.x: javac- and javadoc-related cleanup in components, part 2

akarnokd:
Next round of cleanup, still about 100 cases remain in `Completable` and `Observable`, but those will most likely conflict with open PRs.


artem-zinnatullin:
👍

zsxwing:
👍

Issue #2656, opened 2015-02-13
just(T) : RxJavaObservableExecutionHook not called


ronanM:
When using rx.Observable.just(T) the RxJavaObservableExecutionHook is not call.

This is not like other just(T,...) that call rx.Observable.from(Iterable<? extends T>).

benjchristensen:
Thanks for the report.

benjchristensen:
@abersnaze Since this probably affects work you're doing as well, can you take a look at what we need to fix here? It will be anything using the internal ScalarSynchronousObservable optimization I think.

abersnaze:
sure.

akarnokd:
What is the status of this issue? I've lost track of the proposed changes to `ScalarSynchronousObservable`.

akarnokd:
See #3958.

zsxwing:
Closing via #3958

PR #3958, opened 2016-05-21
1.x: just() construction to call the onCreate execution hook

akarnokd:
This PR adds the call to `RxJavaObservableExecutionHook.onCreate()` to the `just()` - `ScalarSynchronousObservable` construction.

Related: #2656.

### DIFF
--- a/src/main/java/rx/internal/util/ScalarSynchronousObservable.java
+++ b/src/main/java/rx/internal/util/ScalarSynchronousObservable.java
@@ -23,6 +23,7 @@ import rx.functions.*;
 import rx.internal.producers.SingleProducer;
 import rx.internal.schedulers.EventLoopsScheduler;
 import rx.observers.Subscribers;
+import rx.plugins.*;
 
 /**
  * An Observable that emits a single constant scalar value to Subscribers.
@@ -33,6 +34,8 @@ import rx.observers.Subscribers;
  * @param <T> the value type
  */
 public final class ScalarSynchronousObservable<T> extends Observable<T> {
+    /** The execution hook instance. */
+    static RxJavaObservableExecutionHook hook = RxJavaPlugins.getInstance().getObservableExecutionHook();
     /**
      * Indicates that the Producer used by this Observable should be fully
      * threadsafe. It is possible, but unlikely that multiple concurrent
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/util/ScalarSynchronousObservable.java
```diff
@@ -33,6 +34,8 @@
  * @param <T> the value type
  */
 public final class ScalarSynchronousObservable<T> extends Observable<T> {
+    /** The execution hook instance. */
+    static RxJavaObservableExecutionHook hook = RxJavaPlugins.getInstance().getObservableExecutionHook();
```
artem-zinnatullin, line 44: nit: final

src/main/java/rx/internal/util/ScalarSynchronousObservable.java
```diff
@@ -131,6 +127,20 @@ public void call() {
         return create(new ScalarAsyncOnSubscribe<T>(t, onSchedule));
     }
     
+    /** The OnSubscribe callback for the Observable constructor. */
+    static final class JustOnSubscribe<T> implements OnSubscribe<T> {
```
artem-zinnatullin, line 137: :+1:

src/main/java/rx/internal/util/ScalarSynchronousObservable.java
artem-zinnatullin, line 44: Ah, you override it in the test. Then it needs a comment or test should use reflection.

src/test/java/rx/internal/util/ScalarSynchronousObservableTest.java
```diff
@@ -230,4 +234,30 @@ public void onNext(Integer t) {
...
+            };
+            
+            int n = 10;
+            
+            for (int i = 0; i < n; i++) {
+                Observable.just(1).subscribe();
+            }
+            
+            Assert.assertEquals(n, c.get());
+        } finally {
```
artem-zinnatullin, line 259: I guess it's more important to verify that it uses `OnSubscribe` returned by hook rather than verify just invocation.

akarnokd:
Updated javadoc and test.

artem-zinnatullin:
:+1:

ZacSweers:
👍

src/test/java/rx/internal/util/ScalarSynchronousObservableTest.java
artem-zinnatullin, line 259: For reviewers: this is fixed below.

zsxwing:
:+1:

Issue #3667, opened 2016-02-04
`Observable.scan()` overload to take a factory for initial value?


NiteshKant:
Currently the scan operator has [an overload](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L7356) to take an initial value for the accumulation. The problem with this is that the initial value is the same for all subscribers of this observable, which hardly is intended by the user. It is actually a source of subtle bugs for a casual user.

I think there should atleast be an overload that instead of the realized initial value takes a factory for the initial value for each subscription. The underlying [`OperatorScan`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorScan.java#L73) already supports an initial value factory.

I would also suggest to either deprecate the current `scan()` method or if it is usable in some cases, atleast put a warning in the javadoc.

zsxwing:
See #1881 for the type erase issue and #1884 for `collect`.

NiteshKant:
Thanks @zsxwing. But `collect` and `scan` have different behaviors, isn't? 
So, the question here still makes sense?

NiteshKant:
If there is some agreement on what to do here, I can send a PR.

zsxwing:
Technically, you can implement `scan(Func0, accumulator)` using `collect`. I tried in RxScala but the code looks ugly, so I'm very happy to have a `scan` overload. However, we need a new method name due to the ambiguity. Any suggestion for the new method name?

akarnokd:
PRs are always welcome and sometimes the only means to get people moving around here.

akarnokd:
See #3959.

akarnokd:
Due to potential problems with the two parameter variant proposed in #3959, I added some javadoc instead showing how one can use `defer` to avoid sharing the mutable initial value (#4063).

If you have further input on the issue, don't hesitate to reopen this issue or post a new one.

Issue #1881, opened 2014-11-15
Scan/Reduce Overloads & Type Erasure


benjchristensen:
I think the recent additions to `scan`/`reduce` may cause issues

``` java
scan(R, Func2<R, ? super T, R>)
scan(Func0<R>, Func2<R, ? super T, R>)
```

The `Func0` passed in looks like it can be treated like an `Object` and considered ambiguous and match with `R` instead of `Func0`.

Here is a compilation error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project learnrxjava: Compilation failure: Compilation failure:
[ERROR] /Users/benjchristensen/development/github/learnrxjava/src/main/java/learnrxjava/examples/ScanVsReduceExample.java:[10,32] reference to reduce is ambiguous
[ERROR] both method <R>reduce(R,rx.functions.Func2<R,? super T,R>) in rx.Observable and method <R>reduce(rx.functions.Func0<R>,rx.functions.Func2<R,? super T,R>) in rx.Observable match
[ERROR] /Users/benjchristensen/development/github/learnrxjava/src/main/java/learnrxjava/examples/ScanVsReduceExample.java:[10,39] incompatible types: cannot infer type-variable(s) R
[ERROR] (argument mismatch; java.lang.Object is not a functional interface)
[ERROR] /Users/benjchristensen/development/github/learnrxjava/src/main/java/learnrxjava/examples/ScanVsReduceExample.java:[17,32] reference to scan is ambiguous
[ERROR] both method <R>scan(R,rx.functions.Func2<R,? super T,R>) in rx.Observable and method <R>scan(rx.functions.Func0<R>,rx.functions.Func2<R,? super T,R>) in rx.Observable match
[ERROR] /Users/benjchristensen/development/github/learnrxjava/src/main/java/learnrxjava/examples/ScanVsReduceExample.java:[17,37] incompatible types: cannot infer type-variable(s) R
[ERROR] (argument mismatch; java.lang.Object is not a functional interface)
[ERROR] -> [Help 1]
```

Here is example code: https://github.com/jhusain/learnrxjava/blob/master/src/main/java/learnrxjava/examples/ScanVsReduceExample.java

``` java
package learnrxjava.examples;

import java.util.ArrayList;

import rx.Observable;

public class ScanVsReduceExample {

    public static void main(String... args) {
[...]

PR #3959, opened 2016-05-21
1.x: scan with an initial factory callback

akarnokd:
This PR adds an overload to `scan` which allows specifying an initial value factory. The initial value factory is invoked for each incoming Subscriber individually and provides a non-shared starting point for the `scan` operation.

I propose the `scan(Func2 accumulator, Func0 initialFactory)` parameter order because some Java 8+ compilers can't properly disambiguate with `scan(R initialValue, Func2 accumulator)`, thinking that R can be made `Func0`. 

Related: #3667


abersnaze:
Why add the overload if it can be done with defer and scan then?

akarnokd:
Yep, it can be done. A decision has to be made if this convenience should make it into the library or not, and notify @NiteshKant the outcome in #3667.

davidmoten:
I'd like to see this overload in, thanks.

zsxwing:
I'm not strong against adding it but not like the idea of the new parameter order. I would give +1 if giving it a better name instead of changing the parameter order. By the way, there is already a `OperatorScan(Func0<R> initialValueFactory, Func2<R, ? super T, R> accumulator)` for this overload so you don't need to use `defer`.

Probably @benjchristensen should take a look since you removed it before.

benjchristensen:
It was removed due to ambiguity as shown in https://github.com/ReactiveX/RxJava/issues/1881#issue-48878398

As long as that issue is not recreated, I have nothing against the functionality pursued here. I too don't like solving the overload ambiguity by moving parameter order around. But there is no good solution, it's either that or a new name.

akarnokd:
As an alternative, I suggest extending `scan`'s javadoc with the example of the defer trick.

akarnokd:
The argument order bothers me as well. I'll close this and post a javadoc update instead.

Issue #3679, opened 2016-02-08
Actual exception is being hidden by OnErrorFailedException


ktchernov:
Using RxJava 1.10 (and similar issue on 1.0.15), I stumbled upon a mysterious stack trace that made no sense in my app, after a lot of debugging, I found that RxJava will report the wrong exception in some scenarios. For example, if you run this sample code:

    Observable.just(1).doOnNext(new Action1<Integer>() {
			@Override public void call(Integer val) {
				throw new IllegalStateException("doOnNextException");
			}
		}).retryWhen(new Func1<Observable<? extends Throwable>, Observable<?>>() {
			@Override public Observable<?> call(Observable<? extends Throwable> errorObservable) {
				return errorObservable.flatMap(new Func1<Throwable, Observable<?>>() {
					@Override public Observable<?> call(Throwable throwable) {
						return Observable.just(3).delay(1, TimeUnit.SECONDS).doOnNext(
								new Action1<Integer>() {
									@Override public void call(Integer integer) {
										throw new UnsupportedOperationException("retry exception");
									}
								});
					}
				});
			}
		}).subscribe(new Subscriber<Integer>() {
			@Override public void onCompleted() {
				Log.v("EX", "onCompleted");
			}

			@Override public void onError(Throwable e) {
				Log.v("EX", "onError");
				throw new IllegalThreadStateException();
			}

			@Override public void onNext(Integer integer) {
				Log.v("EX", "onNext");
			}
		});

The expected behaviour should be that `IllegalThreadStateException` (thrown within the `onError` handler of the subscriber) is included as the cause. However, the actual stack trace contains `UnsupportedOperationException` as the cause - which is not the problem in the code snippet above. the `UnsupportedOperationException` is totally fine, but the problem arises when `onError` throws.

This leads to a stack trace that points to the wrong cause and does not include the actual cause (which would be the `IllegalThreadStateException`). If I removed the `retryWhen()` block - then the stack trace correctly includes `IllegalThreadStateException` as the exception from within `onError`.

The (incorrect) trace from the code snippet above:

    Process: com.example.mydemoapp, PID: 14118
    java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.Worker thread.
        at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:62)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
[...]

PR #3963, opened 2016-05-25
Composite exception should update cause to match the correct throwable that's first in the chain

premnirmal:
Fixes https://github.com/ReactiveX/RxJava/issues/3679


akarnokd:
Please add an unit test that shows verifies the expected behavior.

premnirmal:
Updated

zsxwing:
@premnirmal This is not correct. I just sent #3977 to fix it. Thanks anyway!

akarnokd:
Thanks for the contribution. Could you check #3977 and verify it fixes your issue?

premnirmal:
Thanks @zsxwing , closing

Issue #3779, opened 2016-03-18
withLatestFrom interface inconsistent with combineLatest


MidnightWonderer:
Why withLatestFrom has only 1 calling signature while combineLatest has plenty of convenient overload for combining multiple observables.  
Should it be the same or this is by design?

akarnokd:
How would you imagine those missing forms of `withLatestFrom`, especially in their behavior?

MidnightWonderer:
Does this make sense?
a$.withLatestFrom(b$, c$, d$, new Func4<a,b,c,d>{
// use a, b, c, d relying on timing emit by a only
});

akarnokd:
Would you like to submit a PR?

MidnightWonderer:
I would love to if I have some spare time.  
I look into the code a bit and found out that combineLatest is just thin wrapper of OnSubscribeCombineLatest factory inside Observable class.  
The test is done on OnSubscribeCombineLatest class itself the factory methods are untested.  
  
It will be trivial if I can just create an overload by the calling exist (current) withLatestFrom method multiple times to aggregate the value one by one in Observable class without additional test on the overloads.  
  
Is this the way to go?  
If yes you can expect me to send the PR eventually.
(Don't know when. If anyone want it fast feel free to pick it up. It is just convenient overloads after all we can continue working without it.)  
If no please give me your opinion about my proposed method. At least I will learn a thing or two from this issue.

akarnokd:
No, you have to work from `withLatestFrom` with more sources, subscribers and current values.

MidnightWonderer:
Yep that's the reply I expected.
By looking at the code I have to integrate Iterable into the operator class.
  
For sure it will be easier than CombineLatest implementation but I don't know concurrency concerns and proper way (exist method/pattern) to handle it inside RxJava that well.
So it will require more time digging into the code I don't even sure if I am able to do it.
  
So I will say don't expect PR from me anytime soon.
Thank you for your feedback.

akarnokd:
Do you have an use case for such an operator? I'd say it is uncommon and I'd guess the original [request](https://github.com/ReactiveX/RxJava/issues/405) for `withLatestFrom` didn't mention a multi-source version.

You can achieve an effect with some tuple types:

```java
source.withLatestFrom(b, (a, b) -> Pair.of(a, b))
.withLatestFrom(c, (ab, c) -> func.call(ab.a, ab.b, c))
```

MidnightWonderer:
[...]

PR #3966, opened 2016-05-26
1.x: add multi-other withLatestFrom operators

akarnokd:
Add `withLatestFrom` operator version that can take 2 to N other sources and combine them with the main source.

Related: #3779.


stevegury:
👍

PR #3969, opened 2016-05-27
1.x fix long-chained xWith() operator stack overflow

akarnokd:
This PR fixes the case when operators of pattern `xWith` chained into a long sequence causes `StackOverflowError` because the subscription call-stack gets deep.

Operators affected and fixes:
- `Observable.ambWith()`
- `Observable.mergeWith()`
- `Observable.concatWith()`
- `Observable.startWith()`
- `Observable.zipWith()`
- `Completable.ambWith()`
- `Completable.mergeWith()`
- `Completable.concatWith()`
- `Completable.startWith()`
- `Single.zipWith()`

The PR contains some tidy-up of existing functionality and introducing short operators that delegate to the "big" operators for the purpose of identification.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -119,63 +119,7 @@ public class Completable {
             return sources[0];
         }
         
-        return create(new CompletableOnSubscribe() {
-            @Override
-            public void call(final CompletableSubscriber s) {
-                final CompositeSubscription set = new CompositeSubscription();
-                s.onSubscribe(set);
-
-                final AtomicBoolean once = new AtomicBoolean();
-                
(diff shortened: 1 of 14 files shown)


akarnokd:
Long running, unrelated tests fail on travis for hours now.

zsxwing:
Not sure if this is worth to fix (Hundreds of codes to fix a rare case). For merging tons of Observable, I would expect that the user uses the static array version instead of writing a for-loop. In addition, this won't fix this issue totally. E.g., people can add an operator like this and will still get StackOverflowError.

```
        Observable<Integer> source = Observable.just(1);
        Observable<Integer> result = source;

        for (int i = 0; i < n; i++) {
            result = result.mergeWith(source).map(x -> x + 1);
        }
```

PR #3971, opened 2016-05-31
1.x: request rebatch operator

akarnokd:
This is a follow-up on #3964 but with a separate operator on `Observable`.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -6692,6 +6692,30 @@ public class Observable<T> {
     }
 
     /**
+     * Requests {@code n} initially from the upstream and then 75% of {@code n} subsequently
+     * after 75% of {@code n} values have been emitted to the downstream.
+     * 
+     * <p>This operator allows preventing the downstream to trigger unbounded mode via {@code request(Long.MAX_VALUE)}
+     * or compensate for the per-item overhead of small and frequent requests.
+     * 
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator expects backpressure from upstream and honors backpressure from downstream.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code publish} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *  
+     * @param n the initial request amount, further request will happen after 75% of this value
+     * @return the Observable that rebatches request amounts from downstream
+     */
+    public final Observable<T> rebatchRequests(int n) {
+        if (n <= 0) {
+            throw new IllegalArgumentException("n > 0 required but it was " + n);
+        }
+        return lift(OperatorObserveOn.<T>rebatch(n));
+    }
+    
+    /**
      * Returns an Observable that applies a specified accumulator function to the first item emitted by a source
      * Observable, then feeds the result of that function along with the second item emitted by the source
      * Observable into the same function, and so on until all items have been emitted by the source Observable,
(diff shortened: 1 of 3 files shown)


src/main/java/rx/Observable.java
```diff
@@ -6692,6 +6692,30 @@ public final void forEach(final Action1<? super T> onNext, final Action1<Throwab
...
+     * 
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator expects backpressure from upstream and honors backpressure from downstream.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code publish} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *  
+     * @param n the initial request amount, further request will happen after 75% of this value
+     * @return the Observable that rebatches request amounts from downstream
```
DavidMGross, line 6709: @beta? @experimental? @since?

src/main/java/rx/Observable.java
```diff
@@ -6692,6 +6692,30 @@ public final void forEach(final Action1<? super T> onNext, final Action1<Throwab
...
+     * after 75% of {@code n} values have been emitted to the downstream.
+     * 
+     * <p>This operator allows preventing the downstream to trigger unbounded mode via {@code request(Long.MAX_VALUE)}
+     * or compensate for the per-item overhead of small and frequent requests.
+     * 
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator expects backpressure from upstream and honors backpressure from downstream.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code publish} does not operate by default on a particular {@link Scheduler}.</dd>
```
DavidMGross: operator name incorrect here

src/main/java/rx/Observable.java
akarnokd, line 6709: Added.

src/main/java/rx/Observable.java
akarnokd: Fixed

zsxwing:
👍

akarnokd:
@abersnaze, @stealthcode you had some use cases for this, any objections?

abersnaze:
the reuse of the observeOn is interesting but couldn't it be done without the allocation of a queue?

akarnokd:
If the downstream request is unbounded and the downstream has caught up then the queue can be skipped. In this case, `observeOn` can't be reused anymore and a custom drain logic has to be implemented.

Otherwise, the upstream emissions have to be stored temporarily for an underrequesting downstream.

stevegury:
👍

stealthcode:
I know that @abersnaze still had reservations about this. I think that this should not be using `observeOn`.

stealthcode:
My concern is this - If @abersnaze implemented the batching functionality then why wouldn't we use that? The queue in observeOn scheduling creates a layer of indirection that seems unnecessary.

akarnokd:
Remember, this started out as a change to `observeOn` to not ignore the `immediate` scheduler but people wanted this behavior exposed behind a proper name.

stealthcode:
Thanks for reminding me of the context of this work. It seems like we have 2 implementations for the same functionality. I think @abersnaze and I agree that the 2 features of request batching and request valve type functionality could be composed. However I think that using `observeOn` for this functionality is not necessarily the best way to accomplish this.

stealthcode:
I personally would be okay with either implementation. I think `observeOn` is a nice choice because of the request management but would prefer it more if there wasn't a queue. But realistically I think the choice is fine. 

Also it's interesting to note that users are gravitating more and more to taking direct control over the `requester-producer` interactions.

abersnaze:
For example this PR does something similar but exactly n (could be modified to have optional 25%) and without a queue #3781.

PR #3975, opened 2016-05-31
1.x: lift into Subject

akarnokd:
See #2458.

This PR shows a way to implement lifting into a Subject. It is possible to lift on the front and the back side. The main idea is to add behavior to a Subject while presenting it to the outside world still as a Subject.

I'm not 100% certain the value of this because one has to write operators for this to work (relying on the `rx.internal.operators` is not recommended). The front-side operator can be simpler since it doesn't have to deal with unsubscription and backpressure as the wrapped Subjects don't have that capability on their front anyway.

### DIFF
--- a/src/main/java/rx/subjects/LiftedSubject.java
+++ b/src/main/java/rx/subjects/LiftedSubject.java
@@ -1,0 +1,183 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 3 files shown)


stevegury:
I'm not convinced either by its value, do you have a example in mind?

akarnokd:
This was @abersnaze 's long lasting dream from #2458. I thought I give an implementation to see if we want it or #2458 can be closed for good.

abersnaze:
I wouldn't go as far as saying it was a dream of my but it has always bugged me that the type `Subject<T, R>` clearly has an `R` that can be different but there isn't a practical way to change it.  If subject is never intended to be used like that then we should change it to `Subject<T>` in 2.0.

PR #3977, opened 2016-06-01
1.x: Use the correct Throwable to set the cause for CompositeException

zsxwing:
The cause of #3679 is we use a wrong Throwable (its cause has been set) to set the cause and `initCause` will throw an exception. Hence, the cause chain is not created correctly. In this PR, it searches the root cause (which doesn't have a cause) and use it to call `initCause`.

### DIFF
--- a/src/main/java/rx/exceptions/CompositeException.java
+++ b/src/main/java/rx/exceptions/CompositeException.java
@@ -147,13 +147,12 @@ public final class CompositeException extends RuntimeException {
                 // we now have 'e' as the last in the chain
                 try {
                     chain.initCause(e);
-                    chain = chain.getCause();
                 } catch (Throwable t) {
                     // ignore
                     // the javadocs say that some Throwables (depending on how they're made) will never
                     // let me call initCause without blowing up even if it returns null
-                    chain = e;
                 }
+                chain = getRootCause(chain);
             }
             cause = _cause;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/rx/exceptions/CompositeException.java
```diff
@@ -293,4 +292,25 @@ public String getMessage() {
...
+     * @param e the {@link Throwable} {@code e}.
+     * @return The root cause of {@code e}. If {@code e.getCause()} returns {@null}, just return {@code e} itself.
+     */
+    private Throwable getRootCause(Throwable e) {
+        Throwable root = e.getCause();
+        if (root == null) {
+            return e;
+        } else {
+            while(true) {
+                if (root.getCause() == null) {
```
JakeWharton: I have seen cases where a `Throwable` will return `this` for `getCause()` which can cause infinite loops in code like this.

I think you want:

``` java
while (true) {
  Throwable cause = root.getCause();
  if (cause == null || cause == root) {
    return root;
  }
  root = cause;
}
```

as a result instead.

zsxwing: I was concerned it too but didn't do it because the default `getCause` will return `null` when the inner `cause` field is `this`.

However, `getCause` can be override so it's still possible. Will update it.

akarnokd:
:+1:

stevegury:
👍

PR #3979, opened 2016-06-01
1.x: remove remaining field updaters

akarnokd:
This PR removes the remaining `AtomicXFieldUpdater`s from the library because reflection is somewhat problematic on Android.

This also deprecates the internal `BackpressureUtils.getAndAddRequest` to indicate `requested` field updater should not be used anymore.

### DIFF
--- a/src/main/java/rx/internal/operators/BackpressureUtils.java
+++ b/src/main/java/rx/internal/operators/BackpressureUtils.java
@@ -45,7 +45,9 @@ public final class BackpressureUtils {
      * @param n
      *            the number of requests to add to the requested count
      * @return requested value just prior to successful addition
+     * @deprecated Android has issues with reflection-based atomics
      */
+    @Deprecated
     public static <T> long getAndAddRequest(AtomicLongFieldUpdater<T> requested, T object, long n) {
         // add n to field but check for overflow
         while (true) {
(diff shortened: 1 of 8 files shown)


zsxwing:
> BackpressureUtils.getAndAddRequest

This is an internal API. Nobody is supposed to use it. Why not just remove it?

akarnokd:
External, custom operators may still depend on this utility method.

src/main/java/rx/internal/operators/BackpressureUtils.java
```diff
@@ -45,7 +45,9 @@ private BackpressureUtils() {
      * @param n
      *            the number of requests to add to the requested count
      * @return requested value just prior to successful addition
+     * @deprecated Android has issues with reflection-based atomics
```
JakeWharton, line 48: s/Android/Samsung/ 😀 

Let's not lump all of Android in with the likes of the insane OS developers at Samsung who modify Java system packages needlessly without rhyme or reason.

akarnokd, line 48: It's one thing those devices can't find fields but how about reflection in general. Desktop Java does well with inlining and intrinsification of AtomicXFieldUpdaters; I'm not sure about Android in general through.

JakeWharton, line 48: Yeah I'm not sure. I would imagine the new runtime on Android 5.0+ could do something about it. I'll have to investigate it more.

zsxwing:
> External, custom operators may still depend on this utility method.

Okey, let's remove it later. :+1:

stevegury:
👍

PR #3980, opened 2016-06-01
2.x: Design.md +extension +fusion

akarnokd:
Added sections from #3935.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -366,15 +366,29 @@ Therefore, new standard factory methods will try to address the common entry poi
   - `create(Consumer<? super FlowEmitter<T>>)` to relay multiple values or error from multi-valued reactive-sources (i.e., button-clicks) while also give flow control options right there (buffer, drop, error, etc.).
   - `create(Consumer<? super CompletionEmitter)>` signal a completion or error from valueless reactive sources
 
-The following table lists where these create methods will be available in respect of the base types:
+The `Flowable` will contain the following `create` methods:
 
-| Method | Flowable | Observable | Single | Completable |
-|--------|----------|------------|--------|-------------|
-| `create(SyncGenerator<T, S>)` | Yes | Yes | No | No |
-| `create(AsyncOnSubscribe<T, S>)` | Yes | No | No | No |
-| `create(Consumer<? super SingleEmitter<T>>)` | Yes | Yes | Yes | No |
-| `create(Consumer<? super FlowEmitter<T>>)` | Yes | Yes | No | No |
-| `create(Consumer<? super CompletionEmitter>)` | Yes | Yes | No | Yes |
+   - `create(SyncGenerator<T, S>)`
+   - `create(AsyncOnSubscribe<T, S>)`
+   - `create(Consumer<? super SingleEmitter<T>>)`
+   - `create(Consumer<? super FlowEmitter<T>>)`
+   - `create(Consumer<? super CompletionEmitter>)`
+   
+The `Observable` will contain the following `create` methods:
+
+   - `create(SyncGenerator<T, S>)`
+   - `create(Consumer<? super SingleEmitter<T>>)`
+   - `create(Consumer<? super FlowEmitter<T>>)`
+   - `create(Consumer<? super CompletionEmitter>)`
+
+The `Single` will contain the following `create` method:
+
+   - `create(Consumer<? super SingleEmitter<T>>)`
+   
+The `Completable` will contain the following `create` method:
+
+   - `create(Consumer<? super CompletionEmitter>)`
+
 
 The first two `create` methods take an implementation of an interface which provides state and the generator methods:
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -325,18 +325,149 @@ In the addition of the previous rules, an operator for `Flowable`:
 
 ### Creation
 
-Creation of the various types should be exposed through factory methods that provide safe construction.
+Unlike RxJava 1.x, 2.x base classes are to be abstract, stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes.
+
+Instead of the indirection of an `OnSubscribe` and `lift`, operators are to be implemented by extending the base classes. For example, the `map`
+operator will look like this:
 
 ```java
-Flowable.create(SyncGenerator generator)
+public final class FlowableMap<T, R> extends Flowable<R> {
+    
+    final Flowable<? extends T> source;
+    
+    final Function<? super T, ? extends R> mapper;
+    
+    public FlowableMap(Flowable<? extends T> source, Function<? super T, ? extends R> mapper) {
+        this.source = source;
+        this.mapper = mapper;
+    }
+    
+    @Override
+    protected void subscribeActual(Subscriber<? super R> subscriber) {
+        source.subscribe(new FlowableMapSubscriber<T, R>(subscriber, mapper));
+    }
+    
+    static final class FlowableMapSubscriber<T, R> implements Subscriber<T>, Subscription {
+        // ...
+    }
+}
+``` 
 
-Flowable.create(AsyncGenerator generator)
+Since Java still doesn't have extension methods, "adding" more operators can only happen through helper methods such as `lift(C -> C)` and `compose(R -> P)` where `C` is the default consumer type (i.e., `rs.Subscriber`), `R` is the base type (i.e., `Flowable`) and `P` is the base interface (i.e., `rs.Publisher`). As before, the library itself may gain or lose standard operators and/or overloads through the same community process.
 
-Observable<T>.create(OnSubscribe<Observer<T>> onSubscribe)
+In concert, `create(OnSubscribe)` will not be available; standard operators extend the base types directly. The conversion of other RS-based libraries will happen through the `Flowable.wrap(Publisher<T>)` static method. 
 
-Single<T>.create(OnSubscribe<Single.Subscriber<T>> onSubscribe)
+(*The unfortunate effect of `create` in 1.x was the ignorance of the Observable contract and beginner's first choice as an entry point. We can't eliminate this path since `rs.Publisher` is a single method functional interface that can be implemented just as badly.*)
 
-Completable<T>.create(OnSubscribe<Completable.Subscriber<T>> onSubscribe)
+Therefore, new standard factory methods will try to address the common entry point requirements: 
+  - `create(SyncGenerator<T, S>)` to safe, synchronous generation of signals, one-by-one
+  - `create(AsyncGenerator<T, S>)` to batch-create signals based on request patterns
+  - `create(Consumer<? super SingleEmitter<T>>)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
+  - `create(Consumer<? super FlowEmitter<T>>)` to relay multiple values or error from multi-valued reactive-sources (i.e., button-clicks) while also give flow control options right there (buffer, drop, error, etc.).
+  - `create(Consumer<? super CompletionEmitter)>` signal a completion or error from valueless reactive sources
+
+The following table lists where these create methods will be available in respect of the base types:
+
+| Method | Flowable | Observable | Single | Completable |
+|--------|----------|------------|--------|-------------|
+| `create(SyncGenerator<T, S>)` | Yes | Yes | No | No |
+| `create(AsyncOnSubscribe<T, S>)` | Yes | No | No | No |
+| `create(Consumer<? super SingleEmitter<T>>)` | Yes | Yes | Yes | No |
+| `create(Consumer<? super FlowEmitter<T>>)` | Yes | Yes | No | No |
+| `create(Consumer<? super CompletionEmitter>)` | Yes | Yes | No | Yes |
+
+The first two `create` methods take an implementation of an interface which provides state and the generator methods:
+
+```java
+interface SyncGenerator<T, S> {
+
+    S createState();
+    
+    S generate(S state, Observer<T> output);
+    
+    void disposeState(S state);
+}
+
+interface AsyncGenerator<T, S> {
+
+    S createState();
+    
+    S generate(S state, long requested, Observer<Observable<T>> output);
+    
+    void disposeState(S state);
+}
+```
+
+These latter three `create` methods will provide the following interaction interfaces to the `java.util.function.Consumer`:
+
+```java
+interface SingleEmitter<T> {
+
+    complete(T value);
+    
+    fail(Throwable error);
+    
+    stop();
+    
+    setDisposable(Disposable d);
+    
+}
+
+interface FlowEmitter<T> {
+
+    void next(T value);
+   
+    void fail(Throwable error);
+   
+    void complete();
+   
+    void stop();
+
+    setDisposable(Disposable d);
+
+    enum BackpressureHandling {
+        IGNORE,
+        ERROR,
+        DROP,
+        LATEST,
+        BUFFER
+    }
+
+    void setBackpressureHandling(BackpressureHandling mode);
+    
+}
+
+interface CompletableEmitter<T> {
+
+    complete();
+    
+    fail(Throwable error);
+    
+    stop();
+    
+    setDisposable(Disposable d);
+    
+}
+
+```
+
+By extending the base classes, operator implementations would loose the tracking/wrapping features of 1.x. To avoid this, the methods `subscribe(C)` will be final and operators have to implement a protected `subscribeActual` (or any other reasonable name).
+
+```java
+@Override
+public final void subscribe(Subscriber<? super T> s) {
+   subscribeActual(hook.onSubscribe(s));
+} 
+
+protected abstract void subscribeActual(Subscriber<? super T> s);
+```
+
+Assembly-time hooks will be moved into the individual standard methods on the base types:
+
+```java
+public final Flowable<R> map(Function<? super T, ? extends R> mapper) {
+    return hook.onAssembly(new FlowableMap<T, R>(this, mapper));
+}
 ```
 
 ### Terminal behavior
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -325,18 +325,37 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+
+In concert, `create(OnSubscribe)` will not be available; standard operators extend the base types directly. The conversion of other RS-based libraries will happen through the `Flowable.wrap()` static method. 
 
-Flowable.create(AsyncGenerator generator)
+(*The unfortunate effect of `create` in 1.x was the ignorance of the Observable contract and beginner's first choice as an entry point. We can't eliminate this path since `rs.Publisher` is a single method functional interface that can be implemented just as badly.*)
 
-Observable<T>.create(OnSubscribe<Observer<T>> onSubscribe)
+Therefore, new standard factory methods will try to address the common entry point requirements: 
+  - `create(SyncOnSubscribe)` to safe, synchronous generation of signals, one-by-one
+  - `create(AsyncOnSubscribe)` to batch-create signals based on request patterns (I'm still not convinced)
```
stealthcode: > I'm still not convinced

I don't think personal opinions belong in a design doc. Can you add a comment or try to articulate this outside of the committed document?

DESIGN.md
```diff
@@ -325,18 +325,37 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+  - `create(SingleEmitter)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
```
stealthcode: What is a `SingleEmitter` and `FlowEmitter`? These aren't mentioned anywhere else.

DESIGN.md
akarnokd: Removed

DESIGN.md
```diff
@@ -325,18 +325,37 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+In concert, `create(OnSubscribe)` will not be available; standard operators extend the base types directly. The conversion of other RS-based libraries will happen through the `Flowable.wrap()` static method. 
 
-Flowable.create(AsyncGenerator generator)
+(*The unfortunate effect of `create` in 1.x was the ignorance of the Observable contract and beginner's first choice as an entry point. We can't eliminate this path since `rs.Publisher` is a single method functional interface that can be implemented just as badly.*)
 
-Observable<T>.create(OnSubscribe<Observer<T>> onSubscribe)
+Therefore, new standard factory methods will try to address the common entry point requirements: 
+  - `create(SyncOnSubscribe)` to safe, synchronous generation of signals, one-by-one
+  - `create(AsyncOnSubscribe)` to batch-create signals based on request patterns
+  - `create(SingleEmitter)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
```
stealthcode: What is a `SingleEmitter` and `FlowEmitter`? Are they variations of a `SyncOnSubscribe` or a `Consumable`?

DESIGN.md
```diff
@@ -325,18 +325,37 @@ In the addition of the previous rules, an operator for `Flowable`:
 
 ### Creation
 
-Creation of the various types should be exposed through factory methods that provide safe construction.
+Unlike RxJava 1.x, 2.x base classes are abstract and stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes. 
 
-```java
-Flowable.create(SyncGenerator generator)
```
stealthcode, line 331: @stevegury Was the `SyncGenerator` wording supposed to be a rename of the `SyncOnSubscribe`?

DESIGN.md
akarnokd:
``` java
/**
 * Interface to generate signals to a bridged {@link Subscriber}.
 * <p>
 * At least one of the methods
 * should be called per invocation of the generator function
 *
 * @param <T> the output value type
 */
public interface SignalEmitter<T> {

    /**
     * @see Subscriber#onComplete()
     */
    void complete();

    /**
     * @see Subscriber#onError(Throwable)
     */
    void fail(Throwable e);

    /**
     * Try emitting, might throw an unchecked exception.
     * @see Subscriber#onNext(Object)
     *
     * @throws RuntimeException
     */
    void next(T t);

    /**
     * Indicate there won't be any further signals delivered by
     * the generator and the operator will stop calling it.
     * <p>
     * Call to this method will also trigger the state consumer.
     */
    void stop();

    void setCancellation(Cancellation c);
}

/**
 * Wrapper API around a downstream Subscriber for emitting any number of
 * onNext signals followed by zero or one onError/onComplete.
 * <p>
 * One should call {@link FlowEmitter#setBackpressureHandling(BackpressureHandling)}
 * before emitting any signal.
 * @param <T> the value type
 */
public interface FlowEmitter<T> extends SignalEmitter<T>  {

    enum BackpressureHandling {
        IGNORE,
        ERROR,
        DROP,
        LATEST,
        BUFFER
    }

    void setBackpressureHandling(BackpressureHandling mode);
}
```

DESIGN.md
akarnokd: Those are there to bridge push-based sources such as classical async callbacks or notifications while ensuring the RS protocol is honored.

DESIGN.md
stealthcode: @akarnokd I've searched for where this is defined and I'm not finding them in the 2.x source code, your branch, 1.x, or the design doc. Where did you copy paste this snippet from? This document should be self contained (i.e. not dependent on any existing code other than reactive streams dependencies) and should stand on its own. So could you include a definition here in the design doc? 

Also you have a typo, `Single` should be `Signal`.

akarnokd: `SyncGenerator` and co weren't defined either.

akarnokd: Added definitions to those types in create()

stealthcode: > `SyncGenerator` and co weren't defined either.

That's a problem too. Not sure if they were a rename or different entity.

DESIGN.md
```diff
@@ -325,18 +325,99 @@ In the addition of the previous rules, an operator for `Flowable`:
 
 ### Creation
 
-Creation of the various types should be exposed through factory methods that provide safe construction.
+Unlike RxJava 1.x, 2.x base classes are abstract and stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes.
```
stealthcode: > base classes are abstract and stateless and generally no longer wrap an `OnSubscribe` callback

This paragraph suggests a design that is already implemented. This design document is intended to capture the requirements for 2.x, what **needs** to be implemented for 2.x without the presumption that an implementation has been chosen or finalized. I'm fine with mentioning this aspect of the design but could you please elaborate this paragraph so that everyone can be on the same page?

DESIGN.md
```diff
@@ -325,18 +325,99 @@ In the addition of the previous rules, an operator for `Flowable`:
 
 ### Creation
 
-Creation of the various types should be exposed through factory methods that provide safe construction.
+Unlike RxJava 1.x, 2.x base classes are abstract and stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes. 
+
+Since Java still doesn't have extension methods, "adding" more operators can only happen through helper methods such as `lift(C -> C)` and `compose(R -> P)` where `C` is the default consumer type (i.e., `rs.Subscriber`), `R` is the base type (i.e., `Flowable`) and `P` is the base interface (i.e., `rs.Publisher`). As before, the library itself may gain or lose standard operators and/or overloads through the same community process.
```
stealthcode, line 356: > "adding" more operators can only happen through

This version of `compose` seems to be very similar to the proposed `extend` method on `Consumable` as suggested in my prototype, as merged on the `Observable.extend` and as we discussed in prior pull requests. Is the intent with the `compose` overload to replace the `extend` method?

DESIGN.md
```diff
@@ -325,18 +325,99 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+
+In concert, `create(OnSubscribe)` will not be available; standard operators extend the base types directly. The conversion of other RS-based libraries will happen through the `Flowable.wrap()` static method.
```
stealthcode: > The conversion of other RS-based libraries will happen through the `Flowable.wrap()` static method. 

Could you add a method signature for `Flowable#wrap` to the `Flowable` class declaration in this design document?

DESIGN.md
```diff
@@ -325,18 +325,99 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+    S generate(S state, long requested, Observer<Observable<T>> output);
+    
+    void disposeState(S state);
+}
+```
+
+These latter two `create` methods will provide the following interaction interfaces to the `java.util.function.Consumer`:
+
+```java
+interface SingleEmitter<T> {
```
stealthcode, line 413: You deleted the mention of how to create a `Single`. Is the intent that a `Consumable<SingleEmitter<T>>` will replace the `Single<T>.create(OnSubscribe<Single.Subscriber<T>> onSubscribe)`? What will be used to replace the creation of a `Completable` as mentioned: `Completable<T>.create(OnSubscribe<Completable.Subscriber<T>> onSubscribe)`?

DESIGN.md
```diff
@@ -325,18 +325,99 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+@Override
+public final void subscribe(Subscriber<? super T> s) {
+   subscribeActual(hook.onSubscribe(s));
+} 
 
-Completable<T>.create(OnSubscribe<Completable.Subscriber<T>> onSubscribe)
+protected abstract void subscribeActual(Subscriber<? super T> s);
+```
+
+Assembly-time hooks will be moved into the individual standard methods on the base types:
```
stealthcode, line 474: There has been no previous mention of the hooks in this document. Is it necessary to include this in the design?

DESIGN.md
```diff
@@ -325,18 +325,99 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+  - `create(SyncOnSubscribe)` to safe, synchronous generation of signals, one-by-one
+  - `create(AsyncOnSubscribe)` to batch-create signals based on request patterns
+  - `create(Consumer<? super SingleEmitter<T>>)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
+  -  `create(Consumer<? super FlowEmitter<T>>)` to relay multiple values or error from multi-valued reactive-sources (i.e., button-clicks) while also give flow control options right there (buffer, drop, error, etc.).
+
+The first two `create` methods take an implementation of an interface which provides state and the generator methods:
 
 ```java
-Flowable.create(SyncGenerator generator)
+interface SyncOnSubscribe<T, S> {
```
stealthcode: @akarnokd I'm pretty sure that the `SyncGenerator` was supposed to be a renamed `SyncOnSubscribe` (and likewise the Asyc variant). Would you mind renaming these interface declarations to match? If anyone remembers otherwise maybe they can correct my memory @abersnaze @benjchristensen @stevegury

DESIGN.md
akarnokd, line 356: compose() is different from extend because it gives the full base type to the function. Substituting R with Flowable<T> the signature looks like this:

``` java
public final <U> Flowable<U> compose(Function<? super Flowable<T>, ? extends Publisher<U>> composer);
```

DESIGN.md
akarnokd, line 474: How would you otherwise explain the need to make `subscribe` final and have a `subscribeActual` instead?

DESIGN.md
akarnokd: Done.

DESIGN.md
akarnokd, line 413: Detailed.

DESIGN.md
stealthcode, line 356: Okay, I would rather only have one method for "extending" functionality? Is that also your goal? In which case we should discuss the merits of compose as a replacement for extend. 

For comparison I'm including the signature of extend.

``` java
<O2, X extends Consumable<O2>> X extend(Function<Consumer<O>, X> f)
```

Which in the case of a Flowable would be as follows (note that I have replaced the `Consumable<O2>` with `Publisher<U>`)

``` java
<O2, X extends<Publisher<U>> X extend(Function<Consumer<Subscriber<? super T>>, X> f)
```

So the only substantial difference between compose and extend is that with compose the function consumes a concrete Flowable type and with extend it consumes an OnSubscribe. 

I think my concern is that the `compose` functions _must_ know the absolute type of the thing that you are converting from (or at least a base class/interface). This makes it more restrictive than what `extend` allows, converting from any variation of a `Flowable` by virtue of unwrapping and accessing the internal callback. So composition functions would be case specific, and would not be reusable for different types of flowable, observable, single, or completable. 

Note that I am **NOT** proposing making conversion functions from a flowable that also work on an observable for instance. What I am proposing is that if you have different variations of an Observable that the same logic could be use to convert it to a different thing. I think that there is the potential for better code reuse with `extend`. Also my intent is that extending functions for the core RxJava types (i.e. the `Flowable`, `Observable`, `Single`, `Completable` and all of there variants which implement their respective `Consumable<S>` for their subscriber/observer type) be written and packaged with RxJava. But this would not be limited to just internal apis as the pattern could be useful for libraries which write their own `Consumable`.

DESIGN.md
```diff
@@ -325,18 +325,149 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+  - `create(AsyncGenerator<T, S>)` to batch-create signals based on request patterns
+  - `create(Consumer<? super SingleEmitter<T>>)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
+  - `create(Consumer<? super FlowEmitter<T>>)` to relay multiple values or error from multi-valued reactive-sources (i.e., button-clicks) while also give flow control options right there (buffer, drop, error, etc.).
+  - `create(Consumer<? super CompletionEmitter)>` signal a completion or error from valueless reactive sources
+
+The following table lists where these create methods will be available in respect of the base types:
+
+| Method | Flowable | Observable | Single | Completable |
+|--------|----------|------------|--------|-------------|
+| `create(SyncGenerator<T, S>)` | Yes | Yes | No | No |
```
stealthcode: A SyncGenerator (the SyncOnSubscribe) could be used to model a Single or a Completable.

DESIGN.md
```diff
@@ -325,18 +325,149 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+| `create(AsyncOnSubscribe<T, S>)` | Yes | No | No | No |
```
stealthcode: The `AsyncGenerator` could also be used to model any of these variants of observables.

DESIGN.md
akarnokd: SyncGenerator allows calling onCompleted which is invalid for `Single` and onNext which is invalid for `Completable`.

DESIGN.md
akarnokd: `AsyncGenerator.generate` takes a request amount which is undefined for `Observable`, `Single` and `Completable`.

DESIGN.md
```diff
@@ -325,18 +325,149 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+    S generate(S state, long requested, Observer<Observable<T>> output);
+    
+    void disposeState(S state);
+}
+```
+
+These latter three `create` methods will provide the following interaction interfaces to the `java.util.function.Consumer`:
+
+```java
+interface SingleEmitter<T> {
```
stealthcode, line 413: Maybe I'm not understanding this correctly. Could you please give us an example of how this would work in practice?

DESIGN.md
```diff
@@ -325,18 +325,149 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+In concert, `create(OnSubscribe)` will not be available; standard operators extend the base types directly. The conversion of other RS-based libraries will happen through the `Flowable.wrap(Publisher<T>)` static method. 
+
+(*The unfortunate effect of `create` in 1.x was the ignorance of the Observable contract and beginner's first choice as an entry point. We can't eliminate this path since `rs.Publisher` is a single method functional interface that can be implemented just as badly.*)
 
-Flowable.create(AsyncGenerator generator)
+Therefore, new standard factory methods will try to address the common entry point requirements: 
+  - `create(SyncGenerator<T, S>)` to safe, synchronous generation of signals, one-by-one
+  - `create(AsyncGenerator<T, S>)` to batch-create signals based on request patterns
+  - `create(Consumer<? super SingleEmitter<T>>)` to relay a single value or error from other reactive sources (i.e., addListener callbacks)
+  - `create(Consumer<? super FlowEmitter<T>>)` to relay multiple values or error from multi-valued reactive-sources (i.e., button-clicks) while also give flow control options right there (buffer, drop, error, etc.).
```
stealthcode: Then I'm not seeing how the `SingleEmitter` can be used to model an Observable? Is this an Observable of just 1 value? I think that the table is not accurately portraying their usage.

DESIGN.md
stealthcode: The CompletionEmitter can only model an empty observable but you list it with the others. Could you please remove the table and just list how you propose that people will create each type? I don't think that these methods are directly comparable in a tabular format.

stealthcode:
@akarnokd since the edits are causing all in-line comments to be hidden I'm posting our discussion on "extension" functionality here. 

> compose() is different from extend because it gives the full base type to the function. Substituting R with Flowable<T> the signature looks like this:
> 
> ``` java
> public final <U> Flowable<U> compose(Function<? super Flowable<T>, ? extends Publisher<U>> composer);
> ```

Okay, I would rather only have one method for "extending" functionality? Is that also your goal? In which case we should discuss the merits of compose as a replacement for extend. 

For comparison I'm including the signature of extend.

``` java
<O2, X extends Consumable<O2>> X extend(Function<Consumer<O>, X> f)
```

Which in the case of a Flowable would be as follows (note that I have replaced the `Consumable<O2>` with `Publisher<U>`)

``` java
<O2, X extends<Publisher<U>> X extend(Function<Consumer<Subscriber<? super T>>, X> f)
```

So the only substantial difference between compose and extend is that with compose the function consumes a concrete Flowable type and with extend it consumes an OnSubscribe. 

I think my concern is that the `compose` functions _must_ know the absolute type of the thing that you are converting from (or at least a base class/interface). This makes it more restrictive than what `extend` allows, converting from any variation of a `Flowable` by virtue of unwrapping and accessing the internal callback. So composition functions would be case specific, and would not be reusable for different types of flowable, observable, single, or completable. 

Note that I am **NOT** proposing making conversion functions from a flowable that also work on an observable for instance. What I am proposing is that if you have different variations of an Observable that the same logic could be use to convert it to a different thing. I think that there is the potential for better code reuse with `extend`. Also my intent is that extending functions for the core RxJava types (i.e. the `Flowable`, `Observable`, `Single`, `Completable` and all of there variants which implement their respective `Consumable<S>` for their subscriber/observer type) be written and packaged with RxJava. But this would not be limited to just internal apis as the pattern could be useful for libraries which write their own `Consumable`.

DESIGN.md
akarnokd, line 413:
``` java
// usage
Consumer<SingleEmitter<T>> c = e -> {
    AutoCloseable close = someAPI.performAsync(new Callback<T>() {
        @Override
        public void onData(T data) {
            e.complete(data);
        }

        @Override
        public void onFailure(Throwable ex) {
            e.fail(ex);
        }
    });
    e.setDisposable(() -> {
        try {
            close.close();
        } catch (Exception ex) {
            RxJavaPlugins.onError(ex);
        }
    });
};

Flowable.create(c).subscribe(System.out::println);
```

DESIGN.md
akarnokd: Since `Flowable`, `Observable` can also be empty, you can trigger completion with all 3 types.

akarnokd: Flattened the table.

DESIGN.md
```diff
@@ -325,18 +325,158 @@ In the addition of the previous rules, an operator for `Flowable`:
...
+    final Function<? super T, ? extends R> mapper;
+    
+    public FlowableMap(Flowable<? extends T> source, Function<? super T, ? extends R> mapper) {
+        this.source = source;
+        this.mapper = mapper;
+    }
+    
+    @Override
+    protected void subscribeActual(Subscriber<? super R> subscriber) {
+        source.subscribe(new FlowableMapSubscriber<T, R>(subscriber, mapper));
```
stealthcode, line 347: Should this be `source.subscribeActual`?

DESIGN.md
```diff
@@ -325,18 +325,158 @@ In the addition of the previous rules, an operator for `Flowable`:
 
 ### Creation
 
-Creation of the various types should be exposed through factory methods that provide safe construction.
+Unlike RxJava 1.x, 2.x base classes are to be abstract, stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes.
+
+Instead of the indirection of an `OnSubscribe` and `lift`, operators are to be implemented by extending the base classes. For example, the `map`
```
stealthcode, line 330: Okay, I see where you are going with this change. I don't think this has been discussed before. If I'm understanding correctly this is a departure from the `Operator` interface as an adapter between subscribers and instead just giving each operator method the full context of the previous observable. 

I would rather find an alternative to class inheritance. What are the advantages of this approach? Also, why is `FlowableMap` declared `final`?

DESIGN.md
akarnokd, line 347: That would bypass the plugin hook in source.

DESIGN.md
akarnokd, line 330: If Java had extension methods, you didn't have to do this inheritance. FlowableMap is final to prevent unintended extension.

stealthcode, line 330: This will not work with `Consumable` because each operator is not reusable for every implementation of a Consumable for a given observer/subscriber type. I do not agree that this should be the way to provide extensions.

akarnokd, line 330: Cross base-type operator reuse can't work. Extending the base types with new methods via inheritance doesn't work. It requires returning <? extends BaseType> and the compiler can't properly chain calls with them without target-typing each stage.

stealthcode, line 330: If I am understanding your terminology correctly, you are saying that a "cross base-type" operator is an operator which functions for both an `Observable` and a `Flowable` for instance. Am I correct? I'm suggesting that an operator which knows how to negotiate backpressure from a `Subscriber` should be capable of supporting any kind of `Consumable` that functions using `Subscriber`. This means that there can be many types which implement `Consumable<Observer<T>>` and any operator that knows how to work with an `Observer` can be reused.

akarnokd, line 330: For `Flowable` base _class_ type we have the `rs.Publisher` as the base _interface_ type. In principle, you could define base interfaces for `Observable`, `Single` and `Completable`, but I don't see why unless you want to encourage other libraries to implement them differently than in RxJava.

For `Flowable`, the `Operator<T, R>` == `Function<Subscriber<R>, Subscriber<T>>`; the `OnSubscribe<T>` == `Consumer<Subscriber<T>>` ~ `rs.Publisher<T>`.

The plan is that standard operator implementations, such as `FlowableMap` may take an `rs.Publisher<T>` at least (and `Flowable<T>` at most), but you won't be able to combine `FlowableMap` and `ObservableMap` because the `actual` output class differs and needs different calls to `onSubscribe`.

stealthcode, line 330: > but you won't be able to combine FlowableMap and ObservableMap because the actual output class differs and needs different calls to onSubscribe.

@akarnokd I am telling you that is not what I am saying! Read my comments carefully! I am saying that a Map operator can be written that works over all Flowables AND a separate map operator can be written that works over all `Observables`! I really don't know why you don't read my comments. 

> but I don't see why unless you want to encourage other libraries to implement them differently than in RxJava.

We talked about this when you tried to implement `Consumable`. I thought you were on board with creating a base interface or class for all kinds of Observables (and likewise a different base for Completables and Singles). Is there something I'm missing?

> The plan

Who's plan? I know that I'm pretty confused about the perceived benefits here. It seems like a micro optimization to save an allocation.

akarnokd, line 330: The drawback of arguing in a design document is that there is no proof that things actually work out properly and not just locally. If we'd work from code PRs instead, we'd be sure the concepts actually compile and how far their effects go.

stealthcode, line 330: That is why I showed that it can be done in a proof of concept project. That you reviewed.

DESIGN.md
```diff
@@ -325,18 +325,158 @@ In the addition of the previous rules, an operator for `Flowable`:
 
 ### Creation
 
-Creation of the various types should be exposed through factory methods that provide safe construction.
+Unlike RxJava 1.x, 2.x base classes are to be abstract, stateless and generally no longer wrap an `OnSubscribe` callback - this saves allocation in assembly time without limiting the expressiveness. Operator methods and standard factories still live as final on the base classes.
```
stevegury, line 328: Can you explain more about the rationale behind having abstract base classes that are stateless?
Is that an optimization?

DESIGN.md
```diff
@@ -359,6 +499,100 @@ This section contains current design work which needs more discussion and elabor
...
+
+##### Micro-fusion
+
+Micro-fusion goes a step deeper and tries to reuse internal structures, mostly queues, in operator pairs, saving on allocation and sometimes on atomic operations. It's property is that, in a way, subverts the standard Reactive-Streams protocol between subsequent operators that both support fusion. However, from the outside world's view, they still work according to the RS protocol.
+
+Currently, two main kinds of micro-fusion opportunities are available.
+
+###### 1) Conditional Subscriber
+
+This extends the RS `Subscriber`interface with an extra method: `boolean tryOnNext(T value)` and can help avoiding small request amounts in case an operator didn't forward but dropped the value. The canonical use is for the `filter()` operator where if the predicate returns false, the operator has to request 1 from upstream (since the downstream doesn't know there was a value dropped and thus not request itself). Operators wanting to participate in this fusion have to implement and subscribe with an extended Subscriber interface:
```
stevegury, line 543: Could you clarify this with an example.

DESIGN.md
stevegury, line 347: Like @stealthcode, I was under the impression that we had decided to go with the `Consumable` interface.

DESIGN.md
akarnokd, line 328: It uses less memory.

DESIGN.md
akarnokd, line 543: Example of what? Rsc has a bunch of implementations with this pattern, for example, [filter](https://github.com/reactor/reactive-streams-commons/blob/master/src/main/java/rsc/publisher/PublisherFilter.java#L57).

DESIGN.md
akarnokd, line 347: If you implement it, then go with `Consumable`.

stealthcode:
My take away from this pull request is that @akarnokd has decided on a plan without communicating it to the other contributors on the project. I would like to hear _much_ more details on how this is to be implemented and I don't think that this PR is the right way to do so. I would prefer to have much better communication with the RxJava Core Committers so that we can try to reach some sort of understanding of our motivations and goals in design. 

We at Netflix have tried to communicate with a proof of concept and many conversations over PRs and issues to no avail. It seems to me that there are a very significant misalignments in the goals and rationale behind design decisions. 

For example, I personally do not see a significant benefit in complicating the code to avoid a minority of memory allocations. I am not considering removing the `Operator` interface to save a 1 time allocation. The chances are very good that the JVM will optimize the runtime anyway so multiple allocations wouldn't be a significant overhead.

DESIGN.md
stealthcode, line 347: That's not how collaborating on a team works @akarnokd.

akarnokd:
The last 9+ months was quite tedious for me regarding this project and Netflix' behavior: extended periods of inactivity, delayed decision making, extensive gating of contributions and being stuck in local design/thinking choices expressed repeatedly (or argued endlessly) by some. In addition, I simply don't have the capacity give tailored explanations to everybody.

Such local thinking is the negligence of allocation count. It might not be an issue on multi-gigabyte Amazon cloud servers, but RxJava has to run on millions of Android devices where memory footprint and allocation does matter.

The proof-of-concepts Netflix provided in this regard were too small to be conclusive; they lacked some non-trivial operators and usage examples, with which the compiler ambiguities or coding inconveniences could have been discovered - apparently, me pointing them out is not enough.

Note that there has been a working RxJava 2.x implementation available as developer preview since September 2015 in the 2.x branch here. Its review never progressed even before this design document was posted by Netflix.

As the primary contributor and maintainer of RxJava, I had to start making decisions on my own due to these. The basis for it is the fact that I've been developing _complete_ reactive libraries before and during my time with RxJava and accumulated considerable amount experience with all things reactive.

The design choices I made for RxJava 2.x are proven in the research project [Reactive-Streams-Commons](https://github.com/reactor/reactive-streams-commons) and in fact forms the basis for the competing library [Reactor-Core](https://github.com/reactor/reactor-core) of Pivotal. Textual descriptions on why can be found in my blog: [RxJava Design Retrospect](http://akarnokd.blogspot.hu/2016/03/rxjava-design-retrospect.html), [Operator Fusion, part 1](http://akarnokd.blogspot.hu/2016/03/operator-fusion-part-1.html) and [part 2](http://akarnokd.blogspot.hu/2016/04/operator-fusion-part-2-final.html).

There are two reasons I'm still contributing to RxJava: 1) it's the only proper reactive library for Android and 2) Reactor-Core decided to provide only a subset of operators. However, any of this can change.

In conclusion, all I can say is to take this PR, propose alterations to it in a separate PR or leave it.

stealthcode:
@akarnokd we have proposed changes in our comments. I would prefer it if you made an attempt to communicate with us. 

> I simply don't have the capacity give tailored explanations to everybody.

You are extremely responsive to the community requests over github, stackoverflow, and email and your lengthy responses show that you do seem to have time to respond with very useful information. However your interactions with the RxJava committers demonstrates that you have little or no patience with your collaborators. We have tried reaching out to you by email, instant message, and video chat to open some form of communication with you to no avail. Being a part of a team does require some level of coordination and communication. We'd like to know that the most prolific person on the team can at least be reasoned with. 

> run on millions of Android devices where memory footprint and allocation does matter.

You haven't made any attempt to view any other possible optimizations other than your own designs. You claim that you'

stealthcode:
I think RxJava core committers team has a lack of cohesion on 2.x designs because there is a lack of communication. @akarnokd we have proposed changes in our comments but you have no patience with our concerns. We have tried reaching out to you by email, instant message, and video chat in the past but you don't seem interested in direct communication outside of Github. Being a part of a team does require a higher level of coordination and communication however I suspect that you have no intention of working with us. 

> run on millions of Android devices where memory footprint and allocation does matter.

Our goal is to develop a generally applicable library that can be safely and productively used by Android as well as other platforms. @stevegury and I believe that the JVM runtime (even on Android) would optimize the code such that there is little to no difference between the source code optimized implementation and our proposed designs. I think its in the best interest of the RxJava community for the contributors to evaluate different options. 

My professional work at Netflix uses and depends heavily on RxJava. I _try_ to contribute my professional expertise to RxJava however when I read responses from @akarnokd I perceive a tone of hostility and a lack of interest. Many others at Netflix choose not to contribute to RxJava because it is too difficult to make progress on pull requests. I have not found an efficient way to collaborate with @akarnokd in design or implementation. RxJava is not benefiting in significant ways from Netflix contributions as many of our pull requests are deadlocked in debate or discouraged.

There are aspects that we don't understand about this approach and when we reach out to him we are met with what I interpret as avoidance and hostility. I can only conclude that @akarnokd is trying to implement what he thinks is best for the community while avoiding any contribution from Netflix.

akarnokd:
Since this is a design document change only, I'm merging this as is, allowing progress on the code side as well as enabling PRs for further clarifications if necessary.

I encourage contributors with architectural ideas to make sure their suggested changes actually work out in the context of the entire 2.x codebase. The current 2.x branch contains a working (although outdated) variant and let's anyone experiment within his/her own environment.

PR #3987, opened 2016-06-03
Add more time to certain tests that failed under heavy CI load

akarnokd:
Sometimes, the heavy Travis CI load (or throttling) makes a few tests fail with timeout. This increases the timeout for some of them to reduce the likelihood.


ZacSweers:
👍

PR #3488, opened 2015-11-03
1.x Remove all instances of Atomic*FieldUpdater

markrietveld:
Replace them all with their respective Atomic* counterparts
For example AtomicLongFieldUpdater -> AtomicLong
Addresses https://github.com/ReactiveX/RxJava/issues/3459

Issue #3459, opened 2015-10-19
NoSuchFieldException: unsubscribed


yyfrankyy:
```
java.lang.NoSuchFieldException: unsubscribed
java.lang.Class.getDeclaredField(Class.java:929)
java.util.concurrent.atomic.AtomicIntegerFieldUpdater$AtomicIntegerFieldUpdaterImpl.<init>(AtomicIntegerFieldUpdater.java:251)
java.util.concurrent.atomic.AtomicIntegerFieldUpdater.newUpdater(AtomicIntegerFieldUpdater.java:49)
rx.subscriptions.BooleanSubscription.<clinit>(SourceFile:32)
rx.subscriptions.Subscriptions.create(SourceFile:75)
rx.subjects.SubjectSubscriptionManager.addUnsubscriber(SourceFile:68)
rx.subjects.SubjectSubscriptionManager.call(SourceFile:58)
rx.subjects.SubjectSubscriptionManager.call(SourceFile:35)
rx.Observable.subscribe(SourceFile:7803)
rx.Observable.subscribe(SourceFile:7531)
```

Happened randomly, I cannot reproduce this.

Machine Info

```
SM-N900;Android 5.0 LRX21V,level 21
```

amitcse:
I too have seen this issue occurring randomly. In my case it was happening while using `HystrixCollapser` inside map/flatMap operator. On debugging I found that `onCompleted` was being called on the observable before `onNext`. I was not able to isolate if it was due to rx operator or Hystrix.

@yyfrankyy What is scenario in your use-case? Can you share your flow of composition ?

akarnokd:
Hi. Do you use ProGuard by any chance? I'm not an expert on it but you might need to change the field retention policy.

artem-zinnatullin:
ProGuard detects and keeps fields used by `AtomicFieldUpdater` since v4.6 (release happened 4 years ago), maybe @yyfrankyy uses another code obfuscator?

yyfrankyy:
@akarnokd @artem-zinnatullin we do use proguard, but keep all the staff that rx needs by `-keep class rx.** { * ; }`

@amitcse It crashed at the very beginning since the `Application#onCreate()`, the code are here, but it works quite well at most of the time.

https://github.com/WeTeX/Watchers/blob/master/src/main/java/com/github/QQMail/Watchers.java#L244

[...]

PR #3991, opened 2016-06-06
Remove instance of Atomic*FieldUpdater

lucas34:
Hi,

Based on the PR #3488 and the issue #3459 
I thought all the instance of Atomic*FieldUpdater has been removed and replaced.


JakeWharton:
This was already changed in #3979.

Issue #3993, opened 2016-06-06
SchedulerLifecycleTest failures


akarnokd:
After merging #3986, these to fail in master (worked as PR):

```
rx.schedulers.SchedulerLifecycleTest > testShutdown FAILED
    java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@73f20bba rejected from java.util.concurrent.ScheduledThreadPoolExecutor@17e01bce[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2048)
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:821)
        at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:325)
        at java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:530)
        at java.util.concurrent.ScheduledThreadPoolExecutor.execute(ScheduledThreadPoolExecutor.java:619)
        at rx.schedulers.SchedulerLifecycleTest.tryOutSchedulers(SchedulerLifecycleTest.java:98)
        at rx.schedulers.SchedulerLifecycleTest.testShutdown(SchedulerLifecycleTest.java:36)
rx.schedulers.SchedulerLifecycleTest > testStartIdempotence STARTED
rx.schedulers.SchedulerLifecycleTest > testStartIdempotence FAILED
    java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@18264a30 rejected from java.util.concurrent.ScheduledThreadPoolExecutor@17e01bce[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2048)
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:821)
        at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:325)
[...]

PR #3996, opened 2016-06-07
1.x: fix reset() shutting down everything other than the schedulers

akarnokd:
The reset shut down all the main schedulers and the helper pools, causing the failure in #3993 .

@hzsweers, you were right with the need for the instance-shutdown method.

### DIFF
--- a/src/main/java/rx/schedulers/Schedulers.java
+++ b/src/main/java/rx/schedulers/Schedulers.java
@@ -48,7 +48,7 @@ public final class Schedulers {
             if (INSTANCE.compareAndSet(null, current)) {
                 return current;
             } else {
-                shutdown();
+                current.shutdownInstance();
             }
         }
     }
@@ -168,8 +168,10 @@ public final class Schedulers {
      */
     @Experimental
     public static void reset() {
-        shutdown();
-        INSTANCE.set(null);
+        Schedulers s = INSTANCE.getAndSet(null);
+        if (s != null) {
+            s.shutdownInstance();
+        }
     }
     
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/rx/schedulers/Schedulers.java
```diff
@@ -201,22 +197,44 @@ public static void reset() {
...
             RxRingBuffer.SPSC_POOL.shutdown();
 
             RxRingBuffer.SPMC_POOL.shutdown();
         }
     }
+    
+    /**
+     * Start the instance-specific schedulers.
+     */
+    synchronized void startInstance() {
```
ZacSweers, line 214: Any reason to not make these private?

ZacSweers:
Minor nit, rest of the code looks good 👍 

Nice catch, sorry for the inadvertent bug and glad it was a relatively simple fix

src/main/java/rx/schedulers/Schedulers.java
akarnokd, line 214: We try to avoid too much `private` modifiers; generally it adds hidden bridge methods, increasing the method count which is bad for Android. Also package level access leaves it open to same-package unit testing.

artem-zinnatullin:
👍

src/main/java/rx/schedulers/Schedulers.java
ZacSweers, line 214: Fair enough, though this case wouldn't generate synthetic accessors I don't think.

akarnokd, line 214: Right, bridges are not in play here; it has become a second nature of me scoping as package-private.

ZacSweers, line 214: Not a bad approach :). LGTM then

stevegury:
👍

Issue #395, opened 2013-09-20
java.util.Comparator is not suitable for distinct and distinctUntilChanged


samuelgruetter:
In C#, `distinct` and `distinctUntilChanged` take an `IEqualityComparer` (which provides `Equals` and `GetHashCode`).
That's not the same as `java.util.Comparator` (which provides `compare`).

Or mathematically speaking: There are two kinds of comparators:
-    Those which define a _total ordering_: `java.util.Comparator` and C#'s `IComparer`
-    Those which define an _equivalence relation_: C#'s `IEqualityComparer`, RxJava's `Func2<T, T, Boolean>`, and unfortunately no standard Java interface

The problem is that for `distinct` and `distinctUntilChanged`, we're not interested in an order on the elements, but only in equality.

So it would be better to use `Func2<? super T, ? super T, Boolean> equality` (just like `sequenceEqual`) instead of `Comparator<U> equalityComparator`.

PR #4001, opened 2016-06-09
1.x: added distinctUntilChanged(comparator)

Nexen23:
For some cases, `Observable.distinctUntilChanged(keySelector)` isn't enough. For example, when data consists of 2+ objects with ID, you cannot simply summarize them and use it as a key. Also often there are stream of arrays of data which can't be distincted easily and generically.
So I think `Observable.distinctUntilChanged(comparator)` would be helpful. I use kinda tricky implementation, but think, it is okey. Firstly I tried to add ctor to OperatorDistinctUntilChanged(), but it was... too bad.

**p.s.** I was really confused by 

> If you would like to contribute code you can do so through GitHub by forking the repository and sending a pull request (on a branch other than master, 1.x, 2.x, or gh-pages).

and spent some time trying to create some sort of temporal branch here (newbie me). But finally looked up into PRs and so that it is okey to PR into 1.x. Is it an error in .md-file?

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -4761,6 +4761,44 @@ public class Observable<T> {
     public final <U> Observable<T> distinctUntilChanged(Func1<? super T, ? extends U> keySelector) {
         return lift(new OperatorDistinctUntilChanged<T, U>(keySelector));
     }
+    
+    /**
+     * Returns an Observable that emits all items emitted by the source Observable that are distinct from their
+     * immediate predecessors, according to a comparator function.
+     * <p>
+     * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/distinctUntilChanged.key.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code distinctUntilChanged} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * 
+     * @param comparator
+     *            a function that compares two successive items and decide whether an item
+     *            is distinct from another one or not
+     * @return an Observable that emits those items from the source Observable whose keys are distinct from
+     *         those of their immediate predecessors
+     * @see <a href="http://reactivex.io/documentation/operators/distinct.html">ReactiveX operators documentation: Distinct</a>
+     */
+    public final Observable<T> distinctUntilChanged(Func2<? super T, ? super T, Boolean> comparator) {
+        return lift(new OperatorDistinctUntilChanged<>(new Func1<T, Object>() {
+			T previousValue;
+
+			@Override
+			public Object call(T t) {
+				if (previousValue != null) {
+					if (comparator.call(previousValue, t)) {
+						return previousValue;
+					} else {
+						previousValue = t;
+						return t;
+					}
+				} else {
+					previousValue = t;
+					return previousValue;
+				}
+			}
+		}));
+    }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/rx/Observable.java
```diff
@@ -4761,6 +4761,44 @@ public Completable toCompletable() {
...
+     * @param comparator
+     *            a function that compares two successive items and decide whether an item
+     *            is distinct from another one or not
+     * @return an Observable that emits those items from the source Observable whose keys are distinct from
+     *         those of their immediate predecessors
+     * @see <a href="http://reactivex.io/documentation/operators/distinct.html">ReactiveX operators documentation: Distinct</a>
+     */
+    public final Observable<T> distinctUntilChanged(Func2<? super T, ? super T, Boolean> comparator) {
+        return lift(new OperatorDistinctUntilChanged<>(new Func1<T, Object>() {
+			T previousValue;
```
akarnokd: This will be shared among all subscribers of the chain and you won't get proper results when multiple subscribers run the chain concurrently.

akarnokd:
Please add unit tests that verify the behavior.

src/main/java/rx/Observable.java
```diff
@@ -4761,6 +4761,44 @@ public Completable toCompletable() {
...
+     * 
+     * @param comparator
+     *            a function that compares two successive items and decide whether an item
+     *            is distinct from another one or not
+     * @return an Observable that emits those items from the source Observable whose keys are distinct from
+     *         those of their immediate predecessors
+     * @see <a href="http://reactivex.io/documentation/operators/distinct.html">ReactiveX operators documentation: Distinct</a>
+     */
+    public final Observable<T> distinctUntilChanged(Func2<? super T, ? super T, Boolean> comparator) {
+        return lift(new OperatorDistinctUntilChanged<>(new Func1<T, Object>() {
```
akarnokd: Our target is Java 6 where the diamond operator is not allowed.

src/main/java/rx/Observable.java
Nexen23: > This will be shared among all subscribers

Oh, I see.
Is it okay to implement it inside `OperatorDistinctUntilChanged`? It would be ugly due to call() will have `if/else` of `new Subscriber`, so I don't know is it permissible to do so.

Nexen23:
How about this one?

akarnokd:
Sorry for the delay. I remembered there was a discussion about this overload a while ago in #395.

The verdict was to not have them and you have to create your own wrapper class with custom `equals` and use it as a keySelector.

The decision may be revised, depending on the other's opinions.

(Otherwise, I would have done it differently, not with that subscriberProvider).

akarnokd:
See #4034 about how I'd implement this instead (easier shown in code than as a comment). Let me know if you want to adjust your PR.

Nexen23:
By `adjust your PR` you mean copy-paste your implementation as successive commit to existed ones?

akarnokd:
> By adjust your PR you mean copy-paste your implementation as successive commit to existed ones?

If you think that my PR doesn't cover everything you wanted then yes. Otherwise, I leave it to you.

Nexen23:
> If you think that my PR doesn't cover everything you wanted then yes

No, I think your PR is just okay and do what I wanted and do it better than mine own :+1: 
So I guess I need to close this PR now?

akarnokd:
Yes, you can close this. Thanks for your contribution anyway! If you have any further improvement suggestions / PRs for the library don't hesitate to post them. (If you need some hints about operator coding styles and structure, try looking at the common and simpler operators implementations.)

PR #4005, opened 2016-06-13
1.x: fix Spsc queues reporting not empty but then poll() returns null

akarnokd:
In the spsc queues, the indexes were written before the actual elements and thus a concurrent `isEmpty` check would report a non-empty queue but a `poll` would still return `null`. The fix swaps the two writes.

Note that this an inconsistency in the original JCTools code and not the lack of keeping up with it.


artem-zinnatullin:
👍

stevegury:
👍

PR #4007, opened 2016-06-14
1.x: new hook management proposal

akarnokd:
This PR adds a new, in-between, hook manager class, `RxJavaHooks` that allows runtime hooking of `Observable`, `Single`, `Completable` and `Schedulers` components and is aimed to be more versatile than the `RxJavaPlugins`.

Since `RxJavaPlugins` is public API, it can't be just removed, therefore, `RxJavaHooks` by default delegates to it but that delegation can be completely disabled via `clear()`.
#### Usage

Call the `setOn` methods with an appropriate function to manipulate the object being hooked:

``` java
RxJavaHooks.setOnComputationScheduler(s -> Schedulers.immediate());

RxJavaHooks.setObservableCreate(o -> { System.out.println("Creating: " + o); return o; });
```

You can also get the current hooks via the `getOn` methods, allowing chaining multiple hooks if necessary. Changing and retrieving the hooks are thread safe, although it is recommended you change them in quiet times.

Operator writers should usually not call the `onXXX` methods on `RxJavaHooks` except `RxJavaHooks.onError()`; It is useful when they have to signal a `Throwable` that can't be delivered to a `Subscriber`.

Calling `reset()` will restore the original delegation hooks.

This PR also features a way of tracking the assembly locations, that is where `Observable.create()` was invoked internally or externally, via `enableAssemblyTracking`. It changes the creation hooks of all 3 base types by adding an `OnSubscribeOnAssemblyX` wrapper. This operator will replace the Throwable flowing through `onError` and wraps it via the new `AssemblyStackTraceException`. 

Both the wrapper and the exception contain a string representation of the captured stacktrace, this helps "printing" out that information in a debugging session by simply viewing the field contents.

This tracking can be enabled at any time and affects sequences created afterwards. To stop the tracking, use `resetAssemblyTracking` to restore the default delegate callbacks to `RxJavaPlugins` or `clearAssemblyTracking` to restore the empty hooks (only affecting the `onCreate` hooks).

For frameworks, the `RxJavaHooks` can be locked down, preventing changing the hooks.

If you are using `RxJavaPlugins` existing features, you don't have to do anything; tests will work as before.
#### Performance impact

A default `RxJavaHooks` incurs a 2-4 level indirection in method calls. `onNext` calls are not affected.

A clear `RxJavaHooks` incurs a volatile read (very cheap) followed by a branch (predictable). `onNext` calls are not affected.

A tracking-enabled RxJavaHooks incurs an estimated 1000-3000 cycles overhead for each source creation and operator application. The `AssemblyStackTraceException` itself doesn't fill in its own stacktrace but takes only the captured stacktrace string and has mostly the cost of an object allocation. Due to the in-between nature of tracking, `onXXX` calls get through another indirection at each operator. For example, `range(1, 5).map(v -> v).filter(v -> true).subscribe()` will have 3 extra layers (one for each `.`). I believe these are acceptable overheads because the tracking feature is for tracking down crashes and not performance problems.
#### Discussion

Not all hook methods have been replicated completely on `RxJavaHooks`. For one, I wanted a minimal working prototype that passes existing tests. Second, those that are left out are not even tested (`onSubscribeError`, `onLift`). If the **concept** of this PR is accepted, those can be added along with their unit tests. Also further Javadoc will be added in the same case.

Names and structure are not set in stone.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeOnAssembly.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeOnAssembly.java
@@ -1,0 +1,128 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.operators;
+
+import rx.*;
+import rx.Observable.OnSubscribe;
+import rx.exceptions.AssemblyStackTraceException;
+
+/**
+ * Captures the current stack when it is instantiated, makes
+ * it available through a field and attaches it to all
+ * passing exception.
+ *
+ * @param <T> the value type
+ */
+public final class OnSubscribeOnAssembly<T> implements OnSubscribe<T> {
+
+    final OnSubscribe<T> source;
+    
+    final String stacktrace;
+
+    /**
+     * If set to true, the creation of PublisherOnAssembly will capture the raw
+     * stacktrace instead of the sanitized version.
+     */
+    public static volatile boolean fullStackTrace;
+    
+    public OnSubscribeOnAssembly(OnSubscribe<T> source) {
+        this.source = source;
+        this.stacktrace = createStacktrace();
+    }
+    
+    static String createStacktrace() {
+        StackTraceElement[] stes = Thread.currentThread().getStackTrace();
+
+        StringBuilder sb = new StringBuilder("Assembly trace:");
+        
+        for (StackTraceElement e : stes) {
+            String row = e.toString();
+            if (!fullStackTrace) {
+                if (e.getLineNumber() <= 1) {
+                    continue;
+                }
+                if (row.contains("RxJavaHooks.")) {
+                    continue;
+                }
+                if (row.contains("OnSubscribeOnAssembly")) {
+                    continue;
+                }
+                if (row.contains(".junit.runner")) {
+                    continue;
+                }
+                if (row.contains(".junit4.runner")) {
+                    continue;
+                }
+                if (row.contains(".junit.internal")) {
+                    continue;
+                }
+                if (row.contains("sun.reflect")) {
+                    continue;
+                }
+                if (row.contains("java.lang.Thread.")) {
+                    continue;
+                }
+                if (row.contains("ThreadPoolExecutor")) {
+                    continue;
+                }
+                if (row.contains("org.apache.catalina.")) {
+                    continue;
+                }
+                if (row.contains("org.apache.tomcat.")) {
+                    continue;
+                }
+            }
+            sb.append("\n at ").append(row);
+        }
+        
+        return sb.append("\nOriginal exception:").toString();
+    }
+
+    @Override
+    public void call(Subscriber<? super T> t) {
+        source.call(new OnAssemblySubscriber<T>(t, stacktrace));
+    }
+    
+    static final class OnAssemblySubscriber<T> extends Subscriber<T> {
+
+        final Subscriber<? super T> actual;
+        
+        final String stacktrace;
+        
+        public OnAssemblySubscriber(Subscriber<? super T> actual, String stacktrace) {
+            super(actual);
+            this.actual = actual;
+            this.stacktrace = stacktrace;
+        }
+
+        @Override
+        public void onCompleted() {
+            actual.onCompleted();
+        }
+
+        @Override
+        public void onError(Throwable e) {
+            e = new AssemblyStackTraceException(stacktrace, e);
+            actual.onError(e);
+        }
+
+        @Override
+        public void onNext(T t) {
+            actual.onNext(t);
+        }
+        
+    }
+}
--- a/src/main/java/rx/plugins/RxJavaHooks.java
+++ b/src/main/java/rx/plugins/RxJavaHooks.java
@@ -15,6 +15,8 @@
  */
 package rx.plugins;
 
+import java.lang.Thread.UncaughtExceptionHandler;
+
 import rx.*;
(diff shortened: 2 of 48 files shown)


src/main/java/rx/internal/operators/OnSubscribeOnAssembly.java
```diff
@@ -0,0 +1,128 @@
...
+import rx.exceptions.AssemblyStackTraceException;
+
+/**
+ * Captures the current stack when it is instantiated, makes
+ * it available through a field and attaches it to all
+ * passing exception.
+ *
+ * @param <T> the value type
+ */
+public final class OnSubscribeOnAssembly<T> implements OnSubscribe<T> {
```
artem-zinnatullin, line 29: 👍

artem-zinnatullin:
👍 // More tests would be great, especially for removing parts of stacktrace, but otherwise LGTM.

akarnokd:
> // More tests would be great, especially for removing parts of stacktrace, but otherwise LGTM.

Do you have tips how to fake stacktrace for `Thread.currentThread().getStacktrace()` ? Otherwise, there might be other common frameworks that could be filtered, such as Android stuff.

artem-zinnatullin:
Accept `StackTraceElement[]` in the function and pass fake `StackTraceElement[]` in tests?

akarnokd:
If this PR gets a go ahead, I'll do that.

artem-zinnatullin:
@hzsweers PTAL

ZacSweers:
Will look when I can, but probably won't be until next week due to work projects.

akarnokd:
Also it is possible to merge RxJavaHooks into RxJavaPlugins. It would avoid some confusion when switching to 2.x plugins API.

artem-zinnatullin:
> Also it is possible to merge RxJavaHooks into RxJavaPlugins. It would avoid some confusion when switching to 2.x plugins API.

Don't you afraid that it'll make API of `RxJavaPlugins` too confusing?

ZacSweers:
How would it make it more confusing?

artem-zinnatullin:
Either we'll need to break current API or add very similar methods to
existing which will make it confusing.

akarnokd:
It's an odd class either way. Plus, I've seen mentioned that few tests actually extend `RxJavaPlugins`. I'm leaning towards having `RxJavaHooks` separate and clearly marked `@Experimental`.

akarnokd:
Updated with more Javadoc.

akarnokd:
I'd like to merge this so we can make progress on the coverage/PMD front (which will indicate what tests are needed for this). Any objections?

ZacSweers:
I'll take a good look at it today if you don't mind waiting a few hours

akarnokd:
Sure, take your time.

src/main/java/rx/plugins/RxJavaHooks.java
```diff
@@ -0,0 +1,944 @@
...
+public final class RxJavaHooks {
+    /** Utility class. */
+    private RxJavaHooks() {
+        throw new IllegalStateException("No instances!");
+    }
+    
+    /**
+     * Prevents changing the hook callbacks when set to true.
+     */
+    /* test */ static volatile boolean lockdown;
```
ZacSweers, line 44: Does this mean it's intended for test-use only?

src/test/java/rx/plugins/RxJavaHooksTest.java
```diff
@@ -0,0 +1,165 @@
...
+                    }
+                    
+                    Object after = getter.invoke(null);
+                    
+                    Assert.assertSame(m.toString(), before, after);
+                }
+            }
+            
+        } finally {
+            RxJavaHooks.lockdown = false;
```
ZacSweers, line 161: It seems like this should have a method for setting in tests, package-local is a little annoying unless this is intended only for RxJava tests. Is it?

ZacSweers:
Could you include an example of what the output from `AssemblyStackTraceException` looks like, for posterity? Doc might be a bit overkill, but it would be nice to see what it looks like

src/main/java/rx/plugins/RxJavaHooks.java
akarnokd, line 44: To test if the lockdown actually works, this has to be locked down but then unlocked to allow other tests setting their hooks after that.

akarnokd:
In the IDE, this is what it looks like when watching the exception (its message):

```
rx.exceptions.AssemblyStackTraceException: Assembly trace:
 at rx.Observable.create(Observable.java:92)
 at rx.Observable.lift(Observable.java:239)
 at rx.Observable.map(Observable.java:6321)
 at rx.plugins.RxJavaHooksTest.createObservable(RxJavaHooksTest.java:31)
 at rx.plugins.RxJavaHooksTest.assemblyTrackingObservable(RxJavaHooksTest.java:45)
 at java.lang.reflect.Method.invoke(Method.java:498)
Original exception:
```

If a test fails, Eclipse's JUnit window can nicely jump to these sources.

akarnokd:
If necessary, we can remove more lines (such as Observable.xxx()) but they might be helpful while diagnosing problems in RxJava itself.

src/test/java/rx/plugins/RxJavaHooksTest.java
akarnokd, line 161: The `lockdown` logic isn't mandatory part of the class; if it feels confusing or want to avoid accidental lockdowns, it can be removed from this PR.

ZacSweers, line 161: My main concern is that a big part of `Schedulers.reset()` was to be able to set them on the fly. As long as things like `setOnComputationScheduler` work without having to change that flag first (which it looks like that's the case), I'm happy :)

ZacSweers:
I think the trace is fine as it is. Good to see it can pick it up as sources that are jumpable as well. We've been using a sort of `TaggedObserver` pattern to help distinguish traces where something bubbles up from an `OnErrorNotImplementedException` and it looks like this could help a lot in tracing the actual assembly.

👍  for me as long as setting schedulers dynamically still works.

akarnokd:
All the old hooks and reset still work as this delegates to those by default (their original unit tests still pass).

ZacSweers:
Would it be possible to use this new API for the same effect though? (changing schedulers on the fly, mainly for testing purposes)

akarnokd:
You can change the schedulers on the fly via `setOnComputationScheduler()`, etc.

ZacSweers:
Sounds good :+1:

akarnokd:
Thanks for the reviews!

PR #4011, opened 2016-06-16
1.x: Deprecate TestObserver, enhance TestSubscriber a bit

akarnokd:
This PR deprecates `TestObserver` in favor of the richer `TestSubscriber`.

In addition, `TestSubscriber` gets 3 new methods and 1 deprecation:
- `getCompletions()` to return the onCompleted count as int instead of Notifications
- `getValueCount()` returns the committed number of onNext events for thread-safe checking of values up to this count
- `awaitValueCount` repeatedly sleeps up to a timeout and waits till the committed onNext count reaches/passes the expected amount.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -20,6 +20,7 @@ import java.util.concurrent.*;
 
 import rx.*;
 import rx.Observer;
+import rx.annotations.Experimental;
 import rx.exceptions.CompositeException;
 
 /**
@@ -29,8 +30,20 @@ import rx.exceptions.CompositeException;
  */
 public class TestSubscriber<T> extends Subscriber<T> {
 
-    private final TestObserver<T> testObserver;
+    private final Observer<T> delegate;
+    
+    private final List<T> values;
+    
+    private final List<Throwable> errors;
+    
+    /** The number of onCompleted() calls. */
+    private int completions;
+    
     private final CountDownLatch latch = new CountDownLatch(1);
+    
+    /** Written after an onNext value has been added to the {@link #values} list. */
+    private volatile int valueCount;
+    
     private volatile Thread lastSeenThread;
     /** The shared no-op observer. */
     private static final Observer<Object> INERT = new Observer<Object>() {
@@ -76,10 +89,13 @@ public class TestSubscriber<T> extends Subscriber<T> {
         if (delegate == null) {
             throw new NullPointerException();
         }
-        this.testObserver = new TestObserver<T>(delegate);
+        this.delegate = delegate;
         if (initialRequest >= 0L) {
             this.request(initialRequest);
         }
+        
+        this.values = new ArrayList<T>();
+        this.errors = new ArrayList<Throwable>();
     }
 
     /**
@@ -180,8 +196,9 @@ public class TestSubscriber<T> extends Subscriber<T> {
     @Override
     public void onCompleted() {
         try {
+            completions++;
             lastSeenThread = Thread.currentThread();
-            testObserver.onCompleted();
+            delegate.onCompleted();
         } finally {
             latch.countDown();
         }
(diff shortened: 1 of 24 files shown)


src/main/java/rx/observers/TestSubscriber.java
```diff
@@ -271,17 +318,74 @@ public void requestMore(long n) {
...
+     * @param timeout the time to wait for the events
+     * @param unit the time unit of waiting
+     * @return true if the expected number of onNext events happened
+     * @throws InterruptedException if the sleep is interrupted
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
+    public final boolean awaitValueCount(int expected, long timeout, TimeUnit unit) throws InterruptedException {
+        while (timeout != 0 && valueCount < expected) {
+            unit.sleep(1);
```
stevegury, line 361: Instead of sleeping for 1 unit, why not using an exponential backoff capped at 1 unit.
This will ensure, fast await when the value is quickly ready, without wasting too much CPU resources.
i.e. something like that:

```
long start = System.nanoTime();
long timeoutNano = unit.convert(timeout, TimeUnit.NANOSECONDS);
long sleepPeriod = 1L;
while (valueCount < expected && System.nanoTime() - start < timeoutNano) {
    TimeUnit.NANOSECONDS.sleep(sleepPeriod);
    sleepPeriod = Math.min(sleepPeriod * 2, TimeUnit.NANOSECONDS.convert(1L, unit));
}
```

stevegury:
Other than that suggestion, LGTM 👍

src/main/java/rx/observers/TestSubscriber.java
artem-zinnatullin, line 361: Another thing is that `sleep(1)` does not guarantee that it'll wake you up back in `1ms`, it may wake you later, so you need to explicitly check that expected timeout is not expired: `System.nanoTime() - start < timeoutNano`

JakeWharton, line 361: I don't agree. If you wake up with the correct number of values but
slightly late that shouldn't warrant being considered a timeout.

artem-zinnatullin, line 361: @JakeWharton well, it depends on use case, I guess somebody who wants to use `await(time)` on `TestSubscriber` either wants it to be:
- "max timeout to fail the test instead of looping infinitely", like @Timeout in JUnit. so in that case value will be someting like `1, MINUTE` while actually data will arrive much faster.
- "precise value to check some concurrency algorithm with expected timeouts" so in that case "success" long (relatively) after actual timeout can be considered as a bug.

I'd be ok with something like this:

``` java
while (true) {
  if (valueCount >= expected) {
    return true;
  }

  if (System.nanoTime() - start > timeoutNano) {
    return false;
  }

  Thread.sleep(1);
}
```

So that only flakiness of last `sleep()` will be amortized.

Current implementation increments `timeout` after each `sleep(1)` and may collect some relatively big error, like you had 150 `sleep(1)` but actually spent `200ms` which is ~25% error and seems possible in real life.

JakeWharton, line 361: The second case does not seem in scope for this class.

akarnokd, line 361: The timeout is there to not wait indefinitely. The sleep(1 unit) is there to not wait too much if the required number of values have been received. For example, testing if an 1M source observed on a scheduler and requested 500k delivers that amount within some reasonable time:

``` java
range(1, 1_000_000).observeOn(Schedulers.computation()).subscribe(ts);

ts.requestMOre(500_000);

ts.awaitValueCount(500_000, 5000, TimeUnit.MILLISECONDS);
```

On a good day, observeOn can stream 20M values per second, thus there would be 25 x 1ms sleep.

akarnokd:
Thanks for the feedback. I'm merging this as is and I'm open for PRs if you think `awaitValueCount` should use a different wait strategy.

PR #4019, opened 2016-06-17
1.x: make the latest CompositeSubscription test more robust.

akarnokd:
Make sure there is enough time for the test and the latch is always counted down.

/cc @artem-zinnatullin


stevegury:
👍

ZacSweers:
👍

Issue #4017, opened 2016-06-17
1.X: Why is CompletableSubscriber not a Subscription?


bryant1410:
I see that `CompletableSubscriber` is not a `Subscription`, but `Subcription` is. So, when I use `Completable#subscribe(CompletableSubscriber)`, I cannot then unsubscribe if I want to.

akarnokd:
`CompletableSubscriber` was designed along the Reactive-Streams pattern where cancellation support is injected via `onSubscribe` call. Consumers extending `CompletableSubscriber` then can decide individually if they want to expose this cancellation support to the outside world by implementing `Subscription` themselves. In contrast, `Subscriber` can never revoke its ability to be unsubscribed externally.

If you are using `subscribe()` or the lambda versions of `subscribe`, you get a `Subscription` back. As for the `subscribe(CompletableSubscriber)`, indeed we don't have a supporting class for it (as far as I remember). I'll post a PR for it today.

akarnokd:
Closing via #4020.

PR #4020, opened 2016-06-17
1.x: add AsyncCompletableSubscriber that exposes unsubscribe()

akarnokd:
This PR adds a default, abstract base class with unsubscription support for `Completable`'s `CompletableSubscriber`.

Pros: direct, slim class

Cons: people probably used `subscriber(Action0)` or `subscribe(Subscriber)` so far.

Related: #4017


stevegury:
👍

PR #4021, opened 2016-06-17
Improve unit test for `publish()`

DavidDTA:


ZacSweers:
You changed logic but you didn't explain why this is an improvement. Could you elaborate?

DavidDTA:
Publish should only subscribe to the underlying observable once, no matter how many subscribers it has.  This ensures that both subscriptions have taken place when verifying that only one subscription to the underlying observable has happened.

akarnokd:
Thanks @hospitalitarian, for the explanation as well.

PR #4023, opened 2016-06-17
1.x: fix timed replay() replaying old data for late subscribers

akarnokd:
In the current master, `replay` only trucates its buffer if an event is received. This means if a subscriber subscribes between two events, the operator may replay old events. This PR skips those events. The operator is now consistent with the recently updated `ReplaySubject`.

The old behavior was somewhat expected by a few unit tests that have been updated.

Related: #3917, #3918.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorReplay.java
+++ b/src/main/java/rx/internal/operators/OperatorReplay.java
@@ -1005,6 +1005,16 @@ public final class OperatorReplay<T> extends ConnectableObservable<T> {
             set(n);
         }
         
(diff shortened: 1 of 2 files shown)


stevegury:
👍 
We should mention this change in the next changelog.

akarnokd:
Definitely. 

@zsxwing Thoughts?

zsxwing:
:+1: Agreed that this is a bug fix instead of a broken change.

akarnokd:
Thanks @zsxwing !

Issue #3007, opened 2015-06-04
Javadoc for cache suggests eager subscribe


mattrjacobs:
Lines 27-28 of OnSubscribeCache:
* This method has similar behavior to {@link Observable#replay()} except that this auto-subscribes
 * to the source Observable rather than returning a connectable Observable.

It appears that the subscribe() to the source is lazy

akarnokd:
Care to submit a PR for this?

akarnokd:
See #4026 for the proposed javadoc changes.

akarnokd:
Closing via #4026.

PR #4026, opened 2016-06-17
1.x: expand and explain cache() in more detail

akarnokd:
This PR rephrases the `cache` documentation a bit and adds workaround for stopping and forgetting the cached values.

Related: #3007.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -4027,27 +4027,48 @@ public class Observable<T> {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stevegury:
👍 
This obviously isn't related to that, but `testReplaySubjectConcurrentSubscriptions` failed in CI.

akarnokd:
Thanks, will look into it before merging this; could be a bug due a pattern they share.

akarnokd:
Nothing obvious so far, the test passes for me locally when run 1000 times. Will investigate it further through #4028

Issue #4025, opened 2016-06-17
Question about onError propagation in Completable#onErrorComplete


vanniktech:
I was just browsing some of the source code to get a better understanding and I stumbled upon [https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Completable.java#L1666](this line). Isn't there a call to `s.onError(e)` missing?

akarnokd:
Good catch!

akarnokd:
See #4027.

akarnokd:
Closing via #4027

PR #4027, opened 2016-06-17
1.x: fix Completable.onErrorComplete(Func1) not relaying function crash

akarnokd:
The catch around the predicate didn't actually signal the CompositeException.

Discovered in #4025


vanniktech:
:+1:

stevegury:
👍

PR #4030, opened 2016-06-18
2.x: first step switching to the reduced-allocation architecture

akarnokd:
This is the first step towards the new architecture where operator implementations are themselves of the base type instead of a 2-3 layer indirection as in 1.x.

This PR removes the instance field `onSubscribe` from the base reactive classes and introduces `subscribeActual` to be overridden with the custom operator logic now on. Existing operators are still working through the classical OnSubscribe/Lift indirection and will be updated in subsequent PRs.

Naming and existence of `subscribeActual` is up for (post-merge) discussion. To recap, the method is needed so that the default `subscribe` method can be made final and ensure plugin hooks can be called (once available). The method is protected and only affects operator implementors that chose to extend the base reactive classes directly. 

The alternative is to leave `subscribe` abstract and add a hook to every operator method (for example, see [this](https://github.com/reactor/reactive-streams-commons/blob/master/src/main/java/rsc/publisher/Px.java#L50) and [this](https://github.com/reactor/reactive-streams-commons/blob/master/src/main/java/rsc/publisher/Px.java#L162).

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/CachedObservable.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/CachedObservable.java
@@ -33,6 +33,7 @@ public final class CachedObservable<T> extends Flowable<T> {
     /** The cache and replay state. */
     private CacheState<T> state;
 
+    private final AtomicBoolean once;
     /**
      * Creates a cached Observable with a default capacity hint of 16.
      * @param <T> the value type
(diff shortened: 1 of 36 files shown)


src/main/java/io/reactivex/internal/operators/flowable/CachedObservable.java
```diff
@@ -33,6 +33,7 @@
     /** The cache and replay state. */
     private CacheState<T> state;
 
+    private AtomicBoolean once;
```
JakeWharton: nit: `final`

JakeWharton:
lgtm :+1:

src/main/java/io/reactivex/internal/operators/flowable/CachedObservable.java
akarnokd: Thanks! Updated.

PR #4034, opened 2016-06-19
1.x: distinctUntilChanged with direct value comparator - alternative

akarnokd:
Alternative implementation to #4001.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
+++ b/src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
@@ -83,7 +83,7 @@ public final class OperatorDistinctUntilChanged<T, U> implements Operator<T, T>,
                 previousKey = key;
                 
                 if (hasPrevious) {
-                    Boolean comparison;
+                    boolean comparison;
                     
                     try {
                         comparison = comparator.call(currentKey, key);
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
```diff
@@ -68,7 +83,16 @@ public void onNext(T t) {
...
+                    Boolean comparison;
+                    
+                    try {
+                        comparison = comparator.call(currentKey, key);
+                    } catch (Throwable e) {
+                        Exceptions.throwOrReport(e, child, key);
+                        return;
+                    }
+                    
+                    if (!comparison) {
```
JakeWharton, line 95: There's a subtle NPE potential here. Not sure if you want to explicitly guard against it from a misbehaving `comparator` implementation.

src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
```diff
@@ -48,6 +50,19 @@
 
     public OperatorDistinctUntilChanged(Func1<? super T, ? extends U> keySelector) {
         this.keySelector = keySelector;
+        this.comparator = this;
```
JakeWharton, line 53: Cute trick

src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
akarnokd, line 95: Good catch! I'll change the comparison to be primitive boolean to trigger NPE inside the try-block if necessary.

JakeWharton:
lgtm :+1:

src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
akarnokd, line 95: Done.

akarnokd:
Thanks Jake for the review!

stevegury:
👍

PR #4045, opened 2016-06-20
2.x: Single operators factored out, headers added

akarnokd:
`Single` operators factored out and made them extend `Single` directly. Added missing headers.

### DIFF
--- a/src/main/java/io/reactivex/internal/subscribers/single/ConsumerSingleSubscriber.java
+++ b/src/main/java/io/reactivex/internal/subscribers/single/ConsumerSingleSubscriber.java
@@ -1,0 +1,58 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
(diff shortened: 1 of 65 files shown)


src/main/java/io/reactivex/internal/subscribers/single/ConsumerSingleSubscriber.java
```diff
@@ -0,0 +1,58 @@
...
+import io.reactivex.disposables.Disposable;
+import io.reactivex.functions.Consumer;
+import io.reactivex.internal.disposables.DisposableHelper;
+
+public final class ConsumerSingleSubscriber<T>
+extends AtomicReference<Disposable>
+implements SingleSubscriber<T>, Disposable {
+
+    /** */
+    private static final long serialVersionUID = -7012088219455310787L;
```
JakeWharton, line 28: When extending types that are `Serializable` we should probably prevent their serialization altogether with:

``` java
private Object writeReplace() {
  throw new UnsupportedOperationException();
}
```

akarnokd, line 28: I don't know. Can you give some justification?

JakeWharton, line 28: Well it's likely to fail anyway since `Consumer` isn't `Serializable`, but it makes explicit the fact that this type isn't a type that's meant to be serialized. It's only `Serializable` as a side-effect of extending another type to save memory.

I just ran over the code and changed my mind because all these instances are internal and while they're exposed in the public API it's only through non-serializable interfaces (like `SingleSubscriber` in this case). You would have to go far out of your way to even try to serialize these instances and it's not worth wasting the methods for something I'm 99.99% sure will never happen.

PR #4050, opened 2016-06-21
1.x: fixes and coverage improvements

akarnokd:
- update to JUnit 4.12
- remove deprecated use of `junit.framework.Assert`
- update to Mockito 1.10.19
- Notification.accept() now uses switch (note that Jacoco can't properly cover enum switches as it only sees the default and impossible path is not taken).
- in `Scheduler.schedulePeriodically`, move the unsubscribe check after the action to prevent unnecessary schedule of the next iteration.
- remove the inner counted loop from `SerializedObserver` as being unnecessary
- increate timeout in `OperatorSwitchTest`
- add `CompletableSubscriberTest`
- extend `Scheduler`, `SerializedObserver`,

### DIFF
--- a/src/main/java/rx/Notification.java
+++ b/src/main/java/rx/Notification.java
@@ -169,7 +169,7 @@ public final class Notification<T> {
             observer.onCompleted();
             break;
         default:
-            break;
+            throw new AssertionError("Uncovered case: " + kind);
         }
     }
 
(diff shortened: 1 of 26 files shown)


src/main/java/rx/Notification.java
```diff
@@ -158,15 +158,24 @@ public boolean isOnNext() {
...
-        } else if (isOnError()) {
+            break;
+        case OnError:
             observer.onError(getThrowable());
+            break;
+        case OnCompleted:
+            observer.onCompleted();
+            break;
+        default:
+            break;
```
JakeWharton: not sure your thoughts, but for exhaustive enum switches I prefer to `throw new AssertionError()` in the default case since it would be a programmer error for this branch to every be taken.

akarnokd: Sure.

PR #4053, opened 2016-06-21
2.x: Convert SchedulerSupport annotation to accept a string value.

### LINKED ISSUE #4052: 2.x: Change SchedulerSupport to string?
In some specific uses of RxJava it's common to have other important schedulers that one cares about. If the `@SchedulerSupport` annotation was changed to use a `String` as its value and provide constants on that type for its first-party values this would allow those other schedulers to be analyzed in the same way.

This will allow static analysis tools to enforce constraints with these "custom" schedulers as actual named-entities instead of having to rely on the otherwise opaque `CUSTOM` value. Additionally, libraries will be able to annotate their factory methods which operate on these schedulers appropriately.

I'm thinking specifically of Android, RxAndroid, and having `"mainThread"` be a value.

Happy to make this change if it sounds like a good idea.

JakeWharton:
Closes #4052.


JakeWharton:
I meant to include in the PR description that I removed the `SINGLE` value. It was unused and the Javadoc didn't make it clear as to what the exact usage would be anyway.

akarnokd:
No problem. Can be added later if necessary, although `SingleScheduler` is not (planned to be) a default for any operators right now.

PR #4058, opened 2016-06-22
Use DisposableHelper and SubscriptionHelper to reduce duplication.

JakeWharton:
A lot of operators contained their own marker instance and class definition of Disposable or Subscription. This changes almost all of them to use the shared instance provided by their respective helper class. Some duplication still exists but can be cleaned up in a subsequent change.


JakeWharton:
There's some actual bug fixes in here too. More than once an AtomicReference was being compared against a Disposable/Subscription using identity without calling get().

JakeWharton:
Maybe I made that up. I can't see them now. It's late...

akarnokd:
:+1: Thanks for this heroic work!

PR #4065, opened 2016-06-22
1.x: ignore .pmd file in root directory

akarnokd:
The PMD tool in local IDE tends to create this custom file. Add a rule to ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ bin/
 # Scala build
 *.cache
 /.nb-gradle/private/
+
+# PMD local rules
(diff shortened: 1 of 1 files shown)


vanniktech:
How about also adding `local.properties`?

akarnokd:
Isn't that Android specific?

vanniktech:
I get it when I'm opening the project via AS so yes could be.

akarnokd:
Sure. I'm merging this so please post a separate PR as I'm into something else right now.

PR #4068, opened 2016-06-22
1.x: Add PMD code checking tool to the build process

akarnokd:
This PR adds a PMD plugin to the build process to check a set of coding rules.

At this point, I don't think rule violations should fail the build.

Unfortunately, I don't know about a service, like codecov-io, that could post the check results as comments under a PR, therefore, the current build just prints the first ~100 violations into the build log:

https://travis-ci.org/ReactiveX/RxJava/builds/139539158#L240

https://travis-ci.org/ReactiveX/RxJava/builds/139539158#L5508


akarnokd:
@artem-zinnatullin @zsxwing your thoughts on this?

build.gradle
```diff
@@ -60,3 +61,42 @@ jacocoTestReport {
...
+    description = 'Run pmd'
+    group = 'verification'
+
+    reports {
+        html.enabled = true
+        xml.enabled = true
+    }
+}
+
+build.dependsOn 'pmd'
```
artem-zinnatullin: I don't think you need this since in the bottom you have `build.dependsOn 'pmdPrint'` which includes `pmd` as far as I understand

artem-zinnatullin:
👍

build.gradle
akarnokd: I'd leave it this way in case the pmdPrint gets replaced with a better feedback mechanisms and the dependency chain won't get broken easily.

build.gradle
```diff
@@ -60,3 +61,42 @@ jacocoTestReport {
...
+
+pmd {
+    toolVersion = '5.4.2'
+    ignoreFailures = true
+    sourceSets = [sourceSets.main]
+    ruleSets = []
+    ruleSetFiles = files('pmd.xml')
+}
+
+task('pmd', type: Pmd, dependsOn: 'assemble') {
```
zsxwing: This task is not necessary. Do you want a task to run manually? If so, just run `./gradlew pmdMain`.

build.gradle
```diff
@@ -60,3 +61,42 @@ jacocoTestReport {
...
+               println(line)
+            }
+        }
+    
+    } else {
+        println("PMD file not found.")
+    }
+}
+
+build.dependsOn 'pmdPrint'
```
zsxwing: nit: using `check.dependsOn pmdPrint` is better.

zsxwing:
Just some minor issues. Otherwise, 👍

akarnokd:
Updated.

build.gradle
```diff
@@ -60,3 +61,40 @@ jacocoTestReport {
...
+}
+
+pmdMain {
+    reports {
+        html.enabled = true
+        xml.enabled = true
+    }
+}
+
+build.dependsOn pmdMain
```
zsxwing, line 81: minor: this line is not necessary because of the task dependency chain: `build -> check -> pmdMain`

pmd.xml
```diff
@@ -0,0 +1,230 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ruleset xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         name="pmd"
+         xsi:schemaLocation="http://pmd.sourceforge.net/ruleset/2.0.0 http://pmd.sourceforge.net/ruleset_2_0_0.xsd">
+   <description>RxJava PMD ruleset</description>
+   <rule ref="rulesets/java/design.xml/AbstractClassWithoutAbstractMethod"/>
+   <rule ref="rulesets/java/design.xml/AbstractClassWithoutAnyMethod"/>
+   <rule ref="rulesets/java/design.xml/AccessorClassGeneration"/>
```
JakeWharton, line 9: :+1: :+1: :+1:

Now why don't they have an `AccessorMethodGeneration` check as well?!?

akarnokd, line 9: No idea, it's not in their design.xml

build.gradle
akarnokd, line 81: Okay, would you like to post a PR?

pmd.xml
JakeWharton, line 9: Filed at https://sourceforge.net/p/pmd/bugs/1496/

build.gradle
zsxwing, line 81: Sure. See #4086

vanniktech:
Should PMD also be added to the 2.x branch? The earlier it's added the earlier warnings could be fixed.

akarnokd:
Sure.

PR #4069, opened 2016-06-22
Remove second type parameter from Subject and FlowProcessor.

### LINKED ISSUE #4056: 2.x: Unify two type parameters from Subject into one?
This came up in passing in an issue or PR before (that I can't find at this moment), but the two type parameters on `Subject` seem to only be a nuisance instead of a benefit.

Should we remove them?

None of the built-in implementations allow consuming a type different than is being emitted, which makes the two type parameters on the supertype seem very out of place.

JakeWharton:
Closes #4056.


akarnokd:
:+1:

PR #4072, opened 2016-06-22
2.x: Dereference atomic ref before identity comparison.

JakeWharton:
This returns false 100% of the time otherwise.


artem-zinnatullin:
👍

akarnokd:
:+1:

PR #4076, opened 2016-06-22
2.x: Some small tweaks to Disposables.

JakeWharton:
- Use BooleanDisposable for Runnables.
- Change BooleanDisposable to use null for disposed instead of keeping a marker instance.
- Eliminate static factory for CompositeDisposable since its constructors are already public.


akarnokd:
:+1:

PR #4080, opened 2016-06-22
2.x: Replace custom empty Disposables with global instance.

JakeWharton:
I keep thinking I found all of these and then more crop up...


artem-zinnatullin:
👍

vanniktech:
There are still some empty Disposable objects in those files:

FlowableThrottleFirstTimed.java
FlowableTimeoutTimed.java
NbpOperatorThrottleFirstTimed.java
NbpOperatorTimeoutTimed.java

Should they also be replaced with EmptyDisposable.INSTANCE?

JakeWharton:
No. Those are marker instances that must be discernable from empty.

PR #4081, opened 2016-06-23
2.x: Do not leak the DISPOSED marker instance into public API.

JakeWharton:
If this were to be used by public code it could have have Very Bad™ effects by causing operators to think that they already disposed resources when they actually had not. Since `Disposable` is stateless it's safe for `disposed()` and `empty()` to be synonymous, although that might change in the future so both methods are retained.


akarnokd:
:+1:

PR #4083, opened 2016-06-23
2.x: Merge CancelledSubscriber into EmptySubscriber.

JakeWharton:


akarnokd:
I think we could remove the two entirely as they are no longer needed by operators (previously, CancelledSubscriber was needed due to lift).

EmptySubscribe -> TestSubscriber (make a local copy)

CancelledSubscriber -> FlowableRefCountTest (make a local copy)

Subscribers -> remove both

akarnokd:
Merging this for now and work out their existence later.

PR #4085, opened 2016-06-23
2.x: Merge NbpCancelledSubscriber into NbpEmptySubscriber.

JakeWharton:
The NBP version of #4083.


akarnokd:
I haven't switched the Nbp stuff to direct-create yet so these have dependencies, but should be removed the same way #4083. I'm merging this to avoid conflicts in Nbp till then. :+1: Thanks!

PR #4087, opened 2016-06-23
2.x: Normalize request validation using SubscriptionHelper.

JakeWharton:


JakeWharton:
This method is super weird with `true` indicating _invalid_ and `false` indicating _valid_.

akarnokd:
I agree, looked like a good idea back then to flip. I think it should be straight: true means valid. Would you like to do the flip?

JakeWharton:
Yep. I'll add it to my list.

PR #4088, opened 2016-06-23
1.x: fix some flaky tests, add more timeout to others

akarnokd:
Some tests are sensitive to Travis hiccups. This adds more timeout to some and reduces heavyness on others.


ZacSweers:
Nit: Maybe worth pulling this out to a constant rather than having to update all the locations each time?

Otherwise 👍

akarnokd:
@hzsweers sure, your PR is welcome at any time.

PR #4090, opened 2016-06-23
Rename Throwable parameter from exception to throwable

ataulm:
- Exception extends Throwable so it's incorrect to generalize the Throwable as `exception`

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -1366,7 +1366,7 @@ public class Observable<T> {
      *  <dd>{@code error} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      * 
-     * @param exception
+     * @param throwable
      *            the particular Throwable to pass to {@link Observer#onError onError}
      * @param <T>
      *            the type of the items (ostensibly) emitted by the Observable
@@ -1374,8 +1374,8 @@ public class Observable<T> {
      *         the Observer subscribes to it
      * @see <a href="http://reactivex.io/documentation/operators/empty-never-throw.html">ReactiveX operators documentation: Throw</a>
      */
-    public static <T> Observable<T> error(Throwable exception) {
-        return create(new OnSubscribeThrow<T>(exception));
+    public static <T> Observable<T> error(Throwable throwable) {
+        return create(new OnSubscribeThrow<T>(throwable));
     }
 
     /**


vanniktech:
There are a lot more of those cases:

```
src/main/java/rx/Completable.java
458:     * Creates a Completable instance that emits the given Throwable exception to subscribers.

src/main/java/rx/exceptions/OnErrorThrowable.java
37:    private OnErrorThrowable(Throwable exception) {
43:    private OnErrorThrowable(Throwable exception, Object value) {

src/main/java/rx/internal/operators/OnSubscribeThrow.java
30:    private final Throwable exception;
32:    public OnSubscribeThrow(Throwable exception) {

src/main/java/rx/internal/operators/OperatorOnBackpressureBuffer.java
163:        public void complete(Throwable exception) {

src/main/java/rx/internal/util/BackpressureDrainManager.java
65:        void complete(Throwable exception);
73:    protected Throwable exception;

src/main/java/rx/Observable.java
1379:    public static <T> Observable<T> error(Throwable exception) {

src/main/java/rx/Single.java
506:    public static <T> Single<T> error(final Throwable exception) {

src/test/java/rx/internal/operators/OperatorDematerializeTest.java
53:        Throwable exception = new Throwable("test");

src/test/java/rx/internal/operators/OperatorMergeTest.java
1048:        Throwable exception = new Throwable();
1061:        Throwable exception = new Throwable();
1085:        Throwable exception = new Throwable();
1099:        Throwable exception = new Throwable();

src/test/java/rx/SingleTest.java
888:        Throwable exceptionFromOnErrorAction = new IllegalStateException();
1012:        Throwable exceptionFromAction = new IllegalStateException();
```

akarnokd:
See the discussion in #1353. Renaming such public parameters may cause problems in other languages.

ataulm:
@akarnokd Thanks for the link.

It seems the discussion there ended saying that it's ok to start renaming:

> that in terms of Java we are not breaking anything and this is RxJava, not RxKotlin or RxScala…

and you mentioned it will be tackled before RC1 of 2.x:

> Let's rename parameters based on @DavidMGross suggestions then.

Should they all be done in the same PR?

akarnokd:
If you are just renaming, we can review it in large batches; plus, saves you the administration cost of setting up PRs. So it would be great you followed the suggestions #1353.

ataulm:
Sounds good, let me close this PR and I'll open one when I have more time this week.

PR #4091, opened 2016-06-23
1.x: apply fixes based on PMD suggestions

akarnokd:
Some notable changes:
- move fields to the beginning of a class
- remove unnecessary modifiers and initializers
- avoid creating objects to early
- update most drain algorithm to use the (more modern) `while (e != r)` pattern
- PMD has a few incorrect checks: complaining about a final local that gets used in an inner class
- adding braces to ifs
- documenting deliberately empty methods
- fix internal field and method namings
- delegate to `Exceptions.propagate`


akarnokd:
Travis doesn't pick up this one for some reason. Reopening.

akarnokd:
https://travis-ci.org/ReactiveX/RxJava/builds/139734219#L240

artem-zinnatullin:
I'll review it in 24 hours if you don't mind

akarnokd:
Sure. I'll return to 2.x till then.

src/main/java/rx/internal/operators/BlockingOperatorNext.java
```diff
@@ -151,9 +151,6 @@ public void remove() {
         private final BlockingQueue<Notification<? extends T>> buf = new ArrayBlockingQueue<Notification<? extends T>>(1);
         final AtomicInteger waiting = new AtomicInteger();
 
-        NextObserver() {
```
JakeWharton, line 154: Removing this adds a synthetic accessor method.

src/main/java/rx/internal/operators/OperatorToObservableSortedList.java
```diff
@@ -100,13 +103,8 @@ public void onNext(T value) {
         child.setProducer(producer);
         return result;
     }
-    // raw because we want to support Object for this default
-    @SuppressWarnings("rawtypes")
-    private static Comparator DEFAULT_SORT_FUNCTION = new DefaultComparableFunction();
 
     private static class DefaultComparableFunction implements Comparator<Object> {
-        DefaultComparableFunction() {
```
JakeWharton, line 108: Removing this adds a synthetic accessor method. Hopefully PMD implements my check for these...

src/main/java/rx/internal/schedulers/ImmediateScheduler.java
```diff
@@ -41,9 +42,6 @@ public Worker createWorker() {
 
         final BooleanSubscription innerSubscription = new BooleanSubscription();
 
-        InnerImmediateScheduler() {
```
JakeWharton, line 44: Another synthetic accessor method if removed

src/main/java/rx/internal/schedulers/TrampolineScheduler.java
```diff
@@ -47,9 +47,6 @@ private TrampolineScheduler() {
         private final BooleanSubscription innerSubscription = new BooleanSubscription();
         private final AtomicInteger wip = new AtomicInteger();
 
-        InnerCurrentThreadScheduler() {
```
JakeWharton, line 50: Synthetic accessor method

src/main/java/rx/internal/util/UtilityFunctions.java
```diff
@@ -118,8 +122,6 @@ public Boolean call(Object o) {
             Func8<T0, T1, T2, T3, T4, T5, T6, T7, R>,
             Func9<T0, T1, T2, T3, T4, T5, T6, T7, T8, R>,
             FuncN<R> {
-        NullFunction() {
```
JakeWharton, line 121: Synthetic accessor method

src/main/java/rx/schedulers/TestScheduler.java
```diff
@@ -54,8 +58,6 @@ public String toString() {
     }
 
     private static class CompareActionsByTime implements Comparator<TimedAction> {
-        CompareActionsByTime() {
```
JakeWharton, line 57: Synthetic accessor method

src/main/java/rx/schedulers/TestScheduler.java
```diff
@@ -135,9 +134,6 @@ public Worker createWorker() {
 
         private final BooleanSubscription s = new BooleanSubscription();
 
-        InnerTestScheduler() {
```
JakeWharton, line 138: Synthetic accessor method

src/main/java/rx/internal/operators/OperatorToObservableSortedList.java
artem-zinnatullin, line 108: If you already have that rule written maybe we can apply it? 
- http://pmd.sourceforge.net/pmd-4.3/howtowritearule.html
- https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.PmdExtension.html#org.gradle.api.plugins.quality.PmdExtension:ruleSetConfig

JakeWharton:
Some of these I really like. Some I don't. Documenting empty methods seems like pure noise.

src/main/java/rx/internal/schedulers/ImmediateScheduler.java
akarnokd, line 44: Would it help if I remove the `private` from the parent class?

JakeWharton, line 44: Yep! Making the whole class package scope will eliminate it as well. I saw
that you did this a few times in this PR as well.

akarnokd, line 44: Great, I'll update the PR.

akarnokd:
Updated.

akarnokd:
@JakeWharton we can iterate on the ruleset after this gets merged.

stevegury:
👍 
I agree with @JakeWharton, adding a one-word comment to empty methods does not add substantial information. We should disable this warning, which doesn't prevent us to document empty methods where it makes sense.

akarnokd:
Okay then. If you don't mind, let's have the comments in this PR as is and have a separate PR disable the rule.

src/main/java/rx/BackpressureOverflow.java
```diff
@@ -24,6 +24,30 @@
 @Experimental
 public final class BackpressureOverflow {
 
+    /**
+     * By default, signal a MissingBackressureException due to lack of requests.
+     */
+    public static final BackpressureOverflow.Strategy ON_OVERFLOW_DEFAULT = Error.INSTANCE;
```
artem-zinnatullin: Would be nice to move this below all other strategies and change assignment to `ON_OVERFLOW_DEFAULT = ON_OVERFLOW_ERROR`  so it'll be more clear that default strategy is same as `ON_OVERFLOW_ERROR`

akarnokd: Maybe, how about in a separate PR after this?

stevegury:
👍 Sound good!

src/main/java/rx/BackpressureOverflow.java
artem-zinnatullin: np

src/main/java/rx/Completable.java
```diff
@@ -377,7 +379,7 @@ public static Completable create(CompletableOnSubscribe onSubscribe) {
         requireNonNull(onSubscribe);
         try {
             return new Completable(onSubscribe);
-        } catch (NullPointerException ex) {
+        } catch (NullPointerException ex) { // NOPMD by akarnokd on 2016.06.23. 10:54
```
artem-zinnatullin: will be great to avoid such comments by tweaking pmd rules

src/main/java/rx/Single.java
```diff
@@ -2589,7 +2593,7 @@ public void call(SingleSubscriber<? super T> singleSubscriber) {
      */
     @SuppressWarnings("unchecked")
     static <T> Single<? extends T>[] iterableToArray(final Iterable<? extends Single<? extends T>> singlesIterable) {
-        final Single<? extends T>[] singlesArray;
+        Single<? extends T>[] singlesArray;
```
artem-zinnatullin, line 2596: PMD complained about final variable?

src/main/java/rx/exceptions/CompositeException.java
```diff
@@ -192,14 +192,14 @@ public void printStackTrace(PrintWriter s) {
      *            stream to print to
      */
     private void printStackTrace(PrintStreamOrWriter s) {
-        StringBuilder bldr = new StringBuilder();
-        bldr.append(this).append("\n");
+        StringBuilder bldr = new StringBuilder(128);
```
artem-zinnatullin, line 195: 👍

src/main/java/rx/Completable.java
artem-zinnatullin: or at least remove "by akarnokd on …", we have git history for that

src/main/java/rx/functions/Functions.java
```diff
@@ -34,7 +34,7 @@ private Functions() {
             @Override
             public R call(Object... args) {
                 if (args.length != 0) {
-                    throw new RuntimeException("Func0 expecting 0 arguments.");
+                    throw new IllegalArgumentException("Func0 expecting 0 arguments.");
```
artem-zinnatullin, line 37: 👍

akarnokd:
Updated: removed my name, swapped the BackpressureStategy constants, disabled empty method/constructor/catch block rules.

artem-zinnatullin:
👍

src/main/java/rx/Single.java
```diff
@@ -2410,10 +2412,12 @@ public void onNext(T t) {
         Observer<T> observer = new Observer<T>() {
             @Override
             public void onCompleted() {
+                // deliberately ignored
             }
 
             @Override
             public void onError(Throwable e) {
```
ZacSweers, line 2419: Not sure if this is covered in PMD, but usually like to also name the param `ignored`. Same for the others. I suppose it's kind of orthogonal to the comment here though, which is that the method itself is deliberately unimplemented. I'd recommend specifying that it's unimplemented rather than "ignored", since "ignored" usually is in reference to the param(s)

src/main/java/rx/internal/util/IndexedRingBuffer.java
```diff
@@ -47,6 +47,10 @@
  * @param <E>
  */
 public final class IndexedRingBuffer<E> implements Subscription {
+    private final ElementSection<E> elements = new ElementSection<E>();
+    private final IndexSection removed = new IndexSection();
+    /* package for unit testing */final AtomicInteger index = new AtomicInteger();
```
ZacSweers, line 52: Maybe worthwhile to import IntelliJ annotations or create a project annotation `@VisibleForTesting`?

src/test/java/rx/util/AssertObservableTest.java
```diff
@@ -31,13 +31,23 @@ public void testPassNull() {
         AssertObservable.assertObservableEqualsBlocking("foo", null, null);
     }
 
-    @Test(expected = RuntimeException.class)
+    @Test
```
ZacSweers, line 34: PMD doesn't like expected failures in test annotations?

ZacSweers:
Some minor nits, LGTM 👍

src/test/java/rx/util/AssertObservableTest.java
JakeWharton, line 34: They don't validate what is actually expected to fail nor let you assert anything about the exception. try/catch is 10x better.

ZacSweers, line 34: Makes sense

akarnokd, line 34: It didn't like throwing RuntimeExceptions elsewhere.

src/main/java/rx/internal/util/IndexedRingBuffer.java
akarnokd, line 52: Don't want to depend on external libs in main.

akarnokd:
Thanks @hzsweers!

src/main/java/rx/internal/util/IndexedRingBuffer.java
ZacSweers, line 52: That's fair. I still think it wouldn't hurt to just make one internally. Cleaner than comments and can still be picked up by IDEs anyway

PR #4093, opened 2016-06-23
1.x: exclude some text files from license checks

akarnokd:
By default, the `license` plugin warns about MDs and text files. In addition, this excluded the JCTools queues and the beta/experimental annotations because they kept their original license header.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,6 +49,10 @@ test{
      maxHeapSize = "2g"
 }
 
+license {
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
Can we configure it to fail build if some file doesn't have license so we won't forget to add it? 

// As far as I understand it's [license-gradle-plugin](https://github.com/hierynomus/license-gradle-plugin).

akarnokd:
Yes, but not now as it will fail 2 classes I fixed in another PR.

artem-zinnatullin:
ok, 👍

PR #4094, opened 2016-06-23
2.x: Remove EmptySubscriber, start to remove NbpEmptySubscriber.

JakeWharton:
Follow up from comments in #4083 and #4085.


src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
```diff
@@ -575,4 +575,21 @@ public void accept(String t1) {
         System.out.println(intervalSubscribed.get());
         assertEquals(6, intervalSubscribed.get());
     }
+
+    private enum CancelledSubscriber implements Subscriber<Integer> {
```
artem-zinnatullin, line 579: `CancellingSubscriber`? Like `DisposingObserver`. It's not cancelled, but it's behavior-subscruber that cancels subscriptions.

My English is not that good so maybe it's false positive from me.

artem-zinnatullin:
👍

akarnokd:
:+1:

PR #4097, opened 2016-06-23
1.x: update map() and filter() to implement OnSubscribe directly

akarnokd:
This change reduces the indirection and allocation count when using `map` or `filter`.


stevegury:
👍

akarnokd:
Thanks @stevegury ! Could you give your opinion about #4098 as well?

artem-zinnatullin:
👍 

// great that we have enough tests and coverage reports to perform such rewrites!

Issue #3164, opened 2015-08-19
Code Coverage and Static Analysis Tools


stealthcode:
The developers on RxJava have historically avoided adding new tools to the project in the interests of keeping the release and build processes as simple as possible. Today this includes Gradle, Nebula, JMH, and JUnit. So, if this impacts you we'd like to know. What tools do you like to use and how can we help make your hacking experience more productive? Can they be easily integrated and optionally executed or would they require integrating into the release process? Note that it's very unlikely that we'll choose to add complexity to the build process without a compelling reason. 

Right now our guidelines for assessing tools related pull requests are roughly the following:

* For changes which integrate a new tool which is not essential to the release process there should be no changes to the `build.gradle`. 
* Auxiliary tools that require gradle changes must remain unlinked to the build system. They should be optionally included as an init script via the `-I` gradle parameter on an as-needed basis by the developer.
* Runtime code should not /suffer/ for the benefit of an auxiliary tool. This includes adding mental overhead (readability, complexity) or reducing the efficiency of crucial computational paths. 

Hope to hear your ideas and suggestions.

akarnokd:
Closing this as we added code coverage to the build process. Apart for indicating what extra unit tests are needed, it already helped discovering bugs by highlighting code sections that never run but should.

vanniktech:
What about static analyze tools though such as PMD / Findbugs / Checkstyle?

akarnokd:
My personal opinion about PMD / Findbugs is that they tend to give too many false positives on RxJava. As for Checkstyle, we didn't have a strict code style so far and were forgiving in many cases.

As with the code coverage, the best would be if you or somebody else posted a PR with the tools added to the build process to see how it works out.

vanniktech:
I could prepare branches with those. Should I target 1.x or 2.x?

akarnokd:
Start with 1.x as it builds faster.

PR #4092, opened 2016-06-23
1.x: add AnimalSniffer to the build process, fix and suppress violations

akarnokd:
This PR adds the AnimalSniffer plugin to check for Java 6 API violations.

Related issue: #4067.

PR #4099, opened 2016-06-23
1.x: Add Findbugs

vanniktech:
Continuing from #3164 .

Parsing the generated Findbugs output file it not as trivial as with PMD since it does contain some meta information. 

Let me know whether this is wanted or not and if so how it can be pursued. 

By default all rules from Findbugs are enabled and one can opt out using `excludeFilter = file('findbugs-filter.xml)` inside the `findbugs` block, which will then contain something alone the lines of [this](http://findbugs.sourceforge.net/manual/filter.html).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,10 @@
 buildscript {
-    repositories { 
-        jcenter() 
+    repositories {
+        jcenter()
     }
-    dependencies { 
+    dependencies {
         classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:4.0.0'
-        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.1.0' 
+        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.1.0'
     }
 }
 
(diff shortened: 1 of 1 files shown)


akarnokd:
Thanks! I'd like to merge #4091 and #4092 first as rebasing this PR is much easier.

akarnokd:
Running FindBugs locally shows me 31 problems. Half of them look valid, the other half can be ignored (although I don't see any obvious way of suppressing a warning at a specific source line).

vanniktech:
Suppressing them is kind of weird AFAIK you can't use `@SuppressWarnings`, however there's a package, `com.google.code.findbugs:annotations:3.0.1` which gives you [`@SuppressFBWarnings`](http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.html).

artem-zinnatullin:
Is there an option to suppress them by XML file? (Yeah it's annoying, but
should help).

vanniktech:
AFAIK you can just use the `excludeFilter` and then inside ignore certain things (the same way we did it in QualityMatters with Butterknife).

akarnokd:
Could you rebase this?

vanniktech:
Done.

akarnokd:
:+1:

PR #4102, opened 2016-06-24
2.x: Expose isDisposed method on Disposable.

### LINKED ISSUE #4075: 2.x: Expose Disposable.isDisposed()?
In `Disposable.java` currently:

``` java
// TODO let's see if we really need this
// boolean isDisposed();
```

I think we should add this since `dispose()` already states:

> Dispose the resource, the operation should be idempotent.

Which means every disposable already needs a mechanism for idempotence which should trivially translate into `isDisposed()`.

There's some massive refactors that can be done if this change is made. `BooleanDisposable` can completely disappear from the public API since `Disposables.from(Runnable)` + `Disposables.empty()` mimic its behavior. Additionally a large portion of the other internal, custom disposables can be rewritten to build on an internal `BooleanDisposable` which manages both idempotence and the `isDisposed()` implementation for you.

I did some of this work last night as an experiment and it looks like a good win. The argument against the refactor would be if there's a reason we don't actually want to expose `isDisposed()` on `Disposable` directly. I'll note that `BooleanDisposable` is not unique in having an `isDisposed()` method. Most other implementations also have this method, they simply don't share it from the interface.

JakeWharton:
This is a fairly "dumb" addition. There's lots of room for follow-up improvement in the utilization of base helper classes and even removing some now-duplicated state management.

Closes #4075.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -877,7 +877,7 @@ public abstract class Observable<T> implements ObservableConsumable<T> {
         return create(new ObservableConsumable<Integer>() {
             @Override
             public void subscribe(Observer<? super Integer> s) {
-                BooleanDisposable d = new BooleanDisposable();
+                Disposable d = Disposables.empty();
                 s.onSubscribe(d);
                 
                 long end = start - 1L + count;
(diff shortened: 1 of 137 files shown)


akarnokd:
There are a bunch of compilation errors: https://travis-ci.org/ReactiveX/RxJava/builds/139948754#L315

JakeWharton:
Probably missing imports. I have to edit those by hand in each file so the
diffs don't go wild.

akarnokd:
:+1: Thanks.

PR #4103, opened 2016-06-24
2.x: prepare the operator-fusion enhancement

akarnokd:
This PR adds 3 operator-fusion related interfaces: 
- `QueueSubscription` will support fusion for `Flowable` - avoid creating queues, avoid cost of multiple enqueue-dequeue, reduce `request()` overhead
- `QueueDisposable` will support fusion for `Observable` - avoid creating queues, avoid cost of multiple enqueue-dequeue
- `ScalarCallable` - indicate a scalar constant source

In addition `ConditionalSubscriber` is marked and will be moved to the new package and its method renamed to `tryOnNext`.

Unlike Reactor-Core (and Rsc), there is no `Fuseable` marker interface. The ability of fusing is indicated by calling `onSubscribe` with the above `QueueX` enabled instances and then called back via `requestFusion`. This allows having less duplicated code compared to Reactor/Rsc - at the cost of usually one extra field to store the `QueueX` value that may be null.

### DIFF
--- a/src/main/java/io/reactivex/internal/fuseable/QueueDisposable.java
+++ b/src/main/java/io/reactivex/internal/fuseable/QueueDisposable.java
@@ -1,0 +1,116 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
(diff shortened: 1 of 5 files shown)


artem-zinnatullin:
Nice :+1:

PR #4106, opened 2016-06-24
2.x: Remove explicit cancel flags where it can be inferred.

JakeWharton:
I suspect there's quite a few of these all over. These are just the ones I noted while doing other changes.


akarnokd:
:+1:

PR #4107, opened 2016-06-24
2.x: make sure the helper returns true if the pre-swap value was null

akarnokd:
The original behavior was somewhat inconsistent: if the target was null before the cancel/dispose, the method didn't return true indicating a successful swap for the current thread. Now that many `cancelled` flag has been removed, it has to consistently trigger the remaining cleanup actions.


JakeWharton:
:+1:

PR #4108, opened 2016-06-24
2.x: Invert subscription validation return type for better semantics.

JakeWharton:


JakeWharton:
Similar to #4095.

akarnokd:
:+1:

PR #4110, opened 2016-06-24
2.x: Remove explicit cancel flags where it can be inferred.

JakeWharton:


stevegury:
Code looks ok, but is the CI failure related to your changes?

JakeWharton:
I see it locally so it must be. Will look later tonight and try to figure out which is the cause.

src/main/java/io/reactivex/internal/operators/flowable/FlowableInterval.java
```diff
@@ -74,19 +72,12 @@ public void request(long n) {
...
-        
-        void disposeResource() {
             DisposableHelper.dispose(resource);
         }
-        
+
         @Override
         public void run() {
-            if (!cancelled) {
+            if (resource.get() == DisposableHelper.DISPOSED) {
```
akarnokd: `!=`

PR #4111, opened 2016-06-24
2.x: Simplify boolean disposables which run code on dispose.

JakeWharton:


JakeWharton:
There's some classes in `internal` which could probably also extend from this type, so we might want to move it to `internal.disposables` and make it public for their use in the future.

JakeWharton:
Another option to make this more widely applicable would be to have a single class and accept a `Consumer<T> disposer`. That might open it up for use in more places internally.

akarnokd:
You can experiment with that if you wish.
:+1:

PR #4113, opened 2016-06-24
2.x: Parallelize unit test execution for speed!

JakeWharton:
This gives me a 57% reduction in execution time.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,8 +39,12 @@ if (project.hasProperty('release.useLastTag')) {
     tasks.prepare.enabled = false
 }
 
-test{
-     maxHeapSize = "2g"
+test {
+    maxHeapSize = "2g"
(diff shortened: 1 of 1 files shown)


JakeWharton:
Looks like Travis can't handle it. I can disable it for CI builds where build speed isn't really important.

akarnokd:
This increases the likelihood of timeouts during heavy times. Maybe later.

akarnokd:
In addition, some tests may conflict because they may change the global hooks. I'm guessing there isn't a JUnit annotation that allows some tests to run in parallel with others while forbidding for others?

JakeWharton:
They run in forked JVMs so any static state won't clash.

akarnokd:
Okay, let's try it.

PR #4119, opened 2016-06-25
1.x: increase coverage of internal utils, remove unused/unnecessary items

akarnokd:
Increase coverage of internal utils, remove unused and unnecessary items.


src/test/java/rx/internal/util/UtilityFunctionsTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.lang.reflect.*;
+
+import org.junit.Test;
+
+public class UtilityFunctionsTest {
+    @Test
+    public void testNotInstantiable() {
+        try {
+            Constructor<?> c = UtilityFunctions.class.getDeclaredConstructor();
```
vanniktech, line 28: For this we could use https://github.com/pushtorefresh/java-private-constructor-checker from @artem-zinnatullin

akarnokd, line 28: There was a discussion about @artem-zinnatullin 's tool in #3130. The verdict was to not include that depencency. If you want, you can implement a minimalistic method that takes a class, but we don't need the full flexibility.

vanniktech, line 28: Personally I don't see the benefit of recreating that mechanism / check instead of just adding a `testCompile` dependency. Also Guava was recently added as a test dependency.

akarnokd, line 28: PR is welcome.

JakeWharton, line 28: I don't see the benefit of testing this at all. This isn't even a public class! We're testing that an internal type is uninstantiable which has no bearing on the usage of the library and is already enforced by PMD.

PR #4125, opened 2016-06-25
1.x: Test all private Constructors

vanniktech:


src/main/java/rx/exceptions/Exceptions.java
```diff
@@ -31,7 +31,7 @@
 
     /** Utility class, no instances. */
     private Exceptions() {
-        throw new IllegalStateException("No instances!");
+        throw new AssertionError("No instances.");
```
akarnokd: Why the change to AssertionError?

vanniktech: I find it that this exception matches the use case a bit better since we're asserting that this constructor shall never be called.

src/test/java/rx/EventStreamTest.java
```diff
@@ -0,0 +1,27 @@
...
+package rx;
+
+import com.pushtorefresh.private_constructor_checker.PrivateConstructorChecker;
+
+import org.junit.Test;
+
+public class EventStreamTest {
+    @Test
+    public void constructorShouldBePrivate() {
+        PrivateConstructorChecker.forClass(EventStream.class).expectedTypeOfException(AssertionError.class).expectedExceptionMessage("No instances.").check();
```
akarnokd: These could be refactored into a `TestHelper` enum:

``` java
public enum TestHelper {
    ;
   public static void check(Class<?> clazz) {
        PrivateConstructorChecker.forClass(clazz)
        .expectedTypeOfException(IllegalStateException.class)
        .expectedExceptionMessage("No instances!").check();
    }
}
```

src/main/java/rx/exceptions/Exceptions.java
akarnokd: I find these as unnecessary changes. The methods throw to those who instantiate the class via reflection and are not test methods themselves to fail with AssertionError like `TestSubscriber.assertX`.

akarnokd:
:+1: Great!

/cc @artem-zinnatullin

artem-zinnatullin:
👍 

Now I'm not that fanatic regarding testing private constructors, but since test coverage causes addiction and constant desire to increase it — let's merge it!

davidmoten:
Can I suggest in the spirit of DRY that you make a static method available somewhere (a new class `TestingUtil`, say) like this:

``` java
public static void checkConstructorPrivate(Class<?> cls)  {
    PrivateConstructorChecker
         .forClass(cls)
        .expectedTypeOfException(IllegalStateException.class)
        .expectedExceptionMessage("No instances!").check();
}
```

and call that instead?

akarnokd:
Thanks!

@davidmoten I'll refactor those checks into a simple call as you and [I](https://github.com/ReactiveX/RxJava/pull/4125#discussion_r68494047) suggested.

PR #4126, opened 2016-06-25
Remove XResource classes

akarnokd:
This PR removes the `XResource` types and replaces them with `XDisposable` or `XSubscription` where necessary.


src/main/java/io/reactivex/disposables/SubscriptionDisposable.java
```diff
@@ -15,6 +15,9 @@
 import org.reactivestreams.Subscription;
 
 final class SubscriptionDisposable extends ReferenceDisposable<Subscription> {
+    /** */
+    private static final long serialVersionUID = -707001650852963139L;
```
JakeWharton, line 19: What are these for? We shouldn't be encouraging people to serialize these instances.

akarnokd, line 19: Eclipse complains about serializable classes (i.e., extending AtomicX) need these fields, even though we don't care about serialization. A small inconvenience.

PR #4130, opened 2016-06-26
2.x: run Findbugs on main only

akarnokd:
Don't run findbugs for the perf and test packages.


vanniktech:
This should also be adjusted in 1.x right?

akarnokd:
Yes, running on test sometimes takes so long travis kills the build.

PR #4133, opened 2016-06-27
Refactor private constructor checker into a utility class

akarnokd:
Move into `TestUtil.checkUtilityClass()`.

### DIFF
--- a/src/test/java/rx/TestUtil.java
+++ b/src/test/java/rx/TestUtil.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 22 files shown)


vanniktech:
👍

src/test/java/rx/TestUtil.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package rx;
+
+import com.pushtorefresh.private_constructor_checker.PrivateConstructorChecker;
+
+/**
+ * Common test utility methods.
+ */
+public enum TestUtil {
```
egor-n, line 24: Just curious - why use an enum here instead of a class? Is it just to forbid inheritance?

akarnokd, line 24: Exactly - but not my invention. I saw this in the JDK source and in fact, saves you testing a private constructor. However, we can't expose such enums through the public API so only some internal tools are defined this way.

davidmoten:
:+1:

PR #4134, opened 2016-06-27
2.x: cleanup and test renames

akarnokd:
This PR adds the missing `@Override` annotations and renames the `Observable` unit tests to conform a single name pattern.


src/main/java/io/reactivex/disposables/Disposable.java
```diff
@@ -23,6 +23,7 @@
 
     /**
      * Returns true if this resource has been disposed.
```
vanniktech, line 25: Is this line then really needed?

akarnokd, line 25: javadoc complains if the `@return` tag is missing or empty.

vanniktech, line 25: Yeah I know that but then `* Returns true if this resource has been disposed.` could be removed, could not it?

akarnokd, line 25: Yes.

Issue #4138, opened 2016-06-27
2.x: FlowableReplayTest > testAsyncComeAndGo 


akarnokd:
Could be an bug with `replay()` or the fact that it goes lockstep and doesn't request in time if the further downstream (observeOn) is slow.

```
io.reactivex.internal.operators.flowable.FlowableReplayTest > testAsyncComeAndGo FAILED
    java.lang.AssertionError: Error(s) present: [io.reactivex.exceptions.MissingBackpressureException: Can't deliver value 236 due to lack of requests]
        at io.reactivex.subscribers.TestSubscriber.fail(TestSubscriber.java:397)
        at io.reactivex.subscribers.TestSubscriber.assertNoErrors(TestSubscriber.java:465)
        at io.reactivex.internal.operators.flowable.FlowableReplayTest.testAsyncComeAndGo(FlowableReplayTest.java:1022)
        Caused by:
        io.reactivex.exceptions.CompositeException
            at io.reactivex.subscribers.TestSubscriber.fail(TestSubscriber.java:398)
            ... 2 more
            Caused by:
            io.reactivex.exceptions.CompositeException$CompositeExceptionCausalChain: Chain of Causes for CompositeException In Order Received =>
                Caused by:
                io.reactivex.exceptions.MissingBackpressureException: Can't deliver value 236 due to lack of requests
```

akarnokd:
Closing via #4139

akarnokd:
It's [back](https://travis-ci.org/ReactiveX/RxJava/builds/144412254#L3408).

akarnokd:
Closing again. Haven't encountered this for a while.

Issue #4109, opened 2016-06-24
2.x: NbpOperatorConcatTest > testIssue2890NoStackoverflow FAILED


akarnokd:
Looks like a timeout issue.

```
io.reactivex.internal.operators.observable.NbpOperatorConcatTest > testIssue2890NoStackoverflow FAILED
    java.lang.AssertionError: expected:<5000> but was:<3797>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at io.reactivex.internal.operators.observable.NbpOperatorConcatTest.testIssue2890NoStackoverflow(NbpOperatorConcatTest.java:730)
```

akarnokd:
Closing via #4139

Issue #4054, opened 2016-06-21
2.x: Flaky test NbpOperatorWindowWithSizeTest.testWindowUnsubscribeOverlappingAsyncSource


JakeWharton:
```
io.reactivex.internal.operators.observable.NbpOperatorWindowWithSizeTest > testWindowUnsubscribeOverlappingAsyncSource FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at io.reactivex.internal.operators.observable.NbpOperatorWindowWithSizeTest.testWindowUnsubscribeOverlappingAsyncSource(NbpOperatorWindowWithSizeTest.java:188)
```

From https://travis-ci.org/ReactiveX/RxJava/builds/139345308

JakeWharton:
https://travis-ci.org/ReactiveX/RxJava/builds/139590872

Also the non-overlapping test failed this time

```
io.reactivex.internal.operators.observable.NbpOperatorWindowWithSizeTest > testWindowUnsubscribeNonOverlappingAsyncSource FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at io.reactivex.internal.operators.observable.NbpOperatorWindowWithSizeTest.testWindowUnsubscribeNonOverlappingAsyncSource(NbpOperatorWindowWithSizeTest.java:145)
```

akarnokd:
A small hiccup and the sourc range runs to completion, failing the assumption it will be cancelled in time. Having no backpressure doesn't really help. I'm thinking about commenting out those two final checks in the methods.

akarnokd:
Didn't happen for some time now.

PR #4139, opened 2016-06-27
2.x: fix flaky tests

akarnokd:
This PR should fix some test timing out otherwise sensitive to thread hiccups.
#4138
#4109
#4054


JakeWharton:
:+1:

PR #4140, opened 2016-06-27
1.x: Change Completable.subscribe(onError, onComplete) to (onComplete, onError)

### LINKED ISSUE #3851: Completable.subscribe(onError, onComplete) design issue
This [particular overload](https://github.com/ReactiveX/RxJava/blob/3c5efaa9442ee24227b63ad039eb965b59aad774/src/main/java/rx/Completable.java#L1916) of `Completable.subscribe()` looks out of design in compare to other RxJava classes like `Observable` and `Single`.

Spent ~5 minutes trying to understand why such code didn't compile:

```java
Completable
  .fromAction(() -> doSomething())
  .subscribe(
    () -> ui.success(), 
    error -> ui.error(error)
  );
```

And the reason is because this overload accepts `error` handler **first** and `complete` handler second.

I do understand that it comes from `Observable.subscribe(onNext, onError, onComplete)`. 

But in compare to many other overloads like:

* `Observable.subscribe(onNext, onError)`
* `Single.subscribe(onSuccess, onError)` // this one has very similar semantic to target overload.
* `Observable.subscribe(onNext)`
* `Single.subscribe(onSuccess)`

Error handler is **never first** parameter in `Observable.subscribe()` and `Single.subscribe()`.

Another point is that compiler error displayed in IDE makes it even worse:
screen shot 2016-04-12 at 00 35 50

-------

`Completable` is still in `@Experimental`, so, we can change this signature as we want. We can `@Deprecate` this overload and add "better" alternative and then delete deprecated overload after one-two releases.

### LINKED ISSUE #4137: Inverted onError and onComplete in Completable
In completable subscribe method there is inverted order of methods (in compare to ex. Observable). In every other classes there is onNext/onSucces first, and in Completable there is onError first.

artem-zinnatullin:
Closes #3851, closes #4137.

### DIFF
--- a/src/test/java/rx/CompletableTest.java
+++ b/src/test/java/rx/CompletableTest.java
@@ -2567,15 +2567,15 @@ public class CompletableTest {
     public void subscribeTwoCallbacksNormal() {
         final AtomicReference<Throwable> err = new AtomicReference<Throwable>();
         final AtomicBoolean complete = new AtomicBoolean();
-        normal.completable.subscribe(new Action1<Throwable>() {
-            @Override
-            public void call(Throwable e) {
-                err.set(e);
-            }
-        }, new Action0() {
+        normal.completable.subscribe(new Action0() {
             @Override
             public void call() {
                 complete.set(true);
+            }
+        }, new Action1<Throwable>() {
+            @Override
+            public void call(Throwable e) {
+                err.set(e);
             }
         });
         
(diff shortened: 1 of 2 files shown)


src/test/java/rx/CompletableTest.java
```diff
@@ -2605,44 +2605,44 @@ public void call() {
...
+        normal.completable.subscribe(null, new Action1<Throwable>() {
             @Override
-            public void call() { }
+            public void call(Throwable throwable) {}
         });
     }
     
     @Test(expected = NullPointerException.class)
     public void subscribeTwoCallbacksSecondNull() {
-        normal.completable.subscribe(null, new Action0() {
```
artem-zinnatullin, line 2616: Didn't catch that on initial PR, test was obviously copied from `subscribeTwoCallbacksFirstNull` and wasn't changed, second parameter should have been `null`.

artem-zinnatullin:
Actually, I can deprecate old one and add new one as overload, it will have less impact on user code because only calls like `subscribe(Actions.empty(), Actions.empty())` will be ambiguous to the compiler.

Thoughts?

akarnokd:
:+1:

akarnokd:
Dynamic languages have trouble with the two methods, let's only have 1 method.

zsxwing:
👍

PR #4143, opened 2016-06-28
Applied skeletal implementation to interface migration refactoring

khatchad:
This is a semantics-preserving refactoring.
- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative and you should not observe many source code changes.
- The source code should be semantically equivalent to the original.
- We have run tests prior to applying to the tool and following the application. All tests pass.

Thank you for your help in this evaluation!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,6 +11,10 @@ buildscript {
 description = 'RxJava: Reactive Extensions for the JVM – a library for composing asynchronous and event-based programs using observable sequences for the Java VM.'
 
 apply plugin: 'java'
+	compileJava {
+		sourceCompatibility = '1.8'
+		targetCompatibility = '1.8'
+	}
 apply plugin: 'pmd'
 apply plugin: 'findbugs'
 apply plugin: 'jacoco'
--- a/src/main/java/rx/Observer.java
+++ b/src/main/java/rx/Observer.java
@@ -34,7 +34,9 @@ public interface Observer<T> {
      * <p>
      * The {@link Observable} will not call this method if it calls {@link #onError}.
      */
-    void onCompleted();
+    default void onCompleted() {
+	
+	}
 
     /**
      * Notifies the Observer that the {@link Observable} has experienced an error condition.
@@ -45,7 +47,9 @@ public interface Observer<T> {
      * @param e
      *          the exception encountered by the Observable
      */
-    void onError(Throwable e);
+    default void onError(Throwable e) {
+	    e.printStackTrace();
+	}
 
     /**
      * Provides the Observer with a new item to observe.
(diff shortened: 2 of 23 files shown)


build.gradle
```diff
@@ -11,6 +11,10 @@ buildscript {
 description = 'RxJava: Reactive Extensions for the JVM – a library for composing asynchronous and event-based programs using observable sequences for the Java VM.'
 
 apply plugin: 'java'
+	compileJava {
+		sourceCompatibility = '1.8'
```
artem-zinnatullin, line 15: RxJava can't target Java 8, it's Java 6+ library, mostly because of Android.

khatchad, line 15: Ah, okay, thanks for the consideration!

src/main/java/rx/Observer.java
```diff
@@ -34,7 +34,9 @@
      * <p>
      * The {@link Observable} will not call this method if it calls {@link #onError}.
      */
-    void onCompleted();
+    default void onCompleted() {
```
akarnokd, line 37: No. We want them to implement `onCompleted`.

src/main/java/rx/Observer.java
```diff
@@ -45,7 +47,9 @@
      * @param e
      *          the exception encountered by the Observable
      */
-    void onError(Throwable e);
+    default void onError(Throwable e) {
+	    e.printStackTrace();
```
akarnokd, line 51: Same as with `onCompleted()`. Plus, this wouldn't be the default we'd choose.

src/main/java/rx/Producer.java
```diff
@@ -47,6 +47,8 @@
      *          want the Producer to produce items at its own pace
      * @throws IllegalArgumentException if the request amount is negative
      */
-    void request(long n);
+    default void request(long n) {
```
akarnokd, line 50: Makes no sense to have a default on this interface method.

akarnokd:
I'm sorry, these are unacceptable changes. RxJava is a library, targeting Java 6, so even beyond the `default` changes, we want those methods to be implemented.

khatchad:
@akarnokd Thank you for the feedback!

src/main/java/rx/Producer.java
khatchad: Is that because you want to force clients to implement this method?

akarnokd: This is a single method interface with the main purpose of getting implemented. Having it default empty will not work with "add unimplemented methods" quick fix for once.

PR #4146, opened 2016-06-29
Fix the initialization of Completable.complete()

akarnokd:
I forgot to invert the noHook -> useHook change.


akarnokd:
The related tests failed for me due to test ordering; if they were the very first encounters of `Completable`.

PR #4151, opened 2016-06-30
2.x: enable scalar and async fuseable sources and intermediates

akarnokd:
This PR enables operator fusion for scalar and knowingly single-element sources.

There are no fusion-consumers yet.


akarnokd:
@JakeWharton, @artem-zinnatullin want to take a look?

artem-zinnatullin:
Yes, in 24 hours if you don't mind

src/test/java/io/reactivex/single/SingleTest.java
```diff
@@ -188,7 +188,10 @@ public void testFlatMap() {
             }
         }
         ).subscribe(ts);
-        ts.awaitTerminalEvent();
+        if (!ts.awaitTerminalEvent(5, TimeUnit.SECONDS)) {
```
JakeWharton, line 191: you can also use `assertTrue(ts.await...)` for handy failures when it doesn't time out

JakeWharton:
lgtm :+1:

src/test/java/io/reactivex/single/SingleTest.java
akarnokd, line 191: I need to call `ts.cancel()` otherwise the test leaks and affects other tests.

stevegury:
👍

src/main/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElements.java
```diff
@@ -27,27 +31,145 @@ public FlowableIgnoreElements(Publisher<T> source) {
...
+        }
+        
+        @Override
+        public void onComplete() {
+            actual.onComplete();
+        }
+        
+        @Override
+        public boolean add(T e) {
+            throw new UnsupportedOperationException("Should not be called!");
```
artem-zinnatullin, line 72: Maybe add `e` to exception so it will be easier to find the problem?

src/main/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElements.java
```diff
@@ -27,27 +31,145 @@ public FlowableIgnoreElements(Publisher<T> source) {
...
+        }
+
+        @Override
+        public boolean offer(T e) {
+            throw new UnsupportedOperationException("Should not be called!");
```
artem-zinnatullin, line 77: same as above

src/main/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElements.java
```diff
@@ -27,27 +31,145 @@ public FlowableIgnoreElements(Publisher<T> source) {
...
+        }
+
+        @Override
+        public boolean isEmpty() {
+            return true;
+        }
+
+        @Override
+        public boolean contains(Object o) {
+            throw new UnsupportedOperationException("Should not be called!");
```
artem-zinnatullin, line 112: same as above

src/main/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElements.java
```diff
@@ -27,27 +31,145 @@ public FlowableIgnoreElements(Publisher<T> source) {
...
+        }
+
+        @Override
+        public Object[] toArray() {
+            throw new UnsupportedOperationException("Should not be called!");
+        }
+
+        @Override
+        public <U> U[] toArray(U[] a) {
+            throw new UnsupportedOperationException("Should not be called!");
```
artem-zinnatullin, line 127: same as above

artem-zinnatullin, line 127: same for functions below

akarnokd, line 127: It doesn't matter what the value was but where the call was attempted.

src/main/java/io/reactivex/internal/operators/single/SingleToFlowable.java
```diff
@@ -33,27 +34,44 @@ public SingleToFlowable(SingleConsumable<? extends T> source) {
...
+
+        @Override
+        public void onError(Throwable e) {
+            actual.onError(e);
+        }
+
+        @Override
+        public void cancel() {
+            super.cancel();
+            d.dispose();
```
artem-zinnatullin, line 74: nit: npe is possible here, though state overall won't be ok, but npe is still seems possibe

akarnokd, line 74: No, because the downstream can only call `cancel` after it received `this` on line 57.

artem-zinnatullin, line 74: got it, ok 👍

src/main/java/io/reactivex/internal/subscriptions/DeferredScalarSubscription.java
```diff
@@ -0,0 +1,158 @@
...
+import org.reactivestreams.Subscriber;
+
+/**
+ * A subscription that signals a single value eventually.
+ * <p>
+ * Note that the class leaks all methods of {@link java.util.concurrent.atomic.AtomicLong}.
+ * Use {@link #complete(Object)} to signal the single value.
+ * @param <T> the value type
+ */
+public class DeferredScalarSubscription<T> extends BasicQueueSubscription<T> {
```
artem-zinnatullin, line 25: This class looks pretty complicated, can you please cover it with tests? https://codecov.io/gh/ReactiveX/RxJava/src/4f2eecee404238b49c4ac0f2f074ed6d63938231/src/main/java/io/reactivex/internal/subscriptions/DeferredScalarSubscription.java

akarnokd, line 25: Half of it only works if there is a fusion consumer (none yet) and the second half is dependent on probabilistic race.

artem-zinnatullin, line 25: But why not just test it in isolation (unit testing)?

artem-zinnatullin:
👍

src/main/java/io/reactivex/internal/subscriptions/DeferredScalarSubscription.java
akarnokd, line 25: New tests will be added before RC1 based on missing tests from 1.x and code coverage.

akarnokd:
Thanks to everyone for the review.

Issue #4115, opened 2016-06-25
Throttling Observables


abersnaze:
We have a problem where multiple code paths are making different calls to a backend service but we want to throttle the number of concurrent requests that each client is making. The problem is that different requests could be different types or without knowledge of the other happening something like this:

``` java
Service.getA()
    .subscribe(processResponseA())
```

I was thinking that it would be cool to have a ThrottlingScheduler and/or a MaxConcurrentScheduler that could limit the number of Observables that are active without having to interfere with the data of any of them and without having to make a new thread pool.

``` java
static Scheduler serviceLimiter = Schedulers.maxConcurrent(2, Schedulers.computation());

Service.getA()
    .subscribeOn(serviceLimiter)
    .subscribe(processResponseA())
```

davidmoten:
So the completion of the subscribe action upstream of the call to `subscribeOn(serviceLimiter)` would release back to the pool?

abersnaze:
I was thinking `rx.Scheduler.Worker.unsubscribe()` would be when the next Observable in the queue would be subscribed too.

akarnokd:
With the existing API, you can submit Observables to a PublishSubject+flatMap with maxConcurrency of 2:

``` java
    Observable<T1> t1 = Observable.fromCallable(() -> 1).observeOn(Schedulers.computation());
    Observable<T1> t2 = ...

    Subject<Observable, Observable> ps = PublishSubject.create().toSerialized(); // raw Observable

    ps.flatMap(v -> v.ignoreElements(), 2).subscribe(() -> { }, e -> log(e));

    ps.onNext(t1);
    ps.onNext(t2);
```

abersnaze:
I was thinking of using mergeMaxConcurrent inside the scheduler implementation. I would map createWorker() to Observable.create and onNext to a subject. Then map worker.unsubscribe() to onComplete(). While this strange Worker + Observable thing is subscribed to it schedules it's actions on the underlying actual Scheduler. I'm not sure if schedule() would be mapped on onNext or if any uncaught exceptions would be mapped to onError. It might be interesting to think about.

Having the user do all the subject stuff would mean significant constraint of all code paths to only make at most one call the service. The result of each service call Observable would need to be returned and sent through a publish subject for throttling. For example how would could this code be rewritten to only run 2 at a time across the whole of the VM.

```
void foo() {
   return zip(Service.getA(), Service.getB(), Service.getC(), (a,b,c) -> {...});
}
```

[...]

PR #4154, opened 2016-06-30
Ability to create custom schedulers with behavior based on composing operators.

abersnaze:
based on #4115

Allows the use of operators for controlling the timing around when actions scheduled on workers are actually done. This makes it possible to layer additional behavior on an existing `Scheduler`. The main parameter is a function that flattens an `Observable<Completable>` into just one `Completable`. There must be a chain of operators connecting the returned value to the source Observable otherwise any work scheduled on the Scheduler will not be executed.

When `scheduler.createWorker()` is invoked a `Completable` is onNext'd to the combinator to be flattened. If the `Completable` is not immediately subscribed to an calls to `worker.schedule(...)` are buffered. Once the `Completable` is subscribed to actions are then scheduled on the actual Scheduler. When the `Worker` is unsubscribed the `Completable` emits an onComplete and triggers any behavior in the flattening operator. The `Observable<Completable>` and all `Completable`s give to the flattening function never onError.

Slowing down the rate to no more than than 1 a second.

``` java
 Scheduler slowSched = new ScheduleWhen(workers -> {
    // use concatenate to make each worker happen one at a time.
    return Completable.concat(workers.map(worker -> {
        // delay the starting of the next worker by 1 second.
        return worker.delay(1, TimeUnit.SECONDS);
    }));
 }, Schedulers.computation());
```

Limit the amount concurrency two at a time without creating a new fix size thread pool:

``` java
 Scheduler limitSched = new ScheduleWhen(workers -> {
    // use merge max concurrent to two at a time
    return Completable.merge(workers, 2);
 }, Schedulers.computation());
```

### DIFF
--- a/src/main/java/rx/Scheduler.java
+++ b/src/main/java/rx/Scheduler.java
@@ -17,7 +17,10 @@ package rx;
 
 import java.util.concurrent.TimeUnit;
 
+import rx.annotations.Experimental;
 import rx.functions.Action0;
+import rx.functions.Func1;
+import rx.internal.schedulers.SchedulerWhen;
 import rx.schedulers.Schedulers;
 import rx.subscriptions.MultipleAssignmentSubscription;
 
@@ -182,4 +185,79 @@ public abstract class Scheduler {
         return System.currentTimeMillis();
     }
 
+	/**
+	 * Allows the use of operators for controlling the timing around when
+	 * actions scheduled on workers are actually done. This makes it possible to
+	 * layer additional behavior on this {@link Scheduler}. The only parameter
+	 * is a function that flattens an {@link Observable} of {@link Observable}
+	 * of {@link Completable}s into just one {@link Completable}. There must be
+	 * a chain of operators connecting the returned value to the source
+	 * {@link Observable} otherwise any work scheduled on the returned
+	 * {@link Scheduler} will not be executed.
+	 * <p>
+	 * When {@link Scheduler#createWorker()} is invoked a {@link Observable} of
+	 * {@link Completable}s is onNext'd to the combinator to be flattened. If
+	 * the inner {@link Observable} is not immediately subscribed to an calls to
+	 * {@link Worker#schedule} are buffered. Once the {@link Observable} is
+	 * subscribed to actions are then onNext'd as {@link Completable}s.
+	 * <p>
+	 * Finally the actions scheduled on the parent {@link Scheduler} when the
+	 * inner most {@link Completable}s are subscribed to.
+	 * <p>
+	 * When the {@link Worker} is unsubscribed the {@link Completable} emits an
+	 * onComplete and triggers any behavior in the flattening operator. The
+	 * {@link Observable} and all {@link Completable}s give to the flattening
+	 * function never onError.
+	 * <p>
+	 * Limit the amount concurrency two at a time without creating a new fix
+	 * size thread pool:
+	 * 
+	 * <pre>
+	 * Scheduler limitSched = Schedulers.computation().when(workers -> {
+	 * 	// use merge max concurrent to limit the number of concurrent
+	 * 	// callbacks two at a time
+	 * 	return Completable.merge(Observable.merge(workers), 2);
+	 * });
+	 * </pre>
+	 * <p>
+	 * This is a slightly different way to limit the concurrency but it has some
+	 * interesting benefits and drawbacks to the method above. It works by
+	 * limited the number of concurrent {@link Worker}s rather than individual
+	 * actions. Generally each {@link Observable} uses its own {@link Worker}.
+	 * This means that this will essentially limit the number of concurrent
+	 * subscribes. The danger comes from using operators like
+	 * {@link Observable#zip(Observable, Observable, rx.functions.Func2)} where
+	 * subscribing to the first {@link Observable} could deadlock the
+	 * subscription to the second.
+	 * 
+	 * <pre>
+	 * Scheduler limitSched = Schedulers.computation().when(workers -> {
+	 * 	// use merge max concurrent to limit the number of concurrent
+	 * 	// Observables two at a time
+	 * 	return Completable.merge(Observable.merge(workers, 2));
+	 * });
+	 * </pre>
+	 * 
+	 * Slowing down the rate to no more than than 1 a second. This suffers from
+	 * the same problem as the one above I could find an {@link Observable}
+	 * operator that limits the rate without dropping the values (aka leaky
+	 * bucket algorithm).
+	 * 
+	 * <pre>
+	 * Scheduler slowSched = Schedulers.computation().when(workers -> {
+	 * 	// use concatenate to make each worker happen one at a time.
+	 * 	return Completable.concat(workers.map(actions -> {
+	 * 		// delay the starting of the next worker by 1 second.
+	 * 		return Completable.merge(actions.delaySubscription(1, TimeUnit.SECONDS));
+	 * 	}));
+	 * });
+	 * </pre>
+	 * 
+	 * @param combine
+	 * @return
+	 */
+    @Experimental
+	public <S extends Scheduler & Subscription> S when(Func1<Observable<Observable<Completable>>, Completable> combine) {
+		return (S) new SchedulerWhen(combine, this);
+	}
 }
--- a/src/main/java/rx/internal/schedulers/SchedulerWhen.java
+++ b/src/main/java/rx/internal/schedulers/SchedulerWhen.java
@@ -1,0 +1,302 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.schedulers;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import rx.Completable;
+import rx.Completable.CompletableOnSubscribe;
+import rx.Completable.CompletableSubscriber;
+import rx.Scheduler.Worker;
+import rx.Observable;
+import rx.Observer;
+import rx.Scheduler;
+import rx.Subscription;
+import rx.annotations.Experimental;
+import rx.functions.Action0;
+import rx.functions.Func1;
+import rx.internal.operators.BufferUntilSubscriber;
+import rx.observers.SerializedObserver;
+import rx.subjects.PublishSubject;
+import rx.subscriptions.Subscriptions;
+
+/**
+ * Allows the use of operators for controlling the timing around when actions
+ * scheduled on workers are actually done. This makes it possible to layer
+ * additional behavior on this {@link Scheduler}. The only parameter is a
+ * function that flattens an {@link Observable} of {@link Observable} of
+ * {@link Completable}s into just one {@link Completable}. There must be a chain
+ * of operators connecting the returned value to the source {@link Observable}
+ * otherwise any work scheduled on the returned {@link Scheduler} will not be
+ * executed.
+ * <p>
+ * When {@link Scheduler#createWorker()} is invoked a {@link Observable} of
+ * {@link Completable}s is onNext'd to the combinator to be flattened. If the
+ * inner {@link Observable} is not immediately subscribed to an calls to
+ * {@link Worker#schedule} are buffered. Once the {@link Observable} is
+ * subscribed to actions are then onNext'd as {@link Completable}s.
+ * <p>
+ * Finally the actions scheduled on the parent {@link Scheduler} when the inner
+ * most {@link Completable}s are subscribed to.
+ * <p>
+ * When the {@link Worker} is unsubscribed the {@link Completable} emits an
+ * onComplete and triggers any behavior in the flattening operator. The
+ * {@link Observable} and all {@link Completable}s give to the flattening
+ * function never onError.
+ * <p>
+ * Limit the amount concurrency two at a time without creating a new fix size
+ * thread pool:
+ * 
+ * <pre>
+ * Scheduler limitSched = Schedulers.computation().when(workers -> {
+ * 	// use merge max concurrent to limit the number of concurrent
+ * 	// callbacks two at a time
+ * 	return Completable.merge(Observable.merge(workers), 2);
+ * });
+ * </pre>
+ * <p>
+ * This is a slightly different way to limit the concurrency but it has some
+ * interesting benefits and drawbacks to the method above. It works by limited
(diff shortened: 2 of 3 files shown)


src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,248 @@
...
+			public Subscription schedule(final Action0 action) {
+				// send a scheduled action to the actionQueue
+				ImmediateAction immediateAction = new ImmediateAction(action);
+				actionQueue.onNext(immediateAction);
+				return immediateAction;
+			}
+		};
+
+		// enqueue the completable that process actions put in reply subject
+		workerQueue.onNext(completable);
```
akarnokd: Workers can be created from multiple threads and this will likely mess up the internal state of the `PublishSubject`+`onBackpressureBuffer`.

src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,248 @@
...
+
+		public DelayedAction(Action0 action, long delayTime, TimeUnit unit) {
+			this.action = action;
+			this.delayTime = delayTime;
+			this.unit = unit;
+		}
+
+		@Override
+		protected Subscription callActual(Worker actualWorker) {
+			return actualWorker.schedule(action, delayTime, unit);
```
akarnokd: This could delay the task way further than intended by the original caller and may lead to timeout in sensible sequences.

akarnokd:
This test changes its timing pattern when run on one of the merging-customization of this scheduler:

``` java
@Test
public void fastTasks() {
    final Scheduler sched = new ScheduleWhen(new Func1<Observable<Completable>, Completable>() {
        @Override
        public Completable call(Observable<Completable> workers) {
            return Completable.merge(workers, 2);
        }
    }, Schedulers.computation());


    Observable.range(1, 4).flatMap(new Func1<Integer, Observable<String>>() {
        @Override
        public Observable<String> call(final Integer v) {
            return Observable.interval(500, TimeUnit.MILLISECONDS, sched)
            .take(10).map(new Func1<Long, String>() {
                @Override
                public String call(Long u) {
                    return v + " refresh " + u;
                }
            });
        }
    }
    )
    .timeInterval()
    .toBlocking()
    .subscribe(new Action1<TimeInterval<String>>() {
        @Override
        public void call(TimeInterval<String> v) {
            System.out.println(v);
        }
    });
}
```

What I get is that for the first 5 seconds, I get `1 refresh x` and `2 refresh x` in 500 ms intervals, but after thet, all the `3 refresh x` and `4 refresh x` gets printed instantly.

```
TimeInterval [intervalInMilliseconds=515, value=2 refresh 0]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 0]
TimeInterval [intervalInMilliseconds=500, value=1 refresh 1]
TimeInterval [intervalInMilliseconds=0, value=2 refresh 1]
TimeInterval [intervalInMilliseconds=500, value=2 refresh 2]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 2]
TimeInterval [intervalInMilliseconds=500, value=2 refresh 3]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 3]
TimeInterval [intervalInMilliseconds=485, value=2 refresh 4]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 4]
TimeInterval [intervalInMilliseconds=516, value=2 refresh 5]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 5]
TimeInterval [intervalInMilliseconds=486, value=1 refresh 6]
TimeInterval [intervalInMilliseconds=0, value=2 refresh 6]
TimeInterval [intervalInMilliseconds=500, value=1 refresh 7]
TimeInterval [intervalInMilliseconds=0, value=2 refresh 7]
TimeInterval [intervalInMilliseconds=499, value=2 refresh 8]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 8]
TimeInterval [intervalInMilliseconds=500, value=2 refresh 9]
TimeInterval [intervalInMilliseconds=0, value=1 refresh 9]
TimeInterval [intervalInMilliseconds=516, value=3 refresh 0]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 0]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 1]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 1]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 2]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 3]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 4]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 5]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 6]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 7]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 8]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 2]
TimeInterval [intervalInMilliseconds=0, value=4 refresh 9]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 3]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 4]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 5]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 6]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 7]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 8]
TimeInterval [intervalInMilliseconds=0, value=3 refresh 9]
```

If I understand the purpose of this scheduler, especially with the `merge(workers, 2)` case, you want to limit the number of concurrently running tasks. However, the implementation blocks out further workers from executing anything until a previous worker is released. 

For example, given a service that accepts 2 connections per client, if I have to periodically execute 3 or more tasks against this service (via the `interval()` above), only the first two would execute periodically. In addition, the other 2 would slowly pile up tasks in the internal queue and if the associated sequence uses timeout, this will make it timeout.

I think the solution is to use a fixed thread pool instead: it limits the number of concurrent tasks while allowing all workers to make progress.

My second reservation is that since the `ScheduleWhen` is a self-contained class (unlike operators that need a method in `Observable`) this can live in any external library (just like `AndroidSchedulers`).

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: I'll add some synchronization.

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: Yes. The only guarantee for scheduling is that it won't happen sooner than asked.

abersnaze:
This isn't intended for long running Observables. Your example is a pathological case. Rather than scheduling on the outer timer if flatMap and schedule on the inner Observables it would not block and all three would be able to run.

There are many was to misuse this but if used in a controlled setting of a service library it can be use to protect service from bursts of traffic. 

The main point is to be able to reuse an existing scheduler with some additional behavior rather than having to create new thread pools to enforce rate limits.

abersnaze:
To address your concern I've changed the API from `Ob<Comp> -> Comp` to `Ob<Ob<Comp>> -> Comp`. This is a more direct mapping of Scheduler of Workers of Actions that give the user complete control how and when the actions are done. In the example above the actions from all workers can be merged into one stream of actions and then the merge max (N) concurrent could be applied to the actions to better simulate running on a N threads.

abersnaze:
@akarnokd I think I've addressed all the concerns regrading this PR. What can I do help get this experiment merged?

akarnokd:
Oh, sorry; was busy with something else.

Okay, let's continue with the experiment. :+1:

/cc @zsxwing @artem-zinnatullin

abersnaze:
Does 👍  mean the PR is merged now? Should we wait for a review from @zsxwing and/or @artem-zinnatullin?

akarnokd:
Let's wait for at least one of them.

src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,263 @@
...
+		Worker worker = new Worker() {
+			private final AtomicBoolean unsubscribed = new AtomicBoolean();
+
+			@Override
+			public void unsubscribe() {
+				// complete the actionQueue when worker is unsubscribed to make
+				// room for the next worker in the workerQueue.
+				if (unsubscribed.compareAndSet(false, true)) {
+					actualWorker.unsubscribe();
+					synchronized (actionQueue) {
```
zsxwing: Potential dead lock here as it will call user's Observer with a lock. You can just use SerializedSubject to wrap `actionQueue`.

zsxwing:
👍  for this experiment api. Just one comment about `synchronized (actionQueue)`.

src/test/java/rx/schedulers/ScheduleWhenTest.java
```diff
@@ -0,0 +1,197 @@
...
+	public void testAsyncMaxConcurrent() {
+		TestScheduler tSched = new TestScheduler();
+		Scheduler sched = maxConcurrentScheduler(tSched);
+		TestSubscriber<Long> tSub = TestSubscriber.create();
+
+		asyncWork(sched).subscribe(tSub);
+
+		tSub.assertValueCount(0);
+
+		tSched.advanceTimeBy(0, TimeUnit.SECONDS);
```
artem-zinnatullin: nit: static import for `SECONDS`

src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,263 @@
...
+				});
+			}
+		});
+
+		// a worker that queues the action to the actionQueue subject.
+		Worker worker = new Worker() {
+			private final AtomicBoolean unsubscribed = new AtomicBoolean();
+
+			@Override
+			public void unsubscribe() {
```
artem-zinnatullin: Needs test (it's very major part of Scheduler's behavior)

src/main/java/rx/schedulers/ScheduleWhen.java
artem-zinnatullin: +1

src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,263 @@
...
+ * 
+ * <pre>
+ * Scheduler limitSched = new ScheduleWhen(workers -> {
+ * 	// use merge max concurrent to two at a time
+ * 	return Completable.merge(workers, 2);
+ * }, Schedulers.computation());
+ * </pre>
+ */
+@Experimental
+public class ScheduleWhen extends Scheduler {
```
artem-zinnatullin: What about hiding the actual class and providing `Schedulers.when()` instead? (Not sure about the name though)

artem-zinnatullin:
Very interesting API, @abersnaze! Few comments and I'm 👍 with it.

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: The name was from following the theme of `retryWhen` and `repeatWhen` that use `Observable` as a control flow and not a data flow.

davidmoten: Nit: call the class what it is (a Scheduler) so ScheduleWhen->SchedulerWhen and +1 to adding a static method to Schedulers for discoverability

davidmoten: If add Schedulers static method then move class to internal package

src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,263 @@
...
+	private final Scheduler actualScheduler;
+	private final PublishSubject<Observable<Completable>> workerQueue;
+
+	public ScheduleWhen(Func1<Observable<Observable<Completable>>, Completable> combine, Scheduler actualScheduler) {
+		this.actualScheduler = actualScheduler;
+		// workers are converted into completables and put in this queue.
+		this.workerQueue = PublishSubject.create();
+		// send it to a custom combinator to pick the order and rate at which
+		// workers are processed.
+		combine.call(workerQueue.onBackpressureBuffer()).subscribe();
```
davidmoten: If subscribed observable emits error then may throw OnErrorNotImplemedException which could confuse. Please add tests for error case.

davidmoten: Do we have 'Scheduler' life cycle concerns with this one? Should this Scheduler have a close method to cancel this subscription so we can always clean up references?

src/main/java/rx/schedulers/ScheduleWhen.java
davidmoten: Would you add calculate the expected run time  using Scheduler.now when adding to actions queue then recalculate delay when passing to actual worker?

src/main/java/rx/schedulers/ScheduleWhen.java
```diff
@@ -0,0 +1,263 @@
...
+					actionQueue.onNext(immediateAction);
+				}
+				return immediateAction;
+			}
+		};
+
+		// enqueue the completable that process actions put in reply subject
+		synchronized (workerQueue) {
+			workerQueue.onNext(actions);
+		}
```
davidmoten: As per @zsxwing comment worker queue should be wrapped by SerializedSubject

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: There are tests that cover this code.

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: will do

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: I was thinking of this a something might be used handful of times in a VM. Should Scheduler implement Subscription?

src/main/java/rx/schedulers/ScheduleWhen.java
abersnaze: in an earlier version i was subtracting out the delay in cured from the queuing but it was a too much but i can add it back.

src/main/java/rx/Scheduler.java
```diff
@@ -182,4 +184,78 @@ public long now() {
...
+	 * 		// delay the starting of the next worker by 1 second.
+	 * 		return Completable.merge(actions.delaySubscription(1, TimeUnit.SECONDS));
+	 * 	}));
+	 * });
+	 * </pre>
+	 * 
+	 * @param combine
+	 * @return
+	 */
+	public <S extends Scheduler & Subscription> S when(Func1<Observable<Observable<Completable>>, Completable> combine) {
```
akarnokd, line 260: Experimental?

abersnaze, line 260: do'h

abersnaze:
I don't know how I would draw the marble diagram for this one.

akarnokd:
@abersnaze If all feedback is resolved, feel free to merge this PR and if you want, post any short description of this addition beyond the ref to this PR.

abersnaze:
Thanks for the feed back

davidmoten:
As far as I'm concerned the feedback cycle hadn't really finished on this. I'm hoping that api changes etc are not out of scope if I raise an issue.

davidmoten:
My comment about the use of `subscribe()` in the constructor was not addressed. Should be `unsafeSubscribe()` I think.

akarnokd:
@davidmoten If you have further concerns, don't hesitate to open an issue or post a PR. The file needs some javadoc tidying up anyway...

PR #4157, opened 2016-07-01
2.x: Enable fusion-consumers

akarnokd:
This PR enables fusion-consumers such as `observeOn`, `flatMap`, `zip`, `concatMap`; makes `UnicastProcessor` async-fuseable (used by `window`); fixes a few hidden issues with operators and has convenience renames of `SubscriptionHelper.validateX`. In addition, it features the new `FlowableFlattenIterable` used by `flatMapIterable` and `concatMapIterable`.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMap.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMap.java
@@ -13,212 +13,647 @@
 package io.reactivex.internal.operators.flowable;
 
 import java.util.Queue;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.*;
 
 import org.reactivestreams.*;
 
-import io.reactivex.Flowable;
 import io.reactivex.functions.Function;
-import io.reactivex.internal.queue.*;
+import io.reactivex.internal.functions.Objects;
+import io.reactivex.internal.fuseable.QueueSubscription;
+import io.reactivex.internal.queue.SpscArrayQueue;
 import io.reactivex.internal.subscriptions.*;
-import io.reactivex.internal.util.Pow2;
+import io.reactivex.internal.util.Exceptions;
 import io.reactivex.plugins.RxJavaPlugins;
-import io.reactivex.subscribers.SerializedSubscriber;
 
-public final class FlowableConcatMap<T, U> extends Flowable<U> {
-    final Publisher<T> source;
-    final Function<? super T, ? extends Publisher<? extends U>> mapper;
-    final int bufferSize;
-    public FlowableConcatMap(Publisher<T> source, Function<? super T, ? extends Publisher<? extends U>> mapper, int bufferSize) {
-        this.source = source;
-        this.mapper = mapper;
-        this.bufferSize = bufferSize;
+public final class FlowableConcatMap<T, R> extends FlowableSource<T, R> {
+final Function<? super T, ? extends Publisher<? extends R>> mapper;
+    
+    final int prefetch;
+    
+    final ErrorMode errorMode;
+    
+    /**
+     * Indicates when an error from the main source should be reported.
+     */
+    public enum ErrorMode {
+        /** Report the error immediately, cancelling the active inner source. */
+        IMMEDIATE,
+        /** Report error after an inner source terminated. */
+        BOUNDARY,
+        /** Report the error after all sources terminated. */
+        END
+    }
+
+    public FlowableConcatMap(Publisher<T> source,
+            Function<? super T, ? extends Publisher<? extends R>> mapper, 
+            int prefetch, ErrorMode errorMode) {
+        super(source);
+        if (prefetch <= 0) {
+            throw new IllegalArgumentException("prefetch > 0 required but it was " + prefetch);
+        }
+        this.mapper = Objects.requireNonNull(mapper, "mapper");
+        this.prefetch = prefetch;
+        this.errorMode = Objects.requireNonNull(errorMode, "errorMode");
+    }
+    
+    public static <T, R> Subscriber<T> subscribe(Subscriber<? super R> s, Function<? super T, ? extends Publisher<? extends R>> mapper, 
+            int prefetch, ErrorMode errorMode) {
+        Subscriber<T> parent = null;
+        switch (errorMode) {
+        case BOUNDARY:
+            parent = new ConcatMapDelayed<T, R>(s, mapper, prefetch, false);
+            break;
+        case END:
+            parent = new ConcatMapDelayed<T, R>(s, mapper, prefetch, true);
+            break;
+        default:
+            parent = new ConcatMapImmediate<T, R>(s, mapper, prefetch);
+        }
+        return parent;
     }
     
     @Override
-    protected void subscribeActual(Subscriber<? super U> s) {
+    public void subscribeActual(Subscriber<? super R> s) {
+        
         if (ScalarXMap.tryScalarXMapSubscribe(source, s, mapper)) {
             return;
         }
-        SerializedSubscriber<U> ssub = new SerializedSubscriber<U>(s);
-        SubscriptionArbiter sa = new SubscriptionArbiter();
-        ssub.onSubscribe(sa);
-        source.subscribe(new SourceSubscriber<T, U>(ssub, sa, mapper, bufferSize));
+        
+        source.subscribe(subscribe(s, mapper, prefetch, errorMode));
     }
-    
-    static final class SourceSubscriber<T, U> extends AtomicInteger implements Subscriber<T> {
+
+    static abstract class BaseConcatMapSubscriber<T, R> 
+    extends AtomicInteger
+    implements Subscriber<T>, ConcatMapSupport<R>, Subscription {
+        
         /** */
-        private static final long serialVersionUID = 8828587559905699186L;
-        final Subscriber<? super U> actual;
-        final SubscriptionArbiter sa;
-        final Function<? super T, ? extends Publisher<? extends U>> mapper;
-        final Subscriber<U> inner;
-        final Queue<T> queue;
-        final int bufferSize;
+        private static final long serialVersionUID = -3511336836796789179L;
+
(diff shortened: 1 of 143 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMap.java
```diff
@@ -13,212 +13,647 @@
...
-    final Publisher<T> source;
-    final Function<? super T, ? extends Publisher<? extends U>> mapper;
-    final int bufferSize;
-    public FlowableConcatMap(Publisher<T> source, Function<? super T, ? extends Publisher<? extends U>> mapper, int bufferSize) {
-        this.source = source;
-        this.mapper = mapper;
-        this.bufferSize = bufferSize;
+
+public final class FlowableConcatMap<T, R> extends FlowableSource<T, R> {
+final Function<? super T, ? extends Publisher<? extends R>> mapper;
```
JakeWharton: nit: indent

src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMap.java
```diff
@@ -13,212 +13,647 @@
...
+        this.prefetch = prefetch;
+        this.errorMode = Objects.requireNonNull(errorMode, "errorMode");
+    }
+    
+    public static <T, R> Subscriber<T> subscribe(Subscriber<? super R> s, Function<? super T, ? extends Publisher<? extends R>> mapper, 
+            int prefetch, ErrorMode errorMode) {
+        Subscriber<T> parent = null;
+        switch (errorMode) {
+        case BOUNDARY:
+            parent = new ConcatMapDelayed<T, R>(s, mapper, prefetch, false);
```
JakeWharton: you can just return from inside these cases and avoid the local var / null initializer problem.

src/main/java/io/reactivex/internal/operators/flowable/FlowableFlattenIterable.java
```diff
@@ -0,0 +1,436 @@
...
+                actual.onSubscribe(this);
+
+                s.request(prefetch);
+            }
+        }
+
+        @Override
+        public void onNext(T t) {
+            if (fusionMode != ASYNC) {
+                if (!queue.offer(t)) {
```
JakeWharton: nested `if`s could be collapsed with `&&`

src/main/java/io/reactivex/internal/operators/flowable/FlowableFlattenIterable.java
```diff
@@ -0,0 +1,436 @@
...
+                    if (m == SYNC) {
+                        fusionMode = m;
+                        this.queue = qs;
+                        done = true;
+
+                        actual.onSubscribe(this);
+
+                        return;
+                    } else
+                        if (m == ASYNC) {
```
JakeWharton: Is this wrap+indent intentional? took me a second to figure out what was going on

akarnokd: Eclipse CTRL+I sometimes does this to `else if`. I'll fix these and the previous onces but wait for you to finish the review

src/main/java/io/reactivex/internal/operators/flowable/FlowableObserveOn.java
```diff
@@ -14,209 +14,504 @@
...
+            }
+            worker.schedule(this);
+        }
+        
+        @Override
+        public final void run() {
+            if (outputFused) {
+                runBackfused();
+            } else
+            if (sourceMode == SYNC) {
```
JakeWharton: The dangling else make the code very easy to confuse with

``` java
if (outputFused) {
  runBackfused();
}
if (sourceModel == SYNC) {
  runSync();
} else {
  runAsync();
}
```

src/main/java/io/reactivex/processors/UnicastProcessor.java
```diff
@@ -38,380 +41,323 @@
...
                 
-                if (getAndIncrement() == 0) {
-                    clear(queue);
+                a.onNext(t);
+                
+                e++;
+            }
+            
+            if (r == e) {
+                if (checkTerminated(done, q.isEmpty(), a, q)) {
```
JakeWharton: nit: nested `if`s could be collapsed with `&&`

JakeWharton:
Wow this was dense! I tried my best to parse a lot of the queue/drain changes, but I definitely didn't get a full mental model going of each one since there were so many. I started pattern matching and just checking for more simple issues. I have confidence in the present and future tests for these as well, which helps.

:+1:

akarnokd:
Thanks @JakeWharton. 

It is really simple concept: instead of creating a queue, operators reuse the queue-view of their upstream, which could be queue views themselves. The drain algorithms hardly change except when `request()` has to be suppressed. The difficulty is the need to know **all** operator internals over the entire library.

src/main/java/io/reactivex/processors/UnicastProcessor.java
vanniktech: There's also a [PMD check](http://pmd.github.io/pmd-4.2.5/rules/basic.html#CollapsibleIfStatements) for this one

stealthcode:
> The difficulty is the need to know all operator internals over the entire library.

I haven't had a chance to go through your changes in any detail. Just have a question. What will this mean for `lift`, `extend`, and general extensibility use cases?

akarnokd:
@stealthcode `lift`, `to` and `compose` are still usable. There is no `extend` in 2.x because the 1.x is just simply too restrictive on its function-input; `to(Function<Flowable, X>)` gives access to all operators and you can have your favorite leave-transform function with it:

``` java
Function<Flowable<Integer>, Integer> conv = f -> f.map(a -> a + 1).toBlocking().last();

System.out.println(source.to(conv))
```

Implementing operators is now generally harder because the lifecycle of Reactive-Streams is different. Operator-fusion adds an extra requirement that you should not forward an upstream `Subscription`:

``` java
    @Override
         child.onSubscribe(s); // <- not allowed, may break the operator or fusion around it
    }
```

stealthcode:
It looks like `to` offers at least all of the functionality of `extend`. Clearly as there is no `OnSubscribe` the same API isn't possible. 

> you should not forward an upstream `Subscription`

In 2.x the disposable model forced changes in this area but not such that subscriptions couldn't be forwarded down the chain. AFAIK this was a common optimization pattern in 1.x for some operators. @abersnaze and I talked and I think my understanding is that a fuseable operator will pass along a subscription but a non-fusable operator will create it's own, is that correct?

akarnokd:
> In 2.x the disposable model forced changes in this area but not such that subscriptions couldn't be forwarded down the chain. 

That was true before operator-fusion came into picture.

> AFAIK this was a common optimization pattern in 1.x for some operators.

The architecture change induced by Reactive-Streams and operator-fusion added a lot of optimizations, surpassing those in 1.x.

> @abersnaze and I talked and I think my understanding is that a fuseable operator will pass along a subscription but a non-fusable operator will create it's own, is that correct?

No, every intermediate `Subscriber` implements `Subscription` should pass itself along, at least. If the operator doesn't want to support fusion, that's the minimum.

``` java
    @Override
         if (SubscriptionHelper.validate(this.s, s)) {
             this.s = s;
             child.onSubscribe(this);
         }
    }

    @Override
    public void request(long n) {
         s.request(n);
    }

    @Override
    public void cancel() {
         s.cancel();
    }
```

Otherwise, if an operator would pass the `Subscription` along, it could cut itself out from the dataflow as the `QueueSubscription.poll` skips over it.

akarnokd:
@artem-zinnatullin I guess you want to review this as well?

artem-zinnatullin:
@akarnokd yup, looking

src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMap.java
```diff
@@ -13,212 +13,644 @@
...
+    final Function<? super T, ? extends Publisher<? extends R>> mapper;
+    
+    final int prefetch;
+    
+    final ErrorMode errorMode;
+    
+    /**
+     * Indicates when an error from the main source should be reported.
+     */
+    public enum ErrorMode {
```
artem-zinnatullin, line 40: Maybe integer constants instead?

akarnokd, line 40: Enum is not a problem here, these are not part of a binary flag unlike the fusion constants.

artem-zinnatullin, line 40: Sure, but it'll save a little bit memory and few cpu cycles in comparisons and allocations. You don't use any runtime benefits of enum here ¯_(ツ)_/¯

akarnokd, line 40: Okay, I'll update this in the next round.

src/main/java/io/reactivex/internal/operators/flowable/FlowableFlattenIterable.java
```diff
@@ -0,0 +1,434 @@
...
+
+        volatile boolean cancelled;
+
+        final AtomicReference<Throwable> error;
+
+        Iterator<? extends R> current;
+
+        int consumed;
+
+        int fusionMode;
```
artem-zinnatullin, line 115: don't you want explicitly initialize it with `NONE`?

akarnokd, line 115: No. Saves a write and one has to be carefur not redefine `QueueSubscription.NONE`.

artem-zinnatullin:
ok, 👍, in general, pattern looks understandable but amount of code and states in it make it hard to understand…

// Don't want to complain, but will be easier to review smaller PRs :) (I know that at least I create backpressure for your PRs so I understand why you're doing such huge amounts of work in one PR, sorry for that)

akarnokd:
Thanks @artem-zinnatullin . 

Most of the time, I start small but then a change activates cascading effects: fixing a base structure exposes a bug in another operator that has to be updated.

PR #4167, opened 2016-07-05
2.x: reduce overhead of blocking first/last/single

akarnokd:
Optimize the reception, less allocation, less overhead in general.

Benchmark comparison (i7 4790, Windows 7 x64, Java 8u92)

image


JakeWharton:
lgtm :+1:

PR #4169, opened 2016-07-06
1.x: Detect concurrent emission in observeOn and throw ISE instead of MBE

artem-zinnatullin:
We've recently faced `MissingBackpressureException` which actually was not related to backpressure. 

After investigation it appeared that we were writing to `PublishSubject` from multiple threads which broke queue in `observeOn`: `queue.offer()` returned false not because it reached capacity limit but because another thread already written to exact same index.

This PR throws `IllegalStateException` instead of `MissingBackpressureException` in case if it has detected concurrent data emission + suggests user to serialize data emission to avoid this problem.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorObserveOn.java
+++ b/src/main/java/rx/internal/operators/OperatorObserveOn.java
@@ -97,6 +97,8 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
         final Queue<Object> queue;
         /** The emission threshold that should trigger a replenishing request. */
         final int limit;
+
+        final int calculatedSize;
         
         // the status of the current stream
         volatile boolean finished;
@@ -121,7 +123,7 @@ public final class OperatorObserveOn<T> implements Operator<T, T> {
             this.recursiveScheduler = scheduler.createWorker();
             this.delayError = delayError;
             this.on = NotificationLite.instance();
-            int calculatedSize = (bufferSize > 0) ? bufferSize : RxRingBuffer.SIZE;
+            calculatedSize = (bufferSize > 0) ? bufferSize : RxRingBuffer.SIZE;
             // this formula calculates the 75% of the bufferSize, rounded up to the next integer
             this.limit = calculatedSize - (calculatedSize >> 2);
             if (UnsafeAccess.isUnsafeAvailable()) {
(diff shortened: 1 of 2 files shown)


akarnokd:
No. :-1:. Don't change `observeOn`. Your detection is fragile anyway.

artem-zinnatullin:
I agree, but MBE is totally misleading in such case. What about adding a
message to MBE that it may be caused by multithreading emission?

akarnokd:
That sounds reasonable, but would require all places that use Spsc queues to be updated; plus it could be still totally misleading.

artem-zinnatullin:
Yeah, I'm open to any suggestions (I'll be out for some time and can
respond with delays).

artem-zinnatullin:
Just found another case where we were writing to `PublishSubject` concurrently **and** consumer had `onBackpressureBuffer().observeOn()`.

In that case `observeOn()` throws MBE because of concurrent write and `onBackpressureBuffer()` handles it as  MBE while it's actually illegal state…

akarnokd:
This is an issue with the use of the library. There is the `Subject.toSerialized()` that allows concurrent calls to its front.

artem-zinnatullin:
Sure, I've fixed it with serializing the subject. Reason why I posted it is to show that `MBE` hides the actual error and in our case we didn't even crash the app (I'd prefer that) and "handled" it as backpressure.

I can add another way to detect concurrent emission to `observeOn`, but it'll add a little bit overhead and I guess @akarnokd won't be happy about that so I'm closing this for now.

abersnaze:
I think the problem is that queuing operator (like observeOn) doesn't know if the chain supplying data supports back pressure or not. If is does then it is a concurrency issue if not then it's a back pressure issue.

In 1.x how about adding a 'boolean supportsBackpressure' to Observable and an overload to create() to allow setting it to true otherwise it's false?

artem-zinnatullin:
Even if chain supports backpressure, source may emit data concurrently
(which is violation of Observable contract, but easily possible in real
life code) and corrupt internal queue of the observeOn.

akarnokd:
I don't think adding those to `Observable` is a good idea; people will happily lie about their `OnSubscribe`'s backpressure support.

Instead, maybe it's time `PublishSubject`, `BehaviorSubject`, `timer()` and `interval()` (anybody else?) to start request tracking and signal `MissingBackpressureException` themselves and not rely on some queueing operator to fail later on.

In addition, depending on how severe the misuse of `PublishSubject` is, I'm willing to force serialization on its input side. Basically, I'd copy `PublishSubject` to `DirectSubject` or something for those who want to keep the overhead low and then retrofit `PublishSubject` to serialize its inputs.

artem-zinnatullin:
Yes, I definitely love your idea about making publish subject serialized
and adding something like SingleThreadPublishSubject

PR #4172, opened 2016-07-07
1.x: add missing backpressure descriptions and update old ones

akarnokd:
This PR adds or updates places where the `Backpressure:` description was missing or outdated.

Unfortunately, GitHub has a hard limit on diff size per file so reviewers have to use their own local diff.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -60,6 +60,8 @@ public class Observable<T> {
     }
 
     /**
+     * <b>This method requires advanced knowledge about building operators and data sources; please consider
+     * other standard methods first;</b> 
      * Returns an Observable that will execute the specified function when a {@link Subscriber} subscribes to
      * it.
      * <p>
@@ -75,6 +77,10 @@ public class Observable<T> {
      * See <a href="http://go.microsoft.com/fwlink/?LinkID=205219">Rx Design Guidelines (PDF)</a> for detailed
      * information.
      * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The {@code OnSubscribe} instance provided is responsible to be backpressure-aware or
+     *  document the fact that the consumer of the returned {@code Observable} has to apply one of
+     *  the {@code onBackpressureXXX} operators.</dd>
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code create} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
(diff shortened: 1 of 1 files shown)


DavidMGross:
"...it &lt;t&gt;may&lt;/i&gt; lead to IllegalStateException being throw..."
=> "...it **&lt;em&gt;**may**&lt;/em&gt;** lead to IllegalStateException being **thrown**..."

...but better yet would be to avoid the passive voice and indicate what entity throws the exception, e.g. "If any of the source Observables violate this, XXX may throw an IllegalStateException when the source Observable completes."

DavidMGross:
"...no backpressue applied to them..." => "...no **backpressure** applied to them..."

DavidMGross:
"The other Observables supplied by the function is consumed..." => "The other Observables supplied by the function **are** consumed..."

DavidMGross:
I notice that some of the `<dd>`s aren't being closed with `</dd>` also.

DavidMGross:
typo: "...apply any of the `onBackpressureXXX` opertors &lt;b&gt;before&lt;/b&gt; applying `observeOn` itself..." => "...apply any of the `onBackpressureXXX` **operators** **&lt;strong&gt;**before**&lt;/strong&gt;** applying `observeOn` itself..."

DavidMGross:
For operators that both accept and return an Observable (most of them), I find the phrase "this Observable" in the documentation to be ambiguous: does it mean the Observable about to be operated on, or the Observable that results from the operation?  I'd recommend replacing it with "the source Observable" or "the resulting Observable" or something like that, depending on which one you're talking about.

DavidMGross:
typo: "...the output's backpressure behavior is derermined by..." => "...the output's backpressure behavior is **determined** by..."

DavidMGross:
```
+     *  <dd>The operator honors backpressure from downstream and consumes this {@code Observable} in an
+     *  unbounded manner (i.e., no backpressure is applied to it).</dd>
+     *  behavior.</dd>
```

That last line seems to be extraneous.

DavidMGross:
Some of those typos appear in multiple places in the file.

It's great to see these sections getting filled out and made more precise!

akarnokd:
Thanks for the review. I've updated the text.

akarnokd:
I'm merging this so work touching `Observable.java` can commence. Anybody let me know if you find something else (or better yet, post a PR).

artem-zinnatullin:
Thank you for this @akarnokd!

davidmoten:
Beaut!

PR #4176, opened 2016-07-07
1.x: optimize collect, reduce and takeLast(1)

akarnokd:
This PR reduces the overhead with `collect`, `reduce` and `takeLast` and introduces the `DeferredScalarSubscriber` to support similar unbounded-in single-out scenarios.

Benchmark comparison (i7 4770K, Windows 7 x64, Java 8u92)

image

Relative to each other: (each practically returning the very last value)

image


src/main/java/rx/internal/operators/DeferredScalarSubscriber.java
```diff
@@ -0,0 +1,177 @@
+/**
+ * Copyright 2015 Netflix, Inc.
```
JakeWharton: 2016? The class this came from had 2016 as well

akarnokd:
Updated header year, @JakeWharton .

JakeWharton:
lgtm :+1:

PR #4178, opened 2016-07-08
1.x: improve coverage of rx.Observable methods

akarnokd:
This PR improves the coverage of `rx.Observable` methods plus 
- fixes a javadoc issue 
- fixes an enum-coverage anomaly in Notification (now it is simply biased towards onNext signals)
- removes `equals()` from `TestException` as it caused anomalies with deduplication inside `CompositeException`
- fixes 3 PMD rule violations (2 suppressed, 1 corrected)
- `timestamp` and `timeInterval` now use the `Schedulers.computation()` as the source for the current time instead of `Schedulers.immediate()` which can't be properly hooked. By default, they both return `System.currentTimeMillis()`.

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -10702,7 +10702,8 @@ public class Observable<T> {
      *  <dd>The operator doesn't interfere with backpressure which is determined by the source {@code Observable}'s backpressure
      *  behavior.</dd>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>{@code timeInterval} operates by default on the {@code computation} {@link Scheduler}.</dd>
+     *  <dd>{@code timeInterval} does not operate on any particular scheduler but uses the current time
+     *  from the {@code computation} {@link Scheduler}.</dd>
      * </dl>
      * 
      * @return an Observable that emits time interval information items
@@ -10722,7 +10723,8 @@ public class Observable<T> {
      *  <dd>The operator doesn't interfere with backpressure which is determined by the source {@code Observable}'s backpressure
      *  behavior.</dd>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>you specify which {@link Scheduler} this operator will use</dd>
+     *  <dd>The operator does not operate on any particular scheduler but uses the current time
+     *  from the specified {@link Scheduler}.</dd>
      * </dl>
      * 
      * @param scheduler
@@ -11006,7 +11008,8 @@ public class Observable<T> {
      *  <dd>The operator doesn't interfere with backpressure which is determined by the source {@code Observable}'s backpressure
      *  behavior.</dd>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>{@code timestamp} operates by default on the {@code computation} {@link Scheduler}.</dd>
+     *  <dd>{@code timestamp} does not operate on any particular scheduler but uses the current time
+     *  from the {@code computation} {@link Scheduler}.</dd>
      * </dl>
      * 
      * @return an Observable that emits timestamped items from the source Observable
@@ -11026,7 +11029,8 @@ public class Observable<T> {
      *  <dd>The operator doesn't interfere with backpressure which is determined by the source {@code Observable}'s backpressure
      *  behavior.</dd>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>you specify which {@link Scheduler} this operator will use</dd>
+     *  <dd>The operator does not operate on any particular scheduler but uses the current time
+     *  from the specified {@link Scheduler}.</dd>
      * </dl>
      * 
      * @param scheduler
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -9825,7 +9825,7 @@ public class Observable<T> {
                 Exceptions.throwIfFatal(e2);
                 // if this happens it means the onError itself failed (perhaps an invalid function implementation)
                 // so we are unable to propagate the error correctly and will just throw
-                RuntimeException r = new RuntimeException("Error occurred attempting to subscribe [" + e.getMessage() + "] and then again while trying to pass to onError.", e2);
+                RuntimeException r = new OnErrorFailedException("Error occurred attempting to subscribe [" + e.getMessage() + "] and then again while trying to pass to onError.", e2);
                 // TODO could the hook be the cause of the error in the on error handling.
                 RxJavaHooks.onObservableError(r);
                 // TODO why aren't we throwing the hook's return value.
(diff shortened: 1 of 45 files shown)


src/test/java/rx/ObservableTests.java
```diff
@@ -1174,4 +1175,304 @@ public Object call(OnSubscribe<Object> onSubscribe) {
...
+            public void onError(Throwable e) {
+                throw new TestException();
+            }
+        };
+        
+        try {
+            new FailingObservable().subscribe(ts);
+            fail("Should have thrown OnErrorFailedException");
+        } catch (OnErrorFailedException ex) {
+            // expected
```
JakeWharton, line 1396: If you really wanted to be comprehensive here you could put `TestException` in a method local and verify `assertSame(ex.getCause(), testException)`.

JakeWharton, line 1396: Or just add a message to `TestException` and assert the cause's message would basically achieve the same thing.

src/test/java/rx/ObservableTests.java
```diff
@@ -1174,4 +1175,304 @@ public Object call(OnSubscribe<Object> onSubscribe) {
...
+        o.subscribe(new Action1<String>() {
+
+            @Override
+            public void call(String v) {
+                assertEquals("one", v);
+                latch.countDown();
+            }
+        });
+
+        if (!latch.await(1000, TimeUnit.MILLISECONDS)) {
```
JakeWharton: A trick I use here is `assertTrue(latch.await(1000, MILLISECONDS))` which gives you the same behavior in less lines.

src/test/java/rx/internal/operators/OnSubscribeFromArrayTest.java
```diff
@@ -64,4 +65,15 @@ public void backpressure() {
...
 
+    @Test
+    public void empty() {
+        Assert.assertSame(Observable.empty(), Observable.from(new Object[0]));
+    }
+
+    @Test
+    public void just() {
+        Observable<Integer> source = Observable.from(new Integer[] { 1 });
+        Assert.assertTrue(source.getClass().toString(), source instanceof ScalarSynchronousObservable);
```
JakeWharton, line 76: I don't like these two tests. It's the kind code coverage erroneously lures you in to. We're testing that an implementation detail is true. I would prefer a test that just asserted 0, 1, and N elements behaves as expected and leave it up to the perf tests to ensure there's no regression in performance on changes.

src/test/java/rx/internal/operators/OperatorBufferTest.java
```diff
@@ -1092,4 +1077,180 @@ public void testPostCompleteBackpressure() {
...
+                    Arrays.asList(1, 2),
+                    Arrays.asList(4)
+            );
+            
+            ts.assertNoErrors();
+            ts.assertCompleted();
+        } finally {
+            RxJavaHooks.reset();
+        }
+    }
```
JakeWharton, line 1255: nice tests in here!

JakeWharton:
lgtm :+1:

src/main/java/rx/Observable.java
```diff
@@ -10699,14 +10702,14 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
      *  <dd>The operator doesn't interfere with backpressure which is determined by the source {@code Observable}'s backpressure
      *  behavior.</dd>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>{@code timeInterval} operates by default on the {@code immediate} {@link Scheduler}.</dd>
+     *  <dd>{@code timeInterval} operates by default on the {@code computation} {@link Scheduler}.</dd>
```
artem-zinnatullin: Actually, this operator does not "operate" on the scheduler, it uses it only as source of time

src/main/java/rx/Observable.java
```diff
@@ -10699,14 +10702,14 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
...
      * </dl>
      * 
      * @return an Observable that emits time interval information items
      * @see <a href="http://reactivex.io/documentation/operators/timeinterval.html">ReactiveX operators documentation: TimeInterval</a>
      */
     public final Observable<TimeInterval<T>> timeInterval() {
-        return timeInterval(Schedulers.immediate());
+        return timeInterval(Schedulers.computation());
```
artem-zinnatullin, line 10713: Hope nobody relied on that

src/test/java/rx/ObservableTests.java
```diff
@@ -1174,4 +1175,304 @@ public Object call(OnSubscribe<Object> onSubscribe) {
...
+    }
+    
+    @Test
+    public void subscribeWithNull() {
+        Action1<Integer> onNext = Actions.empty();
+        Action1<Throwable> onError = Actions.empty();
+        Action0 onCompleted = Actions.empty();
+        try {
+            Observable.just(1).subscribe((Action1<Integer>)null);
+            fail("Should have thrown IllegalArgumentException");
```
artem-zinnatullin, line 1269: AssertJ has nice `failBecauseExceptionWasNotThrown(Exception.class)`, maybe add it as test dependency or just add such `fail` method to reuse between tests?

JakeWharton, line 1269: But why? It offers nothing you don't already get here. I wouldn't even put a message. Just `fail();`.

artem-zinnatullin, line 1269: If reader of the code is familar with `try-fail-cath-assert` patter — `fail()` without message is ok, otherwise it's not very obvious, especially when exception in `catch(Exception e)` is not called `expected`.

Also, `failBecauseExceptionWasNotThrown()` gives good exception message in case of test fail.

Not forcing that of course, I'm fine with `fail()` without message.

JakeWharton, line 1269: > especially when exception in catch(Exception e) is not called expected.

That sounds like the right fix then. AssertJ's method isn't DRY with regard to the exception type and optimizing for the failure case seems like the wrong thing to do.

artem-zinnatullin, line 1269: 👍 for `fail()` without a message + `expected` for exception name.

---

> optimizing for the failure case seems like the wrong thing to do

Can't completely agree with that, 2 reasons why I (and probably others) love AssertJ and similar libraries instead of JUnit's assertions:
1. Fluent API with specialization, makes assertions human-readable
2. Really good fail messages with pointing to what was expected and what has actually happened in a very readable manner

JakeWharton, line 1269: Both of those points are accurate for actual _assertions_ which this is not. Having a nice message for a failure is completely pointless when the test code structure is already 100% unambiguous as to the problem.

src/test/java/rx/internal/operators/OnSubscribeFromArrayTest.java
akarnokd, line 76: No, this tests that the scalar optimization is applied by checking the return type.

JakeWharton, line 76: Right. That optimization is an implementation detail of performance, not of behavior. This means any time you change the optimization you have to change the unit test, despite the observed behavior of the API not changing one bit.

src/main/java/rx/Observable.java
akarnokd, line 10713: It is unlikely somebody tested against real-time with this overload.

akarnokd:
Updated.

akarnokd:
Thanks for the review. I'm merging this as it may contain the cure for the memory-kill on Travis. Let me know if there are tests that need more tidying up.

PR #4179, opened 2016-07-08
1.x: new fromAsync to bridge the callback world with the reactive

akarnokd:
This PR adds a new source operator: `fromAsync()` that let's bridge the callback-style world with the reactive world by providing a push surface and offers options to handle backpressure.

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeFromAsync.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeFromAsync.java
@@ -104,16 +104,18 @@ public final class OnSubscribeFromAsync<T> implements OnSubscribe<T> {
         }
     }
     
-    static abstract class BaseAsyncEmitter<T> implements AsyncEmitter<T>, Producer, Subscription {
+    static abstract class BaseAsyncEmitter<T> 
+    extends AtomicLong
+    implements AsyncEmitter<T>, Producer, Subscription {
+        /** */
+        private static final long serialVersionUID = 7326289992464377023L;
+
         final Subscriber<? super T> actual;
-        
-        final AtomicLong requested;
         
         final SerialSubscription serial;
 
         public BaseAsyncEmitter(Subscriber<? super T> actual) {
             this.actual = actual;
-            this.requested = new AtomicLong();
             this.serial = new SerialSubscription();
         }
 
@@ -159,7 +161,7 @@ public final class OnSubscribeFromAsync<T> implements OnSubscribe<T> {
         @Override
         public final void request(long n) {
             if (BackpressureUtils.validate(n)) {
-                BackpressureUtils.getAndAddRequest(requested, n);
+                BackpressureUtils.getAndAddRequest(this, n);
                 onRequested();
             }
         }
@@ -180,11 +182,14 @@ public final class OnSubscribeFromAsync<T> implements OnSubscribe<T> {
 
         @Override
         public final long requested() {
-            return requested.get();
+            return get();
         }
     }
     
     static final class NoneAsyncEmitter<T> extends BaseAsyncEmitter<T> {
+
+        /** */
+        private static final long serialVersionUID = 3776720187248809713L;
 
         public NoneAsyncEmitter(Subscriber<? super T> actual) {
             super(actual);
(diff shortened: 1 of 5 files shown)


akarnokd:
These errors, i.e., Travis running out of memory and killing tests, gets annoying...

src/test/java/rx/internal/operators/OnSubscribeFromAsyncTest.java
```diff
@@ -0,0 +1,727 @@
...
+            subject.onError(e);
+        }
+        
+        @Override
+        public void onCompleted() {
+            subject.onCompleted();
+        }
+    }
+
+}
```
JakeWharton, line 727: Really awesome tests. Easy to read and understand.

src/main/java/rx/internal/operators/OnSubscribeFromAsync.java
```diff
@@ -0,0 +1,514 @@
...
+            }
+        }
+    }
+    
+    static abstract class BaseAsyncEmitter<T> implements AsyncEmitter<T>, Producer, Subscription {
+        final Subscriber<? super T> actual;
+        
+        final AtomicLong requested;
+        
+        final SerialSubscription serial;
```
JakeWharton, line 115: Looks like you could extend from one of these two to continue saving a few bytes!

src/main/java/rx/internal/operators/OnSubscribeFromAsync.java
```diff
@@ -0,0 +1,514 @@
...
+            // default is no-op
+        }
+        
+        @Override
+        public final void setSubscription(Subscription s) {
+            serial.set(s);
+        }
+
+        @Override
+        public final void setCancellation(rx.AsyncEmitter.Cancellable c) {
```
JakeWharton: nit: shouldn't need qualified with `rx.` since `rx.*` is imported.

akarnokd:
Updated.

JakeWharton:
LGTM :+1: !

I'm surprised you couldn't reuse more of the existing infrastructure for the various backpressure modes here and instead need to have a full implementation of the whole queue/drain stuff.

akarnokd:
The codebase spans over several years now and there is no current "best toolset" for building operators. Besides, this inline saves allocation and overhead from applying other operators.

src/main/java/rx/internal/operators/OnSubscribeFromAsync.java
```diff
@@ -0,0 +1,534 @@
...
+        }
+        default: {
+            emitter = new BufferAsyncEmitter<T>(t, RxRingBuffer.SIZE);
+            break;
+        }
+        }
+
+        t.add(emitter);
+        t.setProducer(emitter);
+        asyncEmitter.call(emitter);
```
JakeWharton, line 71: Shouldn't invoking this be deferred until `request > 0`?

akarnokd, line 71: Why would we do that? The subscribe() itself triggers the execution of the body that sets up the push source.

JakeWharton, line 71: Yep, I guess it depends on what the implementation of the emitter is. I have a few use cases in my head I'm thinking through. Besides you could use `AsyncOnSubscribe` if you need that request coordination.

akarnokd:
@JakeWharton how eager are you about this. I'd really love to merge all remaining PRs so 1.1.7 is as complete API-vise as possible. Otherwise, we may have to wait till 1.1.8 and 1.2 is also delayed.

JakeWharton:
I would, of course, prefer that it made it. Releases are few and far between here so missing the boat might mean 3 months before it sees the light of day.

That said, if no one from Netflix is available to review the API and it's the only thing blocking 1.1.7 then I'm fine with it missing the boat.

akarnokd:
I see low risk as this is a completely new operator.

JakeWharton:
Thanks!

defHLT:
I wonder should this method be used instead of `create` in cases like RxBinding present?

stevegury:
👍 
Sorry for the late response (I was on vacation)

PR #4180, opened 2016-07-08
1.x: change test heap size to avoid 137

akarnokd:
It appears that since we added jacoco to the build, it fails sometimes with error 137 which is the code for the system killing the JVM for running out of memory. It is likely the coverage data plus some test's excessive memory use (I don't know which ones) try to go above the current 2GB limit.

This is an experiment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ if (project.hasProperty('release.useLastTag')) {
 }
 
 test{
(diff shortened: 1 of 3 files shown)


akarnokd:
I will rerun this in 8 hours to see how it behaves then (different travis load).

PR #4182, opened 2016-07-09
1.x: add missing RxJavaHooks options, fix bugs

akarnokd:
These PR adds the remaining hooks which weren't even tested with the original RxJavaPlugins plus a full coverage of the RxJavaHooks itself.

This is a mandatory addition and 1.1.7 can't be released without it (i.e., with an incomplete RxJavaHooks that is).

### DIFF
--- a/src/test/java/rx/plugins/RxJavaHooksTest.java
+++ b/src/test/java/rx/plugins/RxJavaHooksTest.java
@@ -15,7 +15,11 @@
  */
 package rx.plugins;
 
+import static org.junit.Assert.*;
+
+import java.lang.Thread.UncaughtExceptionHandler;
 import java.lang.reflect.Method;
+import java.util.*;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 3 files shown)


src/test/java/rx/plugins/RxJavaHooksTest.java
```diff
@@ -552,4 +603,458 @@ public void onScheduleNewThread() throws InterruptedException {
...
+                    list.add(e);
+                    
+                }
+            });
+            
+            RxJavaHooks.onError(new TestException("Forced failure"));
+
+            Thread.currentThread().setUncaughtExceptionHandler(null);
+
+            // this will be printed on the console and should not crash
```
JakeWharton, line 670: You could actually test this by setting `System.err` to something that records, but it would be a bit much.

JakeWharton:
Lgtm :+1:

akarnokd:
Thanks @JakeWharton . Travis CI doesn't execute the build for like 30 minutes now...

PR #4183, opened 2016-07-09
1.x: no need to run gradle assembe by travis

akarnokd:
By default, Travis runs `gradle assemble` and if it fails, there seems to be no way to specify `--stacktrace` for it other than disabling it entirely. The `buildViaTravis.sh` does a full build anyway.

I'm posting this as Nebula fails in the assembly phase for some reason but by default, nothing else is printed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ jdk:
 sudo: false
 # as per http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 
+# prevent travis running gradle assemble; let the build script do it anyway
+install: true
+
 # script for build and release via Travis to Bintray
 script: gradle/buildViaTravis.sh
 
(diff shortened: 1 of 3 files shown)


JakeWharton:
lgtm :+1:

Their idea is to do full compilation of all modules before running tests, but in a single-module repo it doesn't make much sense anyway.

akarnokd:
```
org.eclipse.jgit.errors.MissingObjectException: Missing commit 4e5f6aae88a0993d528077dce3df19c630d825d0
```

[This](https://netbeans.org/bugzilla/show_bug.cgi?id=257422) says the stash is corrupt; I'm assuming the cache we use for building could be corrupt after today's massive outage.

akarnokd:
Hmm, looks like [limiting the clone](https://docs.travis-ci.com/user/customizing-the-build/#Git-Clone-Depth) depth went past that odd commit number (which is available here on GitHub).

akarnokd:
Now we are killed with out of memory - probably due to the container-backlog.

PR #4187, opened 2016-07-10
Move status icons to top, add maven version badge.

akarnokd:
Plus, this should also fix the snapshot versions which were broken (they are with 0.0.1-SNAPSHOT on bintray) because of yesterday's Nebula plugin problems.


akarnokd:
Looks like the versioning works again: https://travis-ci.org/ReactiveX/RxJava/builds/143695889#L157

PR #4189, opened 2016-07-10
Update 1.1.7 javadoc files

akarnokd:
I've extracted the javadoc.jar from the release into a subdir 1.1.7 and replaced the root files as well.

/cc @DavidMGross


zsxwing:
👍

PR #4190, opened 2016-07-11
2.x: implement concat(Map)Eager(DelayError) and expose concat(Map)DelayError overload

akarnokd:
Started a syncing process between 1.x and 2.x operators (based on [this diff](https://gist.github.com/akarnokd/4a391527e099412cad87d3dffbfef762)). The first set is the addition of the eager version of `concat` and `concatMap`. In addition, all of them get a `XDelayError` overloads as well.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
@@ -127,13 +127,13 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
             
             InnerQueuedSubscriber<R> inner = new InnerQueuedSubscriber<R>(this, prefetch);
             
-            if (!cancelled) {
+            if (cancelled) {
                 return;
             }
 
             subscribers.offer(inner);
             
-            if (!cancelled) {
+            if (cancelled) {
                 return;
             }
             
@@ -141,7 +141,6 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
             
             if (cancelled) {
                 inner.cancel();
-            } else {
                 drainAndCancel();
             }
         }
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
@@ -128,15 +128,21 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
             InnerQueuedSubscriber<R> inner = new InnerQueuedSubscriber<R>(this, prefetch);
             
             if (!cancelled) {
-                subscribers.offer(inner);
-                if (!cancelled) {
-                    p.subscribe(inner);
-                    if (cancelled) {
-                        inner.cancel();
-                    }
-                } else {
-                    drainAndCancel();
-                }
+                return;
+            }
+
+            subscribers.offer(inner);
+            
+            if (!cancelled) {
+                return;
+            }
+            
+            p.subscribe(inner);
+            
+            if (cancelled) {
+                inner.cancel();
+            } else {
+                drainAndCancel();
             }
         }
         
@@ -193,11 +199,11 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
         
         @Override
         public void innerNext(InnerQueuedSubscriber<R> inner, R value) {
-            if (!inner.queue().offer(value)) {
+            if (inner.queue().offer(value)) {
+                drain();
+            } else {
                 inner.cancel();
                 innerError(inner, new MissingBackpressureException());
-            } else {
-                drain();
             }
         }
         
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
@@ -70,7 +70,7 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
         final int maxConcurrency;
         
         final int prefetch;
-
+        
         final ErrorMode errorMode;
 
         final AtomicReference<Throwable> error;
@@ -296,6 +296,7 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
                             inner = null;
                             subscribers.poll();
                             current = null;
+                            s.request(1);
                             continue outer;
                         }
                         
@@ -334,6 +335,7 @@ public class FlowableConcatMapEager<T, R> extends FlowableSource<T, R> {
                             inner = null;
                             subscribers.poll();
                             current = null;
+                            s.request(1);
                             continue outer;
                         }
                     }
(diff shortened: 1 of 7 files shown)


akarnokd:
Also I've expanded Flowable with a `test()` method. Should give a nice answer to the question "How do you test a Flowable?".

src/main/java/io/reactivex/internal/util/QueueDrainHelper.java
```diff
@@ -359,4 +360,29 @@
...
+    /**
+     * Creates a queue: spsc-array if capacityHint is positive and 
+     * spsc-linked-array if capacityHint is negative; in both cases, the
+     * capacity is the absolute value of prefetch.
+     * @param <T> the value type of the queue
+     * @param capacityHint the capacity hint
+     * @return the queue instance
+     */
+    public static <T> Queue<T> createQueue(int capacityHint) {
+        if (capacityHint < 0) {
```
vanniktech, line 373: how about `return new SpscLinkedArrayQueue<T>(Math.abs(capacityHint));`

akarnokd, line 373: From the JDK:

``` java
public static long abs(long a) {
        return (a < 0) ? -a : a;
}
```

Pointless, plus this is an internal utility method.

vanniktech, line 373: In my opinion it makes reading the code a bit easier and 3 lines will be saved too but yeah it's just a nit.

src/test/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEagerTest.java
```diff
@@ -230,4 +230,20 @@ public void mainErrorsImmediate() {
         
         ts.assertFailureAndMessage(TestException.class, "Forced failure", 2);
     }
+    
+    @Test
+    public void longEager() {
+
```
artem-zinnatullin, line 236: nit: el

src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
```diff
@@ -0,0 +1,356 @@
...
+    protected void subscribeActual(Subscriber<? super R> s) {
+        source.subscribe(new ConcatMapEagerDelayErrorSubscriber<T, R>(
+                s, mapper, maxConcurrency, prefetch, errorMode));
+    }
+
+    static final class ConcatMapEagerDelayErrorSubscriber<T, R>
+    extends AtomicInteger
+    implements Subscriber<T>, Subscription, InnerQueuedSubscriberSupport<R> {
+        /** */
+        private static final long serialVersionUID = -4255299542215038287L;
```
artem-zinnatullin, line 64: nit: turn off IDE warning and remove `serialVersionUID`?

src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
```diff
@@ -0,0 +1,356 @@
...
+                s.cancel();
+                onError(ex);
+                return;
+            }
+            
+            InnerQueuedSubscriber<R> inner = new InnerQueuedSubscriber<R>(this, prefetch);
+            
+            if (!cancelled) {
+                subscribers.offer(inner);
+                if (!cancelled) {
```
artem-zinnatullin: nit: can we invert that and make `drainAndCancel` first so it'll be easier to read and harder to make some mistake during refactoring?

src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
```diff
@@ -0,0 +1,356 @@
...
+        public void request(long n) {
+            if (SubscriptionHelper.validate(n)) {
+                BackpressureHelper.add(requested, n);
+                drain();
+            }
+        }
+        
+        @Override
+        public void innerNext(InnerQueuedSubscriber<R> inner, R value) {
+            if (!inner.queue().offer(value)) {
```
artem-zinnatullin: nit: same here about reversing `if-else`. I think PMD complains about it btw

akarnokd:
Updated the `if`s

src/main/java/io/reactivex/internal/subscribers/flowable/InnerQueuedSubscriber.java
```diff
@@ -0,0 +1,147 @@
...
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
+import io.reactivex.internal.util.QueueDrainHelper;
+
+/**
+ * Subscriber that can fuse with the upstream and calls a support interface
+ * whenever an event is available.
+ *
+ * @param <T> the value type
+ */
+public final class InnerQueuedSubscriber<T>
```
artem-zinnatullin, line 31: Maybe add separate unit tests for it and check that behavior of public API is correct and it correctly reacts to different fusion modes? Up to you of course

akarnokd, line 31: Will be part of the test-sync and coverage improvement before RC1.

src/main/java/io/reactivex/subscribers/TestSubscriber.java
```diff
@@ -831,14 +831,58 @@ private String fusionModeToString(int mode) {
...
+    }
+
+    /**
+     * Assert that the upstream signalled the specified values in order
+     * and then failed with a specific class or subclass of Throwable.
+     * @param error the expected exception (parent) class
+     * @param values the expected values, asserted in order
+     * @return this
+     */
+    public final TestSubscriber<T> assertFailure(Class<Throwable> error, T... values) {
```
artem-zinnatullin, line 862: nice

src/main/java/io/reactivex/internal/subscribers/flowable/InnerQueuedSubscriber.java
artem-zinnatullin, line 31: ok

src/test/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEagerTest.java
```diff
@@ -0,0 +1,249 @@
...
+    @Test
+    public void normal() {
+        Flowable.range(1, 5)
+        .concatMapEager(new Function<Integer, Publisher<Integer>>() {
+            @Override
+            public Publisher<Integer> apply(Integer t) {
+                return Flowable.range(t, 2);
+            }
+        })
+        .test()
```
artem-zinnatullin, line 37: oh man, this is awesome 👍

artem-zinnatullin:
I'm thinking if there is a way to test such [`if (!cancelled)`](https://github.com/ReactiveX/RxJava/pull/4190/commits/2bf22b287c78c49709db098aea2dcd4ca858a5a1) errors, java debuggers allows suspending a Thread on a particular breakpoint and allow other threads to run at the same time, maybe such mechanism will allow us to simulate and test concurrent access in particular parts of code (drain loops and so on). 

Something like:

``` java
@Test
public void concurrentCancel() {
  TestThread thread1 = testThread({
     disposable = Flowable.just(1).concatMap(v -> Flowable.just(2)).subscribe();
  });

  TestThread thread2 = testThread({
    disposable.dispose();
  });

  TestThreads.builder()
    .start(thread1) // One or multiple.
    .thenSuspend(thread1, FlowableConcatMapEager.class, 128) // Thread; Source class; Line.
    .thenStart(thread2)
    .thenResume(thread1)
    .run(); // Run and wait all test threads to complete.

  // Assert behavior.
}
```

Just thoughts

artem-zinnatullin:
:+1: for PR

akarnokd:
Thanks @artem-zinnatullin !

PR #4192, opened 2016-07-12
2.x: +fromAsync, distinctUC, skip, take overloads, fix TestSubscriber API

akarnokd:
This PR adds a few operators and overloads from 1.x
- `fromAsync`
- `just` - 10 arguments
- `distinctUntilChanged(BiPredicate<? super T, ? super T> comparer)`
- `skip` timed - default scheduler
- `take` timed - default scheduler
- `toCompletable`

In addition, this PR fixes the API of TestSubscriber by making the initial value primitive `long` and having 0 as no initial request instead of the trickery with `null`.

### DIFF
--- a/src/main/java/io/reactivex/internal/util/BackpressureHelper.java
+++ b/src/main/java/io/reactivex/internal/util/BackpressureHelper.java
@@ -72,24 +72,24 @@ public enum BackpressureHelper {
     }
     
     /**
-     * Atomically subtract the given number (positive, not validated) form the target field.
+     * Atomically subtract the given number (positive, not validated) from the target field.
      * @param requested the target field holding the current requested amount
      * @param n the produced element count, positive (not validated)
      * @return the new amount
      */
     public static long produced(AtomicLong requested, long n) {
         for (;;) {
-            long r = requested.get();
-            if (r == Long.MAX_VALUE) {
+            long current = requested.get();
+            if (current == Long.MAX_VALUE) {
                 return Long.MAX_VALUE;
             }
-            long u = r - n;
-            if (u < 0L) {
-                RxJavaPlugins.onError(new IllegalStateException("More produced than requested: " + u));
-                u = 0L;
+            long update = current - n;
+            if (update < 0L) {
+                RxJavaPlugins.onError(new IllegalStateException("More produced than requested: " + update));
+                update = 0L;
             }
-            if (requested.compareAndSet(r, u)) {
-                return r;
+            if (requested.compareAndSet(current, update)) {
+                return update;
             }
         }
     }
--- a/src/main/java/io/reactivex/internal/util/BackpressureHelper.java
+++ b/src/main/java/io/reactivex/internal/util/BackpressureHelper.java
@@ -14,6 +14,8 @@ package io.reactivex.internal.util;
 
 import java.util.concurrent.atomic.AtomicLong;
 
+import io.reactivex.plugins.RxJavaPlugins;
+
 /**
  * Utility class to help with backpressure-related operations such as request aggregation.
  */
@@ -63,6 +65,29 @@ public enum BackpressureHelper {
                 return Long.MAX_VALUE;
             }
             long u = addCap(r, n);
+            if (requested.compareAndSet(r, u)) {
+                return r;
+            }
+        }
+    }
+    
+    /**
+     * Atomically subtract the given number (positive, not validated) form the target field.
+     * @param requested the target field holding the current requested amount
+     * @param n the produced element count, positive (not validated)
+     * @return the new amount
+     */
+    public static long produced(AtomicLong requested, long n) {
+        for (;;) {
+            long r = requested.get();
+            if (r == Long.MAX_VALUE) {
+                return Long.MAX_VALUE;
+            }
+            long u = r - n;
+            if (u < 0L) {
+                RxJavaPlugins.onError(new IllegalStateException("More produced than requested: " + u));
+                u = 0L;
+            }
             if (requested.compareAndSet(r, u)) {
                 return r;
             }
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChangedTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChangedTest.java
@@ -21,7 +21,8 @@ import org.mockito.InOrder;
 import org.reactivestreams.Subscriber;
 
 import io.reactivex.*;
-import io.reactivex.functions.Function;
+import io.reactivex.functions.*;
+import io.reactivex.internal.fuseable.QueueSubscription;
 
 public class FlowableDistinctUntilChangedTest {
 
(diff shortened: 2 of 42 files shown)


akarnokd:
/cc @artem-zinnatullin @JakeWharton 

If this could me merged soon, I could start on syncing the unit tests in which there are [roughly 1500](https://gist.github.com/akarnokd/693866d6b4453ca8c66b866584163e5e) methods to check - some of them are simple relocations. Once done, I have to sync the flowable-observable tests (which have a more symmetric layout luckily).

artem-zinnatullin:
Will look in several hours.

src/main/java/io/reactivex/internal/util/BackpressureHelper.java
```diff
@@ -68,4 +70,27 @@ public static long add(AtomicLong requested, long n) {
             }
         }
     }
+    
+    /**
+     * Atomically subtract the given number (positive, not validated) form the target field.
```
artem-zinnatullin: nit: from the target field.

src/main/java/io/reactivex/internal/util/BackpressureHelper.java
```diff
@@ -68,4 +70,27 @@ public static long add(AtomicLong requested, long n) {
...
+            if (r == Long.MAX_VALUE) {
+                return Long.MAX_VALUE;
+            }
+            long u = r - n;
+            if (u < 0L) {
+                RxJavaPlugins.onError(new IllegalStateException("More produced than requested: " + u));
+                u = 0L;
+            }
+            if (requested.compareAndSet(r, u)) {
+                return r;
```
artem-zinnatullin: Shouldn't you return `u` here? 

Separate tests for this function would be great, and better var names.

src/test/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChangedTest.java
```diff
@@ -129,4 +130,70 @@ public void testDistinctUntilChangedOfSourceWithExceptionsFromKeySelector() {
         inOrder.verify(w, never()).onNext(anyString());
         inOrder.verify(w, never()).onComplete();
     }
+    
+    @Test
+    public void directComparer() {
+        Flowable.fromArray(1, 2, 2, 3, 2, 4, 1, 1, 2)
+        .distinctUntilChanged(new BiPredicate<Integer, Integer>() {
```
artem-zinnatullin, line 137: nit: maybe rename `BiPredicate` to `Predicate2`? (in a separate PR)

akarnokd:
@artem-zinnatullin fixing what you discovered so far. Notify me when you are done.

`BiPredicate` conforms with (and thus conflicts with) Java 8 functional interface naming conventions. Calling it `Pred2` is possible but then `Func1...FuncN` would conflict with RxJava 1 if in the same project.

src/main/java/io/reactivex/internal/util/BackpressureHelper.java
akarnokd: Right, good catch.

src/test/java/io/reactivex/internal/operators/flowable/FlowableFromAsyncTest.java
```diff
@@ -0,0 +1,729 @@
...
+
+    PublishAsyncEmitter source;
+
+    PublishAsyncEmitterNoCancel sourceNoCancel;
+
+    TestSubscriber<Integer> ts;
+    
+    @Before
+    public void before() {
+        source = new PublishAsyncEmitter();
```
artem-zinnatullin, line 22: nit: you can inline it and make variables final, JUnit creates new instance of test class for each test

src/test/java/io/reactivex/internal/operators/flowable/FlowableFromAsyncTest.java
```diff
@@ -0,0 +1,729 @@
...
+        Flowable.fromAsync(source, AsyncEmitter.BackpressureMode.DROP).subscribe(ts);
+        
+        source.onNext(1);
+
+        ts.request(1);
+
+        source.onNext(2);
+        source.onComplete();
+        
+        ts.assertValues(2);
```
artem-zinnatullin, line 61: better to check values

src/test/java/io/reactivex/internal/operators/flowable/FlowableFromAsyncTest.java
```diff
@@ -0,0 +1,729 @@
...
+        Flowable.fromAsync(source, AsyncEmitter.BackpressureMode.LATEST).subscribe(ts);
+        
+        source.onNext(1);
+
+        source.onNext(2);
+        source.onComplete();
+
+        ts.request(1);
+
+        ts.assertValues(2);
```
artem-zinnatullin, line 79: as above

src/test/java/io/reactivex/internal/operators/flowable/FlowableFromAsyncTest.java
```diff
@@ -0,0 +1,729 @@
...
+    public void errorLatest() {
+        Flowable.fromAsync(source, AsyncEmitter.BackpressureMode.LATEST).subscribe(ts);
+        
+        source.onNext(1);
+        source.onNext(2);
+        source.onError(new TestException());
+        
+        ts.request(1);
+        
+        ts.assertValues(2);
```
artem-zinnatullin, line 158: as above

src/test/java/io/reactivex/internal/operators/flowable/FlowableFromAsyncTest.java
akarnokd, line 61: What do you mean?

artem-zinnatullin:
Done, 👍 but comments.

> BiPredicate conforms with (and thus conflicts with) Java 8 functional interface naming conventions. Calling it Pred2 is possible but then Func1...FuncN would conflict with RxJava 1 if in the same project

They won't conflict because package names are different, though you'll have to use fqn if you'll need both in same source file.

akarnokd:
> They won't conflict because package names are different,

Sure, but when you type in `Function` then CTRL+SPACE in IDE, you get the Java version and RxJava 2 version. Sometimes you copy-paste a `Function f = ...` and get a squiggly line saying `Function` is not compatible with `Function`.

src/test/java/io/reactivex/internal/operators/flowable/FlowableFromAsyncTest.java
artem-zinnatullin, line 61: I mean, if possible better to check values than their count, because
operator may emit them in wrong order or one item twice, etc and your test
will pass.

akarnokd, line 61: It's [assertValues(T... values)](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/observers/TestSubscriber.java#L618).

artem-zinnatullin, line 61: Damn, that means I didn't check test precisely.. will check again when I'll
be at home

artem-zinnatullin, line 61: 👍

akarnokd:
Thanks @artem-zinnatullin !

PR #4195, opened 2016-07-13
2.x: sync unit tests of Completable, add missing features

akarnokd:
Don't worry about not all things tested in this PR, anything extra relative to 1.x will be covered after the sync.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -600,12 +600,12 @@ public abstract class Completable implements CompletableConsumable {
     }
 
     /**
-     * Returns an Observable which will subscribe to this Completable and once that is completed then 
+     * Returns an Flowable which will subscribe to this Completable and once that is completed then 
      * will subscribe to the {@code next} Flowable. An error event from this Completable will be 
      * propagated to the downstream subscriber and will result in skipping the subscription of the 
      * Observable.  
      * 
-     * @param <T> the value type of the next Observable
+     * @param <T> the value type of the next Flowable
      * @param next the Observable to subscribe after this Completable is completed, not null
      * @return Flowable that composes this Completable and next
      * @throws NullPointerException if next is null
@@ -635,7 +635,7 @@ public abstract class Completable implements CompletableConsumable {
     }
 
     /**
-     * Returns a completable that first runs this Completable
+     * Returns a Completable that first runs this Completable
      * and then the other completable.
      * <p>
      * This is an alias for {@link #concatWith(CompletableConsumable)}.
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -185,7 +185,7 @@ public abstract class Completable implements CompletableConsumable {
         try {
             // TODO plugin wrapping onSubscribe
             
-            return new CompletableWrapper(onSubscribe);
+            return RxJavaPlugins.onAssembly(new CompletableWrapper(onSubscribe));
         } catch (NullPointerException ex) { // NOPMD
             throw ex;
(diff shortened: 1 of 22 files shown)


src/main/java/io/reactivex/Completable.java
```diff
@@ -260,6 +260,18 @@ public static Completable fromCallable(final Callable<?> callable) {
         return new CompletableFromFlowable<T>(flowable);
     }
     
+    @SchedulerSupport(SchedulerSupport.NONE)
```
artem-zinnatullin, line 263: Looks awkward because it fact it's supported through `subscribeOn`, maybe rename to `@DefaultScheduler`?

src/main/java/io/reactivex/Completable.java
```diff
@@ -572,6 +584,71 @@ public final Completable compose(CompletableTransformer transformer) {
...
+     * @return Observable that composes this Completable and next
+     * @throws NullPointerException if next is null
+     */
+    public final <T> Observable<T> andThen(Observable<T> next) {
+        Objects.requireNonNull(next, "next is null");
+        return next.delaySubscription(toObservable());
+    }
+
+    /**
+     * Returns an Observable which will subscribe to this Completable and once that is completed then
```
artem-zinnatullin: Returns a Flowable

src/main/java/io/reactivex/Completable.java
```diff
@@ -572,6 +584,71 @@ public final Completable compose(CompletableTransformer transformer) {
...
+     * @param next the Single to subscribe after this Completable is completed, not null
+     * @return Single that composes this Completable and next
+     */
+    public final <T> Single<T> andThen(Single<T> next) {
+        Objects.requireNonNull(next, "next is null");
+        return next.delaySubscription(toObservable());
+    }
+
+    /**
+     * Returns a completable that first runs this Completable
```
artem-zinnatullin: Completable from capital letter

src/main/java/io/reactivex/Completable.java
```diff
@@ -572,6 +584,71 @@ public final Completable compose(CompletableTransformer transformer) {
...
+     * and then the other completable.
```
artem-zinnatullin, line 639: Completable

src/main/java/io/reactivex/Completable.java
```diff
@@ -760,8 +857,11 @@ public final Completable endWith(CompletableConsumable other) {
      * @throws NullPointerException if onAfterComplete is null
      */
     @SchedulerSupport(SchedulerSupport.NONE)
+    @Deprecated
```
artem-zinnatullin, line 860: deprecated api in non-released library :)

akarnokd, line 860: There will be an API pass, until then no API is to be deleted that has a 1.x counterpart (would mess up the diff anyway).

src/main/java/io/reactivex/Completable.java
```diff
@@ -1050,14 +1151,14 @@ public final void subscribe(CompletableSubscriber s) {
...
-     * @param onError the consumer that is called if this Completable emits an error
      * @param onComplete the runnable that is called if the Completable completes normally
+     * @param onError the consumer that is called if this Completable emits an error
+     * 
      * @return the Disposable that can be used for cancelling the subscription asynchronously
      * @throws NullPointerException if either callback is null
      */
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Disposable subscribe(final Consumer<? super Throwable> onError, final Runnable onComplete) {
+    public final Disposable subscribe(final Runnable onComplete, final Consumer<? super Throwable> onError) {
```
artem-zinnatullin, line 1161: 👍

PR #4204, opened 2016-07-14
2.x: test sync + cleanup

akarnokd:
- More unit tests ported;
- `TestObserver` cleanup and sync with `TestSubscriber`;
- fix travis to run with Java 7 instead of 8;
- added AnimalSniffer;
- cleaned up `combineLatest`, introduced `combineLatestDelayError`;
- test names are as in 1.x, please don't complain about the `test` prefix in the method names!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ group = 'io.reactivex.rxjava2'
 description = 'RxJava: Reactive Extensions for the JVM – a library for composing asynchronous and event-based programs using observable sequences for the Java VM.'
 
 apply plugin: 'java'
-apply plugin: 'pmd'
+// apply plugin: 'pmd'
 apply plugin: 'findbugs'
 apply plugin: 'jacoco'
 apply plugin: 'ru.vyarus.animalsniffer'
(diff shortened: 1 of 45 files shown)


akarnokd:
Anybody wants to review this?

artem-zinnatullin:
Yes, will do in couple days (busy, sorry)

akarnokd:
I'm merging this so we can progress with other tasks that would affect the same files. I'll address the review feedback in a separate PR if necessary.

build.gradle
```diff
@@ -1,33 +1,56 @@
...
+  }
 }
 
 group = 'io.reactivex.rxjava2'
 
 description = 'RxJava: Reactive Extensions for the JVM – a library for composing asynchronous and event-based programs using observable sequences for the Java VM.'
 
 apply plugin: 'java'
-apply plugin: 'pmd'
+// apply plugin: 'pmd'
```
artem-zinnatullin, line 14: Why?

akarnokd, line 14: Our build task gets killed most of the time if it is enabled.

src/main/java/io/reactivex/exceptions/OnCompleteFailedException.java
```diff
@@ -18,7 +18,11 @@
     private static final long serialVersionUID = -6179993283427447098L;
 
     public OnCompleteFailedException(Throwable cause) {
-        super(cause);
+        super(cause != null ? cause : new NullPointerException());
```
artem-zinnatullin, line 21: not sure it's a good idea, npe will point here and can confuse reader of the stacktrace.

// Also, creating npes manually is usually a bad practice, it's an exception thrown by VM, but RxJava throws them already so it's up to u

akarnokd, line 21: This has been discussed back then: instead of crashing the code further, it sets an NPE with the current stack trace to give something to the receiver.

src/main/java/io/reactivex/internal/operators/flowable/BlockingFlowableIterator.java
```diff
@@ -0,0 +1,180 @@
...
+    @Override
+    public void onSubscribe(Subscription s) {
+        if (SubscriptionHelper.setOnce(this, s)) {
+            s.request(batchSize);
+        }
+    }
+
+    @Override
+    public void onNext(T t) {
+        if (!queue.offer(t)) {
```
artem-zinnatullin, line 129: nit: reversed if will be easier to read/maintain

build.gradle
artem-zinnatullin, line 14: hm, that's sad

src/main/java/io/reactivex/exceptions/OnCompleteFailedException.java
artem-zinnatullin, line 21: ok, but I though it's ok to have null as cause…

akarnokd, line 21: Here, it is an indication that somewhere the cause wasn't set properly, the NPE will help identify that.

Issue #4206, opened 2016-07-15
Unexpected difference between flatmap and concatmap when the stream has only one element


rrevol:
Hello,

I am encountering an unexpected behavior in a piece of code I wrote. I've tried to narrow down a reproduction case in [this gist](https://gist.github.com/rrevol/b08f7231887321f584bc76e20dc81fa7).

To sum up : 
- It's on RxJava 1.1.7
- I have 2 Observables built in such that I am sure their elements match
- These observables come from a flatmap on another observable that emits only one element
- I zip the 2 observables and see that in the output some elements do not match
- important : I use observeOn/subscribeOn because in real life, my process performs io.

What I observe :
- if I replace flatmap by concatmap, no more problem. The way I understand both operations, they should not be different if the mapped stream emits only one element. Do I miss sth ?
- If I remove the observeOn/subscribeOn, no more problem, so I suspect a concurrency issue
- The way it looks is that flatmap shuffles data, but I did a simpler test (use method `initTestFlowWithOrderCheck` instead of `initTestFlowWithZip` ) that seems to discard this hypothesis
- In the end, it could be a concurrency issue in zip but I've read the operator extensively and could not see any problem with it.

Here I am with my investigation, I may have missed something obvious to someone else so I'm looking forward to have your feedback on that.

Thx

akarnokd:
By definition, `flatMap` doesn't guarantee any ordering of its events - not between different sources but sometimes not even within the same source (solo or otherwise). 

If you drive it in an async flow, signals may skip ahead yielding the inconsistency. Therefore, If you need strict ordering, use `concat(Map)` or `concat(Map)Eager`.

abersnaze:
> but sometimes not even within the same source

are you talking about onNext from one source being emitted out of order with itself? something like this:

```
--A-B--->
--C----->

flatMap

--C-B-A->
```

akarnokd:
Yes.

abersnaze:
That is surprising to me. @benjchristensen I thought that ordering of onNext from one Observable was to be maintained. I don't see in the general rx contract anywhere but I know we've verbally talked about it over the years.

rrevol:
I'm still having a hard time understanding what happens under the hood. @akarnokd I'm not saying you're wrong, I just want to understand.

[...]

PR #4209, opened 2016-07-18
1.x: merge/flatMap to keep scalar/inner element relative order

akarnokd:
This PR changes flatmap to make sure there is no element reordering happening on the fast-paths.

Related: #4206.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -352,7 +352,8 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
                 }
             }
             if (success) {
-                if (subscriber.queue == null || subscriber.queue.isEmpty()) {
+                RxRingBuffer subscriberQueue = subscriber.queue;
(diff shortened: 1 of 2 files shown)


davidmoten:
Thanks @akarnokd for quick fix. I haven't reviewed yet but I'm keen to see the issue resolved in a release soon (I also was suprised `flatMap` didn't honour individual observable order).

src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -352,9 +352,15 @@ void tryEmit(InnerSubscriber<T> subscriber, T value) {
                 }
             }
             if (success) {
-                emitScalar(subscriber, value, r);
+                if (subscriber.queue == null || subscriber.queue.isEmpty()) {
```
davidmoten: assign `subscriber.queue` to local variable to avoid double volatile read?

davidmoten:
reviewed :+1:

akarnokd:
Updated, thanks @davidmoten !

artem-zinnatullin:
👍

stevegury:
👍

Issue #1277, opened 2014-05-28
FuncN generic varargs parameter


torgeir:
As mentioned [[1](https://twitter.com/benjchristensen/status/471312995352457219)] [[2](https://gist.github.com/torgeir/bdfffbdfdc8ed98056b8#file-statistics-java-L25-L26)] it would be nice for `Observable.zip(iterable, funcn)` to handle iterables of the same type, without having to go about casting the object to an integer for making something like this to work

``` java
List<Observable<Integer>> iterable = Arrays.asList(Observable.from(1), Observable.from(2), Observable.from(3));
Observable<List<Integer>> observable = Observable.zip(iterable, Arrays::asList);
```

Making `FuncN<R>` into `FuncN<T, R>` kind of snowballed on me as I reached the `OperatorZip$Zip` which seems to be needing the `Object` type.

bobvanderlinden:
I came across this same issue, but with `combineLatest`. I see how it causes an incompatibility, but I'd like this to be reconsidered for v2.

Currently I'm using the following overload of `combineLatest`:
```
public static <T, R> Observable<R> combineLatest(List<? extends Observable<? extends T>> sources, FuncN<? extends R> combineFunction)
```
The definition of `FuncN` is:
```
R call(Object... args);
```

Firstly, the type information gets lost. A user would presume `T` is passed, but it is actually `Object`. This caused quite a bit of confusion when using RxJava in Kotlin, where all types can be implicitly resolved and the end of an chain would result in some form of `Object`. Going back up the chain to see what function had misbehaving type information resulted in `combineLatest`.

Additionally, I personally find the use of varargs a bit confusing. I passed in a `List` and I assume a `List` is passed to the `combineFunction`. Instead it is an Array/`[]`, but it is not explicitly typed as an array (as it is typed using varargs). I see how varargs would fit `combineFunction` if `combineLatest` was called with varargs, but in this particular overload it is not.

It might be a good idea to introduce the following function definition for V2:
```
public static <T, R> Observable<R> combineLatest(List<? extends Observable<? extends T>> sources, Func1<List<? extends T>, ? extends R> combineFunction)
```
That way a list is passed into `combineLatest` and a list is passed to the `combineFunction`. That way, no type information is lost.

laurilehmijoki:
This is how we solved the problem in Kotlin:

``` kotlin
fun <T, R> List<Observable<T>>.combineLatest(combineFunction: (List<T>) -> R): Observable<R> = 
[...]

PR #4211, opened 2016-07-18
2.x: use generic type instead of Object in combineLatest and zip

bobvanderlinden:
As discussed in https://github.com/ReactiveX/RxJava/issues/1277#issuecomment-233357354

This PR gives a stricter type signature for `combineLatest`, `zip`, `zipArray` and `zipIterable`. It replaces occurrences of `Object` with its generic type `T`. In practice users will not need to type-cast arrays in the combine function anymore.


akarnokd:
:+1:

davidmoten:
:+1:

stevegury:
👍

akarnokd:
Thanks for the contribution.

Issue #4212, opened 2016-07-19
RxJavaHooks.enableAssemblyTracking() breaks error handling


artem-zinnatullin:
Problem:

If user enables assembly tracking then **ALL** exceptions get wrapped into `AssemblyStackTraceException` which breaks all error handling code that expects particular exceptions.

Solution:

Do not change type of original exception and modify its stacktrace directly. I'll work on that soon, both for 1.x and 2.x.

// Spent an hour today trying to figure out what was causing app misbehave, glad we have `git bisect`, was very surprised to see commit where I enabled assembly tracking…

akarnokd:
Yeah, in [Java 8](https://github.com/reactor/reactive-streams-commons/blob/master/src/main/java/rsc/publisher/PublisherOnAssembly.java#L140) I simply add it as a suppressed exception. In Java 8, you have to find an unused slot of cause in a potentially complicated graph of exceptions and composites.

pyricau:
This change likely caused #4737

PR #4213, opened 2016-07-19
1.x: Do not hide original exception with RxJavaHooks.enableAssemblyTracking()

artem-zinnatullin:
Part of #4212.


src/main/java/rx/internal/operators/OnSubscribeOnAssembly.java
```diff
@@ -40,15 +41,14 @@
...
-    static String createStacktrace() {
-        StackTraceElement[] stes = Thread.currentThread().getStackTrace();
+    static StackTraceElement[] assemblyStacktrace() {
+        final StackTraceElement[] in = new Exception().getStackTrace();
+        final List<StackTraceElement> out = new ArrayList<StackTraceElement>(in.length);
 
-        StringBuilder sb = new StringBuilder("Assembly trace:");
-        
-        for (StackTraceElement e : stes) {
+        for (StackTraceElement e : in) {
```
artem-zinnatullin, line 51: I'll open separate PR after this one where I'll move filtering stacktrace to `onError`. This is a hot path while `onError` isn't, so we can save some CPU time on filtering for errorless chains.

src/main/java/rx/internal/operators/OnSubscribeOnAssembly.java
```diff
@@ -30,7 +31,7 @@
 
     final OnSubscribe<T> source;
     
-    final String stacktrace;
```
akarnokd, line 33: This field had a purpose to show the stacktrace in debug mode by viewing the field as string!

artem-zinnatullin, line 33: Idk, it consumes memory and I've never used it during debugging. You really need it? I think you should be able to get ~normal output from `Arrays.toString(assemblyStacktrace)` in the debugger

akarnokd:
See #4215 for a less invasive change.

artem-zinnatullin:
Closing in favor of #4215.

PR #4215, opened 2016-07-19
1.x: fix assembly tracking replacing original exception

### LINKED ISSUE #4212: RxJavaHooks.enableAssemblyTracking() breaks error handling

akarnokd:
Modify the assembly tracking logic to not replace the exception flowing through but to attach the tracking exception to the end of the causal chain (if possible).

Fixes #4212

### DIFF
--- a/src/main/java/rx/exceptions/AssemblyStackTraceException.java
+++ b/src/main/java/rx/exceptions/AssemblyStackTraceException.java
@@ -15,7 +15,10 @@
  */
 package rx.exceptions;
 
+import java.util.*;
+
 import rx.annotations.Experimental;
+import rx.plugins.RxJavaHooks;
 
 /**
  * A RuntimeException that is stackless but holds onto a textual
@@ -28,16 +31,6 @@ public final class AssemblyStackTraceException extends RuntimeException {
     private static final long serialVersionUID = 2038859767182585852L;
 
     /**
-     * Constructs an AssemblyStackTraceException with the given message and
-     * a cause.
-     * @param message the message
-     * @param cause the cause
-     */
-    public AssemblyStackTraceException(String message, Throwable cause) {
-        super(message, cause);
-    }
-
-    /**
      * Constructs an AssemblyStackTraceException with the given message.
      * @param message the message
      */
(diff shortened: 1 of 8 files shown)


stevegury:
👍

src/main/java/rx/exceptions/AssemblyStackTraceException.java
```diff
@@ -49,4 +42,48 @@ public AssemblyStackTraceException(String message) {
     public synchronized Throwable fillInStackTrace() { // NOPMD 
         return this;
     }
+    
+    /**
+     * Finds an empty cause slot and assigns itself to it.
+     * @param exception the exception to start from
+     */
+    public void attach(Throwable exception) {
```
artem-zinnatullin: `attachTo`, otherwise it looks like you're attaching `exception` to `this` while in fact it's vice versa

src/main/java/rx/exceptions/AssemblyStackTraceException.java
```diff
@@ -49,4 +42,48 @@ public AssemblyStackTraceException(String message) {
...
+        }
+    }
+    
+    /**
+     * Locate the first AssemblyStackTraceException in the causal chain of the
+     * given Throwable (or it if it's one).
+     * @param e the input throwable 
+     * @return the AssemblyStackTraceException located or null if not found 
+     */
+    public static AssemblyStackTraceException find(Throwable e) {
```
artem-zinnatullin, line 74: this is test method, please move it to tests

artem-zinnatullin:
👍

src/main/java/rx/exceptions/AssemblyStackTraceException.java
akarnokd, line 74: No, this let's one get this specific exception, for example, in a live app that has a tracking enabled for diagnostic purposes.

artem-zinnatullin, line 74: okay… not sure that anybody will ever look here though

PR #4220, opened 2016-07-20
1.x: Optimize hot path in OnSubscribeOnAssembly

artem-zinnatullin:


artem-zinnatullin:
Sorry @akarnokd, I'm not against keeping `String`, but capturing stacktrace is already heavy operation. 

I've removed allocations for `StringBuilder` and searching through each stack line from hot path.

`RxJavaHooks.enableAssemblyTracking()` is a flag that will be used in production because there is nothing worse than a crash that doesn't point to your own code, performance impact should be as low as possible.

akarnokd:
I still don't approve changing the string to stacktrace array. The large cost is getting the stacktrace, its string representation is low cost then on - probably takes up less memory even.

akarnokd:
Having the string field helps in debugging as it gives immediate info about the trimmed assembly location. Debug watches with expressions don't work sometimes and looking at the unfiltered array of stacktrace elements are even more of an inconvenience.

artem-zinnatullin:
Goal here is to minimize allocations and CPU time in hot path.

With enabled `RxJavaHooks.enableAssemblyTracking()` each creation of `Observable/Single/Completable` will go through this path.

Imagine you're creating 1000s of observables per second and none of them fail, why spend time on allocations, GC and filtering strings? I don't understand.

artem-zinnatullin:
> Having the string field helps in debugging as it gives immediate info about the trimmed assembly location. Debug watches with expressions don't work sometimes and looking at the unfiltered array of stacktrace elements are even more of an inconvenience.

I belive most of users of this API just want better crash reports. Modern IDEs should work fine with "debug watches". I could be wrong of course, but that's my point of view.

akarnokd:
You enable tracking to hunt down bugs at which point app perf is secondary. You don't run with it enabled by default. Plus you are worrying on a ~1kB allocation while getting a current thread stacktrace is like 3000 cycles or so and would take up more RAM if left as array.

This API is there to help figure out which of the similar sequences crashen and link it to your code, not just to Rx operators.

The API was born out of an annoying RxNetty bug hunting session where I had trouble finding which of the dozen concatMap using sequence am I stepping through.

artem-zinnatullin:
> You enable tracking to hunt down bugs at which point app perf is secondary. You don't run with it enabled by default.

I would and will enable it for production to get more meaningful stacktraces for crashes. That's why I'm talking about hot path.

> Plus you are worrying on a ~1kB allocation while getting a current thread stacktrace is like 3000 cycles or so and would take up more RAM if left as array.

It's a tradeoff, of course, but you need to get this array no matter if you convert it to strings or not, so I don't see big problem in holding reference to it.

This PR may eat more memory for long running observables but it keeps code as fast as possible.

akarnokd:
You don't have to use the standard tracking but roll your own hook classes and apply those in `RxJavaHooks`.

artem-zinnatullin:
Closing this since we didn't come to agreement :)

PR #4222, opened 2016-07-21
2.x: subscribeOn allow cancelling before the actual subscription happens (just like 1.x)

akarnokd:
This PR fixes the `subscribeOn` operator to allow cancellation before the actual subscription to the source takes place - just like how 1.x works.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableSubscribeOnTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableSubscribeOnTest.java
@@ -24,8 +24,8 @@ import org.reactivestreams.*;
 import io.reactivex.*;
 import io.reactivex.Flowable.Operator;
 import io.reactivex.disposables.Disposable;
-import io.reactivex.internal.subscriptions.*;
-import io.reactivex.schedulers.Schedulers;
+import io.reactivex.internal.subscriptions.BooleanSubscription;
+import io.reactivex.schedulers.*;
 import io.reactivex.subscribers.*;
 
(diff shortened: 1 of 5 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableSubscribeOnTest.java
```diff
@@ -265,5 +265,20 @@ public void onNext(Integer t) {
         ts.awaitTerminalEvent();
         ts.assertNoErrors();
     }
+    
+    @Test
+    public void cancelBeforeActualSubscribe() {
+        TestScheduler test = Schedulers.test();
+        
+        TestSubscriber<Integer> ts = Flowable.just(1).hide()
+                .subscribeOn(test).test(Long.MAX_VALUE, 0, true);
```
artem-zinnatullin, line 274: Why not `dispose()`? In general, why same tests are different?

artem-zinnatullin:
👍

src/test/java/io/reactivex/internal/operators/flowable/FlowableSubscribeOnTest.java
akarnokd, line 274: `Observable` doesn't have backpressure, and a `test` method yet.

artem-zinnatullin, line 274: Ok got it, so you would use `test()` for both while I'd use `dispose()`, ok

Issue #3850, opened 2016-04-11
Unsubscribed observer receives event


alosich:
This code:

```java
import rx.Observable;
import rx.Subscription;
import rx.subjects.PublishSubject;

final class Unsubscribing {

   private static Subscription mFirst;
   private static Subscription mSecond;

   public static void main(final String[] args) {
      final PublishSubject<Integer> publisher = PublishSubject.create();

      mFirst = publisher.subscribe(i -> {
            mSecond.unsubscribe();
            log("first: " + i);
         } );

      mSecond = publisher.subscribe(i -> {
            log("second: " + i);
         } );


      publisher.onNext(1);
      publisher.onNext(2);
   }


   private static void log(final String msg) {
      System.out.println(msg);
   }

}
```

Unsubscribes second subscription in the first callback. At the time second callback called it is not subscribed (unsubscribe call completed for it). 

It's expected that second callback will not be called, but it is.

Output of that program is

```
first: 1
second: 1
first: 2
```

akarnokd:
This is due to the best-effort way of unsubscribing: PublishSubject, to remain low overhead as possible, doesn't check `isUnsubscribed` before calling `onNext` on a child and the dispatch loop works on a snapshot of the available subscribers. Therefore, if the first unsubscribes the second, that is not becoming visible to `PublishSubject.onNext` but only the next time.

akarnokd:
I'm closing this issue due to inactivity. If you have further input on the issue, don't hesitate to reopen this issue or post a new one.

alosich:
@akarnokd isn't from point of view of observer pattern, which is rx based on, this situation is incorrect?

 I mean "subscribed" and "unsubscribed" states from abstraction's point of view differs only by fact if observer continues to receive events. "Subscribed" == "receives events", "unsubscribed" == "not receives events". 

Is this something to be fixed, or user code needs additional checks to ensure that unsubscribed subscriber will not get events?

akarnokd:
Yes, you should check `mSecond.isUnsubscribed()` in this case.

alosich:
@akarnokd as a consequence this makes "unsubscribe" method non trustworthy.

Every time this method is used we need additinally check if subscriber was unsubscrubed when receiving events.

Imagine next code:

``` java
public final class SomeMechanics {
    private final Subscription mEventSubscription;

    public SomeMechanics(final Observable<Event> eventStream) {
        mEventSubscription = eventStream.subscribe(/* ... */);
    } 

[...]

PR #4225, opened 2016-07-22
1.x: PublishSubject fail-fast when backpressured

akarnokd:
This PR modifies the `PublishSubject` to fail fast if the child Subscriber can't keep up. Therefore, instead of some other operator failing somewhere downstream, the `MissingBackpressureException` now points to the `PublishSubject` instead. 

In addition, there were complaints in #3850 that cross-unsubscription doesn't stop another Subscriber from receiving events if it comes after the unsubscribe() call in the dispatch loop. Since `PublishSubject` now tracks request - which is the main extra overhead - it is trivial to add the necessary eager check for the unsubscribed state.

Benchmark comparison (i7 4790, Windows 7 x64, Java 8u102)

image

As expected, this adds some overhead although most noticeably for the mid-range only. Short-lived publishing is now slightly faster even.

Interestingly, many benchmarks behave oddly in these 1000s range - we could be hitting some JIT threshold. While in other benchmarks, the warmup iteration numbers keep increasing as JIT does its work but here, it starts out quite nicely then drops 25% and stays that way. I'm on windows so JMH -perfasm doesn't work.

If this direction is accepted, I'll update `BehaviorSubject`, `timer()` and `interval()` do do the same tracking.

### DIFF
--- a/src/main/java/rx/subjects/PublishSubject.java
+++ b/src/main/java/rx/subjects/PublishSubject.java
@@ -16,13 +16,13 @@
 package rx.subjects;
 
 import java.util.*;
+import java.util.concurrent.atomic.*;
 
+import rx.*;
 import rx.Observer;
 import rx.annotations.Beta;
-import rx.exceptions.Exceptions;
-import rx.functions.Action1;
-import rx.internal.operators.NotificationLite;
-import rx.subjects.SubjectSubscriptionManager.SubjectObserver;
+import rx.exceptions.*;
+import rx.internal.operators.BackpressureUtils;
 
 /**
  * Subject that, once an {@link Observer} has subscribed, emits all subsequently observed items to the
@@ -50,8 +50,8 @@ import rx.subjects.SubjectSubscriptionManager.SubjectObserver;
  *          the type of items observed and emitted by the Subject
  */
 public final class PublishSubject<T> extends Subject<T, T> {
-    final SubjectSubscriptionManager<T> state;
-    private final NotificationLite<T> nl = NotificationLite.instance();
+
+    final PublishSubjectState<T> state;
 
     /**
      * Creates and returns a new {@code PublishSubject}.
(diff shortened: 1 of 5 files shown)


src/main/java/rx/subjects/PublishSubject.java
```diff
@@ -144,10 +112,218 @@ public boolean hasCompleted() {
...
+        @Override
+        public void onError(Throwable e) {
+            error = e;
+            List<Throwable> errors = null;
+            for (PublishSubjectProducer<T> pp : getAndSet(TERMINATED)) {
+                try {
+                    pp.onError(e);
+                } catch (Throwable ex) {
+                    if (errors == null) {
+                        errors = new ArrayList<Throwable>();
```
artem-zinnatullin: nit: `(1)`

src/main/java/rx/subjects/PublishSubject.java
```diff
@@ -144,10 +112,218 @@ public boolean hasCompleted() {
...
+        public void onNext(T t) {
+            long r = get();
+            if (r != Long.MIN_VALUE) {
+                long p = produced;
+                if (r != p) {
+                    produced = p + 1;
+                    actual.onNext(t);
+                } else {
+                    unsubscribe();
+                    actual.onError(new MissingBackpressureException("PublishSubject: could not emit value due to lack of requests"));
```
artem-zinnatullin, line 308: 👍

src/main/java/rx/subjects/PublishSubject.java
```diff
@@ -144,10 +112,218 @@ public boolean hasCompleted() {
...
+                }
+            }
+        }
+        
+        @Override
+        public void onError(Throwable e) {
+            long r = get();
```
artem-zinnatullin: nit: inline

src/main/java/rx/subjects/PublishSubject.java
```diff
@@ -144,10 +112,218 @@ public boolean hasCompleted() {
...
+            if (r != Long.MIN_VALUE) {
+                actual.onError(e);
+            }
+        }
+        
+        @Override
+        public void onCompleted() {
+            long r = get();
```
artem-zinnatullin: nit: inline

artem-zinnatullin:
👍 

// I'm afraid that this will break some amount of user code, but this has to be done…

akarnokd:
> // I'm afraid that this will break some amount of user code, but this has to be done…

To paraphrase Godzilla (2014): _Let them fail (early)!_ :)

stevegury:
👍

felipecsl:
The recent changes to `ReplaySubject`after v1.1.5 broke my unit tests. 
Not sure if caused by this PR (#4225), #4023 or #3918. Probably #3918 since it started after I upgraded to 1.1.6.
In my specific case, I was reusing the same `TestSubscriber` after unsubscribing and then subscribing again. In this case, I'm no longer getting events on the `TestSubscriber` after resubscribing it. This is roughly what I'm doing:

``` java
ReplaySubject<String> subject = ReplaySubject.create();
TestSubscriber subscriber = new TestSubscriber();
Subscription subscription = subject.subscribe(subscriber);
subscription.unsubscribe();
subject.onNext("foo");
subject.onCompleted();
subject.subscribe(subscriber);
// Assertions below pass with v1.1.5 but fail with 1.1.6 and above
subscriber.assertCompleted();
subscriber.assertValue("foo");
```

Questions: 
1. Is this the expected behavior or a bug?
2. If expected, is there any way I can reuse the same subscriber again after it has been unsubscribed?

Thanks!

felipecsl:
Alright apparently using `TestObserver` instead of `TestSubscriber` fixes the problem, although that one is deprecated. Looks like you're not supposed to reuse a `TestSubscriber`?
I ended up rolling my own `TestObserver` that has the same convenience methods as `TestSubscriber`, like `assertNoErrors()`, `assertCompleted()`, etc.
It seems to me that `TestObserver` should not be deprecated because it seems useful in such cases. What do you think?

akarnokd:
You are not supposed to reuse `Subscriber` ever. Subscribe with a new `TestSubscriber` and assert on that.

mikehearn:
My reading of this change is that it's no longer possible to emit on a PublishSubject that has no observers (yet). If so, isn't that a rather large behaviour/API break - a lot more than just failing early?

This change broke our tests, but I don't see any discussion of this possibility above. Was it considered? I don't see anything in the javadocs that suggests that emitting on an unsubscribed subject is a problem.

akarnokd:
> My reading of this change is that it's no longer possible to emit on a PublishSubject that has no observers (yet)

PublishSubject considers each of the subscribed Subscribers individually and sending an onError if that particular Subscriber is not ready to receive `onNext` events. This doesn't prevent you from emitting `onNext` events when there is no Subscriber as these events are simply dropped by PublishSubject since day 1.

> This change broke our tests, but I don't see any discussion of this possibility above. Was it considered? I don't see anything in the javadocs that suggests that emitting on an unsubscribed subject is a problem.

Without seeing what your unit tests did it's hard to say what's wrong with them.

mikehearn:
Indeed, my mistake, sorry, I got confused reading the diff and thought it was going straight from the outer onNext to the inner producer class onNext, d'oh. I guess something subscribed has missing backpressure somewhere. It's hard to tell from the error where the root cause is though, as the exception is thrown by the PublishSubject and it doesn't say much about which subscriber had problems.

akarnokd:
Enable assembly tracking with `RxJavaHooks.enableAssemblyTracking()` around a unit test that fails and you'll get more detailed exceptions about what failed (most likely). But first make sure you don't call `onNext` concurrently on the `PublishSubject`.

Issue #4216, opened 2016-07-19
1.x: Feature Request Single#flatMapCompletable


vanniktech:
Would be nice to flatMap from Single to Completable unless there is any other way to do this right now. `flatMapObservable` already does exist.

Happy to hear thoughts on that from you guys.

akarnokd:
PR welcome.

vanniktech:
I'm curious how would you start implementing this?

akarnokd:
Subscribe to the `Single` and in the `onSuccess`, call the mapper and subscribe to the returned `Completable`.

akarnokd:
Closing via #4226.

PR #4226, opened 2016-07-22
1.x: Single.flatMapCompletable

vanniktech:
Addresses #4216 

Happy to receive feedback on the implementation. Also should I take the documentation from `flatMapObservable` and adjust it?

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1413,7 +1413,7 @@ public class Single<T> {
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX operators documentation: FlatMap</a>
      */
     public final Completable flatMapCompletable(final Func1<? super T, ? extends Completable> func) {
-        return Completable.create(new CompletableFlatMapSingleToCompletable(this, func));
+        return Completable.create(new CompletableFlatMapSingleToCompletable<T>(this, func));
     }
 
     /**
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1411,7 +1411,10 @@ public class Single<T> {
      *            Completable
      * @return the Completable returned from {@code func} when applied to the item emitted by the source Single
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX operators documentation: FlatMap</a>
+     * @Experimental The behavior of this can change at any time.
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
      */
+    @Experimental
     public final Completable flatMapCompletable(final Func1<? super T, ? extends Completable> func) {
         return Completable.create(new CompletableFlatMapSingleToCompletable<T>(this, func));
     }
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1396,6 +1396,25 @@ public class Single<T> {
         return Observable.merge(asObservable(map(func)));
     }
 
+    public final Completable flatMapCompletable(final Func1<? super T, ? extends Completable> func) {
+        return Completable.create(new Completable.CompletableOnSubscribe() {
+            @Override
+            public void call(final Completable.CompletableSubscriber completableSubscriber) {
+                subscribe(new SingleSubscriber<T>() {
+                    @Override
+                    public void onSuccess(final T value) {
+                        func.call(value).subscribe(completableSubscriber);
+                    }
+
+                    @Override
+                    public void onError(final Throwable error) {
+                        completableSubscriber.onError(error);
+                    }
+                });
+            }
+        });
+    }
+
     /**
      * Returns a Single that applies a specified function to the item emitted by the source Single and
      * emits the result of this function application.
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1396,23 +1396,24 @@ public class Single<T> {
         return Observable.merge(asObservable(map(func)));
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


akarnokd:
The unsubscription is not properly linked, the possible exception thrown by the mapper is not handled.

See [this](https://gist.github.com/akarnokd/260ea8b121ec4071401808c1594f2230).

> Also should I take the documentation from flatMapObservable and adjust it?

Yes.

src/main/java/rx/Single.java
```diff
@@ -1396,6 +1396,25 @@ public R call(Object... args) {
         return Observable.merge(asObservable(map(func)));
     }
 
+    public final Completable flatMapCompletable(final Func1<? super T, ? extends Completable> func) {
```
artem-zinnatullin, line 1418: javadoc

src/test/java/rx/SingleTest.java
```diff
@@ -2005,4 +2005,32 @@ public void unsubscribeComposesThroughAsync() {
...
     }
 
+    @Test
+    public void flatMapCompletableComplete() {
+        TestSubscriber testSubscriber = new TestSubscriber();
+
+        Single.just(1).flatMapCompletable(new Func1<Integer, Completable>() {
+            @Override
+            public Completable call(final Integer integer) {
+                return Completable.complete();
```
artem-zinnatullin: Would be better to test against more complicated `Completable`, for example check that some function was invoked

src/test/java/rx/SingleTest.java
```diff
@@ -2005,4 +2005,32 @@ public void unsubscribeComposesThroughAsync() {
...
+            }
+        }).subscribe(testSubscriber);
+
+        testSubscriber.assertCompleted();
+    }
+
+    @Test
+    public void flatMapCompletableError() {
```
artem-zinnatullin, line 2031: 👍

vanniktech:
Updated the documentation. I hope the wording is okay now. Also I'm pointing to https://raw.githubusercontent.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.flatMapCompletable.png which does not exist yet but  @DavidMGross already tracked it at https://github.com/ReactiveX/reactivex.github.io/issues/289, I hope that's okay.

thanks for the `CompletableFlatMapSingleToCompletable` I knew something was missing. I also created two more tests that cover some of the previously missing behaviour.

src/main/java/rx/Single.java
```diff
@@ -1397,6 +1397,26 @@ public R call(Object... args) {
...
+     * </dl>
+     *
+     * @param func
+     *            a function that, when applied to the item emitted by the source Single, returns a
+     *            Completable
+     * @return the Completable returned from {@code func} when applied to the item emitted by the source Single
+     * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX operators documentation: FlatMap</a>
+     */
+    public final Completable flatMapCompletable(final Func1<? super T, ? extends Completable> func) {
+        return Completable.create(new CompletableFlatMapSingleToCompletable(this, func));
```
akarnokd: You forgot to commit the class itself. Plus, use `new CompletableFlatMapSingleToCompletable<T>(this, func)` as it is a generic class.

vanniktech: Yup I just comitted it. Got confused since I had to stage the files individually because of the star imports ...

src/main/java/rx/internal/operators/CompletableFlatMapSingleToCompletable.java
```diff
@@ -0,0 +1,92 @@
...
+import rx.SingleSubscriber;
+import rx.Subscription;
+import rx.exceptions.Exceptions;
+import rx.functions.Func1;
+
+public final class CompletableFlatMapSingleToCompletable<T> implements CompletableOnSubscribe {
+
+    final Single<T> source;
+
+    final Func1<? super T, Completable> mapper;
```
akarnokd: You have to change this to `? extends Completable` (and update the other places) so it compiles with the operator signature in `Single`

src/main/java/rx/Single.java
```diff
@@ -1397,6 +1397,26 @@ public R call(Object... args) {
...
+     * <dd>{@code flatMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param func
+     *            a function that, when applied to the item emitted by the source Single, returns a
+     *            Completable
+     * @return the Completable returned from {@code func} when applied to the item emitted by the source Single
+     * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX operators documentation: FlatMap</a>
+     */
+    public final Completable flatMapCompletable(final Func1<? super T, ? extends Completable> func) {
```
akarnokd, line 1418: Please tag this as `@Experimental`.

akarnokd:
:+1:

vanniktech:
Would it be possible to get this one into 1.1.8?

akarnokd:
If @zsxwing or anyone from Netflix approves it in time.

src/main/java/rx/Single.java
```diff
@@ -1397,6 +1397,29 @@ public R call(Object... args) {
     }
 
     /**
+     * Returns a Completable that completes based on applying a specified function to the item emitted by the
```
artem-zinnatullin: nit: `{@link Completable}`

src/main/java/rx/Single.java
```diff
@@ -1397,6 +1397,29 @@ public R call(Object... args) {
...
+     * source Completable, where that function returns a Completable.
```
artem-zinnatullin: and here

artem-zinnatullin:
👍

Issue #4228, opened 2016-07-22
OperatorReplay leaks its subscribers


ntoskrnl:
We use `replay(1)` a lot in our Android projects. All subscriptions are kept in `CompositeSubscription` and, of course, we always clear it when we don't need it anymore.

Even though we unsubscribe from observable, subscriber won't get collected by GC, because it is referenced by `OperatorReplay`, which prevents Activity or Fragment from being garbage collected. We do need to keep observable alive and resubscribe to it from new Activity instance, but now we can't use replay() as all our Activities are leaked.

Right now we downgraded back to 1.1.5. Alternatively, one can use `publish()` with behavior subject to work around this issue. I believe the issue was introduced in one of the latest versions of the library (perhaps 1.1.6, not sure though).

akarnokd:
Do you have your replay in a completed state or with infrequent updates?

ntoskrnl:
Infinite stream is "replayed", so, I guess, it is never in a completed state. 
What do you mean by infrequent updates?

ntoskrnl:
Connectable observable with replay example (no error handling to keep it simple): 

``` kotlin
val dataStream = refreshes.startWith(Unit)
        .flatMap { model.getData() }
        .replay(1)
```

When presenter is created we connect to the stream and store subscription until presenter is destroyed:

``` kotlin
val presenterSubscription = CompositeSubscription()
val viewSubscription = CompositeSubscription()

presenterSubscription.add(dataStream.connect())
```

View is attached to presenter:

``` kotlin
// subscribe view to the data stream, subscriber has a strong reference to view
viewSubscription.add(dataStream.subscribe {  view.showContent(it) })
```

View detached from presenter:

``` kotlin
// clear view subscription
// after this there should be no strong references to the view
viewSubscription.clear()
```

When presenter is destroyed (when activity is closed): 

``` kotlin
presenterSubscription.unsubscribe()
```

Throughout the presenter lifecycle, different instances of view are attached and then detached (for instance, during screen rotation when Activity is recreated). However, after the new view instance is attached, the old instance is still referenced from within OperatorReplay. According to heap dump analysis, each time there exist only two instances of the view – no matter how I many times I rotate the screen, there are two instances of view: current and only 1 leaked view.

ntoskrnl:
👍

akarnokd:
[...]

PR #3470, opened 2015-10-26
1.x: replay request coordination reduce overhead

akarnokd:
Reported in #3469.

This change will make sure whenever a new subscriber arrives, the
request coordination doesn't scan every other subscriber unnecessary.
(They will trigger a manageRequests individually anyway).

Unfortunately, I can't make the subscription registration constant (via a HashSet) because that would slow down the dispatching loop (which is quite expensive to for-each over compared to a plain array).

Note also that this will likely conflict with #3454.

I get the following numbers with the program in #3469.

```
subscribing...
subscribed so far 10000; took 54ms to subscribe last 10000 at rate: 185185 per sec...
subscribed so far 20000; took 110ms to subscribe last 10000 at rate: 90909 per sec...
subscribed so far 30000; took 157ms to subscribe last 10000 at rate: 63694 per sec...
subscribed so far 40000; took 267ms to subscribe last 10000 at rate: 37453 per sec...
subscribed so far 50000; took 211ms to subscribe last 10000 at rate: 47393 per sec...
subscribed so far 60000; took 320ms to subscribe last 10000 at rate: 31250 per sec...
subscribed so far 70000; took 331ms to subscribe last 10000 at rate: 30211 per sec...
subscribed so far 80000; took 369ms to subscribe last 10000 at rate: 27100 per sec...
subscribed so far 90000; took 415ms to subscribe last 10000 at rate: 24096 per sec...
subscribed so far 100000; took 462ms to subscribe last 10000 at rate: 21645 per sec...
subscribed so far 110000; took 516ms to subscribe last 10000 at rate: 19379 per sec...
subscribed so far 120000; took 547ms to subscribe last 10000 at rate: 18281 per sec...
subscribed so far 130000; took 647ms to subscribe last 10000 at rate: 15455 per sec...
subscribed so far 140000; took 720ms to subscribe last 10000 at rate: 13888 per sec...
subscribed so far 150000; took 811ms to subscribe last 10000 at rate: 12330 per sec...
subscribed so far 160000; took 806ms to subscribe last 10000 at rate: 12406 per sec...
subscribed so far 170000; took 873ms to subscribe last 10000 at rate: 11454 per sec...
subscribed so far 180000; took 911ms to subscribe last 10000 at rate: 10976 per sec...
subscribed so far 190000; took 986ms to subscribe last 10000 at rate: 10141 per sec...
subscribed so far 200000; took 1003ms to subscribe last 10000 at rate: 9970 per sec...
[...]

PR #4229, opened 2016-07-22
1.x: fix replay() retaining reference to the child Subscriber

akarnokd:
In #3470, a flat cache of Subscribers has been introduced that is versioned and gets only updated if there are new Subscribers coming, old Subscribers leaving _and_ a value comes through - improving performance if lots of Subscribers come and go between main onNext signals. This cache is not cleared but rebuilt therefore child Subscribers unsubscribing were still referenced through it and everything beyond - leading to memory leak on sensitive systems.

The fix is to cut the link to the child Subscriber in `InnerProducer` on unsubscription. A secondary action also empties the cache if all Subscribers are gone.

Reported in #4228 .


akarnokd:
This is a medium impact bug so it would be great if it made into tomorrow's release. /cc @artem-zinnatullin @zsxwing

stevegury:
👍

PR #4231, opened 2016-07-23
CachedThreadScheduler should wait until the previous action (if any) …

### LINKED ISSUE #4230: Schedulers.io() reuses blocked threads
Some actions are not interruptible. Since Schedulers.io() reuses threads after unsubscribe, a previous uninterruptible action might block execution of new actions, causing unexpected delays or in some cases, deadlocks.

csabakos-zz:
…completes before releasing a worker to the pool. Fixes #4230.

### DIFF
--- a/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
+++ b/src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
@@ -190,7 +190,14 @@ public final class CachedThreadScheduler extends Scheduler implements SchedulerL
         public void unsubscribe() {
             if (once.compareAndSet(false, true)) {
                 // unsubscribe should be idempotent, so only do this once
-                pool.release(threadWorker);
+
+                // Release the worker _after_ the previous action (if any) has completed
+                threadWorker.schedule(new Action0() {
+                    @Override
+                    public void call() {
+                        pool.release(threadWorker);
+                    }
+                });
             }
             innerSubscription.unsubscribe();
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/schedulers/CachedThreadScheduler.java
```diff
@@ -190,7 +190,14 @@ public Worker createWorker() {
         public void unsubscribe() {
             if (once.compareAndSet(false, true)) {
                 // unsubscribe should be idempotent, so only do this once
-                pool.release(threadWorker);
+
+                // Release the worker _after_ the previous action (if any) has completed
+                threadWorker.schedule(new Action0() {
```
akarnokd: EventLoopWorker can implement Action0 and schedule itself to release.

csabakos-zz: @akarnokd What would be the advantage of doing so?

akarnokd: It saves an allocation; I believe io() workers are used and unsubscribed frequently enough to save on this allocation.

csabakos-zz: Makes sense, let me change it.

akarnokd:
:+1: 

Given a misbehaving task, this at least prevents other work to be scheduled on the same pool.

PR #4232, opened 2016-07-23
save allocation in OnSubscribeAmb

davidmoten:
`Selection` class can extend `AtomicReference` to save an allocation


akarnokd:
:+1:

PR #4233, opened 2016-07-23
save allocation in OnSubscribeAutoConnect

davidmoten:
Another one. I'll keep rolling through the operators to pick up more. 

@akarnokd Are you ok with `@SuppressWarnings("serial")` instead of `private static final long serialVersionUID = ...`?


akarnokd:
I prefer `serialVersionUID` but I'm not nitpicky like some...

akarnokd:
:+1:

PR #4236, opened 2016-07-23
save allocations in OnSubscribeJoin

davidmoten:
as per title.

Would have been nicer to extend `CompositeSubscription` but is final.


akarnokd:
:+1:

PR #4239, opened 2016-07-25
save allocation in OperatorSkipTimed

davidmoten:
as per title, combined `AtomicBoolean` with an `Action`.


src/main/java/rx/internal/operators/OperatorSkipTimed.java
```diff
@@ -77,4 +72,15 @@ public void onCompleted() {
             }
         };
     }
+    
+    final static class AtomicBooleanAction extends AtomicBoolean implements Action0 {
```
JakeWharton: If you subclass `Subscriber`, implement `Action0`, and use a volatile field for the boolean then you can save even more!

davidmoten:
Good point, thanks @JakeWharton, I'll do that.

davidmoten:
Ah, the trouble with that idea @JakeWharton is that I need to create a non-anonymous class to subclass `Subscriber` and implement `Action0` and that class needs to hold a reference to the `child` subscriber. Thus I end up with the same number of allocations.

src/main/java/rx/internal/operators/OperatorSkipTimed.java
akarnokd: Yes, this is better. You don't use CAS on the boolean thus a regular  volatile field suffices. Then you have 1 allocation. Also switching to OnSubscribe saves another.

davidmoten:
happy to switch to `onSubscribe` while I'm here. I'll do that.

davidmoten:
Got rid of `AtomicBoolean`, subscriber implements `Action0`, converted to `OnSubscribe`

davidmoten:
@JakeWharton  upon reflection I realize there's no cost moving from anonymous to non-anonymous class because the anonymous class would have included at least one implicit external reference to access the `child` object. So thanks for the suggestion.

src/main/java/rx/internal/operators/OnSubscribeSkipTimed.java
```diff
@@ -0,0 +1,93 @@
...
+        this.unit = unit;
+        this.scheduler = scheduler;
+    }
+
+    @Override
+    public void call(final Subscriber<? super T> child) {
+        final Worker worker = scheduler.createWorker();
+        child.add(worker);
+        SkipTimedSubscriber<T> subscriber = new SkipTimedSubscriber<T>(child);
+        worker.schedule(subscriber, time, unit);
```
akarnokd, line 50: `child.add(subscriber)` missing.

davidmoten, line 50: ah oops, thank you @akarnokd . I just need to add super(child) in the `SkipTimedSubscriber` constructor. You mentioning that makes me realize that there is probably a backpressure related bug in the existing operator in as much as when we don't emit to the child we should request another (like `filter`). I'll write a unit test for unsubscription as penance and another to demonstrate the backpressure problem (if it exists).

akarnokd, line 50: That's wrong because it calls `unsubscribe` in the `onError`/`onCompleted` path and unsubscribes the child. The `worker` should be added to `subscriber` instead and `subscriber` added to the `child`. No constructor forwarding.

akarnokd, line 50: It doesn't support backpressure:

``` java
/**
 * Returns an Observable that skips values emitted by the source Observable before a specified time window
 * elapses.
 * <p>
 * <img width="640" height="305" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/skip.t.png" alt="">
 * <dl>
 *  <dt><b>Backpressure:</b></dt>
 *  <dd>The operator doesn't support backpressure as it uses time to skip arbitrary number of elements and
 *  thus has to consume the source {@code Observable} in an unbounded manner (i.e., no backpressure applied to it).</dd>
 *  <dt><b>Scheduler:</b></dt>
 *  <dd>This version of {@code skip} operates by default on the {@code computation} {@link Scheduler}.</dd>
 * </dl>
 * 
 * @param time
 *            the length of the time window to skip
 * @param unit
 *            the time unit of {@code time}
 * @return an Observable that skips values emitted by the source Observable before the time window defined
 *         by {@code time} elapses and the emits the remainder
 * @see <a href="http://reactivex.io/documentation/operators/skip.html">ReactiveX operators documentation: Skip</a>
 */
```

davidmoten, line 50: Constructor forwarding handles requests for us too though.

davidmoten, line 50: Shouldn't share subscriptions but if I use the constructor `super(child, false)` then we get requests passed through from the child.

davidmoten, line 50: Um that's odd about backpressure support in the javadoc. Does the approach I suggest not support backpressure?

akarnokd, line 50: That would lead to hot spinning for the duration of the time skip.

davidmoten, line 50: ah I'm with you, thanks

davidmoten, line 50: actually maybe I'm not with you yet. The alternative is to request `Long.MAX_VALUE` and it might still spin like crazy in an upstream operator. It just doesn't seem different from `filter`...

akarnokd, line 50: Requesting 1 each time incurs an atomic increment costing 22-45 cycle overhead. For filter, the element count is the main deciding factor. For skip its time which can translate to arbitrary element count. The problem is you can't know what the source is.

davidmoten, line 50: > For filter, the element count is the main deciding factor

That's assuming that the predicate doesn't use time.

> For skip its time which can translate to arbitrary element count. The problem is you can't know what the source is.

I'm not sure why we wouldn't accept the atomic increment cost as the price to have this operator support backpressure.

akarnokd:
:+1:

Issue #4175, opened 2016-07-07
testRetryWithBackpressureParallel failed


akarnokd:
Just [failed on Travis](https://travis-ci.org/ReactiveX/RxJava/builds/143126717). Last time it was some time ago.

```
rx.internal.operators.OperatorRetryTest > testRetryWithBackpressureParallel FAILED
    java.lang.AssertionError: Data content mismatch: 15={beginningEveryTime x 128}
        at org.junit.Assert.fail(Assert.java:88)
        at rx.internal.operators.OperatorRetryTest.testRetryWithBackpressureParallel(OperatorRetryTest.java:771)
```

akarnokd:
Closing as appears to be fixed with #4241.

PR #4241, opened 2016-07-25
Fix retryWhen/repeatWhen potential concurrent reentry when resubscribing

akarnokd:
When the function's returned Observable signals the resubscription in the operator `redo` (which is the base for `repeatWhen` and `retryWhen`). It is possible an asyncronous source triggers the function (and thus the same `Observable`) from another thread while the first is still coming back from the resubscription itself.

This PR serializes the dispatching `BehaviorSubject` to prevent such concurrent execution of the when chain.

This may or may not be the source of the failure of #4175 but I can't reproduce the failure with the 1.x branch. /cc @davidmoten.


artem-zinnatullin:
👍

davidmoten:
Worth a shot, gnarly bug :+1:

PR #4244, opened 2016-07-27
OperatorAll - prevent multiple terminal events

davidmoten:

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorAll.java
+++ b/src/main/java/rx/internal/operators/OperatorAll.java
@@ -20,6 +20,7 @@ import rx.Subscriber;
 import rx.exceptions.Exceptions;
 import rx.functions.Func1;
 import rx.internal.producers.SingleDelayedProducer;
+import rx.plugins.RxJavaHooks;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/operators/OperatorAll.java
```diff
@@ -61,7 +61,10 @@ public void onNext(T t) {
 
             @Override
             public void onError(Throwable e) {
-                child.onError(e);
+                if (!done) {
+                    done = true;
+                    child.onError(e);
+                }
```
akarnokd, line 73: Call `RxJavaHooks.onError` so no exception is lost.

vanniktech, line 73: Since you pointed it out in both PRs would not this also apply to some other places?

akarnokd, line 73: Yes, anyplace that would drop the exception due to state/protocol considerations should instead call `RxJavaHooks.onError`.

davidmoten:
- added unit tests for `(onNext, onNext)` and `(onNext, onError)` cases

akarnokd:
:+1:

PR #4245, opened 2016-07-27
OperatorAny - prevent multiple terminal events

davidmoten:


src/main/java/rx/internal/operators/OperatorAny.java
```diff
@@ -64,7 +64,10 @@ public void onNext(T t) {
 
             @Override
             public void onError(Throwable e) {
-                child.onError(e);
+                if (!done) {
+                    done = true;
+                    child.onError(e);
+                }
```
akarnokd, line 76: Call `RxJavaHooks.onError` so no exception is lost.

davidmoten:
- added two unit tests for `(onNext, onNext)` and one for `(onNext, onError)` cases

PR #4246, opened 2016-07-27
OperatorReduce - prevent multiple terminal events

davidmoten:

Also renamed test class to match tested class.


src/main/java/rx/internal/operators/OnSubscribeReduce.java
```diff
@@ -77,19 +79,26 @@ public void onNext(T t) {
...
                 }
             }
         }
         
         @Override
         public void onError(Throwable e) {
-            actual.onError(e);
+            if (!done) {
+                done = true;
+                actual.onError(e);
```
vanniktech, line 95: Isn't here `RxJavaHooks.onError` missing?

davidmoten, line 95: yep, fixing

davidmoten:
- added unit tests including coverage of calling `.reduce` on an empty stream

PR #4249, opened 2016-07-27
Value, Variable and Immutable first implementation

fvasco:
Hi,
there is a simple implementation for an observable value, in the mutable or immutable form.
This helps me to share some state (like configuration) when in some case is important to catch every change (using normal subscription) and in other case is useful get only the current value.

I hope this helps,
thanks for software,
Francesco

### DIFF
--- a/src/main/java/io/reactivex/Immutable.java
+++ b/src/main/java/io/reactivex/Immutable.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * <p>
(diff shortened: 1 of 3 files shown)


akarnokd:
Hi and thanks for the contribution but I'm not convinced that these additions belong to RxJava itself. Since the PR doesn't need operators on `Observable` or `Flowable`, they can live in any other library external to RxJava.

fvasco:
Thank you for your response,
I'm not interested to do that.

I will continue to use my Kotlin personal implementation.

PR #4251, opened 2016-07-28
toMap - prevent multiple terminal events, support backpressure

davidmoten:

Includes three unit tests that failed on the original code. The fix also has the side effect of enabling gc of `map` when the factory fails (`onError` was called on the child, not `this` so `map` was not set to null).

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeToMap.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeToMap.java
@@ -16,13 +16,15 @@
 
 package rx.internal.operators;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.Map;
 
-import rx.Observable.Operator;
+import rx.Observable;
+import rx.Observable.OnSubscribe;
 import rx.Subscriber;
 import rx.exceptions.Exceptions;
-import rx.functions.*;
-import rx.observers.Subscribers;
+import rx.functions.Func0;
+import rx.functions.Func1;
 
 /**
  * Maps the elements of the source observable into a java.util.Map instance and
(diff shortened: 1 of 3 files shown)


akarnokd:
I thought it supported backpressure. Could you rewrite it by using `DeferredScalarSubscription`?

davidmoten:
unrelated test failure:

```
rx.schedulers.ComputationSchedulerTests > testHandledErrorIsNotDeliveredToThreadHandler FAILED
    java.lang.AssertionError: Handler should not have received anything expected:<0> but was:<1>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at rx.schedulers.SchedulerTests.testHandledErrorIsNotDeliveredToThreadHandler(SchedulerTests.java:102)
        at rx.schedulers.ComputationSchedulerTests.testHandledErrorIsNotDeliveredToThreadHandler(ComputationSchedulerTests.java:153)
```

davidmoten:
> I thought it supported backpressure. Could you rewrite it by using DeferredScalarSubscription?

Sure.

davidmoten:
What I might do is fix multiple terminal emissions in  `OnSubscribeCollect` first and make necessary `done` changes to `DeferredScalarSubscriber` in that PR. I'll follow that with the backpressure fix for `OperatorToMap`.

src/test/java/rx/internal/operators/OperatorToMapTest.java
```diff
@@ -281,4 +296,111 @@ public Integer call(Integer v) {
...
+                public Integer call(Integer t) {
+                    throw e1;
+                }
+            }).unsafeSubscribe(ts);
+            ts.assertNoValues();
+            assertEquals(Arrays.asList(e1), ts.getOnErrorEvents());
+            assertEquals(Arrays.asList(e2), list);
+            ts.assertNotCompleted();
+        } finally {
+            RxJavaHooks.setOnError(null);
```
akarnokd: This needs fixing as well.

davidmoten:
@akarnokd 

In terms of releasing a value so can be gc'd I thought to modify `DeferredScalarSubscriber` in this method so that the field `this.value` is set to null just before the call to `a.onNext()`:

``` java
protected final void complete(R value) {
        Subscriber<? super R> a = actual;
        for (;;) {
            int s = state.get();

            if (s == NO_REQUEST_HAS_VALUE || s == HAS_REQUEST_HAS_VALUE || a.isUnsubscribed()) {
                return;
            }
            if (s == HAS_REQUEST_NO_VALUE) {
                R v = value; // <--------------------------------
                value = null; // <--------------------------------
                a.onNext(v);
                if (!a.isUnsubscribed()) {
                    a.onCompleted();
                }
                state.lazySet(HAS_REQUEST_HAS_VALUE);
                return;
            }
            this.value = value;
            if (state.compareAndSet(NO_REQUEST_NO_VALUE, NO_REQUEST_HAS_VALUE)) {
                return;
            }
        }
    }
```

Is that ok?

davidmoten:
don't review latest change, just committed so could work on it from another location

akarnokd:
I'm not sure if it's worth it. Many operators don't really do that because it is also likely the whole chain gets forgotten and GC claims all of them on its own. If you want to make sure there is no leak, use `onTerminateDetach`.

akarnokd:
Could you rebase the whole thing? Somehow, you picked up changes to master and they show up in the diff.

davidmoten:
- Rebased
- Rewrote to use `DeferredScalarSubscriberSafe`
- use singleton of `DefaultMapFactory`
- moved to `OnSubscribe` to save allocations
- added backpressure test

akarnokd:
:+1:

src/main/java/rx/internal/operators/OnSubscribeToMap.java
```diff
@@ -33,35 +35,26 @@
...
      * ToMap with key selector, value selector and default HashMap factory.
      * @param keySelector the function extracting the map-key from the main value
      * @param valueSelector the function extracting the map-value from the main value
      */
-    public OperatorToMap(
+    public OnSubscribeToMap(Observable<T> source,
             Func1<? super T, ? extends K> keySelector,
             Func1<? super T, ? extends V> valueSelector) {
-        this(keySelector, valueSelector, new DefaultToMapFactory<K, V>());
+        this(source, keySelector, valueSelector, DefaultMapFactory.<K, V>instance());
```
akarnokd: A small trick: make the OnSubscribeToMap also implement `Func0<Map<K, V>>` and assign `this` to the map factory function - of course you can't use `this()` constructor delegation but assign to the final fields explicitly. This way, there is no need for an extra class.

davidmoten:
Good idea, I've updated the PR.

akarnokd:
:+1:

PR #4252, opened 2016-07-28
collect, DeferredScalarSubscriber - prevent multiple terminal emissions

davidmoten:
As per discussion in #4242, if an operator maps an `onNext` emission to an `onError` emission downstream then it needs be defensive about another event being sent from upstream even if upstream has been unsubscribed.
- `DeferredScalarSubscriber` has been updated with a `done` flag
- moved tests from `ObservableTests` to new class `OnSubscribeCollectTest`
- added three tests of post error emissions
- added factory failure test

### DIFF
--- a/src/test/java/rx/internal/operators/OnSubscribeCollectTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeCollectTest.java
@@ -1,0 +1,253 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package rx.internal.operators;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+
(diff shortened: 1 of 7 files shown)


src/test/java/rx/internal/operators/OnSubscribeCollectTest.java
```diff
@@ -0,0 +1,253 @@
...
+                        @Override
+                        public void call(List<Integer> t1, Integer t2) {
+                            throw e1;
+                        }
+                    }).unsafeSubscribe(ts);
+            assertEquals(Arrays.asList(e1), ts.getOnErrorEvents());
+            ts.assertNotCompleted();
+            assertEquals(Arrays.asList(e2), list);
+        } finally {
+            RxJavaHooks.setOnError(null);
```
akarnokd: You have to call `reset` otherwise RxJavaHooks wont delegate back to `RxJavaPlugins`.

davidmoten: Ah, ok, I see from the source that your are right thus the javadoc on `setOnError` is wrong (or `setOnError` needs fixing).

src/main/java/rx/internal/operators/DeferredScalarSubscriber.java
```diff
@@ -56,19 +57,31 @@
     /** Value will be emitted. */
     static final int HAS_REQUEST_HAS_VALUE = 3;
     
+    /** prevents multiple terminal emissions. */
+    protected boolean done;
```
akarnokd: The problem with this is that not all uses need this flag or the logic. For example, takeLastOne just stores the last value and there is no function involved that could crash. Operators that may crash should individually add these or introduce another abstract class that adds this feature.

davidmoten: That's fair, I'll rejig.

davidmoten:
- Created a new class `DeferredScalarSubscriberSafe` that extends `DeferredScalarSubscriber` and offers the `done` checks. `done` is available as a protected field and `OnSubscribeCollect` uses it in its `onNext` method.
- Replaced the calls in tests to `RxJavaHooks.setOnError(null)` with calls to `RxJavaHooks.reset()` in this PR and in recent similar PRs for operators `any`, `all` and `reduce`.
- I'll submit another PR to fix the javadoc of `RxJavaHooks.setOnError`

akarnokd:
:+1:

PR #4254, opened 2016-07-28
1.x: fix SingleTest generics warnings

akarnokd:
This fixes the generics warnings in the new `Single.flatMapCompletable` unit tests.


vanniktech:
👍 didn't see those since I have Java 8 as default in my IDE

akarnokd:
I'm using Eclipse which turned out to be an excellent tool to warn about these things. Are you using IntelliJ? I've noticed that people using IntelliJ end up with such hidden problems (while IntelliJ for me complained about almost everything else when I loaded RxJava into it once).

vanniktech:
Yes I'm using Android Studio which is based on IntelliJ. That's weird I had no problems loading the project into AS.

PR #4257, opened 2016-07-29
Fix javadoc on RxJavaHooks.setOnError

davidmoten:
as per title


davidmoten:
Don't know what happened but I suppose it's unrelated:

```
Caused by: org.eclipse.jgit.errors.MissingObjectException: Missing commit 75ffcbb01cae042f5b715060622154794f39a690
```

akarnokd:
:+1:

By default, Travis checks out the last N commits. If that last N doesn't contain a version tag, the Nebula plugins' version inference freaks out. I have to keep increasing that value if there are too many commits between releases.

Issue #4152, opened 2016-06-30
1.x: concatDelayError with multiple args


bryant1410:
I think it'd be great if `concatDelayError` supports multiple parameters as `mergeDelayError`.

Btw, doesn't `mergeDelayError` miss a variation in which a variable number of args is admitted?

akarnokd:
The general problem with these that they expand the API surface (like + 8 methods per operator) which is already 'bloated'.

We don't do variable arguments in Java 6 because it necessitates `@SuppressWarnings` almost always (hence the convenience 2-9 arg operators).

akarnokd:
Closing via #4261.

PR #4261, opened 2016-07-31
1.x: concatDelayError multiple arguments

SherifMakhlouf:
This PR adds multiple arguments to concatDelayError operator

Related to: #4152

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -1453,15 +1453,15 @@ public class Observable<T> {
     }
 
     /**
-     * Concatenates two Observables into a single sequence by subscribing to each Observable,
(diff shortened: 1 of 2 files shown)


src/main/java/rx/Observable.java
```diff
@@ -1453,6 +1453,270 @@ public Completable toCompletable() {
     }
 
     /**
+     * Concatenates two Observables into a single sequence by subscribing to each Observable,
```
akarnokd: Could you update the wording according to the [plain `concat`](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L1132) methods as there are no outer and inner Observables with these methods?

akarnokd:
:+1:

Issue #4260, opened 2016-07-30
OperatorDoEach could do with some work


davidmoten:
While looking for emissions after terminal events I was looking at `OperatorDoEach` which is called by `doOnNext`, `doOnError`, `doOnCompleted`, and `doOnEach` of course. I noticed these things:
- lots of wasted allocations per call (use of `ActionSubscriber` when all that is needed is an `ActionObserver`) and passing three actions to `OperatorDoEach` when for three of the four operators we only need one action.
- it doesn't call `RxJavaHooks.error` for secondary error emissions from upstream
- it does call `Exceptions.throwIfFatal` for all error emissions (do we want to do this generally? I haven't seen it much outside error catching on externally supplied functions or actions)

I'd quite like to see `doOnNext` and `doOnError` in particular more efficient because I use them a lot. Should I create dedicated Operator/OnSubscribe classes for the single action operators or perhaps one shared one that has an extra parameter (to indicate what notification type is being used)?

akarnokd:
You may optimize it into an OnSubscribe. I'm not sure about splitting as it triples the method count for the sake of saving 16 bytes.

davidmoten:
I could also use `ActionObserver` instead of `ActionSubscriber` I suppose (saves 32 bytes). Let me know what you think of the `throwIfFatal` issue and I'll do it.

akarnokd:
The `throwIfFatal` shouldn't be there; the operators should not call onError with a fatal exceptions anymore but throw them immediately - this should be also reviewed and fixed in 2.x as well.

akarnokd:
FYI, I follow [this](https://gist.github.com/akarnokd/693866d6b4453ca8c66b866584163e5e) list when I port tests.

akarnokd:
Closing via #4262.

PR #4262, opened 2016-07-31
doOnEach - reduce allocations, report to RxJavaHooks

davidmoten:
as discussed in #4260 
- call `RxJavaHooks.onError` for secondary errors
- reduced allocations (`ActionSubscriber` -> `ActionObserver`, `Operator` -> `OnSubscribe`)
- added 5 unit tests


src/main/java/rx/Observable.java
```diff
@@ -5509,7 +5509,7 @@ public Completable toCompletable() {
         Action1<Throwable> onError = Actions.empty();
         Observer<T> observer = new ActionSubscriber<T>(onNext, onError, onCompleted);
```
akarnokd: ActionObserver?

src/test/java/rx/internal/operators/OnSubscribeReduceTest.java
```diff
@@ -273,5 +273,9 @@ public Integer call(Integer a, Integer b) {
         ts.assertError(NoSuchElementException.class);
     }
 
-
+    @Test
+    public void test() {
+
```
akarnokd: Empty test?

src/main/java/rx/Observable.java
davidmoten: ta

davidmoten:
fixed ActionSubscriber usage and removed empty test, ta

akarnokd:
:+1:

PR #4265, opened 2016-08-01
remove unnecessary map call in sequenceEquals operator

davidmoten:
This PR is a performance improvement in `sequenceEquals` where I removed a call to `map` that appears unnecessary.


akarnokd:
:+1:

PR #4266, opened 2016-08-01
2.x: add dedicated concat for array of publishers

akarnokd:
This PR adds a dedicated `concat` implementation that works with the source array. The main benefit is that there is no need to serialize emissions because the source array can't throw (unlike `concatMap` where there could be an onError from the main source while an inner source is running). Plus, there is a new `concatArrayDelayError` version.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -373,12 +373,11 @@ public abstract class Flowable<T> implements Publisher<T> {
         return concatArray(p1, p2, p3, p4, p5, p6, p7, p8, p9);
     }
 
-    @SuppressWarnings({ "unchecked", "rawtypes" })
     @BackpressureSupport(BackpressureKind.FULL)
     @SchedulerSupport(SchedulerSupport.NONE)
+    @Deprecated // prefetch is unnecessary when the sources is synchronously available
     public static <T> Flowable<T> concatArray(int prefetch, Publisher<? extends T>... sources) {
-        Objects.requireNonNull(sources, "sources is null");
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -400,7 +398,27 @@ public static int bufferSize() {
...
+     * @param <T> the common base value type
+     * @return the new Flowable instance
+     * @throws NullPointerException if sources is null
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public static <T> Flowable<T> concatArrayDelayError(Publisher<? extends T>... sources) {
+        if (sources.length == 0) {
+            return empty();
+        } else
```
JakeWharton, line 417: I still find the line break between `else` and `if` really weird... although you don't even need the `else` after `return`.

JakeWharton:
lgtm :+1:

src/main/java/io/reactivex/Flowable.java
akarnokd, line 417: If you wish, you can go around and fix these.

JakeWharton, line 417: I'm more worried about the cause than the effect. This seems less readable and less intuitive than any format I've seen before. What's the motivation behind it?

akarnokd, line 417: I've been using this style for a long time now and for me, not seeing an `else` is an indication of a likely unintended fall-through case.

JakeWharton, line 417: But how's that any different from the overwhelmingly more common `else if` on one line?

PR #4270, opened 2016-08-02
toMultimap - prevent post terminal emissions, reduce allocations

davidmoten:
- switched to `OnSubscribe`
- switched to use `DeferredScalarSubscriberSafe`
- previous source failed to handle post terminal emissions 
- added 4 unit tests


akarnokd:
:+1:

PR #4276, opened 2016-08-03
1.x: distinctUntilChanged change erroneous behavior

godardt:
#### 📘 PR summary

Fix behavior change of `distinctUntilChanged`  introduced in the version `1.1.7` of RxJava

Related to: #4034

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
+++ b/src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
@@ -62,7 +62,7 @@ public final class OperatorDistinctUntilChanged<T, U> implements Operator<T, T>,
     
     @Override
     public Boolean call(U t1, U t2) {
-        return (t1 == t2 || (t1 != null && t1.equals(t2)));
+        return (t1 == t2 || (t2 != null && t2.equals(t1)));
     }
 
     @Override


src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
```diff
@@ -62,7 +62,7 @@ public OperatorDistinctUntilChanged(Func2<? super U, ? super U, Boolean> compara
     
     @Override
     public Boolean call(U t1, U t2) {
-        return (t1 == t2 || (t1 != null && t1.equals(t2)));
+        return (t1 == t2 || (t2 != null && t2.equals(t1)));
```
artem-zinnatullin, line 65: Please add a test case that fails without fix and passed with it

artem-zinnatullin:
Omg, I'm from phone at the moment, but it looks very critical, right? 

As far as I see it won't compare old value with new if old is `null`

artem-zinnatullin:
ok, I'm with computer now and I don't see what's the problem with previous code. Can you please explain?

src/main/java/rx/internal/operators/OperatorDistinctUntilChanged.java
godardt, line 65: An example:
Object `t1` and `t2` are of the same type and contains an object `x1`.
Object `x1` is `null` in `t1` and is not null in `t2`
Object `t1` and `t2` override `equals` which in turn calls the `equals` of `x1`.
A `NullpointerException` is thrown from `t1` but not from `t2`.

This is a basic example but what I want to say is that applying `equals` on `t1` does not means that it will result the same the other way around.
That's the type of behavior changes which should be fixed on the user side of the rxjava lib but which can create confusion and can be avoided.

akarnokd, line 65: The current comparison is completely legal, same as Java 7's Objects.equals(). Why did you swap the variables? Implementation of equals should be symmetric: a.equals(b) == b.equals(a).

JakeWharton, line 65: That's a broken `equals` implementation and not this library's problem.

akarnokd:
@artem-zinnatullin My guess is that he has a bug in his `equals` implementation and wants to change the library for a workaround.

godardt:
@akarnokd Yes it's the case, it permitted me to pinpoint a bug on my side and I fixed it. But I though that if we could avoid people this behavior change it would be great

JakeWharton:
After this change, all you need to do is reverse `t1` and `t2` to get the same failure again.

PR #4278, opened 2016-08-03
2.x: Functional interfaces now throw

akarnokd:
It had a bit of a cascading effect.


artem-zinnatullin:
@akarnokd jfyi I'm reviewing this, but it takes a lot of time, will try to finish today

JakeWharton:
Holy long PR batman! :+1:

artem-zinnatullin:
Yeah, I've finished like 60%, looks good yet

akarnokd:
Merged it so Jake & I can progress on other things; just shout out if you found something wrong.

dmitry-weirdo:
Was it really a thought-out change?

Now we have to rewrite each lambda-call to new classes from ``io.reactivex.functions`` with hacks like described in https://www.baeldung.com/java-lambda-exceptions.
Allowing a checked exceptions in functional interfaces is a well-known Java antipattern. Why have you implemented such a drastic change that will prevent people from migrating to RxJava 2?

akarnokd:
@dmitry-weirdo Many have successfully migrated to RxJava 2. Since we support Java 6 which has barely any standard functional interfaces, it doesn't really matter what we introduce, especially if one uses lambda expressions. The throws clause gives some convenience when interacting with exception throwing traditional methods.

PR #4279, opened 2016-08-03
1.x: update gradle to v2.14.1

Mauin:
Was released a while ago and since it's an "important update to v2.14", here it is.

Release notes: https://docs.gradle.org/2.14.1/release-notes


akarnokd:
We had no build issues as far as I know but the known issues on that link worries me.

artem-zinnatullin:
Well, we can try to upgrade, but for example for our Android app written in Kotlin neither 2.14 nor 2.14.1 build correctly in non-clean mode.

Mauin:
If you feel like 2.14 is okay for now since no issues happened feel free to close this one. 
I just saw that an "old" Gradle version was used in the build, but we can surely wait for one of the next Gradle releases.

akarnokd:
Thanks for the contribution but we'll wait out this one.

PR #4285, opened 2016-08-06
2.x: Remove defunct 'Nbp' prefix from Observable interfaces.

JakeWharton:


JakeWharton:
All other 'Nbp' references are in tests which I'll clean up next.

JakeWharton:
Also worth noting that these names are the same now as those in `Flowable`, but you're not only unlikely to use both in the same file, the disambiguation of the two is trivial with an `Observable.` or `Flowable.` qualifier.

akarnokd:
:+1:

vanniktech:
What about `NbpFullArbiter` & `NbpQueueDrain`? They are not in the test scope.

JakeWharton:
Yeah I saw those after commenting.

PR #4287, opened 2016-08-06
2.x: Elide ObservableWrapper for empty and never.

JakeWharton:


JakeWharton:
Not sure whether the use of `create` is intentional or not internally. As of right now it's just a trampoline/indirection to `subscribeActual` so if extending `Observable` is fine, I can go through and eliminate all of these.

akarnokd:
:+1:

Issue #4280, opened 2016-08-04
2.x: Exceptions is in internal package


JakeWharton:
...despite its tests being in the public API package.

It's hard to write libraries without `Exceptions.throwIfFatal`. Should this be in the public API?

akarnokd:
Possibly, what should happen with the rest of the methods?

If you'd review #4278 that would allow the refactoring to happen without interference.

JakeWharton:
I would think they'd split, although having two classes of the same name would obviously be less than ideal. Maybe it could be hoisted onto RxJavaPlugins since its often used with calling `RxJavaPlugins.onError`? There could even be a combination method that does both, since there's a few hundred uses of the pair throughout RxJava's codebase as-is.

JakeWharton:
This can be closed! 🎉

JakeWharton:
Oh wait it's my issue, I can do that!

PR #4288, opened 2016-08-06
2.x: move Exceptions to public area, + ExceptionHelper

akarnokd:
Related: #4280.

The `Exceptions.propagate` and `Exceptions.throwIfFatal` are too useful to be hidden in an internal package so this PR moves them into the public `io.reactivex.exceptions` package. The terminal atomics related to Throwable containers lives in the `ExceptionHelper` enum.


JakeWharton:
lgtm :+1:

PR #4289, opened 2016-08-06
2.x: Elide more create/callback wrappers in favor of direct Observables.

JakeWharton:


src/main/java/io/reactivex/Observable.java
```diff
@@ -450,35 +450,35 @@ public Throwable call() {
 
     public static <T> Observable<T> fromPublisher(final Publisher<? extends T> publisher) {
         Objects.requireNonNull(publisher, "publisher is null");
-        return create(new ObservableConsumable<T>() {
+        return new Observable<T>() {
```
akarnokd, line 453: There should be another run to change such anonymous classes to proper internal classes.

akarnokd:
:+1:

src/main/java/io/reactivex/Observable.java
JakeWharton, line 453: I'll add it to my queue.

akarnokd:
:+1:

PR #4290, opened 2016-08-06
2.x: Eliminate an extra allocation and indirection in periodic scheduling.

JakeWharton:


src/main/java/io/reactivex/Scheduler.java
```diff
@@ -106,13 +106,10 @@ public Disposable schedule(Runnable run) {
...
         public Disposable schedulePeriodically(Runnable run, final long initialDelay, final long period, final TimeUnit unit) {
-            final SerialDisposable first = new SerialDisposable();
-
-            final SerialDisposable sd = new SerialDisposable(first);
-            
             final Runnable decoratedRun = RxJavaPlugins.onSchedule(run);
-            
-            first.replace(schedule(new Runnable() {
+
+            final SerialDisposable sd = new SerialDisposable();
```
akarnokd, line 111: The two SerialDisposables was intentional. When the very first is scheduled, the first invocation may set a newer disposable that is then overwritten by the sd.replace of the first itself.

akarnokd, line 111: You can combine the two into a single class but it will cost 2 AtomicReferences either way:

``` java
class TwoDisposable extends AtomicReference<Disposable> implements Disposable {
   final AtomicReference<Disposable> first;

   @Override
   public void replaceFirst(Disposable d) {
       DisposableHelper.replace(first, d);
   }

   public void replaceSecond(Disposable d) {
       DisposableHelper.replace(this, d);
   }

   @Override
   public void dispose() {
       DisposableHelper.dispose(this); 
       DisposableHelper.dispose(second);
   }
}
```

JakeWharton, line 111: Interesting. All the tests passed so I assumed it was safe. Can you show me
in a test how what that would look like? I don't quite follow how sd is
exposed such that it might be changed.

akarnokd, line 111: I can sketch one: schedule periodically like this: `schedule(task, 0, 1, TimeUnit.DAYS).dispose()` and loop it. There is a non-zero probability the outer replaces the Disposable of the second delayed task with the completed task of the first (immediate) and the task keeps executing until the backing threadpool is shut down.

JakeWharton, line 111: But `sd` isn't returned for the calling code to even call `dispose()` on before we replace its value with the `Disposable` return from delegating to `schedule`.

akarnokd, line 111:
```
Thread 1                                   Thread 2
--------                                   --------
d0 = schedule(task, 0, u);                
                                           task.run();
                                           d = schedule(task, delay, u);
                                           sd.replace(d);
sd.replace(d0);

sd.dispose();  <--- d is not disposed and will run the task after the delay
```

JakeWharton, line 111: Ah, got it. I'll try to write a test.

PR #4291, opened 2016-08-06
2.x: Promote anonymous Observables to top-level types.

JakeWharton:


src/main/java/io/reactivex/internal/operators/observable/ObservableFromPublisher.java
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            @Override
+            public void onNext(T t) {
+                o.onNext(t);
+            }
+
+            @Override
+            public void onSubscribe(Subscription inner) {
+                o.onSubscribe(Disposables.from(inner));
```
akarnokd: If you make the outer a proper class, it can implement `Disposable` and `dispose()` can call `cancel()` on the field `inner`.

src/main/java/io/reactivex/internal/operators/observable/ObservableFromPublisher.java
```diff
@@ -28,29 +30,47 @@ public ObservableFromPublisher(Publisher<? extends T> publisher) {
...
+        }
+
+        @Override
+        public void onNext(T t) {
+            o.onNext(t);
+        }
+
+        @Override
+        public void onSubscribe(Subscription inner) {
+            SubscriptionHelper.setOnce(this, inner);
```
akarnokd: You can set it on a plain field, no need for atomics:

``` java
     this.inner = inner;
     o.onSubscribe(this);
     inner.request(Long.MAX_VALUE);
```

src/main/java/io/reactivex/internal/operators/observable/ObservableFromPublisher.java
```diff
@@ -28,29 +30,47 @@ public ObservableFromPublisher(Publisher<? extends T> publisher) {
...
+            o.onSubscribe(this);
+            inner.request(Long.MAX_VALUE);
+        }
+
+        @Override public void dispose() {
+            SubscriptionHelper.dispose(this);
```
akarnokd: switch to `AtomicBoolean` and:

``` java
if (compareAndSet(false, true)) {
    inner.cancel();
}
```

src/main/java/io/reactivex/internal/operators/observable/ObservableFromPublisher.java
```diff
@@ -0,0 +1,79 @@
...
+
+        @Override public void dispose() {
+            if (compareAndSet(false, true)) {
+                inner.cancel();
+                inner = null;
+            }
+        }
+
+        @Override public boolean isDisposed() {
+            return inner == null;
```
akarnokd: `get()`. (Btw since `inner` is plain, this check may end up hoisted and you'd never see it turning null).

src/main/java/io/reactivex/internal/operators/observable/ObservableRange.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    protected void subscribeActual(Observer<? super Integer> o) {
+        Disposable d = Disposables.empty();
+        o.onSubscribe(d);
+
+        long end = start - 1L + count;
+        for (long i = start; i <= end && !d.isDisposed(); i++) {
+            o.onNext((int)i);
```
vanniktech, line 36: nit: spacing between variable and cast?

JakeWharton, line 36: This was just moved as-is from the anonymous version. There's tons of spaceless casts all over the project so I'm not sure it's worth the effort to standardize.

akarnokd, line 36: I don't use spaces in this case either but IDEs sometimes put that in. Leave it as it is.

JakeWharton:
I think this is ready. I have another one following it to normalize the implementations of never+empty for all types (they vary slightly currently).

akarnokd:
This needs rebasing.

akarnokd:
:+1:

PR #4292, opened 2016-08-07
2.x: Actually apply custom scheduler when applying future.

JakeWharton:
Correct annotatoin declaration as well.


JakeWharton:
Actually I should write a test for this.

PR #4295, opened 2016-08-07
2.x: Add serialized single subscriber wrapper.

JakeWharton:
There's already one for observer, subscriber, and completable subscriber.


JakeWharton:
This is copy/paste from the completable one which is untested at this point. Tests for both can be added later, I just didn't want to forget about it.

akarnokd:
Single and Completable are so simple to serialize that operators do it themselves. For the sake of symmetry, let's have it.

PR #4298, opened 2016-08-07
2.x: Clarify behavior of RxJavaPlugins.reset() in javadoc

artem-zinnatullin:
Noticed in RxAndroid https://github.com/ReactiveX/RxAndroid/pull/316#discussion_r73783555

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -231,7 +231,7 @@ public final class RxJavaPlugins {
     }
 
     /**
-     * Removes all handlers and resets the default behavior.
+     * Removes all handlers and resets to default behavior.
      */
     public static void reset() {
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -231,7 +231,7 @@ public static Scheduler onSingleScheduler(Scheduler defaultScheduler) {
     }
 
     /**
-     * Removes all handlers and resets the default behavior.
+     * Removes all handlers and resets to default behavior.
```
vanniktech, line 234: how about `resets to the default behavior.`

artem-zinnatullin, line 234: Again, not an expert in English, but "the" could be used here if there
would be something concrete to refer to. Or I'm wrong?

Issue #4044, opened 2016-06-20
2.x: New base interface types


akarnokd:
As part of the 2.x design, new base interface types have been recently merged & applied. The aim is to match the style of the Reactive-Streams base types + their fluent base class:

`Flowable` -> `Publisher` -> `Subscriber`
`Observable` -> `ObservableConsumable` -> `Observer`
`Single` -> `SingleConsumable` -> `SingleSubscriber`
`Completable` -> `CompletableConsumable` -> `CompletableSubscriber`

They feel odd to me too so any idea is welcome.

JakeWharton:
Just to be clear before I start thinking about this more, `Publisher` and `Subscriber` are the only RS types here, right? And `Single` and `Completable` are always backpressure-aware types?

akarnokd:
> Publisher and Subscriber are the only RS types 

Yes.

> Single and Completable are always backpressure-aware types?

The other base types (including Observable) don't have a notion for backpressure.

artem-zinnatullin:
Can we please keep names for `Observable`, `Single` and `Completable` since they're not part of [RS](https://github.com/reactive-streams/reactive-streams-jvm)? They're short and well known. 

I'd just rename `Observer` to `ObservableSubscriber` to have same pattern as `SingleSubscriber` and `CompletableSubscriber` and not interfere with `Publisher -> Subscriber`.

JakeWharton:
I wouldn't mind renaming `Observable` since it's a description of the pattern to which all these types conform. It's the least descriptive of the bunch, and you wouldn't chose it today given the others.

I used to like `Flowable`, and I still do, but now that we have a non-BP vs. BP distinction perhaps we should consider making BP-aware variants of all 3 types and using `Flowable` as an adjective instead of a noun? The problem is that gets verbose.

akarnokd:
`Single` and `Completable` don't need backpressure because they emit 1 or 0 items respectively.

The non-backpressured `Observable` was requested to allow pure GUI event handling where the sources are hot and can't be backpressured.

JakeWharton:
Hmm what about the `request(0)` case though? Where work isn't triggered upon subscription but deferred until explicitly requested.

JakeWharton:
Sorry I mean the case where an operation cannot respect `request(0)`. I'm struggling to come up with an example, but I know I've ran into potential cases before but I never wrote them down (because we never had `Flowable` either).

akarnokd:
[...]

PR #4299, opened 2016-08-07
2.x: Rename Flowable Base Interface Types for consistency

vanniktech:
Addresses #4044

### DIFF
--- a/src/main/java/io/reactivex/FlowableOperator.java
+++ b/src/main/java/io/reactivex/FlowableOperator.java
@@ -1,0 +1,28 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


src/main/java/io/reactivex/FlowableOperator.java
```diff
@@ -0,0 +1,15 @@
+package io.reactivex;
```
JakeWharton, line 14: license header

src/main/java/io/reactivex/FlowableTransformer.java
```diff
@@ -0,0 +1,15 @@
+package io.reactivex;
```
JakeWharton, line 14: license header

JakeWharton:
lgtm :+1:

src/main/java/io/reactivex/FlowableOperator.java
vanniktech, line 14: yup just noticed it too when I pushed it. added it

akarnokd:
:+1:

PR #4031, opened 2016-06-19
2.x: Update groupID and readme.md

akarnokd:
This PR updates the readme and changes the group ID.

I'm going to merge this and hope the new group id shows up at https://oss.jfrog.org/artifactory/libs-snapshot/io/reactivex/ (may take several attempts, depending on how cooperative the nebula plugin is.)

PR #4300, opened 2016-08-07
2.x: Rename Observable Base Interface Types for consistency

vanniktech:
Addresses #4044.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableBufferExactBoundary.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableBufferExactBoundary.java
@@ -26,11 +26,11 @@ import io.reactivex.observers.SerializedObserver;
 
 public final class ObservableBufferExactBoundary<T, U extends Collection<? super T>, B> 
 extends Observable<U> {
-    final ObservableConsumable<T> source;
-    final ObservableConsumable<B> boundary;
+    final ObservableSource<T> source;
+    final ObservableSource<B> boundary;
     final Callable<U> bufferSupplier;
     
-    public ObservableBufferExactBoundary(ObservableConsumable<T> source, ObservableConsumable<B> boundary, Callable<U> bufferSupplier) {
+    public ObservableBufferExactBoundary(ObservableSource<T> source, ObservableSource<B> boundary, Callable<U> bufferSupplier) {
         this.source = source;
         this.boundary = boundary;
         this.bufferSupplier = bufferSupplier;
@@ -45,7 +45,7 @@ extends Observable<U> {
     extends QueueDrainObserver<T, U, U> implements Observer<T>, Disposable {
         /** */
         final Callable<U> bufferSupplier;
-        final ObservableConsumable<B> boundary;
+        final ObservableSource<B> boundary;
         
         Disposable s;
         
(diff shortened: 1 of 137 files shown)


src/main/java/io/reactivex/internal/operators/observable/ObservableBufferExactBoundary.java
```diff
@@ -17,6 +17,7 @@
 import java.util.concurrent.Callable;
 
 import io.reactivex.*;
+import io.reactivex.ObservableSource;
```
JakeWharton: There's superfluous imports like this throughout this PR and your other one, but I don't think it's a big deal.

akarnokd:
There are a large amount of compilation errors.

src/main/java/io/reactivex/internal/operators/observable/ObservableDematerialize.java
```diff
@@ -18,9 +18,9 @@
 import io.reactivex.internal.disposables.DisposableHelper;
 import io.reactivex.plugins.RxJavaPlugins;
 
-public final class ObservableDematerialize<T> extends ObservableSource<Try<Optional<T>>, T> {
+public final class ObservableDematerialize<T> extends ObservableSource {
```
JakeWharton: Looks like there already was an `ObservableSource` class that we're now clashing with. I'm tempted to just inline it since all it does is hold a value in a field. As an actual type in the hierarchy it's not used in any special way. Either that or just rename it `ObservableFromUpstream` for now and delete it later.

JakeWharton: It's probably worth doing these as separate PRs. Rename/delete the existing `ObservableSource` and then do this rename. Otherwise it's likely to be extra confusing because of the name conflict.

akarnokd: The original `ObservableSource` has its use by mandating a constructor with an upstream source and may later support some graph discovery by providing a standard way of getting the upstream.

JakeWharton: Makes sense. Does `ObservableFromUpstream` or `ObservableWithUpstream` for that sound good?

akarnokd: `ObservableWithUpstream` is okay.

src/main/java/io/reactivex/internal/operators/observable/ObservableBufferExactBoundary.java
vanniktech: Yup IntelliJ is not smart here and automatically inserts the import. Could also be because I've configured it to avoid \* imports ...

Once I'm done with all of the renaming I could fix all of the imports if this is wanted.

src/main/java/io/reactivex/internal/operators/observable/ObservableDematerialize.java
vanniktech: Created #4031 to do this.

vanniktech:
Rebased against 2.x and re-did the last changes.

JakeWharton:
looks good this time :+1:

akarnokd:
:+1:

PR #4307, opened 2016-08-07
2.x: Promote lift implementations to top-level types.

JakeWharton:


src/main/java/io/reactivex/internal/operators/flowable/FlowableDoOnLifecycle.java
```diff
@@ -0,0 +1,39 @@
...
+package io.reactivex.internal.operators.flowable;
+
+import io.reactivex.Flowable;
+import io.reactivex.functions.Consumer;
+import io.reactivex.functions.LongConsumer;
+import io.reactivex.internal.subscribers.flowable.SubscriptionLambdaSubscriber;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+public final class FlowableDoOnLifecycle<T> extends FlowableSource<T, T> {
```
JakeWharton, line 22: Will rename `FlowableSource` to `FlowableWithUpstream` in a follow-up.

Issue #4306, opened 2016-08-07
2.x: Unused exceptions in public API


JakeWharton:
Currently there's a few exceptions which are unused that are in the public API:
- `OnCompleteFailedException`
- `OnErrorFailedException`
- `OnErrorNotImplementedException`
- `UnsubscribeFailedException`

There are tests which use these for various reasons, but the primary cause of lack of use is that subscribers cannot throw according to the Reactive Streams specification.

This is just a tracking issue for doing something about that or removing them before 2.0.

akarnokd:
Yes, they are there mostly to satisfy some tests that are mostly ignored by the way. Certainly we can't throw them  so they can be removed. If you remove them from tests, please keep the test method with `@Ignored` so they don't show up as missing for the test-sync I'm doing.

PR #4309, opened 2016-08-08
2.x: Remove unused exceptions in public API

vanniktech:
As wanted in #4306


JakeWharton:
:+1: if the test changes are acceptable for your setup

akarnokd:
:+1:

PR #4310, opened 2016-08-08
2.x: fix SubscriptionArbiter reentrancy/cancel bug

akarnokd:
This PR fixes the case when a call to request() prevented cancellation of the arbiter if the call never returned, thus locking out the drain loop from the cancellation call.

The PR makes sure `request()` is only called outside the guarded region.

1.x is not affected because unsubscription and requesting go on separate "channels".

The PR also fixes `FlowableRedo` by serializing the repeat signal `Processor`.


JakeWharton:
:+1:

PR #4311, opened 2016-08-08
1.x: Schedulers.from() to call RxJavaHooks.onScheduleAction

akarnokd:
The `ExecutorScheduler` wrapper did not call the `RxJavaHooks.onScheduledAction()` to decorate actions sent to a custom scheduler.


JakeWharton:
:+1:

PR #4316, opened 2016-08-08
2.x: Normalize empty and never sources.

JakeWharton:


src/main/java/io/reactivex/internal/operators/observable/ObservableEmpty.java
```diff
@@ -27,4 +28,8 @@ protected void subscribeActual(Observer<? super Object> o) {
         o.onSubscribe(EmptyDisposable.INSTANCE);
         o.onComplete();
     }
+
+    @Override public Object call() {
```
artem-zinnatullin: nit: ~Square formatting

artem-zinnatullin:
👍

Issue #4255, opened 2016-07-28
1.x: Rename Observable.fromAsync()


artem-zinnatullin:
`fromAsync()` sounds like it's ok to emit from multiple threads in parallel while in fact it doesn't serialize emission.

Can we rename it? For example to `Observable.fromCallback()` which also seems more obvious to use when you convert callback-styled code to `Observable`.

akarnokd:
The suggested name is too close to `fromCallable`. Besides, the Javadoc of `AsyncEmitter` warns about sequential use:

``` java
/**
 * Abstraction over a RxJava Subscriber that allows associating
 * a resource with it and exposes the current number of downstream
 * requested amount.
 * <p>
 * The onNext, onError and onCompleted methods should be called 
 * in a sequential manner, just like the Observer's methods. The
 * other methods are threadsafe.
 *
 * @param <T> the value type to emit
 */
```

artem-zinnatullin:
> The suggested name is too close to fromCallable.

You're worried about dynamic languages like Groovy? They're not _that_ bad!

> Besides, the Javadoc of AsyncEmitter warns about sequential use

Well, to me it looks ~like `void threadSafe()` which has note in javadoc that it's not thread safe 😁

akarnokd:
They sound to similar and may lead to speaker embarrassment when mixed up, not to mention, users getting confused and Content assist/Intellisense bringing up both of them when typing `fromCall`.

JakeWharton:
I don't think something being asynchronous immediately implies that it's also concurrent.

artem-zinnatullin:
True, but not a lot of people understand that.

Only because of `AsyncTask` a lot of Android devs think that async == bg
thread.

Another point is that `fromAsync()` is probably not what developer would
look into when he/she will need to convert callback-styled api to Rx,
`fromCallback()` looks more user-friendly (in my opinion of course).

JakeWharton:
Using an `AsyncTask` inside `fromAsync` would be perfectly fine since it's not concurrent.

`fromCallback` is clearly not usable for the reasons previously stated. I'm not against a rename, but I'm having a hard time coming up with a better name than what we already have.

JakeWharton:
I would almost rename _both_ `fromCallable` and `fromAsync` to just overloads of `from`. We already have a few overloads of `from` for adapting various sources so I don't see why these are treated any differently than just more sources.

[...]

PR #4317, opened 2016-08-08
2.x: Rename create to unsafeCreate, add "safe" create methods.

### LINKED ISSUE #4286: 2.x: Add fromAsync Observable factory
but potentially with a different name based on what the resolution of #4255 is.

JakeWharton:
This is a proposal to solve #4255 but for 2.x and closes #4286 built around one goal: make the `create` methods actually usable by developers!

Currently the advice given to most developers using 1.x is avoid `create` unless you are absolutely 100% sure you understand backpressure and cancellation, and even then avoid it if possible. This method is a trap that most new developers fall into. _The method is named "create", after all, so it must be how I create these things!_ I want to make that statement not only correct, but advisable (at least when the still-convenient "from" methods are not suitable, such as when wrapping callback-based APIs).

Because `Observable`, `Single`, and `Completable` are backpressure-free, this amounts to just suppressing downstream notifications when disposed. This doesn't seem too controversial.

The most controversial part is that I have renamed `Flowable.fromAsync` to `Flowable.create` thus making it the canonical way for users to create `Flowable`s directly. I'm much less opinionated on this change, but it felt wrong to have an `Flowable.unsafeCreate` with no associated `create` and `fromAsync` is the "safest" version we have.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/completable/CompletableFromSource.java
+++ b/src/main/java/io/reactivex/internal/operators/completable/CompletableFromSource.java
@@ -13,18 +13,69 @@
 
 package io.reactivex.internal.operators.completable;
 
-import io.reactivex.*;
+import io.reactivex.Completable;
+import io.reactivex.CompletableObserver;
+import io.reactivex.CompletableSource;
+import io.reactivex.disposables.Disposable;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public final class CompletableFromSource extends Completable {
-
-    final CompletableSource source;
+    private final CompletableSource source;
 
     public CompletableFromSource(CompletableSource source) {
         this.source = source;
     }
-    
-    @Override
-    protected void subscribeActual(CompletableObserver observer) {
-        source.subscribe(observer);
+
+    @Override protected void subscribeActual(CompletableObserver observer) {
+        source.subscribe(new DisposeAwareCompletableObserver(observer));
+    }
+
+    /**
+     * An observer which does not send downstream notifications once disposed. Used to guard against
+     * naive implementations of {@link CompletableSource} which do not check for this.
+     */
+    static final class DisposeAwareCompletableObserver
+    extends AtomicBoolean
+    implements CompletableObserver, Disposable {
+
+        private final CompletableObserver o;
+        private Disposable d;
+
+        DisposeAwareCompletableObserver(CompletableObserver o) {
+            this.o = o;
+        }
+
+        @Override
+        public void onComplete() {
+            if (!get()) {
+                o.onComplete();
+            }
+        }
+
+        @Override
+        public void onError(Throwable e) {
+            if (!get()) {
+                o.onError(e);
+            }
+        }
+
+        @Override
+        public void onSubscribe(Disposable d) {
+            this.d = d;
+            o.onSubscribe(this);
+        }
+
+        @Override
+        public void dispose() {
+            if (compareAndSet(false, true)) {
+                d.dispose();
+                d = null;
+            }
+        }
+
+        @Override
+        public boolean isDisposed() {
+            return get();
+        }
     }
 }
--- a/src/main/java/io/reactivex/internal/operators/completable/CompletableFromUnsafeSource.java
+++ b/src/main/java/io/reactivex/internal/operators/completable/CompletableFromUnsafeSource.java
@@ -1,0 +1,30 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * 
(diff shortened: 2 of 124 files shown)


src/main/java/io/reactivex/internal/operators/completable/CompletableFromSource.java
```diff
@@ -13,18 +13,62 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
 
+public final class CompletableFromSource extends CompletableFromUnsafeSource {
     public CompletableFromSource(CompletableSource source) {
-        this.source = source;
+        super(source);
+    }
+
+    @Override protected void subscribeActual(CompletableObserver observer) {
+        super.subscribeActual(new DisposeAwareCompletableObserver(observer));
```
akarnokd: Why did you add this `DisposeAwareCompletableObserver`?

src/main/java/io/reactivex/internal/operators/completable/CompletableFromUnsafeSource.java
```diff
@@ -0,0 +1,30 @@
...
+public class CompletableFromUnsafeSource extends Completable {
+
+    final CompletableSource source;
+
+    public CompletableFromUnsafeSource(CompletableSource source) {
+        this.source = source;
+    }
+    
+    @Override
+    protected void subscribeActual(CompletableObserver observer) {
```
akarnokd, line 27: I'd really would like these to remain abstract to force the subclasses to override it.

src/main/java/io/reactivex/internal/operators/observable/ObservableFromUnsafeSource.java
```diff
@@ -0,0 +1,29 @@
...
+public class ObservableFromUnsafeSource<T> extends Observable<T> {
+    final ObservableSource<T> source;
+
+    public ObservableFromUnsafeSource(ObservableSource<T> source) {
+        this.source = source;
+    }
+    
+    @Override
+    protected void subscribeActual(Observer<? super T> observer) {
+        source.subscribe(observer);
```
akarnokd, line 27: I'd really would like these to remain abstract to force the subclasses to override it.

src/main/java/io/reactivex/internal/operators/single/SingleFromSource.java
```diff
@@ -1,29 +1,72 @@
...
+
+    @Override protected void subscribeActual(SingleObserver<? super T> observer) {
+        super.subscribeActual(new DisposeAwareSingleObserver<T>(observer));
     }
-    
-    @Override
-    protected void subscribeActual(SingleObserver<? super T> observer) {
-        source.subscribe(observer);
+
+    static final class DisposeAwareSingleObserver<T>
```
akarnokd, line 36: You could combine the single and completable variants into a single class.

src/main/java/io/reactivex/internal/operators/single/SingleFromUnsafeSource.java
```diff
@@ -0,0 +1,29 @@
...
+public class SingleFromUnsafeSource<T> extends Single<T> {
+    final SingleSource<T> source;
+
+    public SingleFromUnsafeSource(SingleSource<T> source) {
+        this.source = source;
+    }
+    
+    @Override
+    protected void subscribeActual(SingleObserver<? super T> observer) {
+        source.subscribe(observer);
```
akarnokd, line 27: I'd really would like these to remain abstract to force the subclasses to override it.

src/main/java/io/reactivex/internal/operators/single/SingleFromSource.java
akarnokd, line 36: Sorry, forgot about the delegate type.

src/main/java/io/reactivex/internal/operators/observable/ObservableFromUnsafeSource.java
JakeWharton, line 27: Yeah using inheritance here was an early crutch when prototyping. It's not needed in practice so I'll remove it.

src/main/java/io/reactivex/internal/operators/completable/CompletableFromSource.java
JakeWharton: It prevents notifications after disposal for implementations which might not be checking.

src/main/java/io/reactivex/internal/operators/observable/ObservableFromUnsafeSource.java
akarnokd, line 27: Okay.

JakeWharton:
I think it's also important to point out I replaced `Consumer<AsynEmitter<T>>` with `FlowableSource<T>` for parity with other "safe" `create` method parameter types. The unsafe variant still takes a raw `Publisher<T>`, however.

artem-zinnatullin:
:heart::heart::heart:

akarnokd:
:+1:

PR #4322, opened 2016-08-09
1.x: add TestSubscriber.assertValuesAndClear

akarnokd:
This PR adds an ~~`assertAndConsume`~~ `assertValuesAndClear` method to `TestSubscriber` to assert on value content and then clear the internal value list. This helps asserting on received values without the need to list all previously seen values in `assertValues`. 

On a sidenote, `getOnNextEvents()` returns the internal list and one can call `clear()` on it.

### DIFF
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -693,7 +693,7 @@ public class TestSubscriber<T> extends Subscriber<T> {
      * @param expectedFirstValue the expected first value
      * @param expectedRestValues the optional rest values
      */
-    public final void assertAndConsume(T expectedFirstValue, T... expectedRestValues) {
+    public final void assertValuesAndClear(T expectedFirstValue, T... expectedRestValues) {
         int n = 1 + expectedRestValues.length;
         assertValueCount(n);
         assertItem(expectedFirstValue, 0);
--- a/src/main/java/rx/observers/TestSubscriber.java
+++ b/src/main/java/rx/observers/TestSubscriber.java
@@ -328,19 +328,22 @@ public class TestSubscriber<T> extends Subscriber<T> {
         }
 
         for (int i = 0; i < items.size(); i++) {
-            T expected = items.get(i);
-            T actual = values.get(i);
-            if (expected == null) {
-                // check for null equality
-                if (actual != null) {
-                    assertionError("Value at index: " + i + " expected to be [null] but was: [" + actual + "]\n");
-                }
-            } else if (!expected.equals(actual)) {
-                assertionError("Value at index: " + i 
-                        + " expected to be [" + expected + "] (" + expected.getClass().getSimpleName() 
-                        + ") but was: [" + actual + "] (" + (actual != null ? actual.getClass().getSimpleName() : "null") + ")\n");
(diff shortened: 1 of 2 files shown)


src/test/java/rx/observers/TestSubscriberTest.java
```diff
@@ -772,4 +772,47 @@ public void awaitValueCountFails() throws Exception {
...
+            Assert.fail("Should have thrown AssertionError");
+        } catch (AssertionError ex) {
+            // expected
+        }
+        
+        ts.assertValueCount(2);
+        
+        try {
+            ts.assertAndConsume(4, 5, 6);
+            Assert.fail("Should have thrown AssertionError");
```
artem-zinnatullin, line 809: nit: static import

src/main/java/rx/observers/TestSubscriber.java
```diff
@@ -670,4 +673,33 @@ final void assertionError(String message) {
...
+     * 
+     * ts.onNext(2);
+     * ts.onNext(3);
+     * 
+     * ts.assertAndConsume(2, 3); // no mention of 1
+     * </pre></code>
+     * @param expectedFirstValue the expected first value
+     * @param expectedRestValues the optional rest values
+     */
+    public final void assertAndConsume(T expectedFirstValue, T... expectedRestValues) {
```
artem-zinnatullin: method name just doesn't work for me

src/test/java/rx/observers/TestSubscriberTest.java
akarnokd, line 809: Does IntelliJ do that automatically? Eclipse doesn't so I have to go and manually type in the import; writing out `Assert` is more convenient for me. If you want, you can turn all of these into static imports in another PR.

artem-zinnatullin:
👍 but method name, I just can't treat it as good for this API

src/test/java/rx/observers/TestSubscriberTest.java
artem-zinnatullin, line 809: In IntelliJ I usually just write `fail(` and hit `ALT + ENTER` to apply static import, IntellIJ applies it for whole file then

src/main/java/rx/observers/TestSubscriber.java
akarnokd: Any other suggestions?

artem-zinnatullin: `assertAndReset()`?

akarnokd: That implies the whole internal state is wiped out and you want to reuse the same instance on another Observable.

artem-zinnatullin: What about assertValuesAndClear()?

akarnokd: Too long for my taste but okay.

artem-zinnatullin:
:+1:

artem-zinnatullin:
@JakeWharton / @zsxwing / @vanniktech can you PTAL? Would be great if we could include it in 1.1.9.

JakeWharton:
Old method name was better

PR #4325, opened 2016-08-09
2.x: Switch the connected stream types to track their upstream.

JakeWharton:


src/main/java/io/reactivex/flowables/ConnectableFlowable.java
```diff
@@ -34,8 +34,12 @@
  * @param <T>
  *          the type of items emitted by the {@code ConnectableObservable}
  */
-public abstract class ConnectableFlowable<T> extends Flowable<T> {
-    
+public abstract class ConnectableFlowable<T> extends FlowableWithUpstream<T, T> {
+
+    protected ConnectableFlowable(Publisher<T> source) {
```
akarnokd, line 39: This mandates `ConnectableFlowable`s to have a source but it is possible that one writes a connectable that emits on its own. Please introduce an intermediate type and leave the Connectables as they are.

akarnokd:
Also there is a compilation error with this.

JakeWharton:
Oops yeah I forgot to push the latest. But I'll fix the hierarchy in a bit.

src/main/java/io/reactivex/flowables/ConnectableFlowable.java
JakeWharton, line 39: Hmm these are both classes though so there'd be no way to use both in the implementations.

JakeWharton, line 39: Unless we make `FlowableFromSource` an interface exposing `source()` and then switch the current class to `AbstractFlowableFromSource`

akarnokd, line 39: It will be an interface.

JakeWharton, line 39: Ok I'm going to send a new PR with that change for existing operators, and then circle back for these connectable folk.

PR #4328, opened 2016-08-10
1.x: compact MultipleAssignment- and Serial-Subscriptions

akarnokd:
This PR reduces the allocation in `MultipleAssignmentSubscription` and `SerialSubscription` by introducing a shared, compact underlying container `SequentialSubscription`.

I've updated `Worker.schedulePeriodically` to use it directly.

In addition, there is a behavior change with `MultipleAssignmentSubscription`: it no longer retains the last `Subscription` as it was likely to cause retention problems (and otherwise didn't make sense to me).

The `SequentialSubscription` feature weak versions of the mutation methods that don't retry if there was a concurrent mutation: for some operations, not winning such races is fine.

/cc @JakeWharton @davidmoten @artem-zinnatullin

### DIFF
--- a/src/main/java/rx/internal/subscriptions/SequentialSubscription.java
+++ b/src/main/java/rx/internal/subscriptions/SequentialSubscription.java
@@ -61,7 +61,7 @@ public final class SequentialSubscription extends AtomicReference<Subscription> 
     
     /**
      * Atomically sets the contained Subscription to the provided next value and unsubscribes
-     * the previous value.
+     * the previous value or unsubscribes the next value if this container is unsubscribed.
      * <p>(Remark: named as such because set() is final).
      * @param next the next Subscription to contain, may be null
      * @return true if the update succeded, false if the container was unsubscribed
@@ -88,7 +88,8 @@ public final class SequentialSubscription extends AtomicReference<Subscription> 
 
     /**
      * Atomically replaces the contained Subscription to the provided next value but
-     * does not unsubscribe the previous value.
+     * does not unsubscribe the previous value or unsubscribes the next value if this 
+     * container is unsubscribed.
      * @param next the next Subscription to contain, may be null
      * @return true if the update succeded, false if the container was unsubscribed
      */
@@ -111,7 +112,7 @@ public final class SequentialSubscription extends AtomicReference<Subscription> 
     
     /**
      * Atomically tries to set the contained Subscription to the provided next value and unsubscribes
-     * the previous value; but in case of a concurrent update there is no retry.
+     * the previous value or unsubscribes the next value if this container is unsubscribed.
      * <p>
      * Unlike {@link #update(Subscription)}, this doesn't retry if the replace failed
      * because a concurrent operation changed the underlying contained object.
@@ -140,7 +141,8 @@ public final class SequentialSubscription extends AtomicReference<Subscription> 
     
     /**
      * Atomically tries to replace the contained Subscription to the provided next value but
-     * does not unsubscribe the previous value.
+     * does not unsubscribe the previous value or unsubscribes the next value if this container 
+     * is unsubscribed.
      * <p>
      * Unlike {@link #replace(Subscription)}, this doesn't retry if the replace failed
      * because a concurrent operation changed the underlying contained object.
--- a/src/main/java/rx/internal/subscriptions/SequentialSubscription.java
+++ b/src/main/java/rx/internal/subscriptions/SequentialSubscription.java
@@ -1,0 +1,187 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/main/java/rx/internal/subscriptions/SequentialSubscription.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import rx.Subscription;
+import rx.subscriptions.Subscriptions;
+
+/**
+ * A container of a Subscription that supports operations of SerialSubscription 
+ * and MultipleAssignmentSubscription via methods (update, replace) and extends
+ * AtomicReference to reduce allocation count (beware the API leak of AtomicReference!).
+ * @since 1.1.9
```
artem-zinnatullin, line 27: nit: do we really need this for internal classes?

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
```diff
@@ -0,0 +1,187 @@
...
+    
+    /**
+     * Returns the current contained Subscription (may be null).
+     * <p>(Remark: named as such because get() is final).
+     * @return the current contained Subscription (may be null)
+     */
+    public Subscription current() {
+        Subscription current = super.get();
+        if (current == Unsubscribed.INSTANCE) {
+            return Subscriptions.unsubscribed();
```
artem-zinnatullin, line 57: Why not make `Subscriptions.unsubscribed()` return `Unsubscribed.INSTANCE`?

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        return current;
+    }
+    
+    /**
+     * Atomically sets the contained Subscription to the provided next value and unsubscribes
+     * the previous value.
```
artem-zinnatullin: +Unsubscribes new value if previous was unsubscribed.

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
artem-zinnatullin, line 57: Or you want them to be different objects?

JakeWharton, line 57: `Unsubscribed.INSTANCE` is an internal-only marker instance. If it were to leak into the public API it could be set on one of these subscriptions which would then think it was already unsubscribed (even though it wasn't).

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
```diff
@@ -0,0 +1,187 @@
...
+    public SequentialSubscription() {
+        
+    }
+    
+    /**
+     * Create a SequentialSubscription with the given initial Subscription.
+     * @param initial the initial Subscription, may be null
+     */
+    public SequentialSubscription(Subscription initial) {
+        lazySet(initial);
```
JakeWharton, line 46: Any reason to prefer this over `super(initial)`?

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
```diff
@@ -0,0 +1,187 @@
...
+                    current.unsubscribe();
+                }
+                return true;
+            }
+        }
+    }
+
+    /**
+     * Atomically replaces the contained Subscription to the provided next value but
+     * does not unsubscribe the previous value.
```
artem-zinnatullin: +Unsubscribes new value if previous was unsubscribed.

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
artem-zinnatullin, line 57: Would be great if its name somehow show that it's internal only so we could pay more attention to leaking it during code reviews

JakeWharton:
:+1:

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
akarnokd, line 57: The package name and javadoc indicates it.

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
akarnokd, line 46: The super constructor uses a full barrier volatile write.

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
akarnokd: Updated.

src/main/java/rx/internal/subscriptions/SequentialSubscription.java
artem-zinnatullin, line 57: ok, though they're ~invisible during code review

artem-zinnatullin:
👍

Issue #4082, opened 2016-06-23
2.x: Subscribers, Observables classes leak internal types into the public API


JakeWharton:
All of the methods doing this are at present unused. If they're for public consumption we need `DisposableSubscriber` and `NbpDisposableSubscriber` (or equivalent) to be in a public package.

akarnokd:
Yes, we should avoid these.

akarnokd:
I'd like to resolve this; could you point me to the problematic classes and what exactly is leaking?

JakeWharton:
It's the `createDisposable` methods in `Subscribers` and `Observers` which return `DisposableSubscriber` and `DisposableObserver`.

akarnokd:
See #4337.

akarnokd:
I think this has been resolved. Let me know if you find something else (or post a fix PR).

PR #4337, opened 2016-08-10
2.x: move DisposableObserver to public area, add some javadocs

akarnokd:
Move `DisposableObserver` to `io.reactivex.observers` + some documentation.

Related: #4082.


JakeWharton:
:+1:

Having these in the public API make it seem like `AsyncObserver` becomes obsolete in the case where it's not managing any additional resources. Should we rename it to `ResourceObserver` (or something) and have it always manage resources?

akarnokd:
Sure, will you do it?

JakeWharton:
Yep!

PR #4338, opened 2016-08-10
2.x: Remove conditional resource management from async listeners.

JakeWharton:
Rename them to have a 'Resource' prefix.

### DIFF
--- a/src/main/java/io/reactivex/observers/Observers.java
+++ b/src/main/java/io/reactivex/observers/Observers.java
@@ -29,8 +29,8 @@ public final class Observers {
         throw new IllegalStateException("No instances!");
     }
     
-    public static <T> AsyncObserver<T> emptyAsync() {
-        return new AsyncObserver<T>() {
+    public static <T> ResourceObserver<T> emptyResource() {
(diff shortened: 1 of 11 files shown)


JakeWharton:
The value of and distinction between the 'Resource' listeners, 'Default' listeners, and 'Disposable' listeners seems very low. Do we need all these individual types?

akarnokd:
I believe the minimum we require is the ability to dispose a `Subscriber` because that's expected the most. I don't think associating resources with or allowing async requesting is that common.

akarnokd:
I leave it to you to remove the unnecessary; we can add them back later or better yet, they can live in a separate library.

JakeWharton:
Yeah. I need to write something non-trivial with 2.x first. Maybe I'll try
this weekend. I definitely think we should be aggressive in minimizing API
surface since it's easy to add but impossible to take away after release.

akarnokd:
I have a 2.x "dumping ground" for leftovers: https://github.com/akarnokd/RxJava2Extensions so we don't lose things.

PR #4343, opened 2016-08-11
Remove tabs indent and redundant space.

hotchemi:
- Remove tabs indent and redundant space.
  - Currently tabs and spaces are mixed up in the code so just unify them.


akarnokd:
Thanks!

PR #4351, opened 2016-08-15
Use Throwable to catch Error when get Android version

tankery:
When do JUnit testing with RxRingBuffer, we may have Android platform
classes defined, but contains no Android environment.

So when we call PlatformDependent.resolveAndroidApiVersion in JUnit
testing, an UnsatisfiedLinkError will throw when calling
SystemProperties.native_get(String, String).

The UnsatisfiedLinkError is not an Exception, so the test failed because
of an uncatched exception.

Fix this by catch the Throwable instead of Exception to catch Error.

### DIFF
--- a/src/main/java/rx/internal/util/PlatformDependent.java
+++ b/src/main/java/rx/internal/util/PlatformDependent.java
@@ -70,7 +70,7 @@ public final class PlatformDependent {
                     .forName("android.os.Build$VERSION", true, getSystemClassLoader())
                     .getField("SDK_INT")
                     .get(null);
-        } catch (Exception e) { // NOPMD 
+        } catch (Throwable e) { // NOPMD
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
Hm, that's strange. What is your testing configuration? I haven't seen such problem with and without Robolectric.

tankery:
Finally I found out the root cause of my issue.
I'm using RxJava on our System App, which have using some hidden API in framework. So we 'provided' a dependency for the framework jar file to let build pass.

But this breaks the mock of Android Studio when doing JUnit test. Android Studio no longer mock the Build.VERSION.SDK_INT, and calling the real method which will drive to a native method. So the UnsatisfiedLinkError has been throw.

Sorry for the mistake. :)

PR #4354, opened 2016-08-16
2.x: Expose values and errors as unmodifiable snapshots not shared mutable lists.

JakeWharton:


akarnokd:
Why?

JakeWharton:
Expose a list which is not only mutable but whose contents may change while you're iterating and/or asserting its contents seems like a recipe for flaky tests.

akarnokd:
If you assert without first making sure the source terminated is also likely to cause test failures. There is the `getValueCount` in 1.x which gives a volatile view of the available number of onNext values that can be safely read from `values()`.

JakeWharton:
I don't see any reason that one has to wait until a terminal event. A lot
of my tests do not, especially in libraries where behavior around requests
is important.

Exposing a shared, mutable list is rarely a good idea. Since we expect no
one to be mutating these lists and we expect the values to not be changing
underneath them, this change has no functional impact. It merely enforces
that these invariants are always held true.

akarnokd:
I see that you do a snapshot but then why prevent any change to it? Besides, I have tests that have thousands elements which are then [checked for a particular pattern](https://github.com/ReactiveX/RxJava/blob/2.x/src/test/java/io/reactivex/internal/operators/flowable/FlowableFlatMapTest.java#L473) by iterating over the shared list (or just sloppily do a [size() on it](https://github.com/ReactiveX/RxJava/blob/2.x/src/test/java/io/reactivex/internal/operators/flowable/FlowableMergeTest.java#L946)).

JakeWharton:
I had accidentally mutated this shared list in one of my tests because of
similar variable names which is what led me to expose it as unmodifiable. I
was surprised that it wasn't already an unmodifiable snapshot, but the
actual live mutable list. Even with the snapshot, modifying the list of
values seems like an indicator that you are testing poorly (or flat out
incorrectly).

akarnokd:
I'm willing to accept such change if you introduce it as a separate method (or rename-refactor the current `values()` to something else, like `liveValues()` or whatever).

JakeWharton:
What's the use case for exposing the mutable live list? Both of your
previously linked examples work fine with the snapshot model.

akarnokd:
Copying may be expensive -> more GC during tests -> more likely CI will be shot down.

JakeWharton:
I'd rather leave it as a mutable landmine than burden callers with two
methods.

zsxwing:
👍 to not expose a mutable list in a public API. Accidentally mutating this shared list will cause some weird behavior and it's pretty hard to figure out.

@JakeWharton why do you need to copy the internal list? Shouldn't `unmodifiableList` is enough?

zsxwing:
For clarification, I'm not saying that changing them for 1.x. It may break the user's existing tests. But I'm totally 👍 for 2.x.

JakeWharton:
Unmodifiable prevents you from modifying it, but the list might still
change as the result of the observable emitting additional values. The copy
and wrap ensures the values reflect those at the time of the method call
and also prevent any future changes.

zsxwing:
@JakeWharton I see. In most case, I think the user just checks the list at once. So I think changing it to an immutable list should cover most of cases and it's still a good idea.

akarnokd:
You can extend `TestSubscriber` locally or have an utility method that does what you need:

``` java
public static <T> List<T> valuesOf(TestSubscriber<T> ts) {
    return Collections.unmodifiableList(new ArrayList<>(ts.values()));
}
```

I guess IntelliJ will gladly do a static import of the method above.

JakeWharton:
The point is to have the sane behavior as the default so that everyone
benefits. I'm not convinced the GC pressure is a real problem.

artem-zinnatullin:
Aaaaand we just hit it.

Problem:

``` java
ts.assertValues("a", "b")
```

Multithreaded test without proper `await` may result in such message:

```
Number of items does not match. Provided: 2 Actual: 2  (!!!!)
Provided values: a, b
Actual values: a, b (!!!!!)
```

Since we operate over mutable list instead of immutable snapshots we have this very interesting exception message that totally confuses user.

So I'm joining Jake's side here and vote for immutable snapshots.

PR #4357, opened 2016-08-16
2.x: switch to throwing Action, switchMapDelayError, other cleanup

akarnokd:
Notable changes:
- Replace the use of `Runnable` in the base reactive types to `Action` whose `run` method can throw a checked exception.
- Fix javadoc referencing 1.x types no longer available.
- The `switchMap` operator has been enhanced to support delaying errors.
- Update `concat(Iterable)` to use `concatMapDelayError` because Iterable can't throw just anytime but on the boundary, reducing the per-element overhead.
- `FlowableEmitter` has been enhanced with a `serialize()` method that serializes calls to `onXXX` methods.
- Factored out `FlowableEmitter.Cancellable` into `io.reactivex.functions` as it will be used by the other base reactive types with their `XEmitter` implementations.
- Added `AtomicThrowable` with convenience methods that use `ExceptionHelper`'s terminal atomics with `Throwable`s.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -275,6 +275,24 @@ public abstract class Completable implements CompletableSource {
         return new CompletableError(error);
     }
     
+    
+    /**
+     * Returns a Completable instance that runs the given Action for each subscriber and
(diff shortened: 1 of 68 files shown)


JakeWharton:
:+1:

PR #4360, opened 2016-08-16
2.x: factor out inner classes from the base reactive types

akarnokd:
This PR factors out the anonymous inner classes from the base reactive types and introduces the appropriate classes and methods in `*Helper` enums.

### DIFF
--- a/src/main/java/io/reactivex/internal/functions/Functions.java
+++ b/src/main/java/io/reactivex/internal/functions/Functions.java
@@ -12,10 +12,13 @@
  */
 package io.reactivex.internal.functions;
 
-import java.util.Comparator;
-import java.util.concurrent.Callable;
+import java.util.*;
+import java.util.concurrent.*;
 
+import io.reactivex.*;
+import io.reactivex.Optional;
 import io.reactivex.functions.*;
+import io.reactivex.schedulers.Timed;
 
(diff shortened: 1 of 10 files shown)


src/main/java/io/reactivex/internal/functions/Functions.java
```diff
@@ -240,4 +243,367 @@ public int compare(Object a, Object b) {
...
+            this.value = value;
+        }
+        
+        @Override
+        public boolean test(T t) throws Exception {
+            return Objects.equals(t, value);
+        }
+    }
+    
+    public static <T> Predicate<T> equalsWith(T value) {
```
JakeWharton, line 362: Is there any reason to use these static factories? The nested classes could just be public (or promoted to top-level) and referenced directly.

JakeWharton:
:+1:

src/main/java/io/reactivex/internal/functions/Functions.java
akarnokd, line 362: To avoid public namespace cluttering.

akarnokd, line 362: Please don't refactor this yet; working on repeatWhen/retryWhen fixes.

PR #4361, opened 2016-08-17
2.x: test sync, fix retryWhen/redoWhen/publish(Function)

akarnokd:
Notable changes:
- Implement the `onBackpressureBuffer` with overflow strategy; change the strategy to plain enum.
- Additional unit tests from 1.x
- Implement `publish(Function)` properly to support latecommers.
- Split `FlowableRedo` into `FlowableRetryWhen` and `FlowableRepeatWhen`.
- Make sure `take(0)` still triggers upstream subscription that is immediately cancelled.


src/main/java/io/reactivex/Flowable.java
```diff
@@ -9070,9 +9070,12 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
      * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
      */
     @Experimental
-    public final Publisher<T> onBackpressureBuffer(long capacity, Action onOverflow, BackpressureOverflow.Strategy overflowStrategy) {
-        // TODO implement
-        throw new UnsupportedOperationException();
+    public final Publisher<T> onBackpressureBuffer(long capacity, Action onOverflow, BackpressureOverflowStrategy overflowStrategy) {
+        Objects.requireNonNull(overflowStrategy, "starategy is null");
```
hotchemi: strategy?

PR #4364, opened 2016-08-18
2.x collect - handle post terminal events

davidmoten:
Just looking at moving some recent 1.x bug fixes and tests over to 2.x.

This is my first PR in 2.x so before I proceed adding more tests from 1.x (to this PR) I want to get some feedback on my approach. In particular in the test `testCollectorFailureDoesNotResultInTwoErrorEmissions` is the `Flowable` creation method ok?

### DIFF
--- a/src/test/java/io/reactivex/flowable/FlowableCollectTest.java
+++ b/src/test/java/io/reactivex/flowable/FlowableCollectTest.java
@@ -1,0 +1,185 @@
+package io.reactivex.flowable;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.junit.Test;
+
+import io.reactivex.Flowable;
+import io.reactivex.functions.BiConsumer;
+import io.reactivex.functions.Consumer;
+import io.reactivex.plugins.RxJavaPlugins;
+
+public class FlowableCollectTest {
+    
+    @Test
+    public void testCollectToList() {
+        Flowable<List<Integer>> o = Flowable.just(1, 2, 3)
+        .collect(new Callable<List<Integer>>() {
+            @Override
+            public List<Integer> call() {
+                return new ArrayList<Integer>();
+            }
+        }, new BiConsumer<List<Integer>, Integer>() {
+            @Override
+            public void accept(List<Integer> list, Integer v) {
+                list.add(v);
+            }
+        });
+        
+        List<Integer> list =  o.blockingLast();
+
+        assertEquals(3, list.size());
+        assertEquals(1, list.get(0).intValue());
+        assertEquals(2, list.get(1).intValue());
+        assertEquals(3, list.get(2).intValue());
+        
+        // test multiple subscribe
+        List<Integer> list2 =  o.blockingLast();
+
+        assertEquals(3, list2.size());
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/flowable/FlowableCollectTest.java
```diff
@@ -0,0 +1,189 @@
...
+                throw e;
+            }
+        }, new BiConsumer<List<Integer>, Integer>() {
+
+            @Override
+            public void accept(List<Integer> list, Integer t) {
+                list.add(t);
+            }
+        }).subscribe(ts);
+        ts.assertNoValues();
```
akarnokd: I'd instead throw `TestException("Forced failure")` and then simply

``` java
Flowable.just(1).collect(...)
.test()
.assertFailureAndMessage(TestException.class, "Forced failure");
```

src/test/java/io/reactivex/flowable/FlowableCollectTest.java
```diff
@@ -0,0 +1,189 @@
...
+                @Override
+                public void accept(Throwable t) {
+                    list.add(t);
+                }
+            });
+            final RuntimeException e1 = new RuntimeException();
+            final RuntimeException e2 = new RuntimeException();
+            TestSubscriber<List<Integer>> ts = TestSubscriber.create();
+
+            new Flowable<Integer>() {
```
akarnokd: You should rather use `just().concatWith(Flowable.error())`; most of my troubles with porting 1.x tests are the custom `Observable`s that need to be rearchitected. Most of them could have been written with standard API and composition.

akarnokd:
If you'd make it more concise, that would be great. Also note that most Flowable operators have an Observable counterpart so it would be nice if such changes would contain the fix & tests for both.

src/test/java/io/reactivex/flowable/FlowableCollectTest.java
davidmoten: Great, I love the `.test` method on this!

src/test/java/io/reactivex/flowable/FlowableCollectTest.java
davidmoten: This is a special case though because operators like `collect` _may_ unsubscribe from the parent when an error occurs (due to `onNext` arrival). The upstream observable you suggest would unsubscribe happily and not emit the error which is not what I want.

davidmoten:
> If you'd make it more concise, that would be great. Also note that most Flowable operators have an Observable counterpart so it would be nice if such changes would contain the fix & tests for both.

Can do.

akarnokd:
I was trying to avoid this class till now. Will you be able to finish what you wanted till Thursday or may I do it?

davidmoten:
@akarnokd I rebased but io.reactivex.Optional disappeared and I was using it in tests. I'll add it back in internal test package.

davidmoten:
@akarnokd the work is done for `Flowable` and I have done the work for `Observable` as well but is on my laptop at home so will finish PR tonight (in about 11 hours time).

davidmoten:
@akarnokd I'm noticing that RxJava 2.x is really slow to compile (2 minutes) in Eclipse 4.5.1 and 4.6 (say after switching branches). Have you noticed that? Got any tips?

akarnokd:
Please try without `Optional`. Yes Eclipse is slow if you change `Observable` and `Flowable`, it doesn't like the cycle between the base classes and their implementations. Not sure why, maybe collecting metadata makes is much slower.

davidmoten:
I've removed `Optional`. Would you like to assess this PR as it stands without the Observable fix that I can do in another PR? You guys are moving fast (as usual) and keeping PRs open a while is a bit of an obstacle.

akarnokd:
:+1:

Issue #4365, opened 2016-08-18
2.x Design: Try+Optional vs Notification


akarnokd:
In 1.x we have the `Notification<T>` that can hold a value, error or completion.

When I started with 2.0, it was aimed at Java 8+ back then, thus I thought we could leverage Java 8's own `Optional` type plus a `Try<T>` to hold an error or an `Optional<T>`. Since now we aim at Java 6, the `Optional<T>` was manually ported back as well to retain the signatures of operators.

Uses: `materialize()`, `dematerialize()`; pacing out `onComplete()` with `retryWhen` and `repeatWhen`.

The drawback of `Try<Optional<T>>` is that for each normal value, it incurs 2 allocations whereas the 1.x notification required only 1 allocation. The completion signal in 1.x requires no allocation - 2.x requires 1 allocation (the `Try`).

Should we prefer less allocation and thus go back to (something like) `Notification<T>` or do we keep the Java 8 style?

JakeWharton:
I much prefer the more semantic Notification type which maps into the Rx
domain better than Try+Optional.

artem-zinnatullin:
:+1: for Notifications, less allocations, easier to understand.

akarnokd:
See #4370

PR #4370, opened 2016-08-18
2.x: remove Try+Optional, introduce Notification

akarnokd:
This PR removes the `Try` and `Optional` classes and (re)introduces `Notification<T>`.

There are several ways to implement it. This PR uses the abstract class + private subclasses approach that saves on instance size.

Related discussion: #4365.


src/main/java/io/reactivex/Notification.java
```diff
@@ -16,52 +16,226 @@
...
+    
+    /**
+     * Constructs an onNext notification containing the given value.
+     * @param <T> the value type
+     * @param value the value to carry around in the notification, not null
+     * @return the new Notification instance
+     * @throws NullPointerException if value is null
+     */
+    public static <T> Notification<T> createOnNext(T value) {
+        Objects.requireNonNull(value, "value is null");
```
JakeWharton, line 126: This seems unfortunate. Does the RS spec forbid null values?

akarnokd, line 126: Yes, rule §2.13 forbids the use of null values, so do we on all possible entry points.

JakeWharton:
> This PR uses the abstract class + private subclasses approach that saves on instance size.

This seems like it would prevent a lot of inlining of the simple "is"ers and "get"ers because the callsites become polymorphic instead of monomorphic.

When I check the object layout having two fields in a root object doesn't increase the size because of object padding.

```
$ cat Notification.java
final class Notification<T> {
  private static final Object COMPLETED = new Object();

  private final T value;
  private final Throwable error;

  Notification(T value, Throwable error) {
    this.value = value;
    this.error = error;
  }
}

$ javac Notification.java

$ java -cp .:/Users/jw/Downloads/jol-cli-0.5-full.jar org.openjdk.jol.Main internals Notification
objc[95020]: Class JavaLaunchHelper is implemented in both /Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home/bin/java and /Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home/jre/lib/libinstrument.dylib. One of the two will be used. Which one is undefined.
# WARNING: Unable to attach Serviceability Agent. You can try again with escalated privileges. Two options: a) use -Djol.tryWithSudo=true to try with sudo; b) echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
# Running 64-bit HotSpot VM.
# Using compressed oop with 3-bit shift.
# Using compressed klass with 3-bit shift.
# WARNING | Compressed references base/shifts are guessed by the experiment!
# WARNING | Therefore, computed addresses are just guesses, and ARE NOT RELIABLE.
# WARNING | Make sure to attach Serviceability Agent to get the reliable addresses.
# Objects are 8 bytes aligned.
# Field sizes by type: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]
# Array element sizes: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]

VM fails to invoke the default constructor, falling back to class-only introspection.

Notification object internals:
 OFFSET  SIZE      TYPE DESCRIPTION                    VALUE
      0    12           (object header)                N/A
     12     4    Object Notification.value             N/A
     16     4 Throwable Notification.error             N/A
     20     4           (loss due to the next object alignment)
Instance size: 24 bytes
Space losses: 0 bytes internal + 4 bytes external = 4 bytes total
```

src/main/java/io/reactivex/Notification.java
JakeWharton, line 126: Yikes! Good to know. That will affect some libraries I work on.

JakeWharton:
Oops I wrote my comparison wrong (and left it out of the comment).

You're right, the individual classes are only 16 bytes (12 header + 4 field). That said, 24 vs 16 isn't that big of a deal to gain guaranteed monomorphism.

akarnokd:
There are other ways of doing it, i.e., single object field + private wrapper for Throwables: 16 bytes per onNext and 32 bytes per onError

JakeWharton:
True. I'll leave it up to you though.

:+1:

akarnokd:
Or getting at least bimorphic by having value == null represent onComplete.

akarnokd:
Updated. Now we have the smaller size but an instanceof check for every isOnNext() and getValue calls.

JakeWharton:
Looks good!

src/main/java/io/reactivex/Notification.java
artem-zinnatullin, line 126: It will affect a LOT of code during migration from v1 to v2…

Issue #4366, opened 2016-08-18
2.x Design: BlockingFlowable, BlockingObservable


akarnokd:
The `BlockingFlowable` and `BlockingObservable` classes host blocking operations. They were introduced to make absolutely clear someone uses blocking operations.

The drawback is that they wrap a `Flowable` and `Observable` respectively which incurs allocation whenever `toBlocking()` is called on the base class.

I suggest discarding these blocking classes and moving their operators into the base classes, prefixing the operator name with `blockingX`.

artem-zinnatullin:
SGTM

JakeWharton:
I'm supportive of this only to see two types removed from the public API
and seeing the disconnect of having no BlockingSingle and
BlockingCompletable eliminated. It seems a bit ridiculous to worry about a
single allocation, though, when every other operator already allocates at
least once and this operator is already doing something 10x worse by
throwing away the current thread on a lock.

akarnokd:
See #4371.

PR #4371, opened 2016-08-18
2.x: move blocking operators into the base classes

akarnokd:
This PR removes the `BlockingFlowable` and `BlockingObservable` classes and moves the blocking operators into the respective reactive classes and names prefixed with `blocking`.

In addition, many of the mentions of `Flowable` has been replaced with `Observable` in `Observable`.

Related: #4366.


src/main/java/io/reactivex/Flowable.java
```diff
@@ -5056,29 +5056,420 @@ private static void verifyPositive(long value, String paramName) {
...
+     * </dl>
+     *
+     * @param initialValue
+     *            the initial value that the {@link Iterable} sequence will yield if this
+     *            {@code BlockingObservable} has not yet emitted an item
+     * @return an {@link Iterable} that on each iteration returns the item that this {@code BlockingObservable}
+     *         has most recently emitted
+     * @see <a href="http://reactivex.io/documentation/operators/first.html">ReactiveX documentation: First</a>
+     */
+    public final Iterable<T> blockingMostRecent(T initialValue) {
```
JakeWharton, line 5292: What's the difference between "most recent" and "latest"?

src/main/java/io/reactivex/Flowable.java
```diff
@@ -5056,29 +5056,420 @@ private static void verifyPositive(long value, String paramName) {
...
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator consumes the source {@code Flowable} in an unbounded manner
+     *  (i.e., no backpressure applied to it).</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code blockingLatest} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @return an Iterable that always returns the latest item emitted by this {@code BlockingObservable}
+     * @see <a href="http://reactivex.io/documentation/operators/first.html">ReactiveX documentation: First</a>
```
JakeWharton, line 5266: Wrong link?

src/main/java/io/reactivex/Flowable.java
```diff
@@ -5056,29 +5056,420 @@ private static void verifyPositive(long value, String paramName) {
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code blockingMostRecent} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param initialValue
+     *            the initial value that the {@link Iterable} sequence will yield if this
+     *            {@code BlockingObservable} has not yet emitted an item
+     * @return an {@link Iterable} that on each iteration returns the item that this {@code BlockingObservable}
+     *         has most recently emitted
+     * @see <a href="http://reactivex.io/documentation/operators/first.html">ReactiveX documentation: First</a>
```
JakeWharton, line 5290: Wrong link?

JakeWharton, line 5290: Nevermind. I see "latest" under there.

akarnokd, line 5290: It's a copy of 1.x: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/observables/BlockingObservable.java#L293

JakeWharton:
LGTM 👍

src/main/java/io/reactivex/Flowable.java
akarnokd, line 5292: Most recent keeps reemitting the current value while the push side just overwrites the current value. Latest takes the current value and blocks until one becomes available again; the push side just overwrites the current value.

PR #4375, opened 2016-08-19
2.x: added missing ops, cleanup 8/19-1

akarnokd:
Notable changes:
- renamed and added safe `create` support to `Observable`, `Single` and `Completable`
- added some javadoc
- added `test()` method to `Single` and `Completable`
- factored out `ErrorMode` into `internal.util` (used by various `concat` operators)


JakeWharton:
:+1:

PR #4378, opened 2016-08-19
2.x: implement ops, add javadoc, remove unused components 8/19-2

akarnokd:
Notable changes:
- use JMH 1.13
- fix javadoc generator exclude pattern
- add javadoc to many components
- remove `Observers` and `Subscribers` - all their function is available via subscribe() and/or by implementing the abstract `XObserver`/`XSubscriber` provided
- remove `Schedulers.immediate()`
- remove `Single.subscribe(Subscriber)`, `Single.subscribe(Observer)`, `Completable.subscribe(Subscriber)` and `Completable.subscribe(Observer)`
- remove `CompletableSerializedObserver` and `SingleSerializedObserver` and their parent package
- implement `Single.takeUntil`, `Single.flatMapCompletable`
- update the time-shift handling in the default `Scheduler.schedulePeriodically()` to match 1.x
- fix `Single.subscribeOn` disposable management
- renamed `Subject.hasSubscribers` to `Subject.hasObservers`


src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -43,56 +45,105 @@
...
     
-    public static TestScheduler test() {
+    /**
+     * Creates and returns a {@code TestScheduler}, which is useful for debugging. It allows you to test
+     * schedules of events by manually advancing the clock at whatever pace you choose.
+     *
+     * @return a {@code TestScheduler} meant for debugging
+     */
+    public static TestScheduler test() { // NOPMD
         return new TestScheduler();
```
JakeWharton, line 91: As an aside, it's weird that this is a factory and all the others are accessors. Is `TestScheduler`'s constructor `public`? If so we should just remove this static method that implies there's somehow a shared scheduler for tests.

JakeWharton:
:+1: to code changes. Skimmed the Javadoc for formatting only, not content.

src/main/java/io/reactivex/schedulers/Schedulers.java
akarnokd, line 91: This method is traditional

``` java
Schedulers.test();
new TestScheduler();
```

akarnokd:
I'm not worried about javadoc; the community is great at finding errors in there that can be trivially fixed.

src/main/java/io/reactivex/schedulers/Schedulers.java
JakeWharton, line 91: Traditional how? That it's in 1.x? I don't see why that should stop us from removing it.

Doing something like this is wrong:

``` java
timer(1, SECONDS, Schedulers.test())
```

yet it's how you'd use every other static method in this class.

akarnokd, line 91: I see. Will you post a PR?

JakeWharton, line 91: Always!

PR #4383, opened 2016-08-20
2.x: Remove double dash on RxThreadFactory thread names.

JakeWharton:
Also optimize name creation to avoid `StringBuilder` having to expand its underlying `char[]`.

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/RxThreadFactory.java
+++ b/src/main/java/io/reactivex/internal/schedulers/RxThreadFactory.java
@@ -30,10 +30,7 @@ public final class RxThreadFactory extends AtomicLong implements ThreadFactory {
     
     @Override
     public Thread newThread(Runnable r) {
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/internal/schedulers/RxThreadFactory.java
```diff
@@ -30,10 +30,7 @@ public RxThreadFactory(String prefix) {
     
     @Override
     public Thread newThread(Runnable r) {
-        StringBuilder nameBuilder = new StringBuilder();
-        nameBuilder.append(prefix)
-        .append('-')
-        .append(incrementAndGet());
+        StringBuilder nameBuilder = new StringBuilder(prefix).append('-').append(incrementAndGet());
```
JakeWharton, line 33: Using the constructor initializes the underlying `char[]` to length + 16 instead of the default constructors size of 16 which almost always had to do an expansion to fit the prefix.

PR #4384, opened 2016-08-21
2.x: Do not promise a new instance where we return a singleton.

JakeWharton:


src/main/java/io/reactivex/disposables/Disposables.java
```diff
@@ -97,8 +97,8 @@ public static Disposable empty() {
     }
 
     /**
-     * Returns a new, disposed Disposable instance.
-     * @return a new, disposed Disposable instance
+     * Returns a disposed Disposable instance.
```
vanniktech, line 100: is this line even necessary? It's just duplicated

PR #4387, opened 2016-08-21
2.x: Remove unused test methods from safe observer/subscriber.

JakeWharton:


JakeWharton:
I'm not a huge fan of the word "safe" here. What about calling these `ContractObserver` and `ContractSubscriber` to more accurately describe what they do?

PR #4388, opened 2016-08-21
2.x: Remove reference counted Disposable public type.

JakeWharton:
This is a very specific use case to provide as a first-party type which can always be added back later should there be sufficient demand.


JakeWharton:
Happy to hear arguments for keeping it, but I couldn't come up with any that were convincing enough.

akarnokd:
I never saw an use of it; it is a remnant from Rx.NET where it is part of the inner `Observable` management in the `groupBy` and `window` operator. We have inlined custom `Observable`/`Flowable` there.

PR #4396, opened 2016-08-21
2.x: Observable test change dispose to cancel

vanniktech:


vanniktech:
Oh right cancel does not exist on `TestObserver` yet it should though, right?

akarnokd:
Yes, it should be conditionally disposed. Didn't your IDE warn about the missing method?

vanniktech:
It did but I somehow managed to ignore it and didn't run the tests locally. Should I create the cancel method on `TestObserver` or just leave it as is and close this PR?

akarnokd:
Just use `dispose`.

vanniktech:
Alright then this can be closed.

PR #4397, opened 2016-08-21
1.x: Fix multiple values produced by throttleFirst with TestScheduler

tony-root:
When throttleFirst was operating on a TestScheduler, it delivered all items passed to it untill TestScheduler's time would change to a non-zero value.

### DIFF
--- a/src/test/java/rx/internal/operators/OperatorThrottleFirstTest.java
+++ b/src/test/java/rx/internal/operators/OperatorThrottleFirstTest.java
@@ -19,6 +19,8 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
 
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


src/test/java/rx/internal/operators/OperatorThrottleFirstTest.java
```diff
@@ -172,4 +174,44 @@ public void timed() {
         ts.assertNoErrors();
         ts.assertCompleted();
     }
+
+    @Test
+    public void testThrottleWithoutAdvancingTimeOfTestScheduler() {
```
artem-zinnatullin: nit: no need for test prefix

tony-root: Done

akarnokd:
:+1:  and thanks for the fix.

src/test/java/rx/internal/operators/OperatorThrottleFirstTest.java
```diff
@@ -172,4 +174,44 @@ public void timed() {
...
+        o.onNext(3); // skip
+        o.onCompleted();
+
+        verify(observer).onNext(1);
+        verify(observer).onCompleted();
+        verifyNoMoreInteractions(observer);
+    }
+
+    @Test
+    public void testThrottleWithTestSchedulerTimeOfZero() {
```
artem-zinnatullin: Not sure if this test is required

tony-root: The intention was not to rely implicitly on the fact that TestScheduler's advanceTimeBy(0) is a no-op.
I don't think it's absolutely necessary, though.

artem-zinnatullin: Ok, that's reasonable

PR #4403, opened 2016-08-22
2.x: add missing ops, cleanup, fusion fixes

akarnokd:
Notable changes:
- Implement `concatEager` and its variants
- Implement `onTerminateDetach`
- Implement `distinctUntilChanged(BiPredicate)`
- Replace `EmptyDisposable.INSTANCE` with `Disposables.empty()` due to clash with fusion (sending an INSTANCE tells a fusion-enabled source to not expect onNext values yet the test still send those - this was a problem with EmptySubscription.INSTANCE a while back too). Also added javadoc warning about its use
- add `ObservableScalarXMap` optimization to `Observable.xMap(Function)` and their operator's `subscribeActual`
- Make `ObservableJust` `ScalarCallable` and sync-fuseable with `ScalarDisposable` (similar to `Flowable.just()`)
- Make `Observable.bufferSize()` public for convenience.
- Fix `flatMap` fused `Callable` handling
- Renamed `Objects` to `ObjectHelper` to avoid accidental bad imports of `java.util.Objects` of Java 8.

### DIFF
--- a/src/main/java/io/reactivex/observers/TestObserver.java
+++ b/src/main/java/io/reactivex/observers/TestObserver.java
@@ -22,22 +22,20 @@ import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.CompositeException;
 import io.reactivex.internal.disposables.DisposableHelper;
 import io.reactivex.internal.functions.Objects;
+import io.reactivex.internal.fuseable.QueueDisposable;
 
 /**
- * A subscriber that records events and allows making assertions about them.
+ * An Observer that records events and allows making assertions about them.
  *
- * <p>You can override the onSubscribe, onNext, onError, onComplete, request and
+ * <p>You can override the onSubscribe, onNext, onError, onComplete and
  * cancel methods but not the others (this is by desing).
  * 
  * <p>The TestSubscriber implements Disposable for convenience where dispose calls cancel.
  * 
- * <p>When calling the default request method, you are requesting on behalf of the
- * wrapped actual subscriber.
- * 
  * @param <T> the value type
  */
 public class TestObserver<T> implements Observer<T>, Disposable {
-    /** The actual subscriber to forward events to. */
+    /** The actual observer to forward events to. */
     private final Observer<? super T> actual;
     /** The latch that indicates an onError or onCompleted has been called. */
     private final CountDownLatch done;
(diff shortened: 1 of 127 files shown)


src/main/java/io/reactivex/observers/TestObserver.java
```diff
@@ -678,6 +751,65 @@ public final boolean awaitTerminalEvent(long duration, TimeUnit unit) {
     }
 
     /**
+     * Sets the initial fusion mode if the upstream supports fusion.
+     * @param mode the mode to establish, see the {@link QueueDisposable} constants
+     * @return this
+     */
+    public final TestObserver<T> setInitialFusionMode(int mode) {
```
JakeWharton, line 758: All these methods leak an implementation detail of the library into its public API. Granted it's not the Observable API or anything, but it still feels like they don't belong here. Can it be accomplished in an RxJava-specific subclass that lives in the test source set?

akarnokd, line 758: The problem with that is that `Observable.test()` would need to return it or one would need to cast it to the internal implementation. We can move the constants into the public space. The other option would be to move test consumers into the test source set that would remove the `test` operators as well. Damn lack of extension methods.

Also a less ugly version would be to turn `test()` methods into calls of `to(Function)` and provide TestHelper.newTest(int fusionMode)`that return some custom`TestObserver`.

JakeWharton, line 758: Yeah I'm torn here. I understand the convenience offered but I worry
because it's only serving the built-in tests and not any consumers.

I will say that your to() proposal affords a lot of other benefits. It
eliminates the need for extend() (that I think was talked about recently in
an issue), has a name matching toXxx() for parity with the built-in
conversions methods, and allows the use of other custom test observers
(which we tend to use more than built-in TestObserver) with the same
support as the built-in one.

akarnokd: Let me merge this and will work out something.

vanniktech: How about an @Internal annotation? That one basically declares that while this may be in the public API it is not really part of the public one and subject to change at any time. And in the documentation of the annotation the reasons for having it could be explained.

PR #4418, opened 2016-08-24
1.x: Sync doOnEach java doc

vanniktech:
The other `doOnEach` function has the right java doc. Terminal events such as `onCompleted` / `onError` are also included in `doOnEach`.


akarnokd:
:+1:

PR #4419, opened 2016-08-24
1.x: Add extend() for Single and Completable.

JakeWharton:
Matches `Observable.extend()`. These are all `@Experimental` (including `Observable`). Perhaps we want to rename them all `to()` to match 2.x?

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -102,7 +102,22 @@ public class Completable {
             s.onSubscribe(Subscriptions.unsubscribed());
         }
     }, false); // hook is handled in never()
-    
+
+    /**
+     * Passes all emitted values from this Completable to the provided conversion function to be collected and
+     * returned as a single value. Note that it is legal for a conversion function to return a Completable
+     * (enabling chaining).
+     *
+     * @param <R> the output type of the conversion function
+     * @param conversion a function that converts from the source {@code Completable} to an {@code R}
+     * @return an instance of R created by the provided conversion function
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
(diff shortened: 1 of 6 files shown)


src/main/java/rx/Completable.java
```diff
@@ -102,7 +102,22 @@ public void call(CompletableSubscriber s) {
...
+     * returned as a single value. Note that it is legal for a conversion function to return a Completable
+     * (enabling chaining).
+     *
+     * @param <R> the output type of the conversion function
+     * @param conversion a function that converts from the source {@code Completable} to an {@code R}
+     * @return an instance of R created by the provided conversion function
+     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+     */
+    @Experimental
+    public <R> R extend(Func1<? super CompletableOnSubscribe, ? extends R> conversion) {
```
akarnokd, line 117: The problem with this is that `CompletableOnSubscribe` is practucally useless; you'll have to call `Completable.create()` with it to get meaningful opererations. The whole method should be the same as `to` in 2.x. The `Observable.extend` was pushed hard back then disergarding my reservations.

Of course if you can give a compelling case why reduce the options for the conversion, I'm open minded.

JakeWharton: I agree, and would much prefer `to()`'s semantics.

Since `extend()` is still `@Experimental` we have that option...

akarnokd: :+1: sounds much better

JakeWharton: Closing and will make a new PR, since it might be a day or two.

JakeWharton: https://github.com/ReactiveX/RxJava/pull/4423

Issue #4400, opened 2016-08-22
2.x Design: subscribe() that returns


akarnokd:
Listening to the [Fragmented Podcast](http://fragmentedpodcast.com/episodes/053-jake-wharton-on-rxjava-2/) I came to the realization that the Reactive-Streams `void subscribe(Subscriber)` is going to be a problem for many users. Since we can't (nor should) change the RS interfaces, I suggest a new method on each of the base types:

``` java
<E extends Subscriber<? super T>> E subscribeWith(E subscriber);
<E extends Observer<? super T>> E subscribeWith(E observer);
<E extends SingleObserver<? super T>> E subscribeWith(E observer);
<E extends CompletableObserver> E subscribeWith(E observer);
```

Each of these will simply return the input subscriber after and if said consumer implements `Disposable` (i.e., `ResourceSubscriber`) you can add it to the `CompositeDisposable` as before. (Project Reactor already does offer these methods.)

hamen:
We will still cry a little bit migrating to 2.x, but this will definitely help 😂

mrsasha:
Hm, so what was the alternative? Without this, how were you able to get the ex-`Subscription` ie. `Disposable` reference with 2.0/RS calls?

akarnokd:
Currently you need a `Subscriber` upfront:

``` java
ResourceSubscriber<T> rs = new ResourceSubscriber<T>() { ... };

source.subscribe(rs);

rs.cancel();
```

Or use only lambdas where you already get `Disposable`.

hamen:
Most of my subscribers are anonymous:

``` java
Subscription subscription = source.subscribe(new Observer [...])
```

I guess this will need some work, but probably replacing every `.subscribe()` with `. subscribeWith()` will do as first refactoring cycle.

mrsasha:
Ok, then I definitely think that having `.subscribeWith()` will make it easier to find them in the code.  To reduce cognitive overload one can use `subscribe()` for when there's no need for `Disposable` reference and `subscribeWith()` for other cases.

akarnokd:
Closing via #4422

PR #4422, opened 2016-08-24
2.x: add subscribeWith to get back the subscriber/observer fluently

akarnokd:
This PR adds the `subscribeWith` methods to the base reactive types to mainly help with the 1.x to 2.x changeover as 2.x `subscribe(Subscriber)` return void instead of the resource/cancellation/unsubscription type `Disposable`. See the javadocs and the unit test for example use.

Related: #4400

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -1414,6 +1414,31 @@ public abstract class Completable implements CompletableSource {
     protected abstract void subscribeActual(CompletableObserver s);
 
     /**
+     * Subscribes a given CompletableObserver (subclass) to this Completable and returns the given
+     * CompletableObserver as is.
+     * <p>Usage example:
+     * <pre><code>
+     * Completable<Integer> source = Completable.complete().delay(1, TimeUnit.SECONDS);
+     * CompositeDisposable composite = new CompositeDisposable();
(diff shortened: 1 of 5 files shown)


JakeWharton:
:+1:

src/main/java/io/reactivex/Completable.java
```diff
@@ -1414,6 +1414,31 @@ public final void subscribe(CompletableObserver s) {
...
+     * 
+     * composite.add(source.subscribeWith(new ResourceCompletableObserver()));
+     * </code></pre>
+     * @param <E> the type of the CompletableObserver to use and return
+     * @param observer the CompletableObserver (subclass) to use and return, not null
+     * @return the input {@code observer}
+     * @throws NullPointerException if {@code observer} is null
+     * @since 2.0
+     */
+    public final <E extends CompletableObserver> E subscribeWith(E observer) {
```
vanniktech, line 1436: will there also be an overload that only takes `onComplete`? Or something that takes 2 arguments, `onComplete` & `onError`?

akarnokd: See the lines below.

vanniktech: Ups 😕  thanks

PR #4423, opened 2016-08-25
1.x: Add to() conversion function to all stream types.

JakeWharton:
This deprecates extend() on Observable, which is a less powerful version of these functions.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -2222,13 +2222,14 @@ public class Completable {
     }
     
     /**
-     * Allows fluent conversion to another type via a function callback.
-     * @param <U> the output type as determined by the converter function
-     * @param converter the function called with this which should return some other value.
-     * @return the converted value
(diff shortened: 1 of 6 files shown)


src/main/java/rx/Completable.java
```diff
@@ -2222,13 +2222,14 @@ public final Completable timeout0(long timeout, TimeUnit unit, Scheduler schedul
...
-     * @param <U> the output type as determined by the converter function
-     * @param converter the function called with this which should return some other value.
-     * @return the converted value
-     * @throws NullPointerException if converter is null
+     * Calls the specified converter function during assembly time and returns its resulting value.
+     * <p>
+     * This allows fluent conversion to any other type.
+     * @param <R> the resulting object type
+     * @param converter the function that receives the current Single instance and returns a value
+     * @return the value returned by the function
```
JakeWharton, line 2230: This documentation change normalizes it with the others and the versions in 2.x

akarnokd:
:+1:

Issue #4420, opened 2016-08-24
1.x: Base interface types


JakeWharton:
Before `Single` and `Completable` progress to stable, I'd like to address the interface naming inconsistency in a similar fashion to what #4044 did for 2.x.

Here's the current state:

<img width="614" alt="screen shot 2016-08-24 at 1 30 57 pm" src="https://cloud.githubusercontent.com/assets/66577/17940957/3f4ee112-69ff-11e6-8d0f-68dc06df051a.png">

It's very clear that there's a lack of normalization. ~~I'd like to take the same approach as we did in 2.x and move to prefixes and top-level types.~~ I'd like to normalize the names of the nested classes and make the subscribers top-level types. We can `@Deprecate` the existing types and make them extend from these new top-level types to make migration easier, deleting these `@Deprecated` types whenever these APIs turn stable. 👍 👎 ?

Additionally, there's a problem with `CompletableSubscriber` in that it's an interface instead of an abstract class that also implements `Subscription` (as both `Subscriber` and `SingleSubscriber` do). I would like to correct this. 👍 👎 ?

akarnokd:
1) Kind of disruptive at this point, plus `@Deprecated` never get removed from the main versions (since it was promised)

2) It's a modern design just like in 2.x but you can introduce the `ResourceCompletable` similar to 2.x `ResourceSubscriber`.

JakeWharton:
Er, sorry I meant to change that paragraph before hitting create. I want to propose non-prefixed inner types similar to what `Observable` has except for the `Subscribers` which will be top-level.

This will only change non-stable APIs so they could just be renamed.

I was suggesting `@Deprecating` these non-stable old interfaces so that it didn't immediately break people upgrading. Then eventually you could delete those interfaces since they were never stable in the first place. We also could just rename them though and not bother.

The changes would look like this:

<img width="614" alt="screen shot 2016-08-24 at 3 51 00 pm" src="https://cloud.githubusercontent.com/assets/66577/17945620/9e654228-6a12-11e6-96e5-8a3618d9f762.png">

akarnokd:
Let's see a PR so the diff is more telling about the consequences.

akarnokd:
Can we do more about this or can this be closed?

JakeWharton:
I think we're now at a good enough place with 1.x here. Any more changes would only be disruptive for the sake of being disruptive.

akarnokd:
Great!

PR #4425, opened 2016-08-25
1.x: Remove 'Completable' prefix from nested interfaces, move its subscription to top-level.

JakeWharton:
This is only the `Completable` part of #4420. It does not make `CompletableSubscriber` an abstract class nor create `Single.Operator`.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -20,7 +20,6 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import rx.Observable.OnSubscribe;
 import rx.annotations.Experimental;
 import rx.exceptions.*;
 import rx.functions.*;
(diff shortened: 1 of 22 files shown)


akarnokd:
[Compiler error](https://travis-ci.org/ReactiveX/RxJava/builds/155096130#L200).

akarnokd:
Looks okay. What about the `@Deprecated` markers?

JakeWharton:
Oh, sure. I'll add them in real quick. I wasn't sure on your stance on them, but I think it'll make migration easier.

akarnokd:
Yes, having deprecated markers is more forgiveable than lacking the entire type. I don't know how extensively `Completable` is in use by other libraries right now (beyond RxJavaReactiveStreams, Reactor's converter, Retrofit`).

JakeWharton:
Updated!

src/main/java/rx/Completable.java
```diff
@@ -39,66 +38,76 @@
...
+    public interface CompletableOperator extends Operator {
+
+    }
     
     /**
      * Convenience interface and callback used by the lift operator that given a child CompletableSubscriber,
      * return a parent CompletableSubscriber that does any kind of lifecycle-related transformations.
      */
-    public interface CompletableOperator extends Func1<CompletableSubscriber, CompletableSubscriber> {
+    public interface Operator extends Func1<rx.CompletableSubscriber, rx.CompletableSubscriber> {
```
akarnokd, line 89: I'm not certain about this being a compatible change.

So does this still compile?

``` java
CompletableOperator op = new CompletableOperator() {
    @Override
    public Completable.CompletableSubscriber call(Completable.CompletableSubscriber cs) {
         return cs;
    }
}
```

JakeWharton: Hmm yeah it's not possible to provide perfect migration using this strategy. The covariant return type works but not the argument. We could provide overloads of `lift`, `compose`, and `subscribe` which accepted the old types, adapted them to the new type, and were marked as deprecated. That should provide perfect compatibility.

akarnokd: In this case, how about the old ones not extending the new ones and have wrappers inside those methods of the old types?

JakeWharton: Updated with overloads

akarnokd:
https://travis-ci.org/ReactiveX/RxJava/builds/155134046#L200

JakeWharton:
Compilation issue resolved.

akarnokd:
I know,  still waiting for a second approval.

akarnokd:
Okay, let's have this. One can complain anytime later.

PR #4430, opened 2016-08-26
2.x Improve coverage, fix bugs

akarnokd:
The new tests revealed some bugs (nothing major).

### DIFF
--- a/src/main/java/io/reactivex/disposables/CompositeDisposable.java
+++ b/src/main/java/io/reactivex/disposables/CompositeDisposable.java
@@ -225,14 +225,7 @@ public final class CompositeDisposable implements Disposable, DisposableContaine
         }
         if (errors != null) {
             if (errors.size() == 1) {
-                Throwable ex = errors.get(0);
-                if (ex instanceof Error) {
-                    throw (Error)ex;
-                }
-                if (ex instanceof RuntimeException) {
-                    throw (RuntimeException)ex;
-                }
-                throw ExceptionHelper.wrap(ex);
+                throw ExceptionHelper.wrapOrThrow(errors.get(0));
             }
             throw new CompositeException(errors);
         }
--- a/src/main/java/io/reactivex/disposables/CompositeDisposable.java
+++ b/src/main/java/io/reactivex/disposables/CompositeDisposable.java
@@ -17,7 +17,7 @@ import java.util.*;
 import io.reactivex.exceptions.*;
 import io.reactivex.internal.disposables.DisposableContainer;
 import io.reactivex.internal.functions.ObjectHelper;
-import io.reactivex.internal.util.OpenHashSet;
+import io.reactivex.internal.util.*;
 
 /**
  * A disposable container that can hold onto multiple other disposables and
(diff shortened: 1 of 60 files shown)


src/main/java/io/reactivex/disposables/CompositeDisposable.java
```diff
@@ -223,7 +225,14 @@ void dispose(OpenHashSet<Disposable> set) {
         }
         if (errors != null) {
             if (errors.size() == 1) {
-                throw Exceptions.propagate(errors.get(0));
+                Throwable ex = errors.get(0);
+                if (ex instanceof Error) {
+                    throw (Error)ex;
+                }
+                if (ex instanceof RuntimeException) {
```
JakeWharton: `wrap` handles this case

src/main/java/io/reactivex/disposables/CompositeDisposable.java
```diff
@@ -223,7 +225,14 @@ void dispose(OpenHashSet<Disposable> set) {
         }
         if (errors != null) {
             if (errors.size() == 1) {
-                throw Exceptions.propagate(errors.get(0));
+                Throwable ex = errors.get(0);
+                if (ex instanceof Error) {
```
JakeWharton: Move this to `wrap`?

src/main/java/io/reactivex/internal/util/ExceptionHelper.java
```diff
@@ -23,6 +23,20 @@
...
 public enum ExceptionHelper {
     ;
+    
+    /**
+     * Wraps the given error Exception into a RuntimeException if not already
+     * a RuntimeException
+     * @param error the error to wrap
+     * @return the (wrapped) error
+     */
+    public static RuntimeException wrap(Throwable error) {
```
JakeWharton: How about `throwUnchecked` or `throwAsUnchecked`?

src/main/java/io/reactivex/disposables/CompositeDisposable.java
akarnokd: The reason I inlined these here is because if I'm using `propagate`, that shows up in the code coverage analysis: the `throw Exceptions.propagate(ex)` never returns properly and thus the coverage thinks a code path is not covered. It is usually the last roadblock to a 100% covered file. I've been thinking about having `ExceptionHelper.wrapOrThrow()` that throws for non-checked and returns a wrapped checked that will run through throw.

JakeWharton:
:+1:

src/main/java/io/reactivex/disposables/CompositeDisposable.java
JakeWharton: Sounds good to me. These exception helper methods could _always_ return an unchecked exception forcing you to use `throw` and ensuring code coverage catches it.

JakeWharton: Er, I guess that doesn't work for `Error` actually.

akarnokd:
``` java
    public static RuntimeException wrapOrThrow(Throwable error) {
        if (error instanceof Error) {
            throw (Error)error;
        }
        if (error instanceof RuntimeException) {
            return (RuntimeException)error;
        }
        return new RuntimeException(error);
    }
```

akarnokd:
Updated those in the first commit in this PR and all other places that used `Exceptions.propagate`.

JakeWharton:
Looks good

PR #4431, opened 2016-08-26
2.x: coverage, fixes, cleanup 8/27-1

akarnokd:
- updated `Single.cache()` to be lock-free, to allocate less and work properly
- added test coverage to some classes
- remove unused internal `EmptyObserver`
- update `PublishSubject` to support cross-cancellation
- update `TestHelper.race` to use the current thread for one of the tasks


src/main/java/io/reactivex/internal/operators/single/SingleCache.java
```diff
@@ -13,97 +13,166 @@
...
+            if (observers.compareAndSet(a, b)) {
+                return;
             }
-            return;
         }
+    }
+
+    @Override
+    public void onSubscribe(Disposable d) {
+        // not supported by this operator
```
JakeWharton, line 124: throw assertion error to ensure it never happens?

akarnokd: We don't support dispose() once the cache has subscribed but the upstream should still run - this method is a no-op. Throwing here is not allowed anyway.

However, I'm willing to extend the SingleCache (and the others) by implementing `Disposable` and people can then cast and cancel a `cache()` too.

JakeWharton:
:+1:

src/main/java/io/reactivex/internal/operators/single/SingleCache.java
JakeWharton: Ah. I don't think that's worth it. I just misunderstood the comment and thought "not allowed" instead of "not supported".

PR #4434, opened 2016-08-27
2.x: cleanup, bugfixes, coverage 8/27-2

akarnokd:
- Remove unused code
- Improve coverage of `Single`, `Completable` and `Observable`
- Fix minor bugs in operators

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -1743,16 +1743,10 @@ public abstract class Observable<T> implements ObservableSource<T> {
      */
     @SchedulerSupport(SchedulerSupport.CUSTOM)
     public static Observable<Long> interval(long initialDelay, long period, TimeUnit unit, Scheduler scheduler) {
-        if (initialDelay < 0) {
-            initialDelay = 0L;
-        }
-        if (period < 0) {
-            period = 0L;
-        }
         ObjectHelper.requireNonNull(unit, "unit is null");
         ObjectHelper.requireNonNull(scheduler, "scheduler is null");
 
-        return RxJavaPlugins.onAssembly(new ObservableInterval(initialDelay, period, unit, scheduler));
+        return RxJavaPlugins.onAssembly(new ObservableInterval(Math.max(0L, initialDelay), Math.max(0L, period), unit, scheduler));
     }
 
     /**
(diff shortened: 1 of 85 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -1743,16 +1743,10 @@ public static int bufferSize() {
      */
     @SchedulerSupport(SchedulerSupport.CUSTOM)
     public static Observable<Long> interval(long initialDelay, long period, TimeUnit unit, Scheduler scheduler) {
-        if (initialDelay < 0) {
-            initialDelay = 0L;
```
JakeWharton, line 1747: Surprised we aren't throwing here (and others)

src/main/java/io/reactivex/Observable.java
```diff
@@ -10108,7 +10082,7 @@ public final void subscribe(Observer<? super T> observer) {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Observable<T> take(long count) {
         if (count < 0) {
```
vanniktech: also use verifyPositive here?

src/main/java/io/reactivex/Observable.java
```diff
@@ -4701,12 +4685,8 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <U extends Collection<? super T>> Observable<U> buffer(int count, int skip, Callable<U> bufferSupplier) {
-        if (count <= 0) {
-            throw new IllegalArgumentException("count > 0 required but it was " + count);
-        }
-        if (skip <= 0) {
-            throw new IllegalArgumentException("skip > 0 required but it was " + count);
-        }
+        verifyPositive(count, "count");
+        verifyPositive(skip, "skip");
```
JakeWharton, line 4689: I think one of these is wrong because `skip` can be 0 but `count` cannot.

JakeWharton, line 4689: Based on other methods in this diff I would guess `verifyPositive` rejects 0

akarnokd: It creates buffers every `index % skip` so zero not allowed.

src/main/java/io/reactivex/observers/TestObserver.java
```diff
@@ -895,15 +895,16 @@ private String fusionModeToString(int mode) {
      * @param time the waiting time
      * @param unit the time unit of the waiting time
      * @return this
-     * @throws InterruptedException if the wait is interrupted
+     * @throws RuntimeExceptio wrapping an InterruptedException if the wait is interrupted
```
JakeWharton: nit: missing 'n' on RuntimeException

JakeWharton:
:+1:

PR #4435, opened 2016-08-28
2.x: cleanup, behavior clarifications, fixes, coverage 8/28-1

akarnokd:
- Add tests
- fix mistakes in `TestObserver`
- Combine `DisposableHelper` and its inner `Disposed` enum
- Compact `NotificationLite` and its inner `Complete` enum
- Fix NPE in `ListCompositeDisposable` constructor
- Fix `DeferredScalarObserver` fusion and state management
- Turned a few empty enums into classes with private constructor: coverage can't cover them 100% otherwise. These empty enums have a constructor in their bytecode which never gets called and thus there's always 10 instructions / 2 method calls missing.
- Removed `TestObserver.awaitDone()`, one should always await with timeout

### DIFF
--- a/src/main/java/io/reactivex/internal/disposables/ListCompositeDisposable.java
+++ b/src/main/java/io/reactivex/internal/disposables/ListCompositeDisposable.java
@@ -42,6 +42,7 @@ public final class ListCompositeDisposable implements Disposable, DisposableCont
     
     public ListCompositeDisposable(Iterable<? extends Disposable> resources) {
         ObjectHelper.requireNonNull(resources, "resources is null");
+        this.resources = new LinkedList<Disposable>();
         for (Disposable d : resources) {
             ObjectHelper.requireNonNull(d, "Disposable item is null");
             this.resources.add(d);
--- a/src/main/java/io/reactivex/observers/SafeObserver.java
+++ b/src/main/java/io/reactivex/observers/SafeObserver.java
@@ -42,9 +42,6 @@ public final class SafeObserver<T> implements Observer<T>, Disposable {
     
     @Override
     public void onSubscribe(Disposable s) {
-        if (done) {
-            return;
-        }
         if (DisposableHelper.validate(this.s, s)) {
             this.s = s;
             try {
@@ -57,7 +54,7 @@ public final class SafeObserver<T> implements Observer<T>, Disposable {
                     s.dispose();
                 } catch (Throwable e1) {
                     Exceptions.throwIfFatal(e1);
-                    RxJavaPlugins.onError(e1);
+                    RxJavaPlugins.onError(new CompositeException(e, e1));
                 }
                 RxJavaPlugins.onError(e);
             }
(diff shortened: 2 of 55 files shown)


src/main/java/io/reactivex/internal/disposables/ListCompositeDisposable.java
```diff
@@ -42,6 +42,7 @@ public ListCompositeDisposable(Disposable... resources) {
     
     public ListCompositeDisposable(Iterable<? extends Disposable> resources) {
         ObjectHelper.requireNonNull(resources, "resources is null");
+        this.resources = new LinkedList<Disposable>();
```
vanniktech, line 45: nit: initialize with resources size?

src/test/java/io/reactivex/observers/SafeObserverTest.java
```diff
@@ -486,4 +487,433 @@ public void onComplete() {
...
+            SafeObserver<Object> so = cd.toSafe();
+            
+            so.onError(new TestException());
+            
+            TestHelper.assertError(list, 0, CompositeException.class);
+            List<Throwable> ce = TestHelper.compositeList(list.get(0));
+            TestHelper.assertError(ce, 0, TestException.class);
+            TestHelper.assertError(ce, 1, NullPointerException.class, "Subscription not set!");
+        } finally {
+            RxJavaPlugins.reset();
```
vanniktech, line 839: maybe a custom junit rule would be good here? I see that scheme a lot in tests to reset the plugin system once the test is done

src/main/java/io/reactivex/internal/disposables/ListCompositeDisposable.java
akarnokd: LinkedList?!

JakeWharton: Also iterables do not expose their size anyway. Only collections.

src/main/java/io/reactivex/observers/SafeObserver.java
```diff
@@ -80,22 +77,60 @@ public void onNext(T t) {
...
+    void onNextNoSubscription() {
+        
+        Throwable ex = new NullPointerException("Subscription not set!");
+        
+        try {
+            actual.onSubscribe(EmptyDisposable.INSTANCE);
+        } catch (Throwable e) {
+            Exceptions.throwIfFatal(e);
+            // can't call onError because the actual's state may be corrupt at this point
+            RxJavaPlugins.onError(new CompositeException(ex, e));
```
JakeWharton, line 122: Aren't these backwards? Shouldn't it be `e, ex` since the message states something like "order they were received".

JakeWharton, line 122: Looks like in other instances of this PR you have them in the `e, ex` order

JakeWharton, line 122: Ah, crap. The naming is just backwards here. Ignore me.

src/main/java/io/reactivex/observers/TestObserver.java
```diff
@@ -877,21 +865,6 @@ private String fusionModeToString(int mode) {
...
      * Awaits until the internal latch is counted down.
      * <p>If the wait times out or gets interrupted, the TestSubscriber is cancelled.
-     * @return this
-     * @throws InterruptedException if the wait is interrupted
-     */
-    public final TestObserver<T> awaitDone() throws InterruptedException {
-        try {
-            done.await();
-        } catch (InterruptedException ex) {
-            cancel();
```
JakeWharton, line 887: This was missing a re-throw of the exception anyway!

src/main/java/io/reactivex/observers/SafeObserver.java
akarnokd: First is the NPE, the second is the crash of the call; don't want to lose either or drift apart. Also naming is swapped sometimes but should be still in the correct order.

JakeWharton:
> Turned a few empty enums into classes with private constructor

This saves methods for Android as well. I wanted to see how many were being added before pressing the issue about this convention, but nice to see there's another motivation for not using enums here.

JakeWharton:
:+1:

Issue #4321, opened 2016-08-09
2.x changing some Observable operators to return Single. 


abersnaze:
I was looking through the Observable operators and wondering if now would be a good time to change the following methods to return Singles instead of Observables.
- [x] `Single<Boolean> all(Predicate<? super T> predicate)`
- [x] `Single<Boolean> any(Predicate<? super T> predicate)`
- [x] `<U> Single<U> collect(Callable<? extends U> initialValueSupplier, BiConsumer<? super U, ? super T> collector)`
- [x] `<U> Single<U> collectInto(final U initialValue, BiConsumer<? super U, ? super T> collector)`
- [x] `Single<Boolean> contains(final Object o)`
- [x] `Single<Long> count()`
- [x] `Maybe<T> elementAt(long index)`
- [x] `Single<T> elementAt(long index, T defaultValue)`
- [x] `Maybe<T> firstElement()`
- [x] `Single<T> first(T defaultItem)`
- [x] `Completable ignoreElements()`
- [x] `Single<Boolean> isEmpty()`
- [x] `Maybe<T> lastElement()`
- [x] `Single<T> last(T defaultItem)`
- [x] `Maybe<T> reduce(BiFunction<T, T, T> reducer)`
- [x] `<R> Single<R> reduce(R seed, BiFunction<R, ? super T, R> reducer)`
- [x] `<R> Single<R> reduceWith(Callable<R> seedSupplier, BiFunction<R, ? super T, R> reducer)`
- [x] `Maybe<T> singleElement()`
- [x] `Single<T> single(T defaultItem)`
- [x] `Single<List<T>> toList()`
- [x] `Single<List<T>> toList(final int capacityHint)`
- [x] `<K> Single<Map<K, T>> toMap(final Function<? super T, ? extends K> keySelector)`
- [x] `<K, V> Single<Map<K, V>> toMap(final Function<? super T, ? extends K> keySelector, final Function<? super T, ? extends V> valueSelector)`
- [x] `<K> Single<Map<K, Collection<T>>> toMultimap(Function<? super T, ? extends K> keySelector)`
- [x] `<K, V> Single<Map<K, Collection<V>>> toMultimap(Function<? super T, ? extends K> keySelector, Function<? super T, ? extends V> valueSelector)`
- [x] `Single<List<T>> toSortedList()`
- [x] `Single<List<T>> toSortedList(final Comparator<? super T> comparator)`
[...]

PR #4436, opened 2016-08-29
2.x: Maybe for lazy Optional

abersnaze:
In #4321 I hit on the problem of needing a lazy object that could either be zero or one value. In that issue the idea of switching the return value of `reduce(R, Func2<R,T,R>)` from `Observable<R>` to `Single<R>` because that operator can only ever produce an Observable that emits exactly one value. The problem was that `reduce(Func2<T,T,T>)` could not be changed to `Single<T>` because the source Observable could be empty and therefore not produce one and only one value.

To solve this problem I mentioned we could create a lazy type that represents either 1 or no value could fill the gap between `Completable` and `Single<T>`.

This PR is for the introduction of the `Maybe<T>` type.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -27,6 +27,7 @@ import io.reactivex.internal.functions.ObjectHelper;
 import io.reactivex.internal.fuseable.*;
 import io.reactivex.internal.operators.completable.CompletableFromPublisher;
 import io.reactivex.internal.operators.flowable.*;
+import io.reactivex.internal.operators.maybe.MaybeFromPublisher;
 import io.reactivex.internal.operators.observable.ObservableFromPublisher;
 import io.reactivex.internal.operators.single.SingleFromPublisher;
 import io.reactivex.internal.schedulers.ImmediateThinScheduler;
@@ -13596,6 +13597,12 @@ public abstract class Flowable<T> implements Publisher<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Single<T> toSingle() {
         return RxJavaPlugins.onAssembly(new SingleFromPublisher<T>(this));
+    }
+
+    @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Maybe<T> toMaybe() {
+        return new MaybeFromPublisher<T>(this);
     }
 
     /**
--- a/src/main/java/io/reactivex/internal/operators/maybe/MaybeFromCallable.java
+++ b/src/main/java/io/reactivex/internal/operators/maybe/MaybeFromCallable.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.internal.operators.maybe;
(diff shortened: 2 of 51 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -5730,7 +5731,7 @@ public final void blockingSubscribe(Subscriber<? super T> subscriber) {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <B> Flowable<List<T>> buffer(Publisher<B> boundaryIndicator, final int initialCapacity) {
         return buffer(boundaryIndicator, Functions.<T>createArrayList(initialCapacity));
-    }
+            }
```
akarnokd: Why are these misaligned?

src/main/java/io/reactivex/internal/operators/maybe/MaybeFromCallable.java
```diff
@@ -0,0 +1,45 @@
...
+    final Callable<? extends T> callable;
+    
+    public MaybeFromCallable(Callable<? extends T> callable) {
+        this.callable = callable;
+    }
+
+    @Override
+    protected void subscribeActual(MaybeObserver<? super T> s) {
+
+        s.onSubscribe(EmptyDisposable.INSTANCE);
```
akarnokd, line 33: Setting an `EmptyDisposable` should be followed either by an `onError()` or `onComplete()` because a fused-consumer won't expect and won't request a value from it.

akarnokd:
You referenced Java 8 types in your code.

akarnokd:
My problem with adding another base type is the increased maintenance cost. Currently we have 3 + 4 types in the major versions and it is likely adding a new operator, overload or making a fix has to be done for all of them. I'm one-man armying RxJava for a considerable time now but I have my limits. Therefore my suggestion is that you have this `Maybe` as a separate library that depends on RxJava 2 but RxJava 2 doesn't depend on it. Since RxJava 2 offers fluent conversion method(s), it should be only a small inconvenience to convert to `Maybe`:

``` java
Flowable<Integer> source = Flowable.range(1, 10);

Maybe<Integer> mb = source.to(Maybe.reduceFlowable((a, b) -> a + b));

mb.filter(v -> false).subscribe(
    System.out::println, Throwable::printStackTrace, () -> System.out.println("Done"));

mb.toFlowable().subscribe(
    System.out::println, Throwable::printStackTrace, () -> System.out.println("Done"));
```

abersnaze:
As I mentioned my plans were to greatly increase the coupling between Observable & Flowable to Single & Maybe by having the various operators that return exactly 1 or sometimes 1 value to these two types.

src/main/java/io/reactivex/internal/operators/maybe/MaybeFromCallable.java
abersnaze: I copied this from `SingleFromCallable` so the same bug exists there too.

```
        s.onSubscribe(EmptyDisposable.INSTANCE);
        try {
            T v = callable.call();
            if (v != null) {
                s.onSuccess(v);
```

akarnokd: I'll revisit the `Single` implementation for these cases.

akarnokd:
I'm willing to accept `Maybe` as the 5th reactive type but the class needs javadoc and unit tests that verify the most basic behavior of each operator in it.

/cc @artem-zinnatullin @JakeWharton @zsxwing

akarnokd:
How about you just contribute `Maybe` with operators throwing `UnsupportedOperationException` and I'll add optimized implementations and tests for them as I go?

akarnokd:
Also Reactor has the [Mono](https://github.com/reactor/reactor-core/blob/master/src/main/java/reactor/core/publisher/Mono.java) type exactly for this purpose, I suggest you try that first.

JakeWharton:
I'm not strictly opposed because

> my plans were to greatly increase the coupling between Observable & Flowable to Single & Maybe by having the various operators that return exactly 1 or sometimes 1 value to these two types.

appeals to me, but I don't have a really strong feeling without using it more.

akarnokd:
Okay then, let's merge this and see how it works out.

akarnokd:
Please don't post PRs targeting Maybe - I'm working on cleaning it up a bit (plus a few other classes if I find something wrong); the code copied from an older commit of the other types.

PR #4437, opened 2016-08-29
2.x: Subject/Processor improvements & small cleanup

akarnokd:
- open up `verifyPositive` as protected for validation convenience in subclasses
- improve size and state handling of `DeferredScalarSubscription`
- enable async-fusion on `AsyncSubject` and `AsyncProcessor`, compact the classes
- **behavior change** `AsyncX.hasValue()` returns false until the terminal state has been reached
- compact the classes `PublishSubject`, `PublishProcessor`, `BehaviorSubject`, `BehaviorProcessor`, `ReplaySubject` and `ReplayProcessor`
- enable async-fusion on `UnicastSubject`, compact the class
- rename `BaseXQueueDisposable` to `BasicXQueueDisposable` to match `BasicXQueueSubscription`
- introduce `DeferredScalarDisposable`, refactor `DeferredScalarObserver`
- adjust tests that asserted on the former parameter names
- subjects and processors can now be created with newing up their constructor: `new PublishSubject<>()`; there is no mandatory shared state between the `Observer` part and the `Observable` part because both are now stateless on their own.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -3757,12 +3757,14 @@ public abstract class Flowable<T> implements Publisher<T> {
      * the parameter name.
      * @param value the value to validate
      * @param paramName the parameter name of the value
+     * @return the value
      * @throws IllegalArgumentException if bufferSize &lt;= 0
      */
-    private static void verifyPositive(int value, String paramName) {
+    protected static int verifyPositive(int value, String paramName) {
         if (value <= 0) {
             throw new IllegalArgumentException(paramName + " > 0 required but it was " + value);
         }
+        return value;
     }
 
     /**
@@ -3770,12 +3772,14 @@ public abstract class Flowable<T> implements Publisher<T> {
      * the parameter name.
      * @param value the value to validate
      * @param paramName the parameter name of the value
+     * @return the value
      * @throws IllegalArgumentException if bufferSize &lt;= 0
      */
-    private static void verifyPositive(long value, String paramName) {
+    protected static long verifyPositive(long value, String paramName) {
         if (value <= 0L) {
             throw new IllegalArgumentException(paramName + " > 0 required but it was " + value);
         }
+        return value;
     }
 
     /**
--- a/src/main/java/io/reactivex/processors/AsyncProcessor.java
+++ b/src/main/java/io/reactivex/processors/AsyncProcessor.java
@@ -13,13 +13,11 @@
 package io.reactivex.processors;
 
 import java.util.Arrays;
-import java.util.concurrent.atomic.*;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.reactivestreams.*;
 
-import io.reactivex.functions.IntFunction;
-import io.reactivex.internal.subscriptions.SubscriptionHelper;
-import io.reactivex.internal.util.NotificationLite;
+import io.reactivex.internal.subscriptions.DeferredScalarSubscription;
 import io.reactivex.plugins.RxJavaPlugins;
 
 /**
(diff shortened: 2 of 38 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -3757,25 +3757,29 @@ public static int bufferSize() {
      * the parameter name.
      * @param value the value to validate
      * @param paramName the parameter name of the value
+     * @return the value
      * @throws IllegalArgumentException if bufferSize &lt;= 0
      */
-    private static void verifyPositive(int value, String paramName) {
+    protected static int verifyPositive(int value, String paramName) {
```
JakeWharton, line 3763: These aren't high-value APIs and are trivial to replicate with

``` java
if (value < 0) throw new IllegalArgumentException("value < 0: " + value);
```

I would prefer they be moved somewhere in `internal` and made public for use all over the library but not be exposed in the library's public API.

akarnokd: Next PR?

src/main/java/io/reactivex/processors/AsyncProcessor.java
```diff
@@ -28,150 +26,244 @@
...
     }
     
-    protected AsyncProcessor() {
-        this.state = new State<T>();
+    /**
+     * Constructs an AsyncProcessor.
+     * @since 2.0
+     */
+    @SuppressWarnings("unchecked")
+    public AsyncProcessor() {
```
JakeWharton: Seems redundant to have both static factories and public constructors in the public API. Especially when there's multiple overloads such as in some of the other processors. Can we choose one and expose only it?

akarnokd: I'd like to have only the constructor(s) (except `ReplaySubject.create(capacityHint)` as there is no good way to expose it and not get confused with bounded `new ReplaySubject(int)`) and wanted your opinion.

Also I haven't thought out the policies between the release candidates. Should we `@Deprecate` what we don't need in the RC2 and then remove them RC3 or just delete stuff and mention it in the release log?

JakeWharton: Ugh, yeah. I don't really have any good suggestions here for that. The bounded constructor could be `int capacityHint, int size` but that's not a very descriptive for the caller. Are there any other examples that clash?

Personally I prefer the named static methods. They make it more clear as to what the values actually mean. Seeing `new ReplaySubject(16, 16)` doesn't tell me much where `ReplaySubject.createWithSize(16)` is much more explicit. Static methods are used for creating stream types (aside from subclassing) so I think it plays nicely.

> Should we `@Deprecate` what we don't need in the RC2 and then remove them RC3 or just delete stuff and mention it in the release log?

It would be nice to do the `@Deprecate` and remove where it doesn't impede the changes that need to be made in the library. Behavior changes may already be breaking consumers so minimizing the amount of work that needs to be done immediately when you're updating the version is helpful.

I would definitely call them out in the change log regardless.

akarnokd: `ReplaySubject(int, int)` makes no sense (in the current algorithm) because you are either unbounded (but want to avoid reallocating the buffer) or bounded (where you have a linked-list anyway). The benefit of the constructors may be more apparent in Java 7+ with diamond.

JakeWharton: I don't think the generics side of this plays a part in influencing the outcome.

Is there some other reconciliation for the fact that you want two overloads that accept an `int` but mean different things? I don't think having both constructors and static factories (even if they don't overlap) is a good API to expose. I would much prefer they all were constructors or all were static factories.

akarnokd: Updated them to have factory methods as the means of construction.

JakeWharton:
thanks :+1:

PR #4450, opened 2016-08-31
2.x: fixes, cleanup, coverage 8/31-1

akarnokd:
- Improve coverage of internal classes
- Turn more empty enums into classes with private constructors
- Fix `Single.subscribeOn` and `Completable.subscribeOn` disposable management.
- Create a compact `FutureSubscriber` and `FutureObserver` classes, drop their former enums
- Deprecate `Disposables.from` methods, add `Disposables.fromX` methods where X = argument type
- Fix error management in `Completable.mergeX` operators.
- Remove unused methods from internal classes


src/main/java/io/reactivex/Flowable.java
```diff
@@ -5086,7 +5086,7 @@ public final T blockingSingle(T defaultItem) {
      * @see <a href="http://reactivex.io/documentation/operators/to.html">ReactiveX documentation: To</a>
      */
     public final Future<T> toFuture() {
-        return FlowableToFuture.toFuture(this);
+        return subscribeWith(new FutureSubscriber<T>());
```
JakeWharton, line 5089: Cool trick!

JakeWharton:
:+1:

PR #4453, opened 2016-09-01
Wrap InterruptedException with an unchecked exception in TestSubscriber#awaitValueCount().

vy:
In its current form, `awaitValueCount()` is the only `TestSubscriber#await*`method that throws a checked exception (that is, `InterruptedException`), whereas the others wrap it with a `IllegalStateException`. This spreads a try-catch disease throughout the entire code base where `awaitValueCount()` is used. One can argue that why not just declaring the exception in the caller method footprint: Because you might be implementing an interface (e.g. `Runnable`) which does not allow any exceptions in its footprint. This patch wraps the `InterruptedException` with an unchecked exception in `TestSubscriber#awaitValueCount()`.


akarnokd:
:+1:

Issue #4456, opened 2016-09-01
2.x: Naming of .doOnCancel()


Mauin:
As I understand it the `.doOnCancel()` operator is basically the replacement of `.doOnUnsubscribe()` in the 1.x world.

In that case: Just from a naming perspective wouldn't it be clearer to name it `.doOnDispose()` ?

akarnokd:
`Flowable` -> `doOnCancel()` because of `Subscription.cancel()`, the other can be called `doOnDispose()`. Want to submit a PR?

Mauin:
Ah, I was not looking much at Flowable yet, that's why I missed that part. Thanks for explaining!

I'll get started on the PR.

Issue #4451, opened 2016-08-31
2.x: Patch releases to the release candidates


akarnokd:
There were a bunch of low-to-medium bugs and annoyances released with RC1 which have been already fixed in the last couple of days. However, RC2 is scheduled for September 23-24 (3 weeks from now) which is a lot of time to live with said issues.

Therefore, I suggest releasing patch updates to the release candidates as follows:
- Their version tag should be like `2.0.0-RC1-u1` if the Nebula/Maven cooperates on it
- Public API renames should keep the original as `@Deprecated` and the API stays until `2.0.0-RC2` in case some only update after each major RC. `2.0.0-RC2-u1` however can remove those deprecated APIs

This allows dependent libraries to walk with us and not revert back to snapshots of 2.x.

vanniktech:
I also actually wanted to propose this since there have been a few things that prevented me from trying it out. 👍

JakeWharton:
Do you think another intermediate release will be needed between this hypothetical one and the next scheduled RC?

akarnokd:
There is an annoying bug in `Single.subscribeOn` I just fixed, many other components didn't work properly until I added the unit tests and fixed them as well. Naming corrections...

zsxwing:
IMO, RC versions are experimental and I don't expect people use a RC version in production. Why not just release RC2 now? 

Moreover, there is no guarantee for APIs in a RC version, so `Deprecated` is not necessary as well.

JakeWharton:
(Whoops got distracted...)

Yeah that was the point that I was leading to. Obviously I don't think we want RCs every week since the library should be stabilizing over time, but I don't think doing another one a week later to fix the initial flurry of problems is a bad thing.

davidmoten:
Make as many RC candidates as you like as often as you want. My use of RxJava2 in production will be  more driven by awareness of the rate of discovery of significant problems than what label the artifact has.

vanniktech:
So will there be something released before 23/09/2016?

akarnokd:
I you don't mind the half-ready `Maybe` then I can release this today.

vanniktech:
Nope I don't mind at all. The `subscribeOn` bug is biting me.

akarnokd:
Sure. Closing via #4476

PR #4458, opened 2016-09-01
2.x: rename Observable and Single #doOnCancel to #doOnDispose

Mauin:
#4456

Completable already used .doOnDispose() however I renamed the arguments to match Observable and Single

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1579,13 +1579,13 @@ public abstract class Single<T> implements SingleSource<T> {
      * <dt><b>Scheduler:</b></dt>
      * <dd>{@code doOnSubscribe} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
-     * @param onCancel the runnable called when the subscription is cancelled (disposed)
+     * @param onDispose the runnable called when the subscription is disposed
      * @return the new Single instance
      * @since 2.0
      */
-    public final Single<T> doOnCancel(final Action onCancel) {
-        ObjectHelper.requireNonNull(onCancel, "onCancel is null");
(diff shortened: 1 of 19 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -1579,13 +1579,13 @@
...
      * </dl>
-     * @param onCancel the runnable called when the subscription is cancelled (disposed)
+     * @param onDispose the runnable called when the subscription is disposed
      * @return the new Single instance
      * @since 2.0
      */
-    public final Single<T> doOnCancel(final Action onCancel) {
-        ObjectHelper.requireNonNull(onCancel, "onCancel is null");
-        return RxJavaPlugins.onAssembly(new SingleDoOnCancel<T>(this, onCancel));
+    public final Single<T> doOnDispose(final Action onDispose) {
```
vanniktech, line 1586: should not we deprecate `doOnCancel` and then remove it with RC3?

akarnokd:
The diff for observable doesn't show but the offline-comparison only lists 30-40 lines changes. Did you run some whitespace cleanup?

src/main/java/io/reactivex/Single.java
akarnokd: I think the discussion #4451 came to a consensus that no need for going soft but delete what's no longer needed before releasing RC2,  hence no deprecation.

Mauin:
@akarnokd Whoops, you're right. Sorry. Auto formatter seems to have changed some whitespaces. I'll revert that.

src/main/java/io/reactivex/Single.java
vanniktech: Alright works for me too

akarnokd:
:+1:

pwittchen:
This update is missing in release notes on GitHub. Please, remember to mention such changes in the future.

PR #4461, opened 2016-09-01
1.x: Single add doOnEach

### LINKED ISSUE #4457: 1.x: Single and Completable are missing doOnEach
Both are missing that method. Is this one wanted?

vanniktech:
Really not that happy with `onNotification.call(Notification.<T>createOnNext(t));` do you guys have any other way of doing this? There's no way of creating a Notification that has the `onCompleted` plus a value. A new one could be introduced there though. Also the `doOnEachSuccess` test feels clunky.

Also why does Single when using the `do` methods does the job by using an Observable? And later converting it back to a Single again. Is it due to the way Single was implemented in 1.x? With Completable there is no converting of back and forth needed.

Javadoc will follow once we sorted out the few nits here.

Fixes #4457

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -2409,24 +2409,35 @@ public class Single<T> {
      */
     @Experimental
     public final Single<T> doOnError(final Action1<Throwable> onError) {
-        Observer<T> observer = new Observer<T>() {
-            @Override
-            public void onCompleted() {
-                // deliberately ignored
-            }
+        if (onError == null) {
+            throw new IllegalArgumentException("onError is null");
+        }
 
+        return Single.create(new SingleDoOnEvent<T>(this, Actions.empty(), new Action1<Throwable>() {
             @Override
-            public void onError(Throwable e) {
-                onError.call(e);
+            public void call(final Throwable throwable) {
+                onError.call(throwable);
             }
+        }));
+    }
 
+    @Experimental
+    public final Single<T> doOnEach(final Action1<Notification<? extends T>> onNotification) {
+        if (onNotification == null) {
+            throw new IllegalArgumentException("onNotification is null");
+        }
+
+        return Single.create(new SingleDoOnEvent<T>(this, new Action1<T>() {
             @Override
-            public void onNext(T t) {
-                // deliberately ignored
+            public void call(final T t) {
+                onNotification.call(Notification.<T>createOnNext(t));
             }
-        };
-
-        return Observable.create(new OnSubscribeDoOnEach<T>(this.toObservable(), observer)).toSingle();
+        }, new Action1<Throwable>() {
+            @Override
+            public void call(final Throwable throwable) {
+                onNotification.call(Notification.<T>createOnError(throwable));
+            }
+        }));
     }
     
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/rx/Single.java
```diff
@@ -2428,6 +2428,32 @@ public void onNext(T t) {
...
         return Observable.create(new OnSubscribeDoOnEach<T>(this.toObservable(), observer)).toSingle();
     }
+
+    @Experimental
+    public final Single<T> doOnEach(final Action1<Notification<? extends T>> onNotification) {
+        if (onNotification == null) {
+            throw new IllegalArgumentException("onNotification is null");
+        }
+
+        Observer<T> observer = new Observer<T>() {
```
akarnokd: Why not natively subscribe with a subscriber that forwards to the action?

src/main/java/rx/Single.java
```diff
@@ -2428,6 +2428,32 @@ public void onNext(T t) {
...
+                onNotification.call(Notification.<T>createOnError(e));
+            }
+
+            @Override
+            public void onNext(T t) {
+                onNotification.call(Notification.<T>createOnNext(t));
+            }
+        };
+
+        return Observable.create(new OnSubscribeDoOnEach<T>(this.toObservable(), observer)).toSingle();
```
akarnokd: This conversion is quite a waste.

src/main/java/rx/Single.java
vanniktech: I just looked at `doOnError` & `doOnSuccess`. If so those should be changed too, right?

akarnokd: Yes, please.

vanniktech:
I'll continue this on Monday and make all suggested changes

vanniktech:
[ExceptionsTest#testOnErrorExceptionIsThrownFromSingleDoOnSuccess](https://github.com/ReactiveX/RxJava/blob/11343ae031feb01d3d08ba5cd0e5f275ed798e68/src/test/java/rx/exceptions/ExceptionsTest.java#L288) fails although I'm not quite sure what the purpose of that test is. Also it seems a bit off to me. Can you give some insights there?

akarnokd:
Looks like removing the toObservable.toSingle no longer wraps the subscriber into a SafeSubscriber and the test just crashes with a different exception.

vanniktech:
No exception is being thrown at all.

akarnokd:
Can you single step the test? In the orginal DoOnEach, it has throwifFatal an throwOrReport you missed btw.

src/main/java/rx/Single.java
```diff
@@ -2409,24 +2409,35 @@ public final Completable toCompletable() {
...
             @Override
-            public void onCompleted() {
-                // deliberately ignored
+            public void call(final Throwable throwable) {
+                onError.call(throwable);
             }
+        }));
+    }
 
+    @Experimental
```
akarnokd, line 2440: javadoc?

src/main/java/rx/Single.java
```diff
@@ -2406,27 +2406,54 @@ public final Completable toCompletable() {
...
+            public void call(final Throwable throwable) {
+                onError.call(throwable);
             }
+        }));
+    }
 
+    /**
+     * Modifies the source {@link Single} so that it invokes an action when it calls {@code onSuccess} or {@code onError}.
+     * <p>
+     * <img width="640" height="310" src="https://raw.githubusercontent.com/wiki/ReactiveX/RxJava/images/rx-operators/doOnEach.png" alt="">
```
vanniktech, line 2428: we might need a new image here since there's no onNext

akarnokd:
:+1:

PR #4466, opened 2016-09-02
1.x: Deprecate Notification.createOnCompleted

### LINKED ISSUE #4462: 1.x: Unused Notification.createOnCompleted(Class<T> type)
While reading the Notification class due to #4457 I noticed [`Notification.createOnCompleted(Class<T> type)`](https://github.com/ReactiveX/RxJava/blob/30da1aa1e38a5ead8965d1784ab8c6df8111d354/src/main/java/rx/Notification.java#L73) which seems to be only used in one [test](https://github.com/ReactiveX/RxJava/blob/30da1aa1e38a5ead8965d1784ab8c6df8111d354/src/test/java/rx/NotificationTest.java#L89) and that test does not really add any value. So I'm wondering since this is in the public API and not used can be it removed or do you see any value in having this method around?

vanniktech:
Fix #4462 

It might just as well shine some light into which method should be used and explain why there are 2

### DIFF
--- a/src/main/java/rx/Notification.java
+++ b/src/main/java/rx/Notification.java
@@ -67,8 +67,10 @@ public final class Notification<T> {
      *
      * @param <T> the actual value type held by the Notification
      * @param type
+     * @deprecated this method does the same as {@link #createOnCompleted()} and does not use the passed in type hence it's useless.
      * @return an {@code OnCompleted} variety of {@code Notification}
      */
+    @Deprecated
     @SuppressWarnings("unchecked")
     public static <T> Notification<T> createOnCompleted(Class<T> type) {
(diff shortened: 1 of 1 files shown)


src/main/java/rx/Notification.java
```diff
@@ -67,8 +67,10 @@
      *
      * @param <T> the actual value type held by the Notification
      * @param type
+     * @deprecated this method does the same as {@link #createOnCompleted()} and does not use the passed in type hence it's useless.
```
JakeWharton, line 70: The argument allows for explicit type inference without needing to specify it on the method call directly when you already have a typed `Class`.

PR #4467, opened 2016-09-02
2.x: add more Maybe operators, fix a few javadoc mistakes

akarnokd:
- Move `verifyPositive` into `ObjectHelper`
- Fix javadoc of `Flowable.subscribe()` and `Observable.subscribe()` mentioning `Subscription` from 1.x
- `Maybe`: add `amb`, `concat`, `concatArray`, `merge`, `mergeArray`, `concatMap`, `subscribe`, `subscribeWith`, `doOnEvent`


src/main/java/io/reactivex/Flowable.java
```diff
@@ -231,7 +231,7 @@ public static int bufferSize() {
     public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<Object[], ? extends R> combiner, int bufferSize) {
         ObjectHelper.requireNonNull(sources, "sources is null");
         ObjectHelper.requireNonNull(combiner, "combiner is null");
-        verifyPositive(bufferSize, "bufferSize");
+        ObjectHelper.verifyPositive(bufferSize, "bufferSize");
```
JakeWharton, line 234: :+1: thanks!

src/main/java/io/reactivex/Maybe.java
```diff
@@ -82,6 +123,161 @@
...
+     * an Iterable sequence. 
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code concat} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <T> the value type
+     * @param sources the Iterable sequence of MaybeSource instances
+     * @return the new Flowable instance
+     */
+    public static <T> Flowable<T> concat(Iterable<? extends MaybeSource<? extends T>> sources) {
```
JakeWharton, line 136: What's the thinking behind using `Flowable` for all the combination operators vs. `Observable`?

JakeWharton:
:+1:

src/main/java/io/reactivex/Maybe.java
akarnokd: Flowable is backpressure enabled by default and you can unbound it in various ways such as converting it to an `Observable`. If it were non-backpressured, you could only turn that into a backpressured one by doing `onBackpressureBuffer` if you don't want to lose data.

src/main/java/io/reactivex/Maybe.java
```diff
@@ -444,6 +857,25 @@
...
+     * Exceptions thrown from the callback will override the event so the downstream receives the
+     * error instead of the original signal.
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code doOnEvent} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param onEvent the callback to call with the terminal event tuple
+     * @return the new Maybe instance
+     */
+    public final Maybe<T> doOnEvent(BiConsumer<? super T, ? super Throwable> onEvent) {
```
JakeWharton, line 875: I'm not sure this is any more ergonomic than `doOnEach` with a `Notification`. Will `Single` and `Completable` be getting `doOnEach` or this `doOnEvent` method?

src/main/java/io/reactivex/Maybe.java
JakeWharton: Ah, yep. That seems obvious in hindsight.

src/main/java/io/reactivex/Maybe.java
akarnokd: Wasn't @vanniktech doing something like this for the others?

JakeWharton: It feels weird to use `null` as such a strong signal where it's forbidden in most (all?) of the API. Not only do I prefer the readability that `Notification`'s methods provide, but it would allow the re-use of side-effecting operations (such as logging) across all stream types since they all would have the same signature.

akarnokd: This is allocation free right now, `Notification.createOnNext` allocates. You can have a utility function that turns a `Consumer<Notification<T>>` into this `BiConsumer` but you can't un-allocate a `Notification` the other way around.

JakeWharton: But by that argument we should remove `doOnEach` from `Observable` and `Flowable` as well in favor of a method with this signature. At least then there'd be parity for peeking across all stream types.

I would much prefer a `doOnEach` in all the stream types and force those who really need zero allocation peeking to just do a `lift` or use the `doOnEach` overload which accepts individual `Consumer`s.

akarnokd: I'm doing coverage improvement right now on Flowable and Observable but after it, I encourage you to post a PR with your suggestions.

vanniktech: Yup I'll also add those methods to RxJava 2 on Monday

PR #4468, opened 2016-09-02
2.x: coverage, bugfixes, 9/03-1

akarnokd:
- Fixed bugs in `Flowable.sequenceEqual`, `Flowable.zip` when errors are delayed, `Flowable.onBackpressureBuffer` when errors are delayed
- Fixed mentions of `NbpX` in some places
- Synchronized unit tests between `Observable` and `Flowable`

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -17,6 +17,7 @@ import java.util.concurrent.*;
 
 import org.reactivestreams.*;
 
+import io.reactivex.Observable;
 import io.reactivex.annotations.*;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.Exceptions;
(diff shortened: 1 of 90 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -2003,16 +2080,9 @@ public static int bufferSize() {
...
-            initialDelay = 0L;
-        }
-        if (period < 0) {
-            period = 0L;
-        }
         ObjectHelper.requireNonNull(unit, "unit is null");
         ObjectHelper.requireNonNull(scheduler, "scheduler is null");
-
-        return RxJavaPlugins.onAssembly(new FlowableInterval(initialDelay, period, unit, scheduler));
+        return RxJavaPlugins.onAssembly(new FlowableInterval(Math.max(0L, initialDelay), Math.max(0L, period), unit, scheduler));
```
JakeWharton, line 2085: Why massage invalid values here when in other methods we throw IAE?

akarnokd: Because ScheduledExecutorService is forgiving in this regard.

JakeWharton:
👍

PR #4469, opened 2016-09-03
2.x: coverage and fixes 9/03-2

akarnokd:
- enable fusion with `Observable.map`
- remove impossible branches in some classes
- add more tests to cover tool classes
- fix `Observable.range()` entering regular emission if fusion-drain is interrupted
- fix `FullArbiter` and `ObservableFullArbiter` not cancelling/disposing the incoming Subscription/Disposable in some cases


JakeWharton:
👍

PR #4482, opened 2016-09-06
2.x: Fix Generic issue in Single

vanniktech:


src/main/java/io/reactivex/Single.java
```diff
@@ -874,7 +874,7 @@
      * @return the new Single instance
      * @since 2.0
      */
-    public static <T, R> Single<R> zip(final Iterable<? extends SingleSource<? extends T>> sources, Function<? super Object[], ? extends R> zipper) {
+    public static <T, R> Single<? extends R> zip(final Iterable<? extends SingleSource<? extends T>> sources, Function<? super Object[], ? extends R> zipper) {
```
akarnokd, line 877: We never return `? extends X` as it adds burden to the receiver. Why did you do this?

vanniktech: Otherwise I get a compile error in IntelliJ

akarnokd: Could be a bug with IntelliJ as javac and Eclipse was fine with the original.

vanniktech: Yeah I also have no problems when running `./gradlew test`

akarnokd: Are you using the latest IntelliJ? A possible workaround could be that you put in `return (Single<R>)...` and then suppress the invalid cast.

vanniktech: Not quite the latest. I'm running Android Studio beta2. So should I change this PR to cast the value and suppress the warning?

akarnokd: If this error prevents you from working further on RxJava then do the changes.

vanniktech: Nah I can live with terminal for executing tests

PR #4483, opened 2016-09-06
2.x: Reuse SingleDoOnEvent

vanniktech:


src/main/java/io/reactivex/Single.java
```diff
@@ -1588,7 +1588,7 @@
      */
     public final Single<T> doOnSuccess(final Consumer<? super T> onSuccess) {
         ObjectHelper.requireNonNull(onSuccess, "onSuccess is null");
-        return RxJavaPlugins.onAssembly(new SingleDoOnSuccess<T>(this, onSuccess));
+        return RxJavaPlugins.onAssembly(new SingleDoOnEvent<T>(this, onSuccess, Functions.emptyConsumer()));
```
akarnokd, line 1591: It was inlined nicely, this now allocates 2 consumers.

vanniktech: True I just thought of reusing it since that's also the way it is done in Completable

akarnokd: Single and Completable are practically a year old and although re-architected, they still run with extra allocations. If I can get there eventually, I'll tidy them up.

vanniktech: Alright cool then I'll close this and help you out with avoiding the extra costs of allocation

PR #4484, opened 2016-09-06
2.x: Flowable.reduce() to return Single, macro fusion

akarnokd:
This PR adds the baseline for changing `Flowable` and `Observable` operators to return `Single`, `Completable` or `Maybe` but in case the user wants to go back to the original type via `toFlowable` or `toObservable` we switch to a natively `Flowable`/`Observable` operator (macro-fusion).

Benchmark comparison (i7 4790, Windows 7 x64, Java 8u102):

image

No changes were applied to `Observable` yet hence the roughly same result there. `Flowable.reduce()` has been updated plus there is a new `FlowableReduce` instead of `scan().last()`.

### DIFF
--- a/src/main/java/io/reactivex/internal/fuseable/FuseToObservable.java
+++ b/src/main/java/io/reactivex/internal/fuseable/FuseToObservable.java
@@ -21,11 +21,11 @@ import io.reactivex.Observable;
  * for toObservable again:
  * <pre>
  * Single&lt;Integer> single = Observable.range(1, 10).reduce((a, b) -> a + b);
- * Observable&lt;Integer> flowable = single.toObservable();
+ * Observable&lt;Integer> observable = single.toObservable();
  * </pre>
  * 
  * The {@code Single.toObservable()} will check for this interface and call the {@link #fuseToObservable()}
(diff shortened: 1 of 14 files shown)


JakeWharton:
:+1:

src/main/java/io/reactivex/internal/fuseable/FuseToObservable.java
```diff
@@ -21,11 +21,11 @@
  * for toObservable again:
  * <pre>
  * Single&lt;Integer> single = Observable.range(1, 10).reduce((a, b) -> a + b);
- * Observable&lt;Integer> flowable = single.toObservable();
+ * Observable&lt;Integer> observable = single.toObservable();
  * </pre>
  * 
  * The {@code Single.toObservable()} will check for this interface and call the {@link #fuseToObservable()}
- * to return a Flowable which could be the Flowable-specific implementation of reduce(BiFunction).
+ * to return a Observable which could be the Observable-specific implementation of reduce(BiFunction).
```
vanniktech: nit: an Observable

akarnokd: Thanks, fixing it here and all the ~400 errors in `Observable` itself.

PR #4486, opened 2016-09-06
2.x: Completable enhance doOnEvent to reduce allocations

vanniktech:


src/main/java/io/reactivex/Completable.java
```diff
@@ -943,10 +943,9 @@ public final Completable delay(final long delay, final TimeUnit unit, final Sche
      * @throws NullPointerException if onComplete is null
      */
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Completable doOnComplete(Action onComplete) {
-        return doOnLifecycle(Functions.emptyConsumer(), Functions.emptyConsumer(),
-                onComplete, Functions.EMPTY_ACTION,
-                Functions.EMPTY_ACTION, Functions.EMPTY_ACTION);
+    public final Completable doOnComplete(final Action onComplete) {
```
akarnokd: Why? There was no extra allocations here as the others are just static instances.

src/main/java/io/reactivex/Completable.java
```diff
@@ -998,51 +995,10 @@ public final Completable doOnError(Consumer<? super Throwable> onError) {
...
-            public void accept(final Throwable throwable) throws Exception {
-                onEvent.accept(throwable);
-            }
-        }, new Action() {
-            @Override
-            public void run() throws Exception {
-                onEvent.accept(null);
-            }
-        }, Functions.EMPTY_ACTION, Functions.EMPTY_ACTION, Functions.EMPTY_ACTION);
+        return RxJavaPlugins.onAssembly(new CompletableDoOnEvent(this, onEvent));
```
akarnokd, line 1001: You can keep this but the other changes are not okay.

src/main/java/io/reactivex/Completable.java
vanniktech: How is this any different from [this](https://github.com/ReactiveX/RxJava/pull/4483/files#diff-68a5181068fafd25f58f5c4600ac7bcaR1591)?

akarnokd: Looks like that comment was misaligned. You changed `SingleDoOnEvent` to have 2 arguments and then had 2 allocations to work with `BiConsumer`.

vanniktech:
@akarnokd this is the way it is desired now, correct?

akarnokd:
Great! :+1:

PR #4490, opened 2016-09-06
2.x: rename emitters #isCancelled to #isDisposed

Mauin:
Renames Observable/Single/CompletableEmitter#isCancelled to #isDisposed.

This now causes a bit of a weird situation where most (not all) Emitter classes implement `Disposable` as well, so both interfaces expose the same `#isDisposed` method. Before the `#isCancelled` methods were basically just calling `#isDisposed` or had the same implementation. 

Suggestions on how to improve this?

### DIFF
--- a/src/main/java/io/reactivex/CompletableEmitter.java
+++ b/src/main/java/io/reactivex/CompletableEmitter.java
@@ -52,8 +52,8 @@ public interface CompletableEmitter {
     void setCancellable(Cancellable c);
     
     /**
-     * Returns true if the downstream cancelled the sequence.
-     * @return true if the downstream cancelled the sequence
+     * Returns true if the downstream disposed the sequence.
+     * @return true if the downstream disposed the sequence
      */
-    boolean isCancelled();
+    boolean isDisposed();
 }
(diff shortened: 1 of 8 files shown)


JakeWharton:
I'm tempted to remove `setDisposable` from the emitter and keep this method named `isCanceled()`. If you want to set a disposable why not use `unsafeCreate` and `ObservableSource` directly?

akarnokd:
@JakeWharton these are the safe versions that also manages a resource for you without the need to worry about (and lose) backpressure or cancellation in general from `unsafeCreate`.

akarnokd:
Otherwise :+1:

JakeWharton:
I'm not sure how that is an argument against what I said. Having both `setCancelable` and `setDisposable` in the API is ugly and redundant.

akarnokd:
If you believe users don't really need the Disposable overload then I'm open to remove them. Also consider dropping Disposables.from variants not in use and adding Cancellables.from.

Mauin:
As for `setCancelable` and `setDisposable` I agree with @JakeWharton. Would probably be best to somehow merge them or get rid of one to clean the API. But that would be a separate issue/PR, I guess?

akarnokd:
Yes, let's have a separate PR for that.

PR #4491, opened 2016-09-06
2.x: reduce blockingX overhead, move internal observers to common pkg

akarnokd:
This PR compacts the `blockingX` operators by extending `CountDownLatch` and implementing `SingleObserver`, `CompletableObserver` and `MaybeObserver` with the necessary blocking await mechanics.

Benchmark comparison (i7 4770, Windows 7 x64, Java 8u102):

image

In addition, the internal package `subscribers.single` and `subscribers.completable` has been removed and the classes moved into `io.reactivex.internal.observers` package. The old blocking enum helpers were removed as well.

The `BlockingSingleSubscriber` and `BlockingSingleObserver` abstract classes have been renamed to `BlockingBaseX` to avoid name confusion (they have nothing to do with the `Single` type).


JakeWharton:
:+1:

PR #4497, opened 2016-09-07
2.x: Add Checkstyle for Trailing Whitespace

vanniktech:
This will add checkstyle though it should not be executed together with Travis and hence does not fail the PR if there are. Also running `./gradlew checkstyleMain` gives a list of places where trailing whitespace exists.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
   repositories { jcenter() }
-  dependencies { 
-    classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:4.0.0' 
+  dependencies {
+    classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:4.0.0'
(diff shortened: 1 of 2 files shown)


JakeWharton:
> it should not be executed together with Travis and hence does not fail the PR

Seems pointless to add then.

vanniktech:
Yeah having the same opinion but David doesn't want it to fail the PR.

JakeWharton:
Then why fail local builds?

akarnokd:
For now, just run it with `./gradlew checkstyleMain` sometimes.

akarnokd:
Yet it does fail it https://travis-ci.org/ReactiveX/RxJava/builds/158395672#L205

vanniktech:
Alright my bad #4505

I thought the checkstyle tasks would not be executed automatically though `build` executes `check` and `checkstyle` task hooked itself into `check` task.

Issue #4123, opened 2016-06-25
Trailing whitespace


vanniktech:
Often when submitting a PR the trailing whitespace gets automatically deleted because Sublime / IntelliJ was configured that way.

What do you guys think about passively trying to remove those?

In 1.x version PR's containing removal of trailing whitespaces could be accepted rather than letting the PR OP removing it.
In 2.x version there could be one bulk PR that deletes all of those spaces. If this is not wanted it could go also along with every PR.

Personally I'd love to see no file having trailing whitespace. If this is not wanted, also not a problem.

akarnokd:
My grief is when unrelated content changes due to these settings. Otherwise I don't mind if a PR contains extra spaces.

JakeWharton:
I spend a non-trivial amount of time doing `git add -p` and editing diffs so reduce whitespace changes. I've even resorted to turning off "trim trailing whitespace" in my editor, which is something I never thought I'd do (and why isn't trimming the default in _every_ editor?).

vanniktech:
@JakeWharton I've been using the solution proposed [here](http://stackoverflow.com/questions/3515597/add-only-non-whitespace-changes) to not waste my time dividing and adding individual chunks.

`git diff -U0 -w --no-color | git apply --cached --ignore-whitespace --unidiff-zero -` did the job for me. Only stages chunks that contain valid changes and ignores white space only changes.

Nonetheless I feel like trimming trailing whitespace should be the default option and no one would need to go through those magic git tips to create a PR here.

JakeWharton:
Looks like it didn't work on your PR though. I already need to use -p to
"fix" the star imports as well.

vanniktech:
True I first created the PR containing the whitespace changes. After that I fixed it and force pushed it. All whitespaces are preseverd now.

Yes the star imports are another problem ...

akarnokd:
I don't mind whitespaces or if you removed them in PRs anyomore but I won't bother removing them in my PRs.

How about everybody posts with what whitespace settings he/she sees adequate?

vanniktech:
How about configuring your IDE so that it'll strip them out automatically?

vanniktech:
@akarnokd are you willing to merge one PR that removes all of the trailing white space and nothing else in 2.x branch? After that a simple Checkstyle rule could be added to enforce no trailing whitespace anymore. There are also options available in Eclipse to [automate this](http://stackoverflow.com/a/2618521/1979703).

[...]

PR #4500, opened 2016-09-08
1.x: Remove trailing whitespace

vanniktech:
Dunno whether this is wanted or not in 1.x. If not just close this PR.


akarnokd:
I haven't seen IDEs adding them back but only remove them so let's have this to avoid future unneccessary removals from unrelated changes in files.

vanniktech:
Alright sounds good and then #4123 can be closed

Issue #4389, opened 2016-08-21
2.x: Ability/API to set thread priority of shared schedulers


JakeWharton:
In 1.x I implemented this with factory methods that accept user-supplied `ThreadFactory` instances. We could mimic this in 2.x, or we could add per-scheduler properties for the thread priority. A nice advantage of the property-based approach is that we can do platform detection and automatically lower the default when running on Android (which was the initial motivation for the feature).

akarnokd:
PR welcome.

akarnokd:
See #4503.

akarnokd:
Closing via #4503.

PR #4503, opened 2016-09-08
2.x: add system properties to adjust thread priorities of Schedulers

akarnokd:
This PR adds the ability to specify the default thread priorities of `computation()`, `io()`, `newThread()` and `single()` schedulers via system properties. I've added the property names to the `Schedulers` javadoc.

Related: #4389.


JakeWharton:
Thanks! :+1:

PR #4507, opened 2016-09-08
2.x: add Reactive-Streams TCK infrastructure and couple of tests

akarnokd:
This PR sets up the build to run TestNG tests as well and adds the Reactive-Streams TCK to allow validating our operators.

Included tests are: `just`, `fromArray`, `fromIterable`, `concat`, `merge`.

Note that the TCK tests for invalid `request()` amount which, according to the spec, should be reported to the running `Subscriber` via `onError`. Unfortunately, this is a very expensive requirement (requiring half-serialization all the time) and we are not going to support it in RxJava 2. Negative requests are bugs in operators and should be fixed, we only provide a notification about such situations in the `RxJavaPlugins.onError` for those who want to check for such errors. 

Therefore, the `FlowableTCK` has been added with does the required behavior by the TCK and can be used as `FlowableTCK.wrap(flowable)` when returning from the TCK's `createPublisher(long)`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,9 @@ dependencies {
 
     perfCompile 'org.openjdk.jmh:jmh-core:1.13'
     perfCompile 'org.openjdk.jmh:jmh-generator-annprocess:1.13'
+
+    testCompile 'org.reactivestreams:reactive-streams-tck:1.0.0'
+    testCompile group: 'org.testng', name: 'testng', version: '6.9.10'
 }
 
 javadoc {
@@ -71,6 +74,12 @@ test {
         maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
     }
 }
+
+task testng(type: Test) { 
+     useTestNG() 
+} 
+
+check.dependsOn testng 
 
 jacoco {
     toolVersion = '0.7.7.201606060606' // See http://www.eclemma.org/jacoco/.
--- a/src/test/java/io/reactivex/tck/FlowableTck.java
+++ b/src/test/java/io/reactivex/tck/FlowableTck.java
@@ -1,0 +1,113 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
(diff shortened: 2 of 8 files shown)


build.gradle
```diff
@@ -31,6 +31,9 @@ dependencies {
 
     perfCompile 'org.openjdk.jmh:jmh-core:1.13'
     perfCompile 'org.openjdk.jmh:jmh-generator-annprocess:1.13'
+
+    testCompile 'org.reactivestreams:reactive-streams-tck:1.0.0'
+    testCompile group: 'org.testng', name: 'testng', version: '6.9.10'
```
vanniktech, line 36: any reason for not using 'org.testng:testng:6.9.10' ?

akarnokd: I copied that line from the first Google hit about maven testng.

src/test/java/io/reactivex/tck/FlowableTck.java
```diff
@@ -0,0 +1,113 @@
...
+            this.actual = actual;
+            this.error = new AtomicThrowable();
+        }
+
+
+        @Override
+        public void request(long n) {
+            if (n <= 0) {
+                s.cancel();
+                onError(new IllegalArgumentException("§3.9 violated: positive request amount required but it was " + n));
```
JakeWharton, line 78: Isn't this cheating? You're modifying behavior in the compliance tests but the runtime actually behaves differently.

akarnokd: Yes and no. You see, Akka-Streams is not native Reactive-Streams but they have a wrapper that exposes it as Reactive-Streams source which passes the TCK (that was mostly influenced by Akka-Streams implementation strategies anyway).

The spec itself has "bugs" and over-restrictive language in some cases. @smaldini occasionally tries to battle this out with the TCK maintainers but everything is moving really slow over there. So we decided it is not worth following the spec to the letter in this tiny aspect and suffer the massive performance penalty but provide an operator that when needed, ensures this behavior.

I'm totally willing to provide `Flowable.tck()` for anyone who is worried the cross-library composition may not work properly because Akka-Streams occasionally requests -1 (I know Reactor and RxJava doesn't do that because I wrote both essentially).

smaldini, line 78: What about safeSubscribe ?

JakeWharton: Good info, thanks. I don't have a strong opinion on the addition of the operator.

smaldini, line 78: And also this negative error should really be relaxed given the serialisation implication of it

akarnokd: `safeSubscribe` is RxJava specific and not `Publisher` method so other RS libraries won't call it if what they all see is a `Publisher`.

PR #4508, opened 2016-09-08
2.x: Add exception message in SingleFromCallable

vanniktech:


vanniktech:
@akarnokd what do you think about a custom exception that explains why null is not allowed anymore? And if someone goes and wants to emit null that one will be thrown instead of NullPointerException

akarnokd:
You mean like

`The callable returned a null value. Null values are generally not allowed in 2.x operators and sources.`

If you manually change all locations then okay, but don't try to factor this out into some parametric method because that extra amount of conditional code might prevent some JIT optimizations due to dept/code size increase.

PR #4509, opened 2016-09-08
2.x: Adjust some JavaDoc in Emitters

vanniktech:


src/main/java/io/reactivex/FlowableEmitter.java
```diff
@@ -17,7 +17,7 @@
 import io.reactivex.functions.Cancellable;
 
 /**
- * Abstraction over a RxJava Subscriber that allows associating
+ * Abstraction over an RxJava {@link org.reactivestreams.Subscriber} that allows associating
```
akarnokd: Actually it is not RxJava so `Abstraction over an {@link org.reactivestreams.Subscriber}`

src/main/java/io/reactivex/CompletableEmitter.java
```diff
@@ -17,7 +17,7 @@
 import io.reactivex.functions.Cancellable;
 
 /**
- * Abstraction over a RxJava CompletableObserver that allows associating
+ * Abstraction over an RxJava {@link CompletableObserver} that allows associating
```
akarnokd, line 20: You sure these are `an`?

akarnokd, line 20: Looks like it: https://www.englishclub.com/pronunciation/a-an.htm

vanniktech: Yes because it's pronounced differently you can say `an` and it also does sound better. Jake taught me that one too

src/main/java/io/reactivex/FlowableEmitter.java
```diff
@@ -17,7 +17,7 @@
 import io.reactivex.functions.Cancellable;
 
 /**
- * Abstraction over a RxJava Subscriber that allows associating
+ * Abstraction over a {@link org.reactivestreams.Subscriber} that allows associating
```
akarnokd: Then this should be `an`, right?

src/main/java/io/reactivex/FlowableEmitter.java
```diff
@@ -17,7 +17,7 @@
 import io.reactivex.functions.Cancellable;
 
 /**
- * Abstraction over a RxJava Subscriber that allows associating
+ * Abstraction over an Reactive Streams {@link org.reactivestreams.Subscriber} that allows associating
```
akarnokd: I don't think this is pronounced `an R-eactive Streams`.

vanniktech: Ups yeah got confused now

akarnokd:
:+1:

Issue #4512, opened 2016-09-08
[RxJava 2.0.0-RC2] Observable.switchMap doesn't unsubscribe from inner Observable


nsk-mironov:
``` java
public class SwitchMapBug {
  public static void main(String[] args) throws Exception {
    final Disposable disposable = Observable.just(true)
        .subscribeOn(Schedulers.io())
        .switchMap(value -> Observable.timer(1000, TimeUnit.MILLISECONDS)
            .compose(new TraceComposer<>("inner"))
        )
        .compose(new TraceComposer<>("outer"))
        .subscribe();

    Thread.sleep(500);
    disposable.dispose();
    Thread.sleep(2000);
  }

  private static class TraceComposer<T> implements Function<Observable<T>, Observable<T>> {
    private final String prefix;

    private TraceComposer(final String prefix) {
      this.prefix = prefix;
    }

    @Override
    public Observable<T> apply(Observable<T> observable) throws Exception {
      return observable
          .doOnSubscribe(disposable -> System.out.println("[" + prefix + "]" + " " + "doOnSubscribe"))
          .doOnDispose(() -> System.out.println("[" + prefix + "]" + " " + "doOnDispose"))
          .doOnNext(value -> System.out.println("[" + prefix + "]" + " " + "doOnNext" + " " + value))
          .doOnError(error -> System.out.println("[" + prefix + "]" + " " + "doOnError" + " " + error))
          .doOnComplete(() -> System.out.println("[" + prefix + "]" + " " + "doOnComplete"))
          .doOnTerminate(() -> System.out.println("[" + prefix + "]" + " " + "doOnTerminate"))
          .doAfterTerminate(() -> System.out.println("[" + prefix + "]" + " " + "doAfterTerminate"));

    }
  }
}
```

Output:

```
[outer] doOnSubscribe
[inner] doOnSubscribe
[outer] doOnDispose
[inner] doOnNext 0
[inner] doOnComplete
[inner] doOnTerminate
[inner] doAfterTerminate
```

Expected output:

```
[outer] doOnSubscribe
[inner] doOnSubscribe
[inner] doOnDispose
[outer] doOnDispose
```

nsk-mironov:
Looks like `a.cancel()` is missed here:
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java#L179

akarnokd:
[...]

PR #4513, opened 2016-09-08
2.x: fix switchMap bad cancellation

akarnokd:
Both `switchMap` implementations didn't properly cancel the current inner consumer (but instead cancelled the outer 'connection' only).

Reported in #4512.


JakeWharton:
:+1:

Issue #4514, opened 2016-09-08
2.x Should observer be disposed after the source observable has finished emitting items?


VictorAlbertos:
With 1.x this test pass:

``` java
@Test
  public void subscriberShouldBeUnsubscribed() {
    rx.observers.TestSubscriber subscriber = new rx.observers.TestSubscriber();
    rx.Observable.just("").subscribe(subscriber);
    subscriber.awaitTerminalEvent();

    subscriber.assertNoErrors();
    subscriber.assertCompleted();
    subscriber.assertUnsubscribed();
  }
```

But with 2.x this test does not pass:

``` java
  @Test
  public void observerShouldBeDisposed()  {
    TestObserver observer = new TestObserver();
    Observable.just("").subscribe(observer);
    observer.awaitTerminalEvent();

    observer.assertNoErrors();
    observer.assertComplete();
    assertThat(observer.isDisposed(), is(true));
  }
```

I think that dispose is similar to unsubscribe (but maybe I'm wrong). So I was wondering if this is the expected behaviour.

  Thanks.

akarnokd:
It's a small inconsistency. I'll fix that in my next cleanup/coverage PR.

akarnokd:
Closing via https://github.com/ReactiveX/RxJava/commit/1145819b658983807e0cede5ba2c7d5ac1117baa

maksim-m:
@akarnokd, with RxJava 2.0.7 the test still fails:
```java
    @Test
    public void observerShouldBeDisposed()  {
        TestObserver observer = new TestObserver();
        Observable.just("").subscribe(observer);
        observer.awaitTerminalEvent();

        observer.assertNoErrors();
        observer.assertComplete();
        assertTrue(observer.isDisposed());
    }
```

akarnokd:
There was a problem with that change and has been reverted in #4873.

PR #4516, opened 2016-09-08
2.x: fix Single.using, add unit tests and coverage

akarnokd:
I've forgotten to create a branch so the first part of todays coverage work [landed in 2.x direct](https://github.com/ReactiveX/RxJava/commit/1145819b658983807e0cede5ba2c7d5ac1117baa)
- removed unused methods from various classes
- compacted `SpscArrayQueue`
- added null checks to `SpscLinkedArrayQueue.offer()` and `MpscLinkedQueue.offer()`
- fixed `ResourceObserver` not calling `onStart`
- Added direct NotificationLite-emission methods to `AppendOnlyLinkedArrayList`, updated `SerializedObserver` and `SerializedSubscriber` to use it directly instead of the former indirection
- `TestObserver.isDisposed` now reports true if terminal events were received (#4514)
- `ResourceSubscriber` to use `SubscriptionHelper`'s deferred Subscription/requesting management
- added unit tests to cover `DisposableXObserver`s
- `SerializedObserver` and `SerializedSubscriber` now have proper unit tests each

This current PR fixes `Single.using` not properly managing the resource and adds unit tests to verify the behavior along with a few extra coverage improvements.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/single/SingleLift.java
+++ b/src/main/java/io/reactivex/internal/operators/single/SingleLift.java
@@ -14,7 +14,9 @@
 package io.reactivex.internal.operators.single;
 
 import io.reactivex.*;
-import io.reactivex.plugins.RxJavaPlugins;
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.internal.disposables.EmptyDisposable;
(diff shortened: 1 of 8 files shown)


JakeWharton:
:+1:

PR #4518, opened 2016-09-09
2.x: Add missing Resource Observer for Maybe, Completable & Single and adjust some Javadoc

### LINKED ISSUE #4517: 2.x: ResourceSubscriber & ResourceObserver
Why is `ResourceSubscriber` using `CompositeDisposable` and `ResourceObserver` using `ListCompositeDisposable`?

Also the java documentation does not seem to fit of the disposables.
`/** The resource composite, can be null. */`

It's initialized and made final and adding a `null` Disposable will throw a `NullPointerException`.

In addition is there a need for ResourceMaybeObserver and the alike implementation for Single & Completable?

vanniktech:
Fixes #4517

### DIFF
--- a/src/test/java/io/reactivex/observers/ResourceObserverTest.java
+++ b/src/test/java/io/reactivex/observers/ResourceObserverTest.java
@@ -85,7 +85,7 @@ public class ResourceObserverTest {
 
         assertFalse(d.isDisposed());
 
-        ro.cancel();
+        ro.dispose();
 
         assertTrue(ro.isDisposed());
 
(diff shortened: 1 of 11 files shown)


src/test/java/io/reactivex/observers/ResourceObserverTest.java
```diff
@@ -62,10 +65,6 @@ public void onComplete() {
 
             dispose();
         }
-
```
akarnokd, line 65: This was for coverage on the protected cancel method.

akarnokd:
It might be worth considering removing the protected `cancel()` methods as they are duplicates of `dispose`. For `ResourceSubscriber` I don't know because it works with `Subscription.cancel()` but also is a `Disposable` so `dispose()` is there as well.

vanniktech:
So should I remove the `cancel()` method on all `Resource*Observer`?

akarnokd:
@JakeWharton your opinion on the removal?

JakeWharton:
I'm supportive of removing `cancel()`. I much prefer when there's only one way to do something!

akarnokd:
:+1:

PR #4525, opened 2016-09-10
2.x: Fix Generics T[] in Zip & CombineLatest

### LINKED ISSUE #4524: RxJava2 `Observable.zip` taking `Iterable` and a `zipper` is broken
http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#zip-java.lang.Iterable-io.reactivex.functions.Function-

The `zipper` signature is `Function<? super T[]>, ? extends R> zipper`, where use of `T[]` breaks it. Just invoke it with non `Object` T, and a `ClassCastException` is thrown.

I checked with `Single` and `Flowable`, they have different signatures for the comparable `zip` operators. They use `Object[]` where `T[]` is used for `Observable`, which are also consistent with the 1.x zip operator.

I think this is just a bug in the signature.

vanniktech:
Fixes #4524

Test for combineLatest fill follow

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -163,7 +163,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     @BackpressureSupport(BackpressureKind.FULL)
-    public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<? super T[], ? extends R> combiner) {
+    public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<? super Object[], ? extends R> combiner) {
         return combineLatest(sources, combiner, bufferSize());
     }
 
@@ -194,7 +194,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     @BackpressureSupport(BackpressureKind.FULL)
-    public static <T, R> Flowable<R> combineLatest(Function<? super T[], ? extends R> combiner, Publisher<? extends T>... sources) {
+    public static <T, R> Flowable<R> combineLatest(Function<? super Object[], ? extends R> combiner, Publisher<? extends T>... sources) {
         return combineLatest(sources, combiner, bufferSize());
     }
 
@@ -227,7 +227,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     @BackpressureSupport(BackpressureKind.FULL)
-    public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<? super T[], ? extends R> combiner, int bufferSize) {
+    public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<? super Object[], ? extends R> combiner, int bufferSize) {
         ObjectHelper.requireNonNull(sources, "sources is null");
         if (sources.length == 0) {
             return empty();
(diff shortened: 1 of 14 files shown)


src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -1060,4 +1060,19 @@ public void toObservableRange() {
     public void toObservableError() {
         Flowable.error(new TestException()).toObservable().test().assertFailure(TestException.class);
     }
+
+    @Test
+    public void zipIterableObject() {
+        final List<Flowable<Integer>> flowables = Arrays.asList(Flowable.just(1, 2, 3), Flowable.just(1, 2, 3));
+        Flowable.zip(flowables, new Function<Object[], Object>() {
```
akarnokd, line 1067: Yeah, this always passes because it is compatible with both `? super T[]` and `? super Object[]`. I have no idea how to test with `Function<Integer[], Object>` because it doesn't even compile with the latter signature. Leave this as is.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -194,7 +194,7 @@ public static int bufferSize() {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     @BackpressureSupport(BackpressureKind.FULL)
-    public static <T, R> Flowable<R> combineLatest(Function<? super T[], ? extends R> combiner, Publisher<? extends T>... sources) {
+    public static <T, R> Flowable<R> combineLatest(Function<? super Object[], ? extends R> combiner, Publisher<? extends T>... sources) {
```
vanniktech, line 207: this one does not seem necessary to be honest. It just has switched arguments with the one above. Or do I oversee anything?

src/main/java/io/reactivex/Flowable.java
```diff
@@ -163,7 +163,7 @@ public static int bufferSize() {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     @BackpressureSupport(BackpressureKind.FULL)
-    public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<? super T[], ? extends R> combiner) {
+    public static <T, R> Flowable<R> combineLatest(Publisher<? extends T>[] sources, Function<? super Object[], ? extends R> combiner) {
```
akarnokd, line 171: It might be worth adding a sentence to the javadocs since people tend to ask why this isn't `? super T[]`

```
<p>
Note on method signature: since Java doesn't allow creating a generic array with {@code new T[]}, the 
implementation of this operator has to create an {@code Object[]} instead. Unfortunately, a 
{@code Function<Integer[], R>} passed to the method would trigger a {@code ClassCastException}.
```

If you feel so, you could add these two sentences after the first sentence of all the affected methods and before the marble diagram (as people forget to scroll down below the diagram).

src/test/java/io/reactivex/flowable/FlowableTests.java
vanniktech: Yup `Function<Integer[], Object>` didn't work for me so I didn't add a test case.

src/main/java/io/reactivex/Flowable.java
akarnokd: This and other overloads of operators let's you use varargs to specify any number of sources at the expense that varargs has to be the last argument.

src/main/java/io/reactivex/Flowable.java
vanniktech: Will do

src/main/java/io/reactivex/Flowable.java
vanniktech: Ups right didn't see those little dots there too many overloads ...

src/test/java/io/reactivex/maybe/MaybeTest.java
```diff
@@ -2858,4 +2858,18 @@ public void ambWith2SignalsSuccess() {
         ts.assertResult(2);
     }
 
+    @Test
+    public void zipIterableObject() {
```
vanniktech: maybe has still no combineLatest operator hence no test for it

vanniktech:
This should be it now

akarnokd:
Single has exactly 1 and Maybe at most 1 item, so there is only 1 latest that can happen thus there is no difference between combineLatest and zip for these sources.

akarnokd:
:+1:

vanniktech:
Yup I immediately noticed it after I wrote my comment :D that's also why I deleted it

PR #4529, opened 2016-09-12
2.x: Fix MaybeTimber by using scheduler and unit

vanniktech:
Also one question why does `CompletableTimer` check `if (!sd.isDisposed()) {` before replacing but SingleTimer does not? I adjusted MaybeTimber to SingleTimer so I didn't add the check there either.

However I feel like this should be consistent, can you elaborate which one is wanted?

### DIFF
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -1238,7 +1238,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     }
 
     /**
-     * Returns a Flowable that emits one item after a specified delay on a specified Scheduler.
+     * Returns a Maybe that emits one item after a specified delay on a specified Scheduler.
      * <p>
      * <img width="640" height="200" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/timer.s.png" alt="">
      * <dl>
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -1238,7 +1238,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     }
 
     /**
-     * Returns a Maybe that emits one item after a specified delay on a specified Scheduler.
+     * Returns a Flowable that emits one item after a specified delay on a specified Scheduler.
      * <p>
      * <img width="640" height="200" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/timer.s.png" alt="">
      * <dl>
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/Maybe.java
```diff
@@ -1238,7 +1238,7 @@
     }
 
     /**
-     * Returns a Flowable that emits one item after a specified delay on a specified Scheduler.
+     * Returns a Maybe that emits one item after a specified delay on a specified Scheduler.
```
akarnokd: I did a cleanup of bunch of these so you may need to rebase this after.

src/main/java/io/reactivex/internal/operators/maybe/MaybeTimer.java
```diff
@@ -38,40 +38,16 @@ public MaybeTimer(long delay, TimeUnit unit, Scheduler scheduler) {
...
-        @Override
-        public boolean isDisposed() {
-            return DisposableHelper.isDisposed(get());
-        }
+        observer.onSubscribe(sd);
 
-        void setFuture(Disposable d) {
-            DisposableHelper.replace(this, d);
-        }
+        sd.replace(scheduler.scheduleDirect(new Runnable() {
```
akarnokd: It was okay with the named class, we should be moving away from anonymous inner classes.

src/main/java/io/reactivex/Maybe.java
vanniktech: This PR was created on the current 2.x so after you posted that earlier PR in any case if you cleaned this one up already I could also just revert this one change

src/main/java/io/reactivex/internal/operators/maybe/MaybeTimer.java
vanniktech: So should I change it back and adapt SingleTimer & CompletableTimer too?

akarnokd: Yes

vanniktech: Alright and what about the `isDisposed` check in CompletableTimber?

akarnokd: Drop it.

src/main/java/io/reactivex/internal/operators/completable/CompletableTimer.java
```diff
@@ -14,10 +14,17 @@
 package io.reactivex.internal.operators.completable;
 
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
-import io.reactivex.*;
-import io.reactivex.internal.disposables.SequentialDisposable;
+import io.reactivex.Completable;
```
akarnokd: Please update your IDE settings to not unroll \* imports.

PR #4533, opened 2016-09-12
2.x: more cleanup and fixed checker, Maybe.takeUntil

akarnokd:
- Fix missed javadoc mistakes, fix the checker that hid those mistakes in instance methods' javadoc due to a bug
- add `Maybe.takeUntil`


JakeWharton:
:+1:

PR #4537, opened 2016-09-12
2.x: last Maybe operators

akarnokd:
This PR adds the last couple of `Maybe` operators `timeout` and `unsubscribeOn`.

@abersnaze Let me know if I missed an operator or you want some overload.

In the subsequent PRs, I'll change the return types of some classical operators to better indicate the cardinality they have (`Single`, `Completable`).


JakeWharton:
:+1:

PR #4540, opened 2016-09-13
2.x: Add Create test with CancelAble for all types and add more MaybeCreateTest cases

vanniktech:
`setCancellable` was not tested before.


vanniktech:
> Gradle suite > Gradle test > io.reactivex.tck.ObserveOnTckTest. FAILED
>     java.lang.AssertionError: Illegal concurrent access detected (entering critical section)! Thread[RxSingleScheduler-1,5,main] emited onNext(0) signal, before Thread[Test worker,5,main] finished its onSubscribe() signal.

akarnokd:
TCK doesn't seem to like that onSubscribe may trigger emission before it even returns - which is completely legal btw. I'll add a workaround for that in the TCK shortly.

PR #4546, opened 2016-09-13
Add switchIfEmpty(SingleSource) method.

mattinger:
If you have Maybe, and do a switchIfEmpty with a Single,
you are semantically guaranteed to never get an onComplete
event.  Thus the result of the operation should be a Single
rather than a Maybe.  Let's add a convenience method for
doing this for you.

### DIFF
--- a/src/test/java/io/reactivex/maybe/MaybeTest.java
+++ b/src/test/java/io/reactivex/maybe/MaybeTest.java
@@ -2997,4 +2997,15 @@ public class MaybeTest {
             }
         }).test().assertResult(1);
     }
+
+    @Test
+    public void switchIfEmptySingle() {
+        Maybe.empty().switchIfEmpty(Single.just(1)).test().assertResult(1);
+
+        Maybe.just(1).switchIfEmpty(Single.just(2)).test().assertResult(1);
+
+        Throwable throwable = new RuntimeException();
+
+        Maybe.empty().switchIfEmpty(Single.error(throwable)).test().assertError(throwable);
+    }
 }
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -3482,6 +3482,25 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     }
 
     /**
+     * Returns a Single that emits the items emitted by the source Maybe or the items of an alternate
+     * SingleSource if the current Maybe is empty.
+     * <p/>
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code switchIfEmpty} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param other
+     *              the alternate SingleSource to subscribe to if the main does not emit any items
+     * @return  a Single that emits the items emitted by the source Maybe or the items of an
+     *          alternate SingleSource if the source Maybe is empty.
+     */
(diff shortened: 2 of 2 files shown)


JakeWharton:
What about `Observable` or `Flowable` or `Completable`? And what about all the other methods? It's really weird to special case this one combination of method and type.

vanniktech:
Agreed with Jake and even if so shouldn't this get a custom implementation?

src/test/java/io/reactivex/maybe/MaybeTest.java
```diff
@@ -2997,4 +2997,15 @@ public boolean getAsBoolean() throws Exception {
             }
         }).test().assertResult(1);
     }
+
+    @Test
+    public void switchIfEmptySingle() {
+        Maybe.empty().switchIfEmpty(Single.just(1)).test().assertResult(1);
```
akarnokd, line 3003: The aim is to ensure `switchIfEmpty` returns `Single`. This fluent writing works if `switchIfEmpty` returned any other type.

akarnokd:
You could have copied the `MaybeSwitchIfEmpty` to `MaybeSwitchIfEmptySingle`, change the base type to `Single`, implement `HasUpstreamMaybeSource<T>` and `FuseToMaybe<T>` then adjust the rest of the accordingly. [Example](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/maybe/MaybeIsEmptySingle.java) with `isEmpty()`:

mattinger:
I definitely could have, but was going to for the simplest implementation possible.  I can certainly add new classes as described by akarnokd.  However, i'm not going to spend the time unless the project owners are willing to accept the new method into the class.  That's something i would prefer to have answered before spending more time on this.

And to Jake's point, we can certainly add switchIfEmpty for other types, but for me, the point here was switchIfEmpty(Single) has a narrowing effect that is guaranteed.  Something like switchIfEmpty(Observable) doesn't have a specific type it must produce.  It's either a Maybe or an Observable, so you'd be widening to Observable to be able to cover the cases.

JakeWharton:
> Something like switchIfEmpty(Observable) doesn't have a specific type it must produce.

It would have to be `Observable`.

Besides, even if you only focus on narrowing you have tons of other methods on `Maybe` that are not getting these overloads and you have tons of methods on `Observable`, `Flowable`, and `Single` which are not being overloaded for narrower types.

mattinger:
Basically, this is a convenience for me.  We're using this for caching purposes.  We have a custom cache we're using (we don't want to use the build in .cache() method on observable).  So the cache produces a Maybe, but the overall method call for our service produces a Single as such:

``` java
Maybe<T> cacheObservable = ...;
Single<T> loadObservable = ...;

cacheObservable.switchIfEmpty(loadObservable.toMaybe()).toSingle().
```

It s a little ugly, but not too horrible.  I just thought this might be something that other people would need as well, hence the specific method for this case.

JakeWharton:
Yeah I understand why you want the convenience, but I don't think it belongs in RxJava unless we're going to do _all_ of the required overloads. Otherwise we'll have people coming to the issues for the next 2 years asking why there isn't an overload of method `x` on stream type `Y` for stream type `Z`.

akarnokd:
Will you update this PR?

mattinger:
i was under the impression that you were leaning towards not including it, so i didn't do any more work for it.

akarnokd:
Let's skip this for now. Thanks anyway!

PR #4552, opened 2016-09-14
2.x: cleanup of style and other minor things, 9/14-1

akarnokd:
This PR has 781 lines changed across 511 files so GitHub won't display the complete diff. It is advised to check this out and do a diff in your IDE.

Notable changes:
- replace mention of `onCompleted` with `onComplete`
- remove `public` from constructors of package-private classes (some may think the class was intended to be public but usually isn't)
- few minor whitespace addition around operators (so the code doesn't look so cluttered to me when I have view/debug older code or unit tests contributed a while ago)
- add missing `.` from the end of the first sencence in javadocs
- removed empty `/** */` comments above the `serialVersionUID` fields
- use checkstyle tool version 6.19 plus a few more checkstyle rules
- fix ObservableRedo asnyc reentrancy problem by serializing the subject


JakeWharton:
LGTM :+1:

JakeWharton:
Looks like the build OOM'd

akarnokd:
I'll disable the extra checks.

PR #4554, opened 2016-09-14
2.x: make internal naming consistent, refactor some classes 9/14-2

akarnokd:
- Make internal naming of components consistent with the base reactive type
- move `io.reactivex.internal.subscribers.flowable.*` to `io.reactivex.internal.subscribers`
- move `io.reactivex.internal.subscribers.observable.*` to `io.reactivex.internal.observers`


JakeWharton (approved):

abersnaze:
oh god the rebasing.

akarnokd:
@abersnaze I recommend you post one or two operator changes at a time as it is likely they need some extra changes until they match structure I expect (such as in Maybe.isEmpty()). This way, you don't have large interference from cascading fixes in the 2.x branch.

abersnaze:
Its hard because they are so intertwined.

akarnokd:
I've run out of ideas about what to clean up so you should have no interference from me in the following days.

PR #4567, opened 2016-09-20
Add extra Maybe iterable Amb tests for overlapped emissions

rzsombor:
- Asserting that downstream will get the results of the first source that began emission, even if multiple sources have overlapped emissions
- Testing Success and Error paths

### DIFF
--- a/src/test/java/io/reactivex/maybe/MaybeTest.java
+++ b/src/test/java/io/reactivex/maybe/MaybeTest.java
@@ -1702,6 +1702,30 @@ public class MaybeTest {
 
     @SuppressWarnings("unchecked")
     @Test
+    public void ambIterable2SignalsSuccessWithOverlap() {
+        PublishProcessor<Integer> pp1 = PublishProcessor.create();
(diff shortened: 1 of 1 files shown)


rzsombor:
As far as I can see the CI error is coming from a seemingly unrelated test:

```
io.reactivex.internal.operators.maybe.MaybeUnsubscribeOnTest > normal FAILED
java.lang.AssertionError
    at org.junit.Assert.fail(Assert.java:86)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertFalse(Assert.java:64)
    at org.junit.Assert.assertFalse(Assert.java:74)
    at io.reactivex.internal.operators.maybe.MaybeUnsubscribeOnTest.normal(MaybeUnsubscribeOnTest.java:50)
```

I'm a bit confused here - did I really manage to break this test or is this one known to be flaky?

akarnokd:
It was a flaky test: the unsubscription may not reach the processor in time the latch unblocks the main thread and thus there appear to be subscribers there.

:+1: Thanks for the unit tests.

Issue #4566, opened 2016-09-19
Alternative to RxJavaErrorHandler


bobvanderlinden:
I'd like to catch unhandled errors that arise from Observables globally.

The documentation [here](https://github.com/ReactiveX/RxJava/wiki/Plugins) suggests to use `RxJavaPlugins.getInstance().registerErrorHandler`, but both `getInstance` and `RxJavaErrorHandler` are deprecated with no hints as what to use instead.

The documentation on error handling in RxJava [here](https://github.com/ReactiveX/RxJava/wiki/Error-Handling) also has no mention of globally handling errors.

I did find `RxJavaHooks.setOnError`, but I have not found any documentation on this, so I was unsure whether it should be used.

Is this still on-going work or can anyone suggest what to use as an alternative to `registerErrorHandler`?

aaronweihe:
The documentation on the Wiki is probably outdated. If you look at the [java doc](http://reactivex.io/RxJava/javadoc/rx/plugins/RxJavaPlugins.html#getInstance%28%29) of `RxJavaPlugins#getInstance()`, it says 

> Deprecated. use the static methods of RxJavaHooks.

😄

akarnokd:
Hi and thanks for the feedback. Those articles are a bit outdated and will be updated in a few days to better reflect the current status with 1.2.

> I did find RxJavaHooks.setOnError, but I have not found any documentation on this, so I was unsure whether it should be used.

There is not much to that method. By default, undeliverable exceptions are routed to `RxJavaHooks.onError` which then calls an `Action1<Throwable>` you can change via `RxJavaHooks.setOnError`.

Here is the [release](https://github.com/ReactiveX/RxJava/releases/tag/v1.1.7) info and the [PR](https://github.com/ReactiveX/RxJava/pull/4007)

bobvanderlinden:
Thank you both for the feedback. `RxJavaHooks.setOnError` it is! I've made a PR that documents the new alternative (#4569).

akarnokd:
Updated the [Plugins](https://github.com/ReactiveX/RxJava/wiki/Plugins) wiki page. Let me know if it's adequate for you.

bobvanderlinden:
@akarnokd Yes, that makes things a lot more clear. Thank you!

PR #4569, opened 2016-09-20
1.x: mark RxJavaPlugins.registerErrorHandler as deprecated

bobvanderlinden:
This marks `RxJavaPlugins#registerErrorHandler` and `RxJavaPlugins#getErrorHandler` as deprecated and documents `RxJavaHooks#setOnError` and `RxJavaHooks#getOnError` as its alternatives.

Related to #4566

### DIFF
--- a/src/main/java/rx/plugins/RxJavaPlugins.java
+++ b/src/main/java/rx/plugins/RxJavaPlugins.java
@@ -18,6 +18,7 @@ package rx.plugins;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
 import rx.annotations.Experimental;
+import rx.functions.Action1;
 
 /**
  * Registry for plugin implementations that allows global override and handles the retrieval of correct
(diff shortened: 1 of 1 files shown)


akarnokd:
There are legitime uses of the old hooks so I'm not certain they should be deprecated.

bobvanderlinden:
@akarnokd Ah, I gathered from #4566 and from `RxJavaPlugins#getInstance` and `RxJavaErrorHandler#handleError` being deprecated that `RxJavaPlugins#registerErrorHandler` would be deprecated as well.

This PR was mostly for people running into the same problem as me: using `RxJavaPlugins#getInstance` and implementing `RxJavaErrorHandler#handleError` both giving deprecation warnings while it is unclear what to use instead. The deprecation of `RxJavaPlugins#getInstance` does refer to `RxJavaHooks`, but not specifically `setOnError`, like what was suggested in #4566.

bobvanderlinden:
Most purposes of this PR are already solved by updating the documentation. See #4566.

PR #4570, opened 2016-09-20
2.x: last() to return Single

akarnokd:
This PR changes the return type of `last()` to `Single<T>` and updates the relevant locations.

Originally, it was implemented as `takeLast(1).single()` so to reduce impact, all other original uses now have this inlined.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -8877,14 +8877,14 @@ public abstract class Flowable<T> implements Publisher<T> {
      *  <dd>{@code last} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
-     * @return a Flowable that emits the last item from the source Publisher or notifies Subscribers of an
+     * @return a Single that emits the last item from the source Publisher or notifies Subscribers of an
      *         error
      * @see <a href="http://reactivex.io/documentation/operators/last.html">ReactiveX operators documentation: Last</a>
      */
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Flowable<T> last() {
-        return takeLast(1).single();
+    public final Single<T> last() {
+        return RxJavaPlugins.onAssembly(new FlowableLastSingle<T>(this, null));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableLastTest.java
```diff
@@ -54,35 +53,35 @@ public void testLastViaObservable() {
...
 
-        Subscriber<Integer> observer = TestHelper.mockSubscriber();
+        SingleObserver<Integer> observer = TestHelper.mockSingleObserver();
         observable.subscribe(observer);
 
         InOrder inOrder = inOrder(observer);
-        inOrder.verify(observer, times(1)).onNext(3);
-        inOrder.verify(observer, times(1)).onComplete();
+        inOrder.verify(observer, times(1)).onSuccess(3);
+//        inOrder.verify(observer, times(1)).onComplete();
```
JakeWharton, line 63: Why not just remove all these?

JakeWharton (approved):

src/test/java/io/reactivex/internal/operators/flowable/FlowableLastTest.java
akarnokd: Will come in handy when the fuse-back is implemented (by copying these and restoring them to Observer/Subscriber). I'll delete them then

src/main/java/io/reactivex/Flowable.java
```diff
@@ -8877,14 +8877,14 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
-     * @return a Flowable that emits the last item from the source Publisher or notifies Subscribers of an
+     * @return a Single that emits the last item from the source Publisher or notifies Subscribers of an
      *         error
      * @see <a href="http://reactivex.io/documentation/operators/last.html">ReactiveX operators documentation: Last</a>
      */
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Flowable<T> last() {
-        return takeLast(1).single();
+    public final Single<T> last() {
```
abersnaze, line 8886: This should return a `Maybe<T>` because the Flowable can be empty.

src/main/java/io/reactivex/Observable.java
```diff
@@ -7659,8 +7659,8 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, Consumer
      * @see <a href="http://reactivex.io/documentation/operators/last.html">ReactiveX operators documentation: Last</a>
      */
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Observable<T> last() {
-        return takeLast(1).single();
+    public final Single<T> last() {
```
abersnaze, line 7662: This should return a `Maybe<T>` because the Flowable can be empty.

src/main/java/io/reactivex/Flowable.java
akarnokd: The javadoc clearly states that an empty stream results in `NoSuchElementException`; but the former use of `single()` here is also a telltale sign.

abersnaze: An observable being empty is not exceptional and therefore shouldn't result in exceptions.

abersnaze:
P.S. I've pushed my current branch here. https://github.com/abersnaze/RxJava/tree/interop

src/main/java/io/reactivex/Flowable.java
akarnokd: For this operator it is according to the 1.x docs:

https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L7398

> Returns an Observable that emits the last item emitted by the source Observable or notifies observers of
> a `{@code NoSuchElementException}` if the source Observable is empty.

2.x matches the 1.x behavior.

akarnokd:
> P.S. I've pushed my current branch here.

I see you didn't keep the original typed code to help with `toFlowable()`. Also I'm not fond of moving the `blockingX` because it adds an extra step/overhead.

I'd have to redo/reoptimize them anyways so don't really need that.

Issue #4577, opened 2016-09-22
fromEmitter cleanup


davidmoten:
The rename of `Observable.fromAsync` to `Observable.fromEmitter` isn't quite complete in that the signature of `Observable.fromEmitter` still refers to an interface `AsyncEmitter` which I would like to rename to `Emitter`.

 I'm happy to make a PR. It would add a new method 

``` java
@Experimental
public static <T> Observable<T> fromEmitter(Action1<Emitter<T>> emitter, Emitter.BackpressureMode backpressure) ;
```

and I would deprecate this one:

``` java
public static <T> Observable<T> fromEmitter(Action1<AsyncEmitter<T>> emitter, AsyncEmitter.BackpressureMode backpressure);
```

Proceed with PR?

akarnokd:
Yes, PR welcome.

PR #4580, opened 2016-09-22
rename AsyncEmitter to Emitter

davidmoten:
as per discussion in #4577, renaming `AsyncEmitter` to `Emitter` via deprecation.

There is a catch though. `CompletableEmitter` is correctly named now and depends on `AsyncEmitter` via the method:

``` java
 void setCancellation(AsyncEmitter.Cancellable c);
```

When can we break the api of that class?

### DIFF
--- a/src/main/java/rx/Emitter.java
+++ b/src/main/java/rx/Emitter.java
@@ -17,6 +17,7 @@
 package rx;
 
 import rx.annotations.Experimental;
+import rx.functions.Cancellable;
 
 /**
  * Abstraction over a RxJava Subscriber that allows associating
@@ -29,11 +30,9 @@ import rx.annotations.Experimental;
  *
  * @param <T> the value type to emit
  * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
- * @deprecated as of 1.2.1 because Async prefix of this class potentially misleading. Use {@link Emitter} instead.
  */
 @Experimental
-@Deprecated
-public interface AsyncEmitter<T> extends Observer<T> {
+public interface Emitter<T> extends Observer<T> {
 
     /**
      * Sets a Subscription on this emitter; any previous Subscription
@@ -56,24 +55,11 @@ public interface AsyncEmitter<T> extends Observer<T> {
     long requested();
 
     /**
-     * A functional interface that has a single close method
-     * that can throw.
-     */
-    interface Cancellable {
-
-        /**
-         * Cancel the action or free a resource.
-         * @throws Exception on error
-         */
-        void cancel() throws Exception;
-    }
-
-    /**
      * Options to handle backpressure in the emitter.
      */
     enum BackpressureMode {
         /**
-         * No backpressure is applied an the onNext calls pass through the AsyncEmitter;
+         * No backpressure is applied as the onNext calls pass through the Emitter;
          * note that this may cause {@link rx.exceptions.MissingBackpressureException} or {@link IllegalStateException}
          * somewhere downstream.
          */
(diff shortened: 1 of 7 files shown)


akarnokd:
:+1:

davidmoten:
Thanks for review @akarnokd . I take it we have to wait for 1.3 to break the api of experimental `CompletableEmitter`?

akarnokd:
Deleting experimental APIs has to wait one patch version so 1.2.1 will have both and 1.2.2 can remove the AsyncEmitter.

davidmoten:
OK I'll patch again once 1.2.1 is out.

src/main/java/rx/AsyncEmitter.java
```diff
@@ -29,8 +29,10 @@
  *
  * @param <T> the value type to emit
  * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+ * @deprecated as of 1.2.1 because Async prefix of this class potentially misleading. Use {@link Emitter} instead.
  */
 @Experimental
+@Deprecated
 public interface AsyncEmitter<T> extends Observer<T> {
```
JakeWharton, line 36: You could make this extend from `Emitter` to avoid having two implementations of the operator.

JakeWharton (approved):

src/main/java/rx/AsyncEmitter.java
artem-zinnatullin: +1 to this, otherwise I'm :+1: for renaming

davidmoten: cannot extend `Emitter` because of two versions of `Cancellable`. Would break existing `AsyncEmitter` use.

src/main/java/rx/Emitter.java
```diff
@@ -0,0 +1,97 @@
...
+     * <p>This method it thread-safe.
+     * @return the current outstanding request amount
+     */
+    long requested();
+
+    /**
+     * A functional interface that has a single close method
+     * that can throw.
+     */
+    interface Cancellable {
```
akarnokd: You could move this into `rx.functions` to decouple it from the emitter. Also `CompletableEmitter` needs to handle the change in some way.

JakeWharton:
Make the cancelable interfaces extend each other as well.

src/main/java/rx/Emitter.java
davidmoten: seems like a good idea to decouple. `CompletableEmitter` is an interface so I can't just add a method and deprecate the  old one. How about I add a `@Deprecated` annotation to `CompletableEmitter` at the class level now, then after next release I remove the annotation and fix the signature of `setCancellation`?

davidmoten:
> Make the cancelable interfaces extend each other as well.

Good idea, thanks

davidmoten:
Um I don't think it's worth it juggling with inheritance of these interfaces. Straight after next release I'm just going to delete the deprecated stuff.

davidmoten:
I've added a `@Deprecated` annotation to `CompletableEmitter`. Let me know if you guys still want the interface inheritance business.

davidmoten:
Forgot to move `Cancellable` to `rx.functions`. That's done now.

src/main/java/rx/functions/Cancellable.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package rx.functions;
+
+/**
+ * A functional interface that has a single close method that can throw.
+ */
+public interface Cancellable {
```
akarnokd, line 25: Experimental / since tag missing

davidmoten: done, ta

akarnokd:
Thanks for working this out!

PR #4581, opened 2016-09-22
2.x: Remove some dead code

vanniktech:
Also what about IntFunction it's right now in the public API however not used. So the question is does Rx want to expose that interface?

### DIFF
--- a/src/main/java/io/reactivex/internal/observers/InnerQueuedObserver.java
+++ b/src/main/java/io/reactivex/internal/observers/InnerQueuedObserver.java
@@ -42,8 +42,6 @@ implements Observer<T>, Disposable {
 
     volatile boolean done;
 
-    long produced;
-
     int fusionMode;
 
     public InnerQueuedObserver(InnerQueuedObserverSupport<T> parent, int prefetch) {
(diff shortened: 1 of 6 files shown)


src/main/java/io/reactivex/annotations/SchedulerSupport.java
```diff
@@ -28,6 +28,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Target({ElementType.METHOD, ElementType.TYPE})
+@SuppressWarnings("unused") // Part of Public API.
 public @interface SchedulerSupport {
```
akarnokd, line 31: I guess IntelliJ complains about this? Now Eclipse will.

akarnokd (approved):
:+1:

src/main/java/io/reactivex/annotations/SchedulerSupport.java
vanniktech: Well `IO` & `NEW_THREAD` are unused since they are not used internally. However they're part of the public API so I suppressed that warning that they're not used. Why will Eclipse complain about this one?

akarnokd: Because it doesn't see anything, like a private method that is not in use.

vanniktech: Oh that's really bad. You want me to remove this one or leave it as is?

akarnokd: Please remove this.

vanniktech: Done

Issue #4572, opened 2016-09-21
2.x: Defer Scheduler initialization in hook init API


JakeWharton:
An issue was [filed on RxAndroid](https://github.com/ReactiveX/RxAndroid/issues/332) asking for the default "main thread" scheduler to be lazily initialized. This would allow support for unit testing on the JVM by swapping out the scheduler for one that works where Android's main thread doesn't exist.

The hook API provides a means of replacing the scheduler during initialization, but the problem is that in order to call these "init" methods the instance needs to be eagerly created.

What I'm proposing is that we change the "init" functions from `Function<Scheduler, Scheduler>` to `Function<Scheduler, Callable<Scheduler>>` which would allow ignoring the default scheduler instance by never delegating to `Callable.call()` to create it.

The implementation of the initializer would change to something like:

``` java
Scheduler initFoo(Callable<Scheduler> defaultScheduler);
Function f = initFunc;
if (f == null) {
  return defaultScheduler.call();
}
return apply(f, defaultScheduler);
```

Now we can make this change _only_ in RxAndroid to support this case. I think that there's a large value in having as much symmetry between the two libraries as possible, however.

Is this something that would be acceptable to change in RxJava? Or are there any alternative APIs that anyone can think of for this?

akarnokd:
I don't think RxJava needs this change. We have standard Java ExecutorServices backing the Schedulers so their creation should always succeed in a reasonable JDK implementation.

Maybe you could combine `AndroidSchedulers` the plugin and use a holder class to defer the creation of the looper/handler if the function happens to be null the first time `main` is invoked. I can see this needs to change the callback to `Callable` and not rely on the default anymore.

``` java
class AndroidSchedulers {
    static final class DefaultHolder {
        static final Scheduler DEFAULT_MAIN = new HandlerScheduler(
            new Handler(Looper.getMainLooper())));
    }

    static volatile Callable<Scheduler> mainHook;

    public static Scheduler main() {
        Function f = mainHook;
        if (f == null) {
            return DEFAULT_HOLDER.DEFAULT_MAIN; // initialized once and lazily
        }
        try {
            return mainHook.call();
        } catch (Throwable ex) {
            throw ExceptionHelper.wrapOrThrow(ex);
        }
    }
}
```

JakeWharton:
That doesn't allow decorating the default scheduler which is a common use
of these init hooks on Android.

I know RxJava doesn't need this, what I'm asking is if the change would be
[...]

PR #4585, opened 2016-09-22
2.x: Evaluate Schedule initialization via Callable

peter-tackage:
This implements the solution proposed in #4572 - to initialize the Schedulers via a Callable, rather than directly via a value.

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -184,7 +184,7 @@ public final class RxJavaPlugins {
 
     /**
      * Calls the associated hook function.
-     * @param defaultScheduler the hook's input value
+     * @param defaultScheduler a {@link Callable} which returns the hook's input value
      * @return the value returned by the hook
      */
     public static Scheduler initComputationScheduler(Callable<Scheduler> defaultScheduler) {
@@ -197,7 +197,7 @@ public final class RxJavaPlugins {
 
     /**
      * Calls the associated hook function.
-     * @param defaultScheduler the hook's input value
+     * @param defaultScheduler a {@link Callable} which returns the hook's input value
      * @return the value returned by the hook
      */
     public static Scheduler initIoScheduler(Callable<Scheduler> defaultScheduler) {
@@ -210,7 +210,7 @@ public final class RxJavaPlugins {
 
     /**
      * Calls the associated hook function.
-     * @param defaultScheduler the hook's input value
+     * @param defaultScheduler a {@link Callable} which returns the hook's input value
      * @return the value returned by the hook
      */
     public static Scheduler initNewThreadScheduler(Callable<Scheduler> defaultScheduler) {
@@ -223,7 +223,7 @@ public final class RxJavaPlugins {
 
     /**
      * Calls the associated hook function.
-     * @param defaultScheduler the hook's input value
+     * @param defaultScheduler a {@link Callable} which returns the hook's input value
      * @return the value returned by the hook
      */
     public static Scheduler initSingleScheduler(Callable<Scheduler> defaultScheduler) {
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -190,7 +190,7 @@ public final class RxJavaPlugins {
      * @throws NullPointerException if the callable parameter or its result are null
      */
     public static Scheduler initComputationScheduler(Callable<Scheduler> defaultScheduler) {
-        ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable cannot be null.");
+        ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable can't be null");
         Function<Scheduler, Scheduler> f = onInitComputationHandler;
         if (f == null) {
             return call(defaultScheduler);
@@ -205,7 +205,7 @@ public final class RxJavaPlugins {
      * @throws NullPointerException if the callable parameter or its result are null
      */
     public static Scheduler initIoScheduler(Callable<Scheduler> defaultScheduler) {
-        ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable cannot be null.");
+        ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable can't be null");
         Function<Scheduler, Scheduler> f = onInitIoHandler;
         if (f == null) {
             return call(defaultScheduler);
@@ -220,7 +220,7 @@ public final class RxJavaPlugins {
      * @throws NullPointerException if the callable parameter or its result are null
      */
     public static Scheduler initNewThreadScheduler(Callable<Scheduler> defaultScheduler) {
-        ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable cannot be null.");
+        ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable can't be null");
         Function<Scheduler, Scheduler> f = onInitNewThreadHandler;
         if (f == null) {
             return call(defaultScheduler);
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -953,6 +971,33 @@ public static Completable onAssembly(Completable source) {
...
+    static <T> T call(Callable<T> t) {
+        try {
+            return t.call();
+        } catch (Throwable ex) {
+            throw ExceptionHelper.wrapOrThrow(ex);
+        }
+    }
+
+    /**
+     * Wraps the call to the callable in try-catch and propagates thrown
```
vanniktech: copy paste error

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -186,10 +187,10 @@ public static boolean isLockdown() {
      * @param defaultScheduler the hook's input value
      * @return the value returned by the hook
      */
-    public static Scheduler initComputationScheduler(Scheduler defaultScheduler) {
+    public static Scheduler initComputationScheduler(Callable<Scheduler> defaultScheduler) {
```
vanniktech, line 192: Should not it throw on a null callable? What's the point of calling with null?

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -953,6 +971,33 @@ public static Completable onAssembly(Completable source) {
...
 
+    /**
+     * Wraps the call to the callable in try-catch and propagates thrown
+     * checked exceptions as RuntimeException.
+     * @param <T> the input type
+     * @param <T> the output type
+     * @param t the callable, not null (not verified)
+     * @return the result of the callable call
+     */
+    static <T> T call(Callable<T> t) {
```
vanniktech: inline this method?

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
peter-tackage: This was done for consistency with the existing expectations in `RxJavaPluginsTest.clearIsPassthrough()`, specifically:

``` java
assertNull(RxJavaPlugins.initComputationScheduler(null));
assertNull(RxJavaPlugins.initIoScheduler(null));
assertNull(RxJavaPlugins.initNewThreadScheduler(null));
assertNull(RxJavaPlugins.initSingleScheduler(null));
```

Should this be changed to only return null if the `Callable` returns null (and throw if the `Callable` itself is null)?

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -953,6 +971,33 @@ public static Completable onAssembly(Completable source) {
...
+
+    /**
+     * Wraps the call to the callable in try-catch and propagates thrown
+     * checked exceptions as RuntimeException.
+     * @param <T> the input and output type
+     * @param t the callable, nullable
+     * @return the callable result if the callable is nonnull, null otherwise.
+     */
+    static <T> T callOrNull(Callable<T> t) {
+        return t == null ? null : call(t);
```
JakeWharton: null should never be allowed

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
JakeWharton: Null should not be allowed as a return value from the `Callable` nor from the init `Function`.

peter-tackage: This would be a change, because previously the `Scheduler` value for `RxJavaPlugins.init*` was allowed to be null, as per - `assertNull(RxJavaPlugins.initSingleScheduler(null));`. 

I will add an additional set of tests for the new behavior (something along the lines of `assemblyHookCrashes`).

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
peter-tackage: Not sure what you mean. I've tried to be consistent with the abstraction of the similar `apply` method. Do you think that should change?

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
peter-tackage: Fixed.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
peter-tackage: Done.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -936,6 +946,26 @@ public static Completable onAssembly(Completable source) {
...
+     * @param <R> the output type
+     * @param f the function to call, not null (not verified)
+     * @param t the {@link Callable} parameter value to the function, not null (not verified). Cannot return null.
+     * @return the result of the function call, not null
+     * @throws NullPointerException if the callable returns null
+     */
+    static <T, R> R apply(Function<T, R> f, Callable<T> t) {
+        try {
+            T value = t.call();
+            ObjectHelper.requireNonNull(t, "Callable result can't be null");
```
vanniktech: I think you mean to check `value` here instead of `t`

vanniktech: Also you could just `T value = Object.requireNonNull(t.call(), "messge")`

peter-tackage: Yep, good find.

peter-tackage: Yes, that would be better. I didn't notice the return value in the signature.

src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -45,15 +46,35 @@
...
     static {
-        SINGLE = RxJavaPlugins.initSingleScheduler(new SingleScheduler());
-
-        COMPUTATION = RxJavaPlugins.initComputationScheduler(new ComputationScheduler());
-
-        IO = RxJavaPlugins.initIoScheduler(new IoScheduler());
+        SINGLE = RxJavaPlugins.initSingleScheduler(new Callable<Scheduler>() {
+            @Override
+            public Scheduler call() throws Exception {
+                return new SingleScheduler();
```
akarnokd: Just in case call() is called multiple times by the hook, these could actually return an field of an inner class (singleton holder pattern or what's the name):

``` java
static final class SingleHolder {
    static final Scheduler DEFAULT = new SingleScheduler();
}

SINGLE = RxJavaPlugins.initiSingleScheduler(new Callable<Scheduler>() {
    @Override
    public Scheduler call() {
        return SingleHolder.DEFAULT;
    }
});
```

peter-tackage:
@akarnokd, Correct me if I'm wrong, but in order to avoid the evaluation of the default Scheduler instance when it is being overridden, I still need to change the `onInit[single|io|new|computation]Handler` signatures to be `Function<Callable<Scheduler>, Scheduler>` as per -

```
    static volatile Function<Callable<Scheduler>, Scheduler> onInitSingleHandler;
```

Otherwise the invocation of `initSingleScheduler` will cause the evaluation of the default value to pass it as a parameter to the overriding/transforming function: `onInitSingleHandler`.

akarnokd:
Yes, you still need `Callable` as the indirection but the inner class will be evaluated only if `call` is actually invoked.

peter-tackage:
I've added the remaining lazy initialization.

I've also fairly aggressively enforced non-null in the associated functions, for example:

``` java
public static Scheduler initIoScheduler(Callable<Scheduler> defaultScheduler) {
    ObjectHelper.requireNonNull(defaultScheduler, "Scheduler Callable can't be null");
    Callable<Scheduler>, Scheduler> f = onInitIoHandler;
    if (f == null) {
        return callRequireNonNull(defaultScheduler);
    }
    return applyRequireNonNull(f, defaultScheduler);
}
```

However, to me, this seems slightly out of place / over the top. Is that enforcement necessary or should the resultant null Scheduler be left unasserted and left to the eventual NullPointerException when the Scheduler is used? Either way, I'm happy to keep or remove that based upon review feedback.

JakeWharton:
Aggressive input validation is never over the top. If you defer checking
then the stacktrace tells you nothing about what actually caused the broken
state.

akarnokd (approved):

peter-tackage:
@akarnokd, all done as far as I am concerned. Do I need to anything else for this to be merged?

akarnokd:
I was waiting for you to settle with the implementation. Thanks for the contribution.

PR #4590, opened 2016-09-23
2.x: Report errors from onError to Plugin when done

vanniktech:
For the first one that I found I even wrote a test. Then I started noticing more missing spots. If wanted I could port that one test to every operator that I touched. Though I didn't really find tests testing that errors are propagated through the Plugin API.


akarnokd:
`TestHelper` has a few generic test methods to check `Maybe` and `Single`. The same pattern could be extended to test these without duplicating that long method. I was planning that as part of another run of coverage improvements but I'm behind with performance work too.

PR #4592, opened 2016-09-23
2.x: UnicastSubject fix onTerminate

vanniktech:


src/main/java/io/reactivex/subjects/UnicastSubject.java
```diff
@@ -184,7 +191,7 @@ public void onNext(T t) {
             return;
         }
         if (t == null) {
-            onError(new NullPointerException());
+            onError(new NullPointerException("onNext called with null. Null values are generally not allowed in 2.x operators and sources."));
```
vanniktech, line 194: should not this throw immediately rather than trying to go through onError?

akarnokd: Yeah, the spec mandates throwing on null input and I think the Processor TCK also expects it. I thought signalling NPE is more graceful with Subjects and FlowableProcessors and shuts down the streams as well whereas a thrown NPE may leave everybody hanging.

akarnokd (approved):

src/main/java/io/reactivex/subjects/UnicastSubject.java
vanniktech: Alright I'll leave the decision up to you. I'd prefer the fail early approach however delegating it down the stream also works. If you want me to change it I'll do so in a follow up PR.

akarnokd:
Leave this as is and I'll think about it.

PR #4600, opened 2016-09-25
2.x: don't show started unit tests

akarnokd:
Unfortunately, test reporting can't be coalesced so our almost 6000 tests overflow the Travis' window and making it more difficult to see what went wrong (especially from mobile). 

However, Travis times the build out if it doesn't detect standard output emission for more than 10 minutes; our tests take 12 minutes on average.

By not showing started but showing skipped (of which we have 200 scattered around), the timeout should not happen and the output is not bloated.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,21 +68,45 @@ if (project.hasProperty('release.useLastTag')) {
 }
 
 test {
+    
+    testLogging  {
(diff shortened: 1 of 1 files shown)


vanniktech:
Very nice that Travis output was always way too long 👍

akarnokd:
If I can get it working. With the settings above, it still prints STARTED.

PR #4604, opened 2016-09-25
2.x: add flattenAs{Observable,Flowable} to Single and Maybe

akarnokd:
This PR adds `Single.flattenAsObservable`, `Single.flattenAsFlowable`, `Maybe.flattenAsObservable` and `Maybe.flattenAsFlowable` to map a success value into an `Iterable` sequence and stream it as `Flowable` or `Observable`. It is more efficient than `Single.toFlowable().flatMapIterable()`.


akarnokd:
`flattenAsFlowable`, `flattenAsObservable`?

PR #4607, opened 2016-09-26
2.x: TestSubscriber & TestObserver add assertValue(Predicate)

vanniktech:


src/main/java/io/reactivex/observers/TestObserver.java
```diff
@@ -533,6 +533,43 @@ private AssertionError fail(String message) {
...
+        int s = values.size();
+        if (s == 0) {
+            throw fail("No values");
+        }
+
+        boolean found = false;
+
+        for (T value : values) {
+            try {
+                if (valuePredicate.test(value)) {
```
akarnokd: This is unnecessary, just test values.get(0) and then report if s > 1 besides.

akarnokd (changes requested):

src/main/java/io/reactivex/observers/TestObserver.java
vanniktech: Yeah thought of that too I just took the same implementation as assertError(predicate). I'll change it there too, ok?

akarnokd: Don't change assertError.

akarnokd (approved):

src/main/java/io/reactivex/observers/TestObserver.java
vanniktech: Any reason why?

akarnokd: Because the first error may come from the lack of onSubscribe call or some other unexpected exception. Knowing that the exception is there but not the only one is different from expecting exactly 1 element but then getting n.

vanniktech: Oh that makes sense. Thanks David for the explanation

akarnokd (approved):

PR #4609, opened 2016-09-26
2.x: Factor out typical TestObserver/TestSubscriber methods

akarnokd:
This PR introduces a new abstract internal class `BaseTestConsumer` which now hosts the `assertX` methods of `TestSubscriber` and `TestObserver` which were duplicates before.


JakeWharton (approved):

akarnokd:
The drawback is that the generated javadoc doesn't have the method documentation and doesn't reference `BaseTestConsumer` because it's in the `internal` zone. I can move it to the public zone so the javadoc links work:

image

PR #4614, opened 2016-09-27
2.x: Add Maybe.flatMapSingle

vanniktech:
Gave that implementation a try. Feedback is welcome I think there are improvements that can be done.

### DIFF
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -13,6 +13,7 @@
 
 package io.reactivex;
 
+import java.util.NoSuchElementException;
 import java.util.concurrent.*;
 
 import org.reactivestreams.*;
@@ -2554,6 +2555,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     /**
      * Returns a {@link Single} based on applying a specified function to the item emitted by the
      * source {@link Maybe}, where that function returns a {@link Single}.
+     * When this Maybe completes a {@link NoSuchElementException} will be thrown.
      * <p>
      * <img width="640" height="305" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Maybe.flatMapSingle.png" alt="">
      * <dl>
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -2570,9 +2570,9 @@ public abstract class Maybe<T> implements MaybeSource<T> {
      * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX operators documentation: FlatMap</a>
      */
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Single<T> flatMapSingle(final Function<? super T, ? extends SingleSource<T>> mapper) {
+    public final <R> Single<R> flatMapSingle(final Function<? super T, ? extends SingleSource<? extends R>> mapper) {
         ObjectHelper.requireNonNull(mapper, "mapper is null");
-        return RxJavaPlugins.onAssembly(new MaybeFlatMapSingle<T>(this, mapper));
+        return RxJavaPlugins.onAssembly(new MaybeFlatMapSingle<T, R>(this, mapper));
     }
 
     /**
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -2552,6 +2552,28 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     }
 
     /**
+     * Returns a {@link Single} based on applying a specified function to the item emitted by the
+     * source {@link Maybe}, where that function returns a {@link Single}.
+     * <p>
+     * <img width="640" height="305" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Maybe.flatMapSingle.png" alt="">
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code flatMapSingle} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
```diff
@@ -0,0 +1,133 @@
...
+    extends AtomicReference<Disposable>
+    implements MaybeObserver<T>, Disposable {
+
+        private static final long serialVersionUID = 4827726964688405508L;
+
+        final SingleObserver actual;
+
+        final Function<? super T, ? extends SingleSource<T>> mapper;
+
+        final SingleObserver<T> singleObserver = new SingleObserver<T>() {
```
vanniktech: Not sure whether this is the right approach

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
```diff
@@ -0,0 +1,133 @@
...
+        public void onError(Throwable e) {
+            actual.onError(e);
+        }
+
+        @Override
+        public void onComplete() {
+            SingleSource<T> ss;
+
+            try {
+                ss = ObjectHelper.requireNonNull(mapper.apply(null), "The mapper returned a null SingleSource");
```
vanniktech: Here is the question whether the mapper should be used or not. I thought it might make sense that `Maybe.empty()` should also go through `flatMapSingle`

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
```diff
@@ -0,0 +1,133 @@
...
+        private static final long serialVersionUID = 4827726964688405508L;
+
+        final SingleObserver actual;
+
+        final Function<? super T, ? extends SingleSource<T>> mapper;
+
+        final SingleObserver<T> singleObserver = new SingleObserver<T>() {
+            @Override
+            public void onSubscribe(final Disposable d) {
+
```
akarnokd: This doesn't compose the disposable from the `Single`.

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
akarnokd: Please avoid anonymous inner classes such as this.

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
akarnokd: We don't map `onComplete` and `mapper` shouldn't anticipate `null` from the library anyaway.

akarnokd (changes requested):

vanniktech:
@akarnokd could you check again? I think I've figured it out now how this should all work

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        @Override
+        public void onError(Throwable e) {
+            actual.onError(e);
+        }
+
+        @Override
+        public void onComplete() {
+            // Ignored.
```
akarnokd: This will hang the output `Single`. You should signal a `NoSuchElementException` instead.

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
```diff
@@ -0,0 +1,132 @@
...
+
+    public MaybeFlatMapSingle(MaybeSource<T> source, Function<? super T, ? extends SingleSource<T>> mapper) {
+        this.source = source;
+        this.mapper = mapper;
+    }
+
+    @Override
+    protected void subscribeActual(SingleObserver<? super T> s) {
+        FlatMapMaybeObserver<T> parent = new FlatMapMaybeObserver<T>(s, mapper);
+        s.onSubscribe(parent);
```
akarnokd: No need to pull this out.

src/main/java/io/reactivex/internal/operators/maybe/MaybeFlatMapSingle.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        @Override
+        public boolean isDisposed() {
+            return DisposableHelper.isDisposed(get());
+        }
+
+        @Override
+        public void onSubscribe(Disposable d) {
+            DisposableHelper.replace(this, d);
```
akarnokd: setOnce() + actual.onSubscribe(this)

akarnokd (changes requested):

src/main/java/io/reactivex/Maybe.java
```diff
@@ -2552,6 +2553,29 @@ public final T blockingGet(T defaultValue) {
...
+     * </dl>
+     *
+     * @param mapper
+     *            a function that, when applied to the item emitted by the source Maybe, returns a
+     *            Single
+     * @return the Single returned from {@code mapper} when applied to the item emitted by the source Maybe
+     * @see <a href="http://reactivex.io/documentation/operators/flatmap.html">ReactiveX operators documentation: FlatMap</a>
+     */
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Single<T> flatMapSingle(final Function<? super T, ? extends SingleSource<T>> mapper) {
```
akarnokd: `Function<? super T, ? extends SingleSource<? extends R>>`

akarnokd (changes requested):

akarnokd (approved):

vanniktech:
Thanks for bearing with me on this one ... :D

PR #4616, opened 2016-09-27
2.x: Add Completable.andThen(MaybeSource)

vanniktech:


akarnokd:
There is a compilation error.

vanniktech:
Ah yes import is missing. I still manually stage them and sometimes forget that one is needed since usually every package is already included with `*`.

src/main/java/io/reactivex/internal/operators/maybe/MaybeDelayWithCompletable.java
```diff
@@ -0,0 +1,115 @@
...
+        final MaybeSource<T> source;
+
+        OtherObserver(MaybeObserver<? super T> actual, MaybeSource<T> source) {
+            this.actual = actual;
+            this.source = source;
+        }
+
+        @Override
+        public void onSubscribe(Disposable d) {
+            if (DisposableHelper.set(this, d)) {
```
akarnokd: setOnce

akarnokd (changes requested):

src/main/java/io/reactivex/internal/operators/maybe/MaybeDelayWithCompletable.java
vanniktech: Then [this](https://github.com/ReactiveX/RxJava/blob/d85d40ecb1d4c3e02a4bc9589dd7af2900752259/src/main/java/io/reactivex/internal/operators/single/SingleDelayWithCompletable.java#L56) would need to be changed too, right?

akarnokd: Yes.

vanniktech: Alright done

akarnokd (approved):

PR #4619, opened 2016-09-28
2.x: Make CompositeExcpetion thread-safe like 1.x and also fix some issues

zsxwing:
Right now CompositeExcpetion has several issues:
- `CompositeException(Throwable... exceptions)` doesn't deduplicate exceptions and flatten CompositeExceptions like `CompositeException(Iterable<? extends Throwable> errors)`
- If using `CompositeException(Iterable<? extends Throwable> errors)` to create CompositeException, `suppress` cannot be used.
- `suppress` doesn't update `cause`.
- `suppress` doesn't deduplicate exceptions and flatten CompositeExceptions.
- `suppress` and `Throwable.addSuppressed` are pretty confusing for Java 7+ users. Without looking at the implementation, it's hard to figure out the differences.

This PR made the following changes:
- Remove `CompositeException.suppress` so that it's easy to make CompositeException thread-safe.
  - This may cause some performance lost in some path rarely happening, e.g., an excpetion is thrown from `onError`, but that's not a big deal.
  - Since `suppress` is removed, it doesn't make sense to create an empty CompositeException, so `isEmpty` is removed and defense codes are added.
- Defense codes for bad exceptions.
- Deduplicate excepctions and flatten CompositeExceptions for `CompositeException(Throwable... exceptions)`.

### DIFF
--- a/src/main/java/io/reactivex/exceptions/CompositeException.java
+++ b/src/main/java/io/reactivex/exceptions/CompositeException.java
@@ -41,36 +41,23 @@ public final class CompositeException extends RuntimeException {
     private Throwable cause;
 
     /**
-     * Constructs an empty CompositeException.
-     */
-    public CompositeException() {
-        this.exceptions = new ArrayList<Throwable>();
-        this.message = null;
-    }
-
-    /**
      * Constructs a CompositeException with the given array of Throwables as the
      * list of suppressed exceptions.
      * @param exceptions the Throwables to have as initially suppressed exceptions
+     *
+     * @throws IllegalArgumentException if <code>exceptions</code> is empty.
      */
     public CompositeException(Throwable... exceptions) {
-        this.exceptions = new ArrayList<Throwable>();
-        if (exceptions == null) {
-            this.message = "1 exception occurred. ";
-            this.exceptions.add(new NullPointerException("exceptions is null"));
-        } else {
-            this.message = exceptions.length + " exceptions occurred. ";
-            for (Throwable t : exceptions) {
-                this.exceptions.add(t != null ? t : new NullPointerException("One of the exceptions is null"));
-            }
-        }
+        this(exceptions == null ?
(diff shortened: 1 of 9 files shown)


src/main/java/io/reactivex/exceptions/CompositeException.java
```diff
@@ -266,15 +244,16 @@ public String getMessage() {
     private List<Throwable> getListOfCauses(Throwable ex) {
         List<Throwable> list = new ArrayList<Throwable>();
         Throwable root = ex.getCause();
-        if (root == null) {
+        if (root == null || root == ex) {
```
akarnokd, line 247: Do you have a test for this? I removed it because code coverage showed it as never taken.

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -130,11 +130,8 @@ protected final AssertionError fail(String message) {
         ;
 
         AssertionError ae = new AssertionError(b.toString());
-        CompositeException ce = new CompositeException();
-        for (Throwable e : errors) {
-            ce.suppress(e);
-        }
-        if (!ce.isEmpty()) {
+        if (!errors.isEmpty()) {
+            CompositeException ce = new CompositeException(errors);
```
akarnokd, line 134: Might be worth changing this so when there is only one exception, it doesn't wrap and lenghten the stacktrace unnecessarily.

akarnokd (approved):

src/main/java/io/reactivex/exceptions/CompositeException.java
zsxwing: Yes. See https://github.com/ReactiveX/RxJava/pull/4619/files#diff-f0be6c6bea0d800fccc9e3fe721c23e4R347

src/main/java/io/reactivex/observers/BaseTestConsumer.java
zsxwing: Fixed in #4631

PR #4625, opened 2016-09-28
2.x: Replace 'resource' observers with plain 'disposable' observers in tests.

JakeWharton:
This makes the resource observers unused internally and I'm curious as to whether they deserve to stick around in the public API. Any thoughts?


akarnokd:
They are there to provide the resource support similar to 1.x `Subscriber` and `SingleSubscriber`. We could move them to RxJava2Extensions.

JakeWharton:
Ah, yes. That makes sense.

akarnokd (approved):

PR #4634, opened 2016-09-29
Upgrade Mockito to 2.1.0-RC.2

TimvdLippe:
The second release candidate of Mockito 2 has been released, which is mostly backwards compatible with your code base. I had to remove the deprecated usage of `Matchers` and replace it with the new `Mockito` equivalent matchers. This solved several ambiguity issues where both `Mockito` and `Matchers` were imported in the same file. Other than that all tests were passing locally, so let's see if CI agrees too.

Thanks a lot for using Mockito and if you have any feedback, we are eager to hear from you!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ dependencies {
     signature 'org.codehaus.mojo.signature:java16:1.1@signature'
 
     testCompile 'junit:junit:4.12'
-    testCompile 'org.mockito:mockito-core:1.10.19'
+    testCompile 'org.mockito:mockito-core:2.1.0-RC.2'
     testCompile 'com.google.guava:guava:19.0'
     testCompile 'com.pushtorefresh.java-private-constructor-checker:checker:1.2.0'
 
--- a/src/test/java/rx/CompletableTest.java
+++ b/src/test/java/rx/CompletableTest.java
@@ -17,7 +17,6 @@
 package rx;
 
 import static org.junit.Assert.*;
-import static org.mockito.Matchers.*;
 import static org.mockito.Mockito.*;
 
 import java.util.*;
--- a/src/test/java/rx/NotificationTest.java
+++ b/src/test/java/rx/NotificationTest.java
@@ -16,7 +16,6 @@
 
 package rx;
 
-import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.*;
 import static org.junit.Assert.*;
 
--- a/src/test/java/rx/ObservableTests.java
+++ b/src/test/java/rx/ObservableTests.java
@@ -16,7 +16,6 @@
 package rx;
 
 import static org.junit.Assert.*;
-import static org.mockito.Matchers.*;
 import static org.mockito.Mockito.*;
 
 import java.lang.Thread.UncaughtExceptionHandler;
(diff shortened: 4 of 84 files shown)


akarnokd:
Any particular reason we should upgrade? Mockito 1.10.19 worked for us so far; also why RC2 and not wait for final?

bric3:
@akarnokd This PR allows us to verify the behaviour of mockito in the wild before the final.

Here's features that may be of interest : https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#improvements

TimvdLippe:
@akarnokd thank you for your response. Mockito 2 should be in general compatible with 1.10.19. The main change in Mockito 2 is using ByteBuddy as mock maker engine. ByteBuddy is future proof and works on more recent Java versions, while it is also (almost) ready for Java 9. CGLib was sadly becoming outdated and had several issues we could not work around.

This PR was mainly for us to investigate the impact of upgrading, especially since we have published our release candidate. We expect to release final in the next week; I understand if you would rather wait a couple of days for the final release.

Apart from the import changes, it does seem some tests started failing on CI, which we will take a look if this is an issue with Mockito 2 in particular, or a breaking change in its usage.

Hopefully this addresses your concerns right now, if you have more questions we are happy to answer them.

akarnokd:
> Other than that all tests were passing locally.

Strange, I get 4 deterministic failures when I run your code locally.
- It seems that `any(T.class)` no longer matches `null`
- Test that matched Exception such as `onError(any(NullPointerException.class))` now get a different exception. Even stranger, the tests expected the wrong exception all along yet never complained.

akarnokd:
Ah, I see `any(T)` no longer accepts nulls and `any(Class<T>)` now actually checks for the type.

TimvdLippe:
@akarnokd Yes they are failing for me locally now too. I sadly have no explanation for why they weren't earlier.

Yes `any` rejects null now. Users were missing bugs, because the matchers accepted too many values. This also seems the case for the second point where Types are actually checked.

akarnokd:
For 1.x the allowing null is important so unless there's a way to accept nulls with Mockito, we can't move to 2.1 yet.

TimvdLippe:
@akarnokd Luckily checking for null is as easy as `verify(observer, times(1)).onNext(null);` :smile: 

The four failing tests should now be fixed, with explicit checks for null as well as the correct types of exceptions.

bric3:
Generally speaking it is not an issue to disallow `null`, also this allow more control. API wise any family was broken in design in the 1.x to avoid casts, because of this we felt it was an important step to rework the any matchers.

Here's the change about any family matchers
- `<T> T any()` will matches anything including `null`
- `T anyT()` / `<T> any(Class<T>)` are aliases of `<T> isA(T)`, will matches non-`null` object of given type. e.g. :
  - `int anyInt()` won't match `null` (for Integers) anymore
  - `int anyString()` won't match `null` (for Strings) anymore, and will check the object is indeed a `String`
  - `int anyMap()` won't match `null` (for Maps) anymore, and will check the object is indeed a `Map`
  - `<T> List<T> anyListOf(Class<T>)` won't match `null`, and will check the object is indeed a `List`

Signature have been reworked and is balanced between the worlds of JDK6 and JDK8, although favoring JDK8 compiler.

We welcome your feed back on this !

akarnokd:
Thanks for looking into this but I believe we'd be better experimenting with this in the 2.x branch and then apply the lessons learned in 1.x with the final/GA of Mockito 2.1. Could you post a PR targeting 2.x?

akarnokd:
Thanks for looking into this. I'm closing this for now as may become conflicting in the meantime anyway. Once you release 2.1, I'm gladly redoing these changes on the fresh base.

PR #4636, opened 2016-09-29
Update Mockito to 2.1.0-RC.2

TimvdLippe:
Equivalent of #4634 , but now for the 2.x branch.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ dependencies {
     compile 'org.reactivestreams:reactive-streams:1.0.0'
 
     testCompile 'junit:junit:4.12'
-    testCompile 'org.mockito:mockito-core:1.10.19'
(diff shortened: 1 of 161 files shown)


TimvdLippe:
The test `io.reactivex.maybe.MaybeTest > onTerminateDetach` seems to be failing, claiming there is a memory leak. I sadly have no experience with this test to fix it.

akarnokd:
That test is a bit flaky (maybe the new Mockito increased the base memory usage?). I'll fix that test and rerun this one.

akarnokd (approved):

TimvdLippe:
Thanks a lot and enjoy :smile: !

TimvdLippe:
FYI Mockito 2.1.0 has been released which is equivalent to 2.1.0-RC.2. Once again thanks a lot for using Mockito :smile:

Issue #4332, opened 2016-08-10
RxJavaHooks.onError() catches all Throwables (not just undeliverable ones)


dlew:
The `RxJavaHooks.onError()` documentation says:

> Consume undeliverable Throwables (acts as a global catch).

However, it consumes deliverable `Throwables` just as much. For example, `SafeSubscriber._onError()` immediately calls it ([source](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/observers/SafeSubscriber.java#L150)) whether there's an onError set for the subscriber or not.

Either the documentation is wrong (and it should say it consumes _all_ `Throwables`) or the behavior is wrong.

akarnokd:
The behavior of `SafeSubscriber` is wrong but many rely on this property already.

dlew:
Would it make sense to have a second hook that is _really_ just for undelivered `Throwables`? That would be handy.

akarnokd:
Maybe `SafeSubscriber` could call `RxJavaPlugins...handleError()` again, skipping `RxJavaHooks` entirely.

dlew:
That would work for me - want a PR?

akarnokd:
Sure.

akarnokd:
See #4641

akarnokd:
Closing via #4641

PR #4641, opened 2016-09-29
1.x: SafeSubscriber not to call RxJavaHooks before delivering the original error.

akarnokd:
Before the introduction of `RxJavaHooks`, the `SafeSubscriber._onError` called the original error handler with the exception it received which was by default an empty handler. The default `RxJavaHooks.onError` behavior, however is to signal errors to the uncaught exception handler of the caller thread which leads to unnecessary logging or app crashes even though the error itself is to be handled properly.

This PR restores the `SafeSubscriber._onError` to skip the `RxJavaHooks` and call the original handler directy so old tracking code should still get all safe error while newer hooking doesn't get called.

Related: #4332.

### DIFF
--- a/src/main/java/rx/observers/SafeSubscriber.java
+++ b/src/main/java/rx/observers/SafeSubscriber.java
@@ -19,7 +19,7 @@ import java.util.Arrays;
 
 import rx.Subscriber;
 import rx.exceptions.*;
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):

dlew:
LGTM. Thanks for doing this.

PR #4648, opened 2016-09-30
1.x: rework Single internals to reduce overhead and stack depth

akarnokd:
This PR makes the `Single.OnSubscribe` as the primary means to execute the subscribe action in `Single` and former couple of reused `Observable` operators are now recreated with `Single.OnSubscribe`.

Because `Single` is now marked final, we can't remove the former `Observable.Operator` and `Observable.OnSubscribe` entry points (but may be deprecated in a separate PR).

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -20,7 +20,6 @@ import rx.annotations.*;
 import rx.exceptions.*;
 import rx.functions.*;
 import rx.internal.operators.*;
-import rx.internal.producers.SingleDelayedProducer;
 import rx.internal.util.*;
 import rx.observers.*;
 import rx.plugins.RxJavaHooks;
@@ -51,7 +50,7 @@ import rx.subscriptions.Subscriptions;
  */
 public class Single<T> {
 
-    final Observable.OnSubscribe<T> onSubscribe;
+    final OnSubscribe<T> onSubscribe;
 
     /**
      * Creates a Single with a Function to execute when it is subscribed to (executed).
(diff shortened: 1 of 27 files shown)


src/main/java/rx/internal/operators/SingleFromCallable.java
```diff
@@ -0,0 +1,51 @@
...
+
+    final Callable<? extends T> callable;
+
+    public SingleFromCallable(Callable<? extends T> callable) {
+        this.callable = callable;
+    }
+
+    @Override
+    public void call(SingleSubscriber<? super T> t) {
+        // TODO Auto-generated method stub
```
JakeWharton: delete

src/main/java/rx/Single.java
```diff
@@ -60,40 +59,25 @@
...
+        this.onSubscribe = RxJavaHooks.onCreate(f);
     }
 
-    private Single(final Observable.OnSubscribe<T> f) {
-        this.onSubscribe = RxJavaHooks.onCreate(f);
+    /**
+     * Creates a Single with a Function to execute when it is subscribed to (executed).
+     * <p>
+     * <em>Note:</em> Use {@link #create(OnSubscribe)} to create a Single, instead of this constructor,
+     * unless you specifically have a need for inheritance.
```
JakeWharton, line 73: We could also `@Deprecate` this constructor.

JakeWharton (approved):

src/main/java/rx/Single.java
```diff
@@ -60,40 +59,28 @@
...
```
akarnokd, line 73: Updated.

PR #4651, opened 2016-10-01
Switch Maybe and Single to use their Transformers in compose()

### LINKED ISSUE #4650: 2.x: Should Maybe and Single have a corresponding Transformers?
Currently, `Maybe.compose()` requires a `Function<? super Maybe<T>, ? extends MaybeSource<R>>` and `Single.compose()` requires something similar

While other types have `FlowableTransformer` and `CompletableTransformer`. Would it make sense to bring them into the same pattern? Aside from consistency, early testing for me shows that the generics gymnastics involved in `Maybe` and `Single`'s `Function` approaches is a major headache, as any parameterized stream type gets lost and requires manual specification, forfeiting the ability to  reuse APIs.

Consider RxLifecycle, which historically has just returned an implementation of the required Transformer.

``` java
someSingle
    .compose(RxLifecycle.bind(this).forSingle())
    .subscribe();
```

This works for simple unparameterized types, but breaks down when parameterized types come in.

screenshot 2016-10-01 02 21 10

``` java
// Same result with both Single and Maybe. Screenshot is Maybe, example below is Single.
// The type is now a List<T> of some type T
someListSingle
    .compose(Confine.to(this).forSingle())  // Compiler error because R type instance not found
    .subscribe();
```

Where `Confine.to` returns a bridging helper like this:

``` java
public static class LifecycleTransformer2<T> implements FlowableTransformer<T, T> {

  // Delegate transformer, what's currently returned by RxLifecycle.bind()
  private LifecycleTransformer<T> delegate;

  public static <T> LifecycleTransformer2<T> create(@NonNull LifecycleTransformer<T> delegate) {
    return new LifecycleTransformer2<>(delegate);
  }

  private LifecycleTransformer2(@NonNull LifecycleTransformer<T> delegate) {
    this.delegate = delegate;
  }

  public Function<Maybe<T>, MaybeSource<T>> forMaybe() {
    return source -> {
      Observable<T> o = toV1Observable(source.toFlowable());
      o = delegate.call(o);
      return RxJavaInterop.toV2Flowable(o).singleElement();
    };
  }

  public Function<? super io.reactivex.Single<T>, ? extends SingleSource<T>> forSingle() {
    return new Function<io.reactivex.Single<T>, SingleSource<T>>() {
      @Override
      public SingleSource<T> apply(io.reactivex.Single<T> source) throws Exception {
        Single<T> o = toV1Single(source);
        o = (Single<T>) delegate.forSingle().call((Single<Object>) o);
        return RxJavaInterop.toV2Single(o);
      }
    };
  }

  public CompletableTransformer forCompletable() {
    return source -> {
      Completable o = toV1Completable(source);
      o = delegate.forCompletable().call(o);
      return RxJavaInterop.toV2Completable(o);
    };
  }

  @Override
  public Publisher<? extends T> apply(Flowable<T> source) throws Exception {
    Observable<T> o = toV1Observable(source);
    o = delegate.call(o);
    return RxJavaInterop.toV2Flowable(o);
  }
}
```

`FlowableTransformer` and `CompletableTransformer` work fine, but the other two have been really tricky to nail down.

ZacSweers:
Resolves #4650

### DIFF
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -1967,13 +1967,12 @@ public abstract class Maybe<T> implements MaybeSource<T> {
      * </dl>
      *
      * @param <R> the value type of the Maybe returned by the transformer function
-     * @param transformer
-     *            implements the function that transforms the source Maybe
+     * @param transformer the transformer function, not null
      * @return a Maybe, transformed by the transformer function
      * @see <a href="https://github.com/ReactiveX/RxJava/wiki/Implementing-Your-Own-Operators">RxJava wiki: Implementing Your Own Operators</a>
      */
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final <R> Maybe<R> compose(Function<? super Maybe<T>, ? extends MaybeSource<R>> transformer) {
+    public final <R> Maybe<R> compose(MaybeTransformer<T, R> transformer) {
         return wrap(to(transformer));
     }
 
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1467,13 +1467,12 @@ public abstract class Single<T> implements SingleSource<T> {
      * </dl>
      *
      * @param <R> the value type of the single returned by the transformer function
-     * @param transformer
-     *            implements the function that transforms the source Single
+     * @param transformer the transformer function, not null
      * @return the source Single, transformed by the transformer function
      * @see <a href="https://github.com/ReactiveX/RxJava/wiki/Implementing-Your-Own-Operators">RxJava wiki: Implementing Your Own Operators</a>
      */
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final <R> Single<R> compose(Function<? super Single<T>, ? extends SingleSource<R>> transformer) {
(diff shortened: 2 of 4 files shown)


src/main/java/io/reactivex/Maybe.java
```diff
@@ -1967,13 +1967,12 @@ public final T blockingGet(T defaultValue) {
      * </dl>
      *
      * @param <R> the value type of the Maybe returned by the transformer function
-     * @param transformer
-     *            implements the function that transforms the source Maybe
+     * @param transformer the transformer function, not null
```
ZacSweers, line 1970: Note that I just borrowed from the `Completable.compose()` doc wording here

ZacSweers:
Whoops forgot to update the tests, will update in a sec

akarnokd:
IntelliJ user I presume?

ZacSweers:
Indeed... Does (I presume 😉) eclipse do something to protect against this?

akarnokd:
Eclipse has an error view which doesn't obscure the project structure, plus when one shows the breadcrumbs, they get error icons indicating there's still something wrong (also the project tree gets error icons) so it is much harder to miss these kinds of errors.

ZacSweers:
Interesting, I'll have to check that out!

Issue #4653, opened 2016-10-01
2.x: Possible 6.x/Android compatibility issue


ZacSweers:
If I have an Android app targeting source compatibility 1.8, I get the following exception at runtime on a 7.0 device:

```
FATAL EXCEPTION: RxSchedulerPurge-1
     Process: io.sweers.catchup.debug, PID: 29250
     java.lang.NoSuchMethodError: No virtual method keySet()Ljava/util/concurrent/ConcurrentHashMap$KeySetView; in class Ljava/util/concurrent/ConcurrentHashMap; or its super classes (declaration of 'java.util.concurrent.ConcurrentHashMap' appears in /system/framework/core-libart.jar)
         at io.reactivex.internal.schedulers.SchedulerPoolFactory$1.run(SchedulerPoolFactory.java:69)
         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:428)
         at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:278)
         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:273)
         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
         at java.lang.Thread.run(Thread.java:761)
```

The exact line is:

``` java
for (ScheduledThreadPoolExecutor e : new ArrayList<ScheduledThreadPoolExecutor>(POOLS.keySet())) {  // CHM.keySet returns KeySetView in Java 8+; false positive here
```

When I look at it in the debugger, it is indeed a KeySetView, but I'm not sure what the issue is here. Let me know if this would be better served posting in RxAndroid.

CC @JakeWharton

ZacSweers:
Interestingly, if I breakpoint this line, the rest of the app appears to run just fine and isn't blocked. As soon as I let it pass though, it crashes.

akarnokd:
Java 8 changed the return type of `keySet` and it trips AnimalSniffer as well. In theory, by compiling with javac 7, there should be no binding to Java 8's type there.

ZacSweers:
doesn't that come at the expense of being able to target Java 8 entirely though?

akarnokd:
Let's see if http://stackoverflow.com/a/32955708/61158 works for us as well.

[...]

PR #4654, opened 2016-10-01
Upcast ConcurrentHashMap to Map to avoid compatibility issue

ZacSweers:
Resovles #4653

See http://stackoverflow.com/a/32955708/61158

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/SchedulerPoolFactory.java
+++ b/src/main/java/io/reactivex/internal/schedulers/SchedulerPoolFactory.java
@@ -16,12 +16,14 @@
 
 package io.reactivex.internal.schedulers;
 
-import java.util.*;
-import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicReference;
-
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/internal/schedulers/SchedulerPoolFactory.java
```diff
@@ -16,13 +16,15 @@
...
 package io.reactivex.internal.schedulers;
 
-import java.util.*;
-import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicReference;
-
 import io.reactivex.internal.util.SuppressAnimalSniffer;
 import io.reactivex.plugins.RxJavaPlugins;
 
+import java.util.ArrayList;
```
ZacSweers: I'll  fix the imports

ZacSweers: a6a6fd6

akarnokd (approved):

PR #4660, opened 2016-10-02
Update gradle to v3.1

SUPERCILEX:


akarnokd:
Looks like the tools we use are not compatible with 3.1. Besides, the gradle files need regeneration as well.

PR #4672, opened 2016-10-05
2.x Remove Function from transformer interfaces to allow a single obj…

dalewking:
This pull request removes Function from the super type of the various Transformer interfaces. While these are technically functions, the problem is that if they all extend Function then you cannot have a single object instance that can implement multiple Transformer interfaces. The goal is to be able to call a function that returns an object that can be passed to Observable.compose or Single.Compose and so on.

This was an issue with RxLifecycle project. See this for more info: https://github.com/trello/RxLifecycle/issues/39

### DIFF
--- a/src/main/java/io/reactivex/FlowableTransformer.java
+++ b/src/main/java/io/reactivex/FlowableTransformer.java
@@ -15,14 +15,12 @@ package io.reactivex;
 
 import org.reactivestreams.Publisher;
 
-import io.reactivex.functions.Function;
-
 /**
  * Interface to compose Flowables.
  *
  * @param <Upstream> the upstream value type
  * @param <Downstream> the downstream value type
  */
-public interface FlowableTransformer<Upstream, Downstream> extends Function<Flowable<Upstream>, Publisher<? extends Downstream>> {
-
+public interface FlowableTransformer<Upstream, Downstream> {
+    Publisher<? extends Downstream> apply(Flowable<Upstream> flowable);
 }
--- a/src/main/java/io/reactivex/FlowableTransformer.java
+++ b/src/main/java/io/reactivex/FlowableTransformer.java
@@ -22,5 +22,5 @@ import org.reactivestreams.Publisher;
  * @param <Downstream> the downstream value type
  */
 public interface FlowableTransformer<Upstream, Downstream> {
-    Publisher<? extends Downstream> apply(Flowable<Upstream> flowable);
+    Publisher<? extends Downstream> apply(Flowable<Upstream> flowable) throws Exception;
 }
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -1974,7 +1974,12 @@ public abstract class Maybe<T> implements MaybeSource<T> {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <R> Maybe<R> compose(MaybeTransformer<T, R> transformer) {
-        return wrap(to(transformer));
+        try {
+            return wrap(transformer.apply(this));
+        } catch (Throwable ex) {
+            Exceptions.throwIfFatal(ex);
+            throw ExceptionHelper.wrapOrThrow(ex);
+        }
     }
 
     /**
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -942,7 +942,12 @@ public abstract class Completable implements CompletableSource {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Completable compose(CompletableTransformer transformer) {
-        return wrap(to(transformer));
+        try {
+            return wrap(transformer.apply(this));
+        } catch (Throwable ex) {
+            Exceptions.throwIfFatal(ex);
+            throw ExceptionHelper.wrapOrThrow(ex);
+        }
     }
 
     /**
--- a/src/main/java/io/reactivex/CompletableTransformer.java
+++ b/src/main/java/io/reactivex/CompletableTransformer.java
@@ -13,12 +13,10 @@
 
 package io.reactivex;
 
-import io.reactivex.functions.Function;
-
 /**
  * Convenience interface and callback used by the compose operator to turn a Completable into another
  * Completable fluently.
  */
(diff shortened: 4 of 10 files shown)


src/main/java/io/reactivex/FlowableTransformer.java
```diff
@@ -15,14 +15,12 @@
...
 /**
  * Interface to compose Flowables.
  *
  * @param <Upstream> the upstream value type
  * @param <Downstream> the downstream value type
  */
-public interface FlowableTransformer<Upstream, Downstream> extends Function<Flowable<Upstream>, Publisher<? extends Downstream>> {
-
+public interface FlowableTransformer<Upstream, Downstream> {
+    Publisher<? extends Downstream> apply(Flowable<Upstream> flowable);
```
akarnokd: We should avoid covariant return types (drop `? extends`).

src/main/java/io/reactivex/Maybe.java
```diff
@@ -1974,7 +1974,12 @@ public final T blockingGet(T defaultValue) {
      */
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <R> Maybe<R> compose(MaybeTransformer<T, R> transformer) {
-        return wrap(to(transformer));
+        try {
+            return wrap(transformer.apply(this));
```
akarnokd, line 1978: There is no need to try-catch it if it doesn't declare `throws Exception`. So either each transformer should specify `apply(...) throws Exception` or these are not needed.

src/main/java/io/reactivex/FlowableTransformer.java
```diff
@@ -15,14 +15,12 @@
...
 /**
  * Interface to compose Flowables.
  *
  * @param <Upstream> the upstream value type
  * @param <Downstream> the downstream value type
  */
-public interface FlowableTransformer<Upstream, Downstream> extends Function<Flowable<Upstream>, Publisher<? extends Downstream>> {
-
+public interface FlowableTransformer<Upstream, Downstream> {
+    Publisher<? extends Downstream> apply(Flowable<Upstream> flowable) throws Exception;
```
akarnokd, line 25: `? extends` should be removed as covariant return types are usually inconvenience to the consumer.

akarnokd (changes requested):

src/main/java/io/reactivex/FlowableTransformer.java
dalewking: It was there before in the code I changed, so I left it there. Whether it should be there or not is orthogonal to this request

akarnokd: This is also an opportunity to fix the signatures, saves 1 PR.

dalewking:
Another cleanup is that some of the compose methods use paramater name transformer and some use composer, should settle on one or the other.

src/main/java/io/reactivex/FlowableTransformer.java
dalewking: I am not certain whether the ? extends should be there or not. I know that I have long had generic type problems with compose and RxLifecycle and am not sure whether the fault is RxJava or RxLifecycle. And the results vary between Java7, Java 8, and Kotlin regarding this.

See this discussion in RxLifecycle: https://github.com/trello/RxLifecycle/issues/58

src/main/java/io/reactivex/Maybe.java
dalewking: It does throw exception see later commit

akarnokd (approved):

ZacSweers:
What's the reasoning for the exception in the signatures?

dalewking:
The reasoning for the exception was that the previous version where they implemented Function supported throwing exceptions so I wasn't trying to change the method signature (there was actually a test for it on one of the types which is why the first build failed).

For the more general reasoning for why all the functional interfaces in RxJava2 declare throws Exception see this section: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#functional-interfaces

ZacSweers:
Cool, thanks for the explanation!

ZacSweers (approved):
LGTM pending resolution of the covariant return type comment

vanniktech:
Should not a unit test be added to guarantee this behavior?

akarnokd:
It seems I have to do that myself.

vanniktech:
How do you guys feel if all of the operators (CompletableOperator, FlowableOperator, ...) would change to this behavior too for consistency and for the same reasons @dalewking already stated.

akarnokd:
They can be standalone, just don't forget the javadoc.

dalewking:
I am also thinking about this for the "to" method by changing the XxxxxTransformer interfaces to be more general purpose and usable by both compose and to. Here is what I am suggesting as applied to Observable and if it sounds good to you I will do a PR:

The new ObservableTransformer:

```
/**
 * Interface to transform Observables, used for composing and "to" transformations.
 *
 * @param <Upstream> the upstream value type
 * @param <Result> the transformation result type
 */
public interface ObservableTransformer<Upstream, Result> {
    /**
     * Applies a function to the upstream Observable and returns a new result with
     * optionally different element type.
     * @param upstream the upstream Observable instance
     * @return the transformed result
     * @throws Exception in case the transformation throws, checked exceptions will be wrapped
     * into a RuntimeException
     */
    Result apply(Observable<Upstream> upstream) throws Exception;
}
```

compose and to methods:

```
public final <R> Observable<R> compose(ObservableTransformer<T, ObservableSource<R>> composer) {
        return wrap(to(composer));
}

public final <R> R to(ObservableTransformer<T, R> converter) {
    try {
        return converter.apply(this);
    } catch (Throwable ex) {
        Exceptions.throwIfFatal(ex);
        throw ExceptionHelper.wrapOrThrow(ex);
    }
}
```

zhoutianxing:
yes

dalewking:
After having tried implementing what I showed above and seeing how many changes that required in the unit test, I am not sure it is the best approach. It may be better to introduce interfaces like `ObservableFunction<T, R>` which signature wise is like `Function<Observable<T>, R>` but not extending Function that is used for the "to" method and ObservableTransformer extends that interface, where R is `ObservableSource<DownStream>`. Interested in feedback.

PR #4674, opened 2016-10-06
2.x: add javadoc and unit test to the recently changed XTransformers

akarnokd:
This PR adds the javadoc and crash-testing for each XTransformer interface changed by #4672.

### DIFF
--- a/src/main/java/io/reactivex/FlowableTransformer.java
+++ b/src/main/java/io/reactivex/FlowableTransformer.java
@@ -23,7 +23,7 @@ import org.reactivestreams.Publisher;
  */
 public interface FlowableTransformer<Upstream, Downstream> {
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/FlowableTransformer.java
```diff
@@ -22,5 +22,13 @@
  * @param <Downstream> the downstream value type
  */
 public interface FlowableTransformer<Upstream, Downstream> {
-    Publisher<? extends Downstream> apply(Flowable<Upstream> flowable) throws Exception;
+    /**
+     * Applies a function to the upstream flowable and returns a Publisher with
```
vanniktech: nit: uppercase Flowable

akarnokd:
We may have to drop jacoco as it kills builds 1 out of 3.

JakeWharton (approved):

PR #4681, opened 2016-10-07
2.x: Fix Flowable + Single elementAt and elementAtOrError operators on empty sources

### LINKED ISSUE #4680: 2.x: ElementAt Operators in Observable and Flowable don't behave correctly on empty sources
`elementAtOrError(1)` on an empty source just leaves the new Single hanging without signalling `onError()`:

``` java
Observable.empty()
    .elementAtOrError(1)
    .test()
    .assertFailure(NoSuchElementException.class);
```

`elementAt(1)` on an empty source just leaves the Maybe hanging without signalling `onComplete()`:

```
Observable.empty()
    .elementAt(1)
    .test()
    .assertResult();
```

Just tested this against RC4. Happy to fix this.

vanniktech:
This will fix #4680
- elementAtOrError(1) on an empty source just leaves the new Single hanging without signalling onError()
- elementAt(1) on an empty source just leaves the Maybe hanging without signalling onComplete()
- I also noticed that elementAt() with a default value on an empty source did not work either.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableElementAtMaybe.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableElementAtMaybe.java
@@ -97,7 +97,7 @@ public final class FlowableElementAtMaybe<T> extends Maybe<T> implements FuseToF
         @Override
         public void onComplete() {
             s = SubscriptionHelper.CANCELLED;
-            if (index <= count && !done) {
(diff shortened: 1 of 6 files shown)


akarnokd (approved):

akarnokd:
You may have to check the build and restart it a few times if it fails with `jacocoTestReport`.

PR #4687, opened 2016-10-11
2.x: Add Observable.rangeLong & Flowable.rangeLong

### LINKED ISSUE #4683: 2.x: Observable.range(long start, long count)
I'd like to have a range of long values as an Observable however only a range method with ints do exist. Is there a reason for this?

The only way I could imagine doing this now would be by using `intervalRange` with a low period however that does not seem right.

vanniktech:
Will fix #4683
- adds Observable.rangeLong()
- adds Flowable.rangeLong()

both implementations are copied same for the unit tests.

What should we do about the range overflow check? We could use BigDecimal to check that.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -3491,8 +3491,9 @@ public abstract class Flowable<T> implements Publisher<T> {
             return just(start);
         }
 
-        if (start + (count - 1) > Long.MAX_VALUE) {
-            throw new IllegalArgumentException("Long overflow");
+        long end = start + (count - 1);
+        if (start > 0 && end < 0) {
+            throw new IllegalArgumentException("Overflow! start + count is bigger than Long.MAX_VALUE");
         }
 
         return RxJavaPlugins.onAssembly(new FlowableRangeLong(start, count));
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -3456,6 +3456,49 @@ public abstract class Flowable<T> implements Publisher<T> {
     }
 
     /**
+     * Returns a Flowable that emits a sequence of Longs within a specified range.
+     * <p>
+     * <img width="640" height="195" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/range.png" alt="">
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator honors backpressure from downstream and signals values on-demand (i.e., when requested).</dd>
+     *  <dt><b>Scheduler:</b></dt>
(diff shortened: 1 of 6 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -3456,6 +3456,49 @@ public static int bufferSize() {
...
+
+        if (count == 0) {
+            return empty();
+        }
+
+        if (count == 1) {
+            return just(start);
+        }
+
+        if (start + (count - 1) > Long.MAX_VALUE) {
```
akarnokd: That is always false. You should check if start + count - 1 flips to negative.

src/main/java/io/reactivex/internal/operators/observable/ObservableRangeLong.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        void run() {
+            if (fused) {
+                return;
+            }
+            Observer<? super Long> actual = this.actual;
+            long e = end;
+            for (long i = index; i != e && get() == 0; i++) {
+                actual.onNext((long)i);
```
akarnokd: unnecessary cast to long

src/main/java/io/reactivex/Observable.java
```diff
@@ -2993,6 +2993,46 @@ public static int bufferSize() {
...
+
+        if (count == 0) {
+            return empty();
+        }
+
+        if (count == 1) {
+            return just(start);
+        }
+
+        if (start + (count - 1) > Long.MAX_VALUE) {
```
akarnokd: Same overflow check here as in Flowable.

akarnokd (changes requested):
Please also make sure the coverage of Flowable and Observable itself remains 100%

src/main/java/io/reactivex/Flowable.java
vanniktech: I know that's why I wrote `What should we do about the range overflow check? We could use BigDecimal to check that.`

src/main/java/io/reactivex/Observable.java
vanniktech: Same as above

src/main/java/io/reactivex/Flowable.java
JakeWharton: You can do `Long.MAX_VALUE - count > start` or check just if it overflows negative like @akarnokd said.

akarnokd: See `intervalRange`:

``` java
        long end = start + (count - 1);
        if (start > 0 && end < 0) {
            throw new IllegalArgumentException("Overflow! start + count is bigger than Long.MAX_VALUE");
        }
```

vanniktech: Oh cool didn't know that. Will change it thanks.

akarnokd (approved):

PR #4692, opened 2016-10-11
2.x: coverage, fixes and cleanup 10/11-2

akarnokd:
- cover remaining `Maybe` operators
- cover `Single` operators
- cover some of the `Completable` operators
- fix missing `onSubscribe` calls
- compact a few operator implementations
- fix the order of inner Throwables on certain `CompositeException` emissions

PR #4695, opened 2016-10-12
2.x: CompositeException fix order of Exceptions

vanniktech:
After looking at #4692 I found a few more exceptions which were passed in the wrong order to CompositeException.


akarnokd:
I'm working on completable. Please skip those classes.

vanniktech:
Alright did that.

akarnokd (approved):

PR #4708, opened 2016-10-15
2.x: coverage, fixes, cleanup 10/15-1

akarnokd:
- improve coverage of `Observable` operators
- remove unnecessary code paths
- fix `publish(Function)` latecommer behavior
- remove `bufferSize` overloads of `publish` as there is no need to buffer anything for an `Observable`
- simplify `Observable.publish`

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableBuffer.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableBuffer.java
@@ -183,7 +183,7 @@ public final class ObservableBuffer<T, U extends Collection<? super T>> extends 
                 U b;
 
                 try {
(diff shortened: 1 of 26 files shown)


src/main/java/io/reactivex/internal/operators/observable/ObservableBuffer.java
```diff
@@ -185,21 +183,14 @@ public void onNext(T t) {
                 U b;
 
                 try {
-                    b = bufferSupplier.call();
+                    b = ObjectHelper.requireNonNull(bufferSupplier.call(), "onNext called with null. Null values are generally not allowed in 2.x operators and sources.");
```
JakeWharton: "onNext" --> "bufferSupplier"

JakeWharton (approved):

Issue #4715, opened 2016-10-17
Subscription#isUnsubscribed of Single keep false after onSuccess/onError


chibatching:
I updated from 1.2.0 to 1.2.1.

Until 1.2.0, `Subscription#isUnsubscribed` return `true` after `onSuccess`.
But from 1.2.1, `Subscription#isUnsubscribed` keeps `false` even if `onSuccess` / `onError` is called.

Is this a expected behavior change? If so, sorry to file rude issue.

This can reproduce with below simple code. 

``` java
public class MainActivity extends AppCompatActivity {

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        final Subscription subscription = Single.create(new Single.OnSubscribe<Boolean>() {
            @Override
            public void call(SingleSubscriber<? super Boolean> singleSubscriber) {
                try {
                    Thread.sleep(1000);
                } catch (InterruptedException e) {
                    e.printStackTrace();
                } finally {
                    singleSubscriber.onSuccess(true);
                }
            }
        })
                .subscribeOn(Schedulers.io())
                .observeOn(AndroidSchedulers.mainThread())
                .subscribe(
                        new Action1<Boolean>() {
                            @Override
                            public void call(Boolean aBoolean) {
                                Toast.makeText(MainActivity.this, "success", Toast.LENGTH_SHORT).show();
                            }
                        },
                        new Action1<Throwable>() {
                            @Override
                            public void call(Throwable throwable) {
                                Toast.makeText(MainActivity.this, "error", Toast.LENGTH_SHORT).show();
                            }
                        }
                );

        findViewById(R.id.check_subscription_button).setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View v) {
                boolean status = subscription.isUnsubscribed();
                Toast.makeText(MainActivity.this, "unsubscribed status: " + status,
                        Toast.LENGTH_SHORT).show();
            }
        });
    }
}
```

Thanks

akarnokd:
There were internal changes to `Single` that got rid of the `SafeSubscriber` and with it the reporting on `isUnsubscribed`. If you don't call `unsubscribe` it doesn't report `true` on termination. I'll try to hunt down such cases.

akarnokd:
Closing via #4716

PR #4716, opened 2016-10-17
1.x: fix subscribe(Action1 [, Action1]) to report isUnsubscribed

akarnokd:
The lambda version didn't report `isUnsubscribed()` as before because of the removal of the `SafeSubscriber` wrapping. This PR makes sure it reports consistently again.

Repored in: #4715

Note that there is no `unsafeSubscribe(SingleSubscriber)` so a regular `subscribe(SingleSubscriber)` by default won't report `isUnsubscribed() == true` unless the implementation of `onSuccess` and `onError` - controlled by the user - doesn't call `unsubscribe` on itself.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1535,19 +1535,7 @@ public class Single<T> {
      * @see <a href="http://reactivex.io/documentation/operators/subscribe.html">ReactiveX operators documentation: Subscribe</a>
      */
     public final Subscription subscribe() {
-        return subscribe(new SingleSubscriber<T>() {
-
-            @Override
-            public final void onError(Throwable e) {
-                throw new OnErrorNotImplementedException(e);
-            }
-
-            @Override
-            public final void onSuccess(T args) {
-                // do nothing
-            }
-
-        });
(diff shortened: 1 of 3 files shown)


JakeWharton (approved):

artem-zinnatullin (approved):

marwinxxii:
Hi everyone,

Seems like `subscribe(SingleSubscriber)` overload has the same problem, because `Single.just(1).subscribe(new SingleSubscriber<Integer>(){...}).isUnsubscribed()` returns `false`.

I checked version 1.2.2-SNAPSHOT. In 1.2.0 it returns `true` as expected.

akarnokd:
@marwinxxii Single has no notion of `SafeSingleSubscriber` and the internals got changed to improve performance. Therefore, you are responsible calling `unsubscribe` in your implementation of `onSuccess` and `onCompleted`.

marwinxxii:
Sorry, I missed this part in PR description.

May be this should be added to SingleSubscriber documentation (or may be `Single.subscribe(SingleSubscriber)`?

akarnokd:
PR welcome to change the docs.

marwinxxii:
Done - #4769

PR #4718, opened 2016-10-17
2.x: Distinct Operator - delegate null Collection down to onError

vanniktech:
- returning null as a Collection in the callable didn't go to onError
- adopted flowable & observable

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableDistinct.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableDistinct.java
@@ -44,7 +44,7 @@ public final class FlowableDistinct<T, K> extends AbstractFlowableWithUpstream<T
         Collection<? super K> collection;
 
         try {
-            collection = collectionSupplier.call();
+            collection = ObjectHelper.requireNonNull(collectionSupplier.call(), "The collectionSupplier returned a null collection. Null values are generally not allowed in 2.x operators and sources.");
         } catch (Throwable ex) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableDistinct.java
```diff
@@ -44,7 +44,7 @@ protected void subscribeActual(Subscriber<? super T> observer) {
         Collection<? super K> collection;
 
         try {
-            collection = collectionSupplier.call();
+            collection = ObjectHelper.requireNonNull(collectionSupplier.call(), "The returned Collection is null. Null values are generally not allowed in 2.x operators and sources.");
```
JakeWharton: This wording seems different than what's used everywhere else. Should it be "The collectionSupplier returned null."?

vanniktech: I think it should even be 'The collectionSupplier returned a null Collection. ..'

JakeWharton: I don't have a preference aside from preferring consistency. A collection supplier by definition can only return collections so it seems redundant to me.

vanniktech: True although according to usages it tends to be a pattern of `returned a null <Type>`.

akarnokd (approved):

Issue #4727, opened 2016-10-19
2.x Why is FlowableEmitter.BackpressureMode and BackpressureStrategy not merged


adipascu:
Why are the two enums not merged into one, they both serve the same purpose and both places where they are used should have the same features.

Also because 
`BackpressureStrategy.NONE` does not exist, I currently can not use a `BufferStrategy` on an `Observable`

I don't think this code is simple easy

``` java
Observable.toFlowable(BackpressureStrategy.NONE)
.onBackpressureBuffer(10, Functions.EMPTY_ACTION, BackpressureOverflowStrategy.DROP_OLDEST)
```

akarnokd:
If you can make it work, then PR welcome.

akarnokd:
Closing via #4729

PR #4729, opened 2016-10-19
Merge FlowableEmitter.BackpressureMode into BackpressureStrategy

adipascu:
This is the implementation of #4727

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -9735,12 +9735,6 @@ public abstract class Flowable<T> implements Publisher<T> {
         return RxJavaPlugins.onAssembly(new FlowableOnBackpressureLatest<T>(this));
     }
 
-    @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
-    @SchedulerSupport(SchedulerSupport.NONE)
-    /*package*/ Flowable<T>  onBackpressureError() {
-        return RxJavaPlugins.onAssembly(new FlowableOnBackpressureError<T>(this));
-    }
-
     /**
      * Instructs a Publisher to pass control to another Publisher rather than invoking
      * {@link Subscriber#onError onError} if it encounters an error.
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -1525,7 +1525,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      *
      *     emitter.setCancellable(c::close);
      *
-     * }, BackpressureMode.BUFFER);
+     * }, BackpressureStrategy.BUFFER);
      * </code></pre>
      * <p>
      * You should call the FlowableEmitter onNext, onError and onComplete methods in a serialized fashion. The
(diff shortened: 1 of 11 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -9735,6 +9735,12 @@ public final Completable ignoreElements() {
         return RxJavaPlugins.onAssembly(new FlowableOnBackpressureLatest<T>(this));
     }
 
+    @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
```
akarnokd: This is the default behavior of most Flowable sources. Unnecessary.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9735,6 +9735,12 @@ public final Completable ignoreElements() {
...
+    @SchedulerSupport(SchedulerSupport.NONE)
+    /*package*/ Flowable<T>  onBackpressureError() {
+        return RxJavaPlugins.onAssembly(new FlowableOnBackpressureError<T>(this));
```
akarnokd: Should be inlined in `Observable.toFlowable()`.

src/main/java/io/reactivex/internal/operators/flowable/FlowableCreate.java
```diff
@@ -61,10 +61,13 @@ public void subscribeActual(Subscriber<? super T> t) {
             emitter = new LatestAsyncEmitter<T>(t);
             break;
         }
-        default: {
+        case BUFFER: {
```
akarnokd: Why changing this?

src/test/java/io/reactivex/internal/operators/observable/ObservableToXTest.java
```diff
@@ -48,4 +49,31 @@ public void toFlowableLatest() {
         ts
         .assertResult(5);
     }
+
+    @Test
+    public void toFlowableError1() {
+        TestSubscriber<Integer> ts = Observable.range(1, 5)
+        .toFlowable(BackpressureStrategy.ERROR)
+        .test(1)
+        .assertError(MissingBackpressureException.class);
```
akarnokd: Why not `assertFailure(MissingBackpressureException.class, 1)` to make sure `1` got through.

src/main/java/io/reactivex/Observable.java
```diff
@@ -11707,12 +11707,18 @@ public final void subscribe(Observer<? super T> observer) {
...
-            return o.onBackpressureBuffer();
+            case DROP:
+                return o.onBackpressureDrop();
+            case LATEST:
+                return o.onBackpressureLatest();
+            case NONE:
+                return o;
+            case ERROR:
+                return o.onBackpressureError();
+            case BUFFER:
```
akarnokd: Please leave these as `default:` because otherwise the default case can't be coverage-tested.

akarnokd (changes requested):

adipascu:
I have implemented the changes from the code comments.
I'll keep in mind to rely on code coverage in the future.
Should I reply to the review comments directly ?

akarnokd (approved):

akarnokd:
No need in this case. Looks good to me.

PR #4731, opened 2016-10-19
2.x: Delegate null Collections down to onError in toList

vanniktech:
- if null from a collection supplier is returned delegate to onError
- adapted Flowable & Observable
- unified exception message

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableToListSingle.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableToListSingle.java
@@ -13,6 +13,7 @@
 
 package io.reactivex.internal.operators.flowable;
 
+import io.reactivex.internal.functions.ObjectHelper;
 import java.util.Collection;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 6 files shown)


akarnokd (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableToListSingle.java
```diff
@@ -47,7 +48,7 @@ public FlowableToListSingle(Publisher<T> source, Callable<U> collectionSupplier)
     protected void subscribeActual(SingleObserver<? super U> s) {
         U coll;
         try {
-            coll = collectionSupplier.call();
+            coll = ObjectHelper.requireNonNull(collectionSupplier.call(), "The collectionSupplier returned a null collection. Null values are generally not allowed in 2.x operators and sources.");
```
JakeWharton, line 51: Out of curiosity, is there a place where null values _are_ allowed? The use of "generally" seems to imply such.

JakeWharton (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableToListSingle.java
JakeWharton: (and yes I know we use this wording everywhere)

akarnokd: `using`'s resource supplier can return null, we don't complain about null resources.

PR #4736, opened 2016-10-20
2.x: cleanup, fixes, coverage 10/20-1

akarnokd:
- Finished covering `Observable` operators to a reasonable level (the remaining are either impossible or only reachable by probabilistic race)
- Synced operator coverage between `Observable` and `Flowable`
- Started covering `Flowable` operators
- Cleaning up a few `Flowable` operators (cache, concatMap)
- fix `onBackpressureBuffer(long, Action, BufferOverflowStrategy)` return type
- fix `concatMapDelayError` wrong barrier mode selected

### DIFF
--- a/src/test/java/io/reactivex/flowable/FlowableTests.java
+++ b/src/test/java/io/reactivex/flowable/FlowableTests.java
@@ -271,10 +271,7 @@ public class FlowableTests {
         verify(w).onNext(10);
     }
 
-    /**
-     * A reduce should fail with an NoSuchElementException if done on an empty Observable.
-     */
-    @Test(expected = NoSuchElementException.class)
+    @Test
     public void testReduceWithEmptyObservable() {
         Flowable<Integer> observable = Flowable.range(1, 0);
         observable.reduce(new BiFunction<Integer, Integer, Integer>() {
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableRangeLongTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableRangeLongTest.java
@@ -24,7 +24,8 @@ import org.junit.Test;
 import org.reactivestreams.Subscriber;
 
 import io.reactivex.*;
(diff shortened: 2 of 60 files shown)


src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -271,10 +271,7 @@ public Integer apply(Integer t1, Integer t2) {
         verify(w).onNext(10);
     }
 
-    /**
-     * A reduce should fail with an NoSuchElementException if done on an empty Observable.
-     */
-    @Test(expected = NoSuchElementException.class)
+    @Test
```
JakeWharton, line 274: Shouldn't the body of this test change too? Maybe to `test().assertComplete()` or whatever?

JakeWharton (approved):

src/test/java/io/reactivex/flowable/FlowableTests.java
akarnokd: Yeah, `.assertResult()` would be more apt here. Will do in the next batch.

src/test/java/io/reactivex/internal/operators/flowable/FlowableRangeLongTest.java
```diff
@@ -323,4 +324,216 @@ public void noOverflow() {
...
+    public void conditionalSlowPathRebatch() {
+        Flowable.rangeLong(1L, 5L)
+        .filter(Functions.alwaysTrue())
+        .rebatchRequests(1)
+        .test()
+        .assertResult(1L, 2L, 3L, 4L, 5L);
+    }
+
+    @Test
+    public void slowPathRebatch() {
```
vanniktech, line 379: how is this one different from the one above?

akarnokd: Copy-paste error, should not have filter there

Issue #4737, opened 2016-10-20
AssemblyStackTraceException doesn't account for unknown causes


pyricau:
[AssemblyStackTraceException#attachTo](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/exceptions/AssemblyStackTraceException.java) method does this:

```
    /**
     * Finds an empty cause slot and assigns itself to it.
     * @param exception the exception to start from
     */
    public void attachTo(Throwable exception) {
        Set<Throwable> memory = new HashSet<Throwable>();

        for (;;) {
            if (exception.getCause() == null) {
                exception.initCause(this);
                return;
            }

            exception = exception.getCause();
            if (!memory.add(exception)) {
                // in case we run into a cycle, give up and report this to the hooks
                RxJavaHooks.onError(this);
                return;
            }
        }
    }
```

This assumes that there are two types of exceptions: either it doesn't have a cause and we can set one, or there is a cycle in the chain and it bails.

However, if you look at the JDK Throwable implementation, you'll see this:

```
public class Throwable  {

    /**
     * The throwable that caused this throwable to get thrown, or null if this
     * throwable was not caused by another throwable, or if the causative
     * throwable is unknown.  If this field is equal to this throwable itself,
     * it indicates that the cause of this throwable has not yet been
     * initialized.
     */
    private Throwable cause = this;

   public Throwable(String message, Throwable cause) {
        fillInStackTrace();
        detailMessage = message;
        this.cause = cause;
    }

    /**
     * Initializes the <i>cause</i> of this throwable to the specified value.
     * (The cause is the throwable that caused this throwable to get thrown.)
     *
     * <p>This method can be called at most once.  It is generally called from
     * within the constructor, or immediately after creating the
     * throwable.  If this throwable was created
     * with {@link #Throwable(Throwable)} or
     * {@link #Throwable(String,Throwable)}, this method cannot be called
     * even once.
     *
     * @param  cause the cause (which is saved for later retrieval by the
     *         {@link #getCause()} method).  (A {@code null} value is
     *         permitted, and indicates that the cause is nonexistent or
     *         unknown.)
     * @return  a reference to this {@code Throwable} instance.
     * @throws IllegalArgumentException if {@code cause} is this
     *         throwable.  (A throwable cannot be its own cause.)
     * @throws IllegalStateException if this throwable was
     *         created with {@link #Throwable(Throwable)} or
     *         {@link #Throwable(String,Throwable)}, or this method has already
[...]

PR #4740, opened 2016-10-21
1.x: Error when tracking exception with unknown cause

pyricau:
#4737


akarnokd (approved):

src/main/java/rx/exceptions/AssemblyStackTraceException.java
```diff
@@ -52,7 +52,15 @@ public void attachTo(Throwable exception) {
 
         for (;;) {
             if (exception.getCause() == null) {
-                exception.initCause(this);
+                try {
+                    exception.initCause(this);
+                } catch (IllegalStateException e) {
```
pyricau, line 57: Should probably rename `e` to `ignored`

Issue #4702, opened 2016-10-13
1.x: Replaying a `Single` to future subscribers. cache() missing?


JakeWharton:
As far as I can tell, there isn't a way to replay a `Single`'s result to future subscribers without re-subscribing upstream or round-tripping through an `Observable` for `cache()`. This has come up 3 times in various channels with no good answer found.

Is `cache()` missing? Or is there some other way of doing this that I can't see?

akarnokd:
So far, nobody complained about it for 1.x. PR welcome.

akarnokd:
Closing via #4757.

PR #4757, opened 2016-10-24
1.x: Add cache() to Single

ekchang:
#4702

Uses the same test as the Observable one.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1271,7 +1271,7 @@ public class Single<T> {
     }
 
     /**
-     * Returns a Single that subscribes to this Single lazily, caches its next event
+     * Returns a Single that subscribes to this Single lazily, caches its success or error event
      * and replays it to all the downstream subscribers.
      * <p>
      * <img width="640" height="410" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/cache.png" alt="">
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -21,6 +21,7 @@ import rx.exceptions.*;
 import rx.functions.*;
 import rx.internal.operators.*;
 import rx.internal.util.*;
+import rx.observables.ConnectableObservable;
 import rx.observers.*;
 import rx.plugins.RxJavaHooks;
 import rx.schedulers.Schedulers;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/Single.java
```diff
@@ -1270,6 +1271,61 @@ public R call(Object... args) {
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code cache} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @return a Single that, when first subscribed to, caches its response for the
+     *         benefit of subsequent subscribers
+     * @see <a href="http://reactivex.io/documentation/operators/replay.html">ReactiveX operators documentation: Replay</a>
+     */
+    public final Single<T> cache() {
+        return toObservable().cacheWithInitialCapacity(1).toSingle();
```
JakeWharton, line 1326: I suppose this is a start. Ideally we would have a dedicated operator implementation that was low-overhead since we know the only thing that needs cached is a single item or complete.

src/main/java/rx/Single.java
```diff
@@ -1270,6 +1271,61 @@ public R call(Object... args) {
...
+     *  of each downstream Subscriber individually.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code cache} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @return a Single that, when first subscribed to, caches its response for the
+     *         benefit of subsequent subscribers
+     * @see <a href="http://reactivex.io/documentation/operators/replay.html">ReactiveX operators documentation: Replay</a>
+     */
+    public final Single<T> cache() {
```
JakeWharton, line 1325: New operators need `@Experimental` annotation.

src/main/java/rx/Single.java
ekchang: Yeah, I figured I'd go for the low hanging fruit of just getting a working API out there rather than try and tackle some implementation of `CachedSingle` all at once.

src/main/java/rx/Single.java
```diff
@@ -1270,6 +1271,62 @@ public R call(Object... args) {
     }
 
     /**
+     * Returns a Single that subscribes to this Single lazily, caches its next event
```
akarnokd: "Returns a Single that subscribes to a this Single lazily, caches its success or error event"

src/main/java/rx/Single.java
```diff
@@ -1270,6 +1271,62 @@ public R call(Object... args) {
...
+     *  <dd>{@code cache} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @return a Single that, when first subscribed to, caches its response for the
+     *         benefit of subsequent subscribers
+     * @see <a href="http://reactivex.io/documentation/operators/replay.html">ReactiveX operators documentation: Replay</a>
+     */
+    @Experimental
+    public final Single<T> cache() {
+        return toObservable().cacheWithInitialCapacity(1).toSingle();
```
akarnokd, line 1326: I thought @JakeWharton wanted a native `Single.cache` and not this type of reuse.

akarnokd (commented):
Javadoc needs rephrasing at least.

akarnokd (approved):

martinbonnin:
Sorry I'm late to the party here. This seems to be rxjava1.x only. Is this going to reach 2.x ?

JakeWharton:
It's in 2

martinbonnin:
Uupps, right ! Thanks !

PR #4760, opened 2016-10-24
2.x: Add Single.fromObservable(ObservableSource)

vanniktech:
- adds an overload to create a Single from an ObservableSource

### DIFF
--- a/src/test/java/io/reactivex/single/SingleTest.java
+++ b/src/test/java/io/reactivex/single/SingleTest.java
@@ -543,6 +543,11 @@ public class SingleTest {
         }).intValue());
     }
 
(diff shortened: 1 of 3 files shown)


src/test/java/io/reactivex/single/SingleTest.java
```diff
@@ -22,6 +22,7 @@
 import org.junit.*;
 
 import io.reactivex.*;
+import io.reactivex.Observable;
```
vanniktech, line 25: needed to add this import otherwise:

`SingleTest.java:562: error: reference to Observable is ambiguous Single.fromObservable(Observable.just(1))`

akarnokd (approved):

PR #4763, opened 2016-10-24
2.x: cleanup & coverage 10/24-2

akarnokd:
- better coverage of `Flowable` operators
- cleanup of operator internals
- better coverage of standard Schedulers.
- rewrite of `takeUntil` to avoid `onSubscribe()` races.

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
@@ -22,7 +22,6 @@ import java.util.concurrent.atomic.*;
 import io.reactivex.Scheduler;
 import io.reactivex.disposables.*;
 import io.reactivex.internal.disposables.EmptyDisposable;
-import io.reactivex.plugins.RxJavaPlugins;
 
 /**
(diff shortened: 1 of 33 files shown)


src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
```diff
@@ -75,23 +74,18 @@
...
-                                public void run() {
-                                    evictExpiredWorkers();
-                                }
-                            }, this.keepAliveTime, this.keepAliveTime, TimeUnit.NANOSECONDS
-                    );
-                } catch (RejectedExecutionException ex) {
-                    RxJavaPlugins.onError(ex);
-                }
+                task = evictor.scheduleWithFixedDelay(this, this.keepAliveTime, this.keepAliveTime, TimeUnit.NANOSECONDS
+                );
```
JakeWharton, line 78: nit: weird wrap

JakeWharton (approved):

src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
akarnokd: Will fix it in the next PR.

Issue #4735, opened 2016-10-20
fromEmitter same-pool deadlock


akarnokd:
This affects both 1.x and 2.x.

There is a question on [StackOverflow](http://stackoverflow.com/questions/40145203/rxjava-observable-fromemitter-odd-backpressure-behaviour) where the example program hangs after it delivered the default bufferSize elements until the `fromEmitter` source completes.

The problem is that `subscribeOn` schedules the downstream's replenishing `request` on the same thread that is busy/sleeping and has no opportunity to update its internal requested amount. We had similar problems with `onBackpressureBlock` which had to be dropped.

We could introduce a `subscribeOn` mode in which it doesn't reschedule the `request` from downstream but it will confuse some and make wonder others.

akarnokd:
This was resolved in 2.x by making sure a `subscribeOn` after `create` doesn't reschedule downstream requests.

bobvanderlinden:
I think I'm running into this problem as well. `fromEmitter` suddenly stops emitting `onNext` items whereas the emitter itself is still being called.

I was using a single-thread-scheduler for the emitter, so I thought (from your explanation) that moving to a threadpool would resolve the problem. It apparently does not for my case. I'm probably misunderstanding what is going on behind the scenes.

Do you have any suggestions on how to workaround this issue (without moving to 2.x).

akarnokd:
@bobvanderlinden Yes, use the NONE policy and apply `onBackpressureXXX` *after* `subscribeOn`.

bobvanderlinden:
@akarnokd That works like a charm! Thanks a lot!

akarnokd:
Closing via #5091 for 1.x  and #4770 for 2.x.

PR #4770, opened 2016-10-26
2.x: allow subscribeOn to work with blocking create

akarnokd:
When running a blocking emission in `Flowable.create` with `subscribeOn`, the default behavior of `subscribeOn` is to schedule downstream requests onto the same thread. Unfortunately, the blocking emission prevents the backing threadpool the `create` from accumulating that request amount leading to either buffer bloat or dropped emissions till the very end of the sequence.

This PR introduces a `nonScheduledRequests` parameter to `subscribeOn` that simply calls the upstream's `request()` from the caller's thread without scheduling it. The parameter is not exposed to the surface API but the `subscribeOn` operator does an `instanceof` check to see if the upstream type is `FlowableCreate`.

Note that this requires a direct upstream `create` and any intermediate operator re-enables the default behavior. It is possible to traverse the upstream graph to locate a `FlowableCreate` but it has relatively high cost and thus penalizing all `subscribeOn` usages.

Related: #4735

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -12251,7 +12251,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @SchedulerSupport(SchedulerSupport.CUSTOM)
     public final Flowable<T> subscribeOn(Scheduler scheduler) {
         ObjectHelper.requireNonNull(scheduler, "scheduler is null");
-        return RxJavaPlugins.onAssembly(new FlowableSubscribeOn<T>(this, scheduler));
+        return RxJavaPlugins.onAssembly(new FlowableSubscribeOn<T>(this, scheduler, this instanceof FlowableCreate));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


JakeWharton (approved):
I understand the change, but I have trouble understand the larger implications of this. Especially when you apply an operator between create and subscribeOn it seems like you're completely out of luck.

akarnokd:
options:
- ignore the issue
- teach users to have `create(..., MISSING).subscribeOn().onBackpressureDrop()` instead
- add explicit parameter to subscribeOn() and teach users
- add more extensive lookup

This impacts only those who have busy loop inside `create` instead of using `generate`. Users converting other async API or callback style sources are not affected as those don't block the create().

LalitMaganti:
This seems like a patchwork fix for what is a deeper problem which really should be fixed in a more user-visible manner. It's is very possible that users are using other operators between create and subscribeOn and hardcoding to create seems to be simply deferring the problem to those other cases.

akarnokd:
Blocking and backpressure don't play nice together. Another fix could forcefully include a `Scheduler` parameter into `create` and as such can't be overridden by applying another `subscribeOn`.

PR #4777, opened 2016-10-28
1.x: support Observable.test()

davidmoten:
The new method chaining `TestSubscriber` in 2.x is too good not to have in 1.x as well (while I wait for 2.x to mature and to make my existing unit tests less verbose).

``` java
Observable
  .just(1)
  .test()
  .assertValue(1)
  .assertCompleted();
```

`Observable.test()` returns a `TestSubscriber2` that wraps a `TestSubscriber` and enables method chaining where it can.

I've added tests to get to 100% test coverage though I've been pretty slap-dash about testing some of the details (for example I haven't tested that `awaitXXX` really does wait for stuff).

I've noted it as `@since 1.2.3` but can update to 1.2.2 if you are happy to squeak it into the release.

### DIFF
--- a/src/main/java/rx/Completable.java
+++ b/src/main/java/rx/Completable.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import rx.annotations.*;
 import rx.exceptions.*;
 import rx.functions.*;
+import rx.internal.observers.AssertableSubscriberObservable;
 import rx.internal.operators.*;
 import rx.internal.util.*;
 import rx.observers.*;
@@ -2363,5 +2364,27 @@ public class Completable {
                 });
             }
         });
+    }
+    
+    // -------------------------------------------------------------------------
+    // Fluent test support, super handy and reduces test preparation boilerplate
+    // -------------------------------------------------------------------------
+    /**
+     * Creates an AssertableSubscriber that requests {@code Long.MAX_VALUE} and subscribes
+     * it to this Observable.
+     * <dl>
+     *  <dt><b>Backpressure:</b><dt>
+     *  <dd>The returned AssertableSubscriber consumes this Observable in an unbounded fashion.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code test} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @return the new AssertableSubscriber instance
+     * @since 1.2.3
+     */
+    @Experimental
+    public final AssertableSubscriber<Void> test() {
+        AssertableSubscriberObservable<Void> ts = AssertableSubscriberObservable.create(Long.MAX_VALUE);
+        subscribe(ts);
+        return ts;
     }
 }
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -19,6 +19,7 @@ import rx.Observable.Operator;
 import rx.annotations.*;
 import rx.exceptions.*;
 import rx.functions.*;
+import rx.internal.observers.AssertableSubscriberObservable;
 import rx.internal.operators.*;
 import rx.internal.util.*;
 import rx.observables.ConnectableObservable;
@@ -2674,5 +2675,27 @@ public class Single<T> {
             throw new NullPointerException();
         }
         return create(new SingleOnSubscribeDelaySubscriptionOther<T>(this, other));
+    }
+
+    // -------------------------------------------------------------------------
+    // Fluent test support, super handy and reduces test preparation boilerplate
+    // -------------------------------------------------------------------------
+    /**
+     * Creates an AssertableSubscriber that requests {@code Long.MAX_VALUE} and subscribes
+     * it to this Observable.
+     * <dl>
+     *  <dt><b>Backpressure:</b><dt>
+     *  <dd>The returned AssertableSubscriber consumes this Observable in an unbounded fashion.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code test} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @return the new AssertableSubscriber instance
+     * @since 1.2.3
+     */
+    @Experimental
+    public final AssertableSubscriber<T> test() {
+        AssertableSubscriberObservable<T> ts = AssertableSubscriberObservable.create(Long.MAX_VALUE);
+        subscribe(ts);
+        return ts;
     }
 }
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -18,10 +18,12 @@ import java.util.concurrent.*;
 import rx.annotations.*;
 import rx.exceptions.*;
 import rx.functions.*;
+import rx.internal.observers.AssertableSubscriberObservable;
 import rx.internal.operators.*;
 import rx.internal.util.*;
 import rx.observables.*;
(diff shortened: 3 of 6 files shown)


akarnokd:
I have mixed feelings about this. 2.x required massive amount of unit tests, often with the same structure across the base types where this was a valuable addition. Maybe you could flesh it out in rxjava-extras and provide a function for the `to` operator:

``` java
import static TestConsumer.*;

Observable.just(1)
.to(newTest(0L))
.assertEmpty()
.requestMore(1)
.assertResult(1);
```

You can base it on `Subscriber` and implement `CompletableSubscriber` but I wish we introduced a base interface for SingleSubscriber before going 1.2.

MyDogTom:
Just for reference. There are at least two libraries that aim to provide similar functionality (chaining assertions):
- [RxTestWrapper](https://github.com/MyDogTom/RxTestWrapper) - just wraps `TestSubsriber` in a similar manner as this PR
- [rxassertions](https://github.com/ubiratansoares/rxassertions) - provides additional assertions, but also introduce some limitations.

davidmoten:
@akarnokd, @MyDogTom 
I have been using this functionality in _rxjava-extras_ using `.to(TestingHelper.test())` for about five weeks. It just seems too useful to miss out on natively (don't have to add extra library to pom.xml, add import, put up with slightly less readable line in code).  Discoverability is sorted for new users too.

For anyone writing tests for their observables I think this is very attractive functionality to have natively and I was stoked to use it with rxjava 2.

Note that my motivation for this is for writing tests in my projects (not rxjava 1.x core, though that would be nice too). 

I forgot to consider `Completable` and `Single` because I don't use them (the majority of our rxjava based code predates their existence). Do they need special treatment (other than addition of a `test` method there)?

akarnokd:
How about having an interface with the get and assert methods and the 3 test()s return it. In an internal package then you can implement them and have delegates as necessary.

davidmoten:
> How about having an interface with the get and assert methods and the 3 test()s return it. In an internal package then you can implement them and have delegates as necessary.

Yep I can do that.

davidmoten:
> How about having an interface with the get and assert methods and the 3 test()s return it. In an internal package then you can implement them and have delegates as necessary.

Having extracted an interface the one implementation can serve all `test()` methods by subscribing to `Completable`, `Single` or `Observable` instances using `subscribe(Subscriber)`. How do you want the implementations to differ?

src/main/java/rx/Completable.java
```diff
@@ -16,19 +16,48 @@
 
 package rx;
 
-import java.util.*;
-import java.util.concurrent.*;
+import java.util.Arrays;
```
akarnokd: Please make sure you don't unroll `*` imports.

src/main/java/rx/observers/TestSubscriber2.java
```diff
@@ -18,184 +18,70 @@
...
-    public TestSubscriber2<T> assertValues(T... values) {
-        ts.assertValues(values);
-        return this;
-    }
-
-    public TestSubscriber2<T> assertValue(T value) {
-        ts.assertValue(value);
-        return this;
-    }
+public interface TestSubscriber2<T> extends Observer<T> {
```
akarnokd: I'd call this `TestConsumer` because that `2` at the end will make people wonder and question endlessly.

src/main/java/rx/observers/TestSubscriber2.java
```diff
@@ -18,184 +18,70 @@
...
-    }
-
-    @Override
-    public String toString() {
-        return ts.toString();
-    }
+    void unsubscribe();
     
-}
+    boolean isUnsubscribed();
```
akarnokd: Reinforce with `implements Subscription`?

akarnokd (changes requested):

MyDogTom:
I think `test()` should be added also for `BlockingObservable`.

akarnokd:
Unnecessary. TestConsumer should have awaitTerminalEvent that blocks.

davidmoten:
TestConsumer maybe not ideal because sounds like it's a significantly different beast from a TestSubscriber. I prefer to retain the word Subscriber in the name, what about TestSubscriberFluent?

akarnokd:
`AssertableSubscriber`

MyDogTom:
+1 for `TestSubscriberFluent` because it reflects the main difference from `TestSubscriber` - fluent API
 -1 for `AssertableSubscriber` it doesn't provide any additional info (`TestSubscriber` is also "assertable").

@akarnokd I still think that `BlockingObservable` should have `test()` method. Let's assume that I encounter some legacy code that use `BlockingObservable` (regardless it's not recommended). I would like to be able write test first and refactor it afterwards (get rid of `BlockingObservable`).

davidmoten:
@MyDogTom  puts it well I think, after all `AssertableSubscriber` would provoke as many "what's the difference?" questions as `TestSubscriber2`. `TestSubscriberFluent` describes exactly what the class is.

davidmoten:
fixed imports, renamed `TestSubscriber2` to `TestSubscriberFluent` (though I'm not assuming the naming discussion has finished).

squashed commits and rebased on latest 1.x

akarnokd:
`AssertableSubscriber` should be an interface with all the relevant `assert` methods that any actual internal test class can implement on top of the respective receiver type. 

> I think, after all AssertableSubscriber would provoke as many "what's the difference?"

If this comes up we will answer that it is the base API interface for asserting a subscriber's state for all 3 reactive types.

I see confusion going the other way around: why `TestSubscriber2` or `TestSubscriberFluent`; its prefix `TestSubscriber` matches, well, `TestSubscriber` itself.

davidmoten:
Righto, let's go with `AssertableSubscriber`. Modified code and squashed commits

src/main/java/rx/Completable.java
```diff
@@ -2364,4 +2365,45 @@ public void call() {
...
+     *  <dt><b>Backpressure:</b><dt>
+     *  <dd>The returned AssertableSubscriber requests the given {@code initialRequest} amount upfront.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code test} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @return the new AssertableSubscriber instance
+     * @since 1.2.3
+     */
+    @Experimental
+    public final AssertableSubscriber<Void> test(long initialRequestAmount) {
```
akarnokd: `Completable` doesn't need backpressure so this method has no use.

src/main/java/rx/Single.java
```diff
@@ -2675,4 +2676,45 @@ public void call(SingleSubscriber<? super T> singleSubscriber) {
...
+     *  <dt><b>Backpressure:</b><dt>
+     *  <dd>The returned AssertableSubscriber requests the given {@code initialRequest} amount upfront.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code test} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @return the new AssertableSubscriber instance
+     * @since 1.2.3
+     */
+    @Experimental
+    public final AssertableSubscriber<T> test(long initialRequestAmount) {
```
akarnokd: Single doesn't support backpressure either.

akarnokd (changes requested):

davidmoten:
Thanks @akarnokd, made fixes and squashed commits.

akarnokd (approved):

Issue #4762, opened 2016-10-24
No subscriber methods are called at all, when having npe because of illegal switchIfEmpty usage


sobersanta:
The following code consists obvious error: null is passed to switchIfEmpty (actually it's a common mistake to use switchIfEmpty instead of defaultIfEmpty).
But the problem is that the consequent NPE is not emitted to subscribers. Actually, none of the subscirber's methods is called: neither onNext, nor onError or onCompleted. This code does not produce any output to the system.out!

That of course could be fixed by simple null verification on the switchIfEmpty operator creation.
But probably its worth looking more deep into the sequence of producers in this chain, and something could be improved?
This situation caused by the described behavior, when no Observable's events are raised on the event loop,  was quite difficult to find and costed us few nervous cells :)

I guess its worth investigating and fixing.

```
// problematic code

Observable.just(null)
        .filter(Objects::nonNull)
        .flatMap(o -> Observable.just(1))
        .switchIfEmpty(null)
        .subscribe(o -> System.out.println(o)
                , Throwable::printStackTrace
                , () -> System.out.println("completed"));

```

akarnokd:
Would you like to submit a PR?

sobersanta:
I'm not ready to provide a PR - probably just a simple null verification on the switchIfEmpty creation.
I think its worth a deeper investigation by someone from the core team.

akarnokd:
I don't have the capacity to investigate this right now. Indeed there is a null check missing in `Observable.switchIfEmpty()`.

moontails:
@akarnokd so what would this entail? Does it just need a null check and throw an NPE?

akarnokd:
The fix is trivial, yes. Figuring out why the NPE never popped out from the internals is another story.

akarnokd:
Closing via #4791

PR #4791, opened 2016-10-31
1.x: Add null check to Observable.switchIfEmpty

ortex:
#4762


JakeWharton (approved):
A test would be lovely

akarnokd (changes requested):
Please add a unit test.

ortex:
@akarnokd need more tests?

akarnokd (approved):

PR #4792, opened 2016-10-31
Fixed detection of TestObserver.assertNotSubscribed()

### LINKED ISSUE #4790: TestObserver.assertNotSubscribed() can never become false after subscription
Test demonstrates the issue:

```
@Test
public void test() throws Exception {
  TestObserver<String> testObserver = new TestObserver<>();
  testObserver.assertNotSubscribed();

  PublishSubject<String> subject = PublishSubject.create();
  subject.subscribe(testObserver);
  testObserver.assertSubscribed();

  testObserver.dispose();
  testObserver.assertNotSubscribed();  // <-- assertion fails
}
```

I would expect that after calling `dispose()` that it would not be subscribed anymore, but since `TestObserver.assertNotSubscribed()` only checks nullity it doesn't realize that it's actually been set to `DisposableHelper.DISPOSED`.

dlew:
It wasn't working post-subscription, but now it checks for DISPOSED

Fixes #4790

### DIFF
--- a/src/main/java/io/reactivex/observers/TestObserver.java
+++ b/src/main/java/io/reactivex/observers/TestObserver.java
@@ -255,7 +255,8 @@ implements Observer<T>, Disposable, MaybeObserver<T>, SingleObserver<T>, Complet
      */
     @Override
     public final TestObserver<T> assertNotSubscribed() {
-        if (subscription.get() != null) {
+        Disposable disposable = subscription.get();
+        if (disposable != null && disposable != DisposableHelper.DISPOSED) {
             throw fail("Subscribed!");
         } else
         if (!errors.isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/observers/TestObserver.java
```diff
@@ -255,7 +255,8 @@ public final boolean hasSubscription() {
      */
     @Override
     public final TestObserver<T> assertNotSubscribed() {
-        if (subscription.get() != null) {
+        Disposable disposable = subscription.get();
+        if (disposable != null && disposable != DisposableHelper.DISPOSED) {
```
vanniktech, line 259: If this gets changed [TestSubscriber](https://github.com/ReactiveX/RxJava/blob/ff282b9647d0accc0cb55780c9bdb97f4ed03053/src/main/java/io/reactivex/subscribers/TestSubscriber.java#L316) should be changed too for consistency

akarnokd:
It was meant to check if the `onSubscribe` has not been called or not but not for detecting cancellation.

dlew:
Ah, so you're just supposed to use `assertTrue(testObserver.isDisposed())` then?

akarnokd:
If you dispose then yes.

PR #4826, opened 2016-11-09
Schedule when bug fix

abersnaze:
In using this in production we found a bug where some actions were dropped.  I've tracked it down to the premature onCompleted and unsubscription of the completable that represents the scheduled action causes the future to get canceled before it is started.

The fix was to delay the onCompleted until the action was truly done.

I did the PR as two commits because the first one is changing all the tabs to spaces! Look at the second commit for the actual functional difference.

### DIFF
--- a/src/main/java/rx/internal/schedulers/SchedulerWhen.java
+++ b/src/main/java/rx/internal/schedulers/SchedulerWhen.java
@@ -143,8 +143,7 @@ public class SchedulerWhen extends Scheduler implements Subscription {
                     @Override
                     public void call(CompletableSubscriber actionCompletable) {
                         actionCompletable.onSubscribe(action);
-                        action.call(actualWorker);
-                        actionCompletable.onCompleted();
+                        action.call(actualWorker, actionCompletable);
                     }
                 });
             }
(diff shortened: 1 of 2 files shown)


src/main/java/rx/internal/schedulers/SchedulerWhen.java
```diff
@@ -104,195 +104,213 @@
...
+        @Override
+        public boolean isUnsubscribed() {
+            return false;
+        }
+    };
+
+    static final Subscription UNSUBSCRIBED = Subscriptions.unsubscribed();
+
+    @SuppressWarnings("serial")
+    private static abstract class ScheduledAction extends AtomicReference<Subscription>implements Subscription {
```
akarnokd, line 209: `private` creates extra accessors in an Android unfriendly way. Could you change these into package-private?

akarnokd (approved):

src/main/java/rx/internal/schedulers/SchedulerWhen.java
abersnaze: Do you want the changes to package-private in a new PR? For both 1.x and 2.x?

PR #4827, opened 2016-11-09
Porting the Scheduler.when operator from 1.x to 2.x

abersnaze:
In fixing the Scheduler.when in 1.x I noticed that it hadn't been ported to 2.x. This PR tries to fix that translating Observable to Flowable and Subscription to Disposable. This also includes the fix from 1.x

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/SchedulerWhen.java
+++ b/src/main/java/io/reactivex/internal/schedulers/SchedulerWhen.java
@@ -1,0 +1,319 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.reactivex.internal.schedulers;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.reactivex.Completable;
+import io.reactivex.CompletableObserver;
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/internal/schedulers/SchedulerWhen.java
```diff
@@ -0,0 +1,328 @@
...
+    public boolean isDisposed() {
+        return disposable.isDisposed();
+    }
+
+    @Override
+    public Worker createWorker() {
+        final Worker actualWorker = actualScheduler.createWorker();
+        // a queue for the actions submitted while worker is waiting to get to
+        // the subscribe to off the workerQueue.
+        ReplaySubject<ScheduledAction> actionSubject = ReplaySubject.create();
```
akarnokd: You can use a `ReplayProcessor` instead.

src/main/java/io/reactivex/internal/schedulers/SchedulerWhen.java
```diff
@@ -0,0 +1,328 @@
...
+        final Observer<ScheduledAction> actionObserver = new SerializedObserver<ScheduledAction>(actionSubject);
+        // convert the work of scheduling all the actions into a completable
+        Flowable<Completable> actions = actionSubject.toFlowable(BackpressureStrategy.BUFFER).map(new Function<ScheduledAction, Completable>() {
+            @Override
+            public Completable apply(final ScheduledAction action) {
+                return Completable.unsafeCreate(new CompletableSource() {
```
akarnokd: No need for such indirection, just extend `Completable`:

``` java
return new Completable() {
  @Override
  public void subscribeActual(CompletableObserver actionCompletable) {
    actionCompletable.onSubscribe(action);
    action.call(actualWorker, actionCompletable);
   }
};
```

akarnokd (approved):

src/main/java/io/reactivex/internal/schedulers/SchedulerWhen.java
```diff
@@ -0,0 +1,328 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.reactivex.internal.schedulers;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.openjdk.jmh.runner.RunnerException;
```
akarnokd: This is not available in the main build.
https://travis-ci.org/ReactiveX/RxJava/builds/174626634#L169

PR #4832, opened 2016-11-10
2.x: add doFinally to the rest of the reactive base classes

akarnokd:
This PR adds the `doFinally` operator to the rest of the base reactive classes.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/completable/CompletableDoFinally.java
+++ b/src/main/java/io/reactivex/internal/operators/completable/CompletableDoFinally.java
@@ -1,0 +1,106 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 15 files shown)


src/main/java/io/reactivex/internal/operators/completable/CompletableDoFinally.java
```diff
@@ -0,0 +1,106 @@
...
+            runFinally();
+        }
+
+        @Override
+        public boolean isDisposed() {
+            return d.isDisposed();
+        }
+
+        void runFinally() {
+            if (compareAndSet(0, 1)) {
```
JakeWharton, line 96: Is there reasoning behind why int over boolean was used?

JakeWharton (approved):

src/main/java/io/reactivex/internal/operators/completable/CompletableDoFinally.java
akarnokd: So the pattern stays the same. Internally, AtomicBoolean does the same thing.

JakeWharton: The half serialization pattern?

akarnokd: This is an "enter once" pattern.

PR #4833, opened 2016-11-10
2.x: add `Flowable.doAfterNext` operator + 3 doX TCKs

akarnokd:
This PR adds the `Flowable.doAfterNext` operator which calls a `Consumer` after the call to `downstream.onNext()` with the same value.

In addition, `doOnNext`, `doAfterNext` and `doFinally` received their Reactive-Streams TCK tests.

### DIFF
--- a/src/test/java/io/reactivex/tck/DoOnNextTckTest.java
+++ b/src/test/java/io/reactivex/tck/DoOnNextTckTest.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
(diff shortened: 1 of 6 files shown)


src/test/java/io/reactivex/tck/DoOnNextTckTest.java
```diff
@@ -0,0 +1,31 @@
...
+import io.reactivex.Flowable;
+import io.reactivex.internal.functions.Functions;
+
+@Test
+public class DoOnNextTckTest extends BaseTck<Integer> {
+
+    @Override
+    public Publisher<Integer> createPublisher(long elements) {
+        return FlowableTck.wrap(
+                Flowable.range(0, (int)elements).doOnNext(Functions.emptyConsumer())
```
JakeWharton, line 28: I don't really understand the purpose of these three tests. If the side-effecting operator just swallows the event it's not really validating anything.

JakeWharton (approved):

src/test/java/io/reactivex/tck/DoOnNextTckTest.java
akarnokd: These test verify that the operators and the chain still honors the RS specification. The regular JUnit test verify if the operator does what it supposed to do.

PR #4839, opened 2016-11-11
2.x: javadoc: fix wording of some operators, remove @throws implications

akarnokd:
This PR fixes some javadoc mistakes with operators such as `single`, `first`, `last` and co and removes `@throws` tags from operators since those don't throw it when the method is invoked.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -7718,8 +7718,8 @@ public abstract class Flowable<T> implements Publisher<T> {
     }
 
     /**
-     * Returns a Maybe that emits the single item at a specified index in a sequence of emissions from a
-     * source Publisher.
+     * Returns a Maybe that emits the single item at a specified index in a sequence of emissions from
+     * this Flowable or completes if this Flowable sequence has fewer elements than index.
      * <p>
      * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/elementAt.png" alt="">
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -7746,7 +7746,7 @@ public final void blockingSubscribe(Subscriber<? super T> subscriber) {
     }
 
     /**
-     * Returns a Flowable that emits the item found at a specified index in a sequence of emissions from a
+     * Returns a Flowable that emits the item found at a specified index in a sequence of emissions from
```
JakeWharton, line 7749: Revert this one?

JakeWharton: Actually or use "this Flowable" like the others

src/main/java/io/reactivex/Observable.java
```diff
@@ -11344,6 +11328,7 @@ public final void subscribe(Observer<? super T> observer) {
      *         within time windows defined by the timeout selectors
      * @throws NullPointerException
      *             if {@code itemTimeoutIndicator} is null
+     *             if {@code other} is null
```
JakeWharton, line 11331: You might want to look at the rendered Javadoc here. I'm betting these get smashed together with only a single space. Might be better to switch to `{@code itemTimeoutIndicator} or {@code other} is null`.

JakeWharton (approved):

akarnokd:
Thanks @JakeWharton, fixed those problems.

Issue #4756, opened 2016-10-23
1.x and 2.x Proposal: Send a subclass of CancellationException in Single.takeUntil()


ZacSweers:
We've run into a case where a `CancellationException` is actually emitted somewhere else in the chain, so checking the throwable's class type isn't enough to know if it was from `takeUntil()` or something else. I'd like to propose a subclass of it, like `TakeUntilCancellation` to make the type explicit. Thoughts?

akarnokd:
Why does it matter?

ZacSweers:
If you use takeUntil() for lifecycle handling, most of the time you want to ignore the cancellation in that event because it just means the lifecycle ended. Similar to how most just ignore the onCompleted event for Observable's takeUntil(). Part of the original PR with it was to be able to better differentiate what caused the error event, and this just takes it a step further. I'll make a PR since you've added the label for it :). Any preference on where that exception class should live?

akarnokd:
Before you do, why not have your source of "until" signal the exception of your choice (instead of signalling `onNext` or `onCompleted`) or map a value to exception:

``` java
PublishSubject<Integer> until = PublishSubject.create();

singleSource.takeUntil(until);

until.onError(new MyCancellation());

// or

singleSource.takeUntil(until.map(v -> { throw new MyCancellation(); }));
```

It also gives you the opportunity to know where the until was triggered as the stacktrace will point to your java file instead of `takeUntil`.

Second, instead of introducing a new exception type, you could just add a message `Single::takeUntil` to the current `CancellationException` and check for that message.

ZacSweers:
Good idea on the message, though I'm not sure how that would fit into the takeUntil() API

Regarding using a subject - most of these `lifecycle` Observables are backed by relays and never emit error events. Even if they were subjects, I don't think making them emit onError is a good solution, especially since this issue only applies to `Single`/`Completable`, whereas `Observable`s bound to that lifecycle have up to this point not expected or needed to handle errors

JakeWharton:
It wouldn't be exposed. It just would be a standard message.

ZacSweers:
That sounds good to me.

akarnokd:
See #4852

ZacSweers:
Going to close this since it has a message now

PR #4852, opened 2016-11-15
1.x: Single::takeUntil CancellationException message enh.

akarnokd:
This PR modifies the `CancellationException` message to include the operator name and source type.

Relate: #4756


ZacSweers (approved):
Nice

JakeWharton (approved):
Not a fan of this, but it's better than adding a new public exception type.

Issue #4867, opened 2016-11-18
2.x: assertForEach in TestSubscriber


erickok:
I am a happy user of `TestSubscriber` (via the very elegant `.test()` 'operator') in RxJava2. I especially like the fluent interface, such that we can write:
```java
Flowable.just(1, 2, 3)
		.filter(integer -> integer > 1)
		.test()
		.assertValues(2, 3);
```

But sometimes I do not which to test for specific values, rather that all values adhere to some predicate. As there is no such `assertXxx` method, I have to either make assertions as a side effect (such as via a `doOnNext`) or grab the result and test it, such as via `values()`, which breaks the fluent interface.

Is it perhaps an idea to add an `assertForEach` on `TestSubscriber`, such that we can write:
```java
Flowable.range(1, 100)
		.filter(integer -> integer > 1)
		.test()
		.assertForEach(integer -> integer > 1 && integer <= 100);
```

JakeWharton:
Does this block until a terminal event then?

erickok:
It would work as `assertValues`. An example implementation (as would be found in `BaseTestConsumer`):

``` java
public final U assertForEach(Predicate<T> forEachPredicate) {
    for (T v : this.values) {
        try {
            if (!forEachPredicate.test(v)) {
                throw fail("Value " + valueAndClass(v) + " fails predicate assertion");
            }
        } catch (Exception ex) {
            throw ExceptionHelper.wrapOrThrow(ex);
        }
    }
    return (U)this;
}
```

JakeWharton:
Got it. I would prefer a different name since forEach in both streams and
Rx consume their entirety. Might also be worth considering passing the
index of the value under test to the predicate.

akarnokd:
I'd suggest exhausting the available options first: doOnNext, operator all
and the assertResult(true), assertOf and a logic wrapper function like your
for loop over values().
1. nov. 18. du. 4:00 ezt írta ("Jake Wharton" <notifications@github.com
   >):

Got it. I would prefer a different name since forEach in both streams and
Rx consume their entirety. Might also be worth considering passing the
index of the value under test to the predicate.

On Fri, Nov 18, 2016 at 9:54 AM Eric Kok notifications@github.com wrote:

> It would work as assertValues. An example implementation (as would be
> found in BaseTestConsumer):
> 
> public final U assertForEach(Predicate<T> forEachPredicate) {
> for (T v : this.values) {
> try {
[...]

PR #4868, opened 2016-11-18
Added assertForEachValue for TestSubscriber usage

erickok:
Example implementation for a proposed `testForEachValue` method, rf #4867, which can be applied to fluently test whether all emitted values adhere to some predicate.

If there is interest and the project maintainers/contributors see use in its addition, I will also write tests for the method.


akarnokd:
I'm still not convinced we'd need this test method; you can inline this into the sequence being tested or use `assertOf` with an utility lambda that does this for your cases. Don't be afraid creating your own tools and wrappers around RxJava components.

akarnokd:
I'm closing this issue due to inactivity. If you have further input on the issue, don't hesitate to reopen this issue or post a new one.

Issue #4878, opened 2016-11-23
2.x: @CheckReturnValue for public API


vanniktech:
How do you guys feel if the `@CheckReturnValue` annotation was added to certain public API methods?

The one that would probably benefit the most in my opinion would be the `subscribe` method where you get a `Subscription` or `Disposable` back. In 99% of the cases you want to handle it.

Also static analyze tools such as [Error Prone](http://errorprone.info/bugpattern/CheckReturnValue) could emit an error / warning.

akarnokd:
I frequently fire up `subscribe(Consumer)` and don't care about cancellation so it would just bother me. Also `@CheckReturnValue` sounds like part of some extra dependency and we don't want that. The only dependency is the Reactive-Streams API.

vanniktech:
Yes either get the annotation from JSR 305 or I believe that creating our own annotation with the same name would also work.

akarnokd:
I think `subscribeWith` is a reasonable target here as its purpose is to give you back what you gave to it so you can continue with that value. Also methods returning `Flowable` etc. shouldn't be ignored. If you can make it without external dependencies, I'm fine with the extra annotations.

vanniktech:
Alright will give it a try soon.

vanniktech:
Closing via #4881

markelliot:
What's the rationale behind forcing handling of a `Disposable` return value? In most of the cases I've encountered I've preferred a setup where my subscription runs indefinitely. Is it unsafe to ignore the return here? If so, why?

akarnokd:
@markelliot It was decided anything that returns `Disposable` should not have `@CheckReturnValue`. Did you find method(s) that do?

markelliot:
I'm on 2.0.5 and the source for subscribe looks like this:

```java
    @CheckReturnValue
    @SchedulerSupport(SchedulerSupport.NONE)
    public final Disposable subscribe(Consumer<? super T> onNext) {
        return subscribe(onNext, Functions.ERROR_CONSUMER, Functions.EMPTY_ACTION, Functions.emptyConsumer());
    }
```

markelliot:
Can't link to the source since GH is truncating at ~10k lines, but it looks like on master the code is the same.

markelliot:
#5244

vanniktech:
Why would you want to not handle the returned Disposable?

markelliot:
[...]

PR #4881, opened 2016-11-24
Add CheckReturnValue annotation

vanniktech:
First draft on this.

So far I have only annotated `subscribeWith()` and `test()` methods.

I checked this against Error Prone and their [check](http://errorprone.info/bugpattern/CheckReturnValue) is able of picking up the RxJava annotation.

How do you feel if I extend `BaseTypeAnnotations` to test for `@CheckReturnValue` too?

Related #4878

### DIFF
--- a/src/main/java/io/reactivex/annotations/CheckReturnValue.java
+++ b/src/main/java/io/reactivex/annotations/CheckReturnValue.java
@@ -21,7 +21,7 @@ import java.lang.annotation.Target;
 
 /**
  * Marks methods whose return values should be checked.
- * 
+ *
  * @since 2.0.2 - experimental
  */
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/main/java/io/reactivex/annotations/CheckReturnValue.java
+++ b/src/main/java/io/reactivex/annotations/CheckReturnValue.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks methods whose return values should be checked.
+ * 
+ * @since 2.0.2 - experimental
+ */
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/annotations/CheckReturnValue.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks methods whose return values should be checked.
+ * 
+ * @since 2.0.2 - experimental
+ */
+@Retention(RetentionPolicy.RUNTIME)
```
artem-zinnatullin, line 27: nit: static import makes it a little bit better

src/main/java/io/reactivex/annotations/CheckReturnValue.java
```diff
@@ -0,0 +1,33 @@
...
+@Documented
+@Target(ElementType.METHOD)
```
artem-zinnatullin, line 29: And here

artem-zinnatullin (approved):
Gooooooood. Afaik Findbugs also able to process annotation with this name correctly.

src/main/java/io/reactivex/annotations/CheckReturnValue.java
vanniktech: Yes I know however for RxJava it's common to not static import them (look at BackpressureSupport & SchedulerSupport) so I left it that way. I leave the final decision up to David.

src/main/java/io/reactivex/annotations/CheckReturnValue.java
vanniktech: same as above

vanniktech:
I quickly checked findbugs and it does not seem like they pick up our annotation since they have their [own](http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/CheckReturnValue.html)

src/main/java/io/reactivex/annotations/CheckReturnValue.java
akarnokd: It's fine as is.

akarnokd (approved):

akarnokd:
Would you manually ignore methods like `Disposable subscribe()`?

artem-zinnatullin:
@vanniktech see https://github.com/findbugsproject/findbugs/blob/264ae7baf890d2b347d91805c90057062b5dcb1e/findbugs/src/java/edu/umd/cs/findbugs/detect/BuildCheckReturnAnnotationDatabase.java#L120

vanniktech:
@artem-zinnatullin weird somehow findbugs didn't pick it up though. Could be a configuration error on my side.

I'd argue that not annotating `Diposable subscribe()` is fine however I feel like `Disposable subscribe(Consumer<T>)` and the others should be annotated. I personally almost never use subscribeWith and prefer the variations with lambdas.

We're on the same point that static factory methods and operators such as flatMap, first(), last() etc should be annotated right? If so I'd apply those and then we can see which might need an annotation too.

vanniktech:
@akarnokd added the `@CheckReturnValue` annotation to all public API methods of `Flowable`, `Maybe`, `Single`, `Observable` and `Completable` except for private methods, void methods and the `Disposable subscribe()` method. Is that okay now?

akarnokd (approved):

akarnokd:
Yes.

/cc @JakeWharton

Issue #4883, opened 2016-11-25
TestScheduler and/or TestSubscriber doesn't work for nanoseconds


piotrturski:
when millisecond resolution is used, everything works correctly. but the same test fails when all units are changed to nanosecond. the following test fails but should pass

``` 
import org.junit.Test;
import rx.Observable;
import rx.observers.TestSubscriber;
import rx.schedulers.TestScheduler;
import java.util.concurrent.TimeUnit;

public class NanosecondTest {

    @Test
    public void shoud_schedule_correctly_with_nanosecond_resolution() {

        TestScheduler scheduler = new TestScheduler();
        TestSubscriber<Long> testSubscriber = new TestSubscriber<>();

        TimeUnit unit = TimeUnit.NANOSECONDS; // change to TimeUnit.MILLISECONDS and test will pass

        Observable.interval(30, unit, scheduler).subscribe(testSubscriber);
        scheduler.advanceTimeTo(60, unit);

        testSubscriber.assertValues(0L, 1L);
    }
}
```

`compile 'io.reactivex:rxjava:1.2.2'`

piotrturski:
with `MICROSECONDS` also doesn't work

akarnokd:
Did you try with 2.x?

piotrturski:
nope, i stick to 1.x, on 2.x it doesn't even compile (after fixing imports) so i haven't tried

akarnokd:
Closing via #4884.

PR #4884, opened 2016-11-25
1.x: enable TestScheduler with nanosecond periodic scheduling

akarnokd:
The default periodic scheduling code accessed the current worker time as milliseconds which prevents running microsecond and nanosecond scale unit tests with TestScheduler as its internal nanosecond resolution time is always converted, losing precision.

This PR introduces the internal `SchedulePeriodicHelper` class, moves the default scheduling code into a static method and defines a `NowNanoSupplier` interface that is now implemented by TestSchedulerWorker to return the current nano time. The default `Worker.schedulePeriodically` delegates to this static method and given a `null` `noNanoSupplier` the `Worker.now()` is converted to nanoseconds as usual.

Related: #4883

### DIFF
--- a/src/test/java/rx/schedulers/TestSchedulerTest.java
+++ b/src/test/java/rx/schedulers/TestSchedulerTest.java
@@ -17,25 +17,18 @@ package rx.schedulers;
 
 import static org.junit.Assert.assertEquals;
 import static org.mockito.Matchers.anyLong;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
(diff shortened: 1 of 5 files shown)


src/test/java/rx/schedulers/TestSchedulerTest.java
```diff
@@ -222,4 +215,24 @@ public void call() {
             inner.unsubscribe();
         }
     }
+    
+    @Test
+    public void resolution() {
+        for (final TimeUnit unit : TimeUnit.values()) {
```
artem-zinnatullin, line 221: :+1:

artem-zinnatullin (approved):

PR #4885, opened 2016-11-26
2.x: dedicated reduce() op implementations

akarnokd:
This PR adds dedicated operator implementations to

  - `Flowable.reduce(seed, reducer)` (returns `Single`),
  - `Flowable.reduceWith(seedSupplier, reducer)`  (returns `Single`),
  - `Observable.reduce(reducer)` (returns `Maybe`),
  - `Observable.reduce(seed, reducer)`  (returns `Single`) and
  - `Observable.reduceWith(seedSupplier, reducer)`  (returns `Single`)

instead of using `scan().takeLast(1)` (`Flowable.reduce(reducer)` already had a dedicated operator).

Comparison (Celeron 1005M, 4GB RAM, Windows 7 x64, Java 8u112, JMH 1.16):

image

The new `ReducePerf` benchmark does a simple sum over a list of integer values. Unfortunately, this creates a lot of garbage for longer sequences (plus the CPU/RAM is not really suited for such benchmarks; the `flowMaybe` lines should be roughly the same since the code didn't change but there is a significant run-to-run variance).

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableReduceSeedSingle.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableReduceSeedSingle.java
@@ -79,13 +79,14 @@ public final class FlowableReduceSeedSingle<T, R> extends Single<R> {
         @Override
         public void onNext(T value) {
             R v = this.value;
-
-            try {
-                this.value = ObjectHelper.requireNonNull(reducer.apply(v, value), "The reducer returned a null value");
-            } catch (Throwable ex) {
-                Exceptions.throwIfFatal(ex);
-                s.cancel();
-                onError(ex);
+            if (v != null) {
+                try {
+                    this.value = ObjectHelper.requireNonNull(reducer.apply(v, value), "The reducer returned a null value");
+                } catch (Throwable ex) {
+                    Exceptions.throwIfFatal(ex);
+                    s.cancel();
+                    onError(ex);
+                }
             }
         }
 
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableReduceSeedSingle.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableReduceSeedSingle.java
@@ -1,0 +1,126 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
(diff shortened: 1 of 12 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableReduceSeedSingle.java
```diff
@@ -0,0 +1,126 @@
...
+        
+        R value;
+        
+        Subscription s;
+
+        public ReduceSeedObserver(SingleObserver<? super R> actual, BiFunction<R, ? super T, R> reducer, R value) {
+            this.actual = actual;
+            this.value = value;
+            this.reducer = reducer;
+            this.value = value;
```
jakeloo: 2x `this.value = value`? L64 & L66.

akarnokd: Thanks, fixing...

src/main/java/io/reactivex/internal/operators/observable/ObservableReduceMaybe.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        @Override
+        public void onError(Throwable e) {
+            if (done) {
+                RxJavaPlugins.onError(e);
+                return;
+            }
+            value = null;
```
davidmoten, line 98: needs `done=true`

src/main/java/io/reactivex/internal/operators/observable/ObservableReduceMaybe.java
```diff
@@ -0,0 +1,122 @@
...
+            actual.onError(e);
+        }
+
+        @Override
+        public void onComplete() {
+            T v = value;
```
davidmoten, line 108: check `done`

davidmoten: I guess the real problem here is you don't have those `done` related unit tests that I said I'd provide! I'll have a look at them soon.

src/main/java/io/reactivex/internal/operators/observable/ObservableReduceSeedSingle.java
```diff
@@ -0,0 +1,118 @@
...
+            if (DisposableHelper.validate(this.d, d)) {
+                this.d = d;
+
+                actual.onSubscribe(this);
+            }
+        }
+
+        @Override
+        public void onNext(T value) {
+            R v = this.value;
```
davidmoten, line 77: `done` checks in this class?

akarnokd: No need for separate done flag as this.value can't be initially null and becomes null if onError is called. Updated the PR with more checks.

davidmoten: agreed

davidmoten:
:+1: with the qualifier that I'm not up on RxJava2/ReactiveStreams lifecycle in terms of subscription and disposal yet (that's for another day sorry!).

PR #4893, opened 2016-11-28
2.x: fix compilation errors when using Java 8

akarnokd:
This PR fixes a few compilation errors that happen when the project is compiled with Java 8 target (Eclipse is even slower then).

Some covariant casts don't work in 8 and some type arguments inferred as Object get inferred as a more concrete type in 8. The slight drawback is that with the changes, Java 6 compilation needs extra suppressions.


vanniktech (approved):
Nice those kept me from executing tests from intellij

JakeWharton (approved):
Was this a change in new revisions of JDK 8 or to support older ones? I haven't had any problems on the CLI or in IntelliJ.

akarnokd:
This is added in order to support Java 8 or 9 compile targets (the default is Java 6 in the project).

Issue #4894, opened 2016-11-28
onError throws Exception after Flowable has been unsubscribed


flowsprenger:
We have recently run up against this in RxJava 2.0.x, and I would like to clarify that this is the intended behaviour.

We are creating a Flowable that connects to a remote server and first needs to resolve the host. If host cannot be resolved, the UnknownHostException is propagated via onError:
``` 
val messages = Flowable.create<SourcedMessage>({ subscriber ->
        this.subscriber = subscriber
        try{
            addr = InetAddress.getByName(host)
        }catch (e:UnknownHostException){
            subscriber.onError(e)
            return@create
        }
       .....
}
```

This seems to work fine (we can use an invalid hostname and the retry logic wrapping it fires just as expected), but we have seen crashes in our crashlogs, where the Android app crashes with an uncaught UnknownHostException, which we did not expect.
The crash goes away as soon as we check subscriber.isCancelled before calling onError as in:

```
val messages = Flowable.create<SourcedMessage>({ subscriber ->
        this.subscriber = subscriber
        try{
            addr = InetAddress.getByName(host)
        }catch (e:UnknownHostException){
            if(!subscriber.isCancelled){
                    subscriber.onError(e)
            }
            return@create
        }
       .....
}
```

For what it matters the Flowable is further chained in a subscribeOn(Schedulers.io()).retryWhen().publish.refCount() chain that is observedOn and subscribed to on the AndroidScheduler.mainThread.

We would have expected the onError just being silently ignored after unsubscription (which occurs when the app goes into the background).
Maybe #4863 and #4807 are related too.

akarnokd:
2.x reports undeliverable exceptions, such as ones that happen around a cancellation to the RxJavaPlugins error handler which then forwards it to the current thread's uncaught exception handler. On desktop, this just prints the stacktrace to the console whereas some Android setups treat it as a fatal exception and stop the app. I suggest installing a custom global error handler via `RxJavaPlugins.setErrorHandler`. I'll think about how to deal with the situation by using system parameters.

flowsprenger:
Thanks for clarification

alouanemed:
@akarnokd, Can't find any guide to use `RxJavaPlugins.setOnError` in the wiki page https://github.com/ReactiveX/RxJava/wiki/Plugins

akarnokd:
Sorry, mixed up the two versions. Otherwise, it is straightforward: you have to specify a callback `Action1`/`Consumer`:

```java
[...]

Issue #4871, opened 2016-11-22
Rx2: Multiple errors in zip'ed observables throw exception


TillSimon:
I execute multiple HTTP calls in parallel by zipping them up. Now 0..n of them can obviously fail. The thing is, when e.g. 2 calls fail at almost the exact time, error 1 is processed in `doOnError` and `onErrorReturnItem` but error 2 slips through and causes an exception. I think this happens if error 2 happens to come in while error 1 is still not processed entirely in those two onError operators.

It can be reproduced with the following simplified code:

```
buttonClick.flatMapSingle(action -> Single.zip(
                    Single.error(new Throwable("error 1")).delay(500, TimeUnit.MILLISECONDS),
                    Single.error(new Throwable("error 2")).delay(500, TimeUnit.MILLISECONDS),
                    (integer, integer2) -> true
            )
                    .doOnError(t -> Timber.d(t.getMessage()))
                    .onErrorReturnItem(true))
                    .subscribe();
```
So every once in a while this would throw:
```
D/Main: error 1
W/System.err: java.lang.Throwable: error 2
```

Shouldn't the `Single.zip` stream only be allowed to emit **one** error?

akarnokd:
`Zip` emits only a single error, the others go into the `RxJavaPlugins.onError` which default prints to `System.err`. It works this way to allow hooking the `RxJavaPlugins.onError` and capture errors that can't be otherwise delivered due to the protocol restrictions on a flow. There should be an overload which lets you delay all errors so you really get one `CompositeException` if there were more than one.

PS.: please avoid [cross-posting](http://stackoverflow.com/questions/40745897/rxjava2-multiple-errors-in-ziped-observables-throw-exception).

TillSimon:
So I'd have to wait for all zipped Singles to emit even if I can proceed on the first error that occurs (let's say it's all or nothing) ? Would be nicer to fail-fast of course.

TillSimon:
It doesn't seem to be happening with `Observable.zip()`, so I don't know if that behavior makes sense...

akarnokd:
Your example has a probabilistic race between emissions which may or may not manifest in a composite exception. RxJava 2 tries to avoid losing exceptions which could be important to the developer even if it happens after the natural lifecycle of a flow.

TillSimon:
[...]

Issue #4863, opened 2016-11-18
Get RxCachedThreadScheduler-n when calling Disposable.dispose()


enginebai:
### Requirement
I migrate from 1.x to 2.x, replace `Subscription` to `Disposable`, and I'd like to **cancel the subscription before a new subscription starts**. But I got `RxCachedThreadScheduler-n` when calling `Disposable.dispose()`. I've check #4807 and found that it may timeout problem, but I'm sure that my api is pretty fast and won't be timeout at all. How can I resolve this problem??

### Exception
```
E/AndroidRuntime: FATAL EXCEPTION: RxCachedThreadScheduler-1
                  Process: com.machipopo.swag, PID: 30241
                  java.io.InterruptedIOException: thread interrupted
                      at okio.Timeout.throwIfReached(Timeout.java:145)
                      at okio.Okio$2.read(Okio.java:136)
                      at okio.AsyncTimeout$2.read(AsyncTimeout.java:211)
                      at okio.RealBufferedSource.request(RealBufferedSource.java:71)
                      at okio.RealBufferedSource.require(RealBufferedSource.java:64)
                      at okio.RealBufferedSource.readHexadecimalUnsignedLong(RealBufferedSource.java:270)
                      at okhttp3.internal.http.Http1xStream$ChunkedSource.readChunkSize(Http1xStream.java:441)
                      at okhttp3.internal.http.Http1xStream$ChunkedSource.read(Http1xStream.java:422)
                      at okio.RealBufferedSource.read(RealBufferedSource.java:50)
                      at okio.RealBufferedSource.exhausted(RealBufferedSource.java:60)
                      at okio.InflaterSource.refill(InflaterSource.java:101)
                      at okio.InflaterSource.read(InflaterSource.java:62)
                      at okio.GzipSource.read(GzipSource.java:80)
                      at okio.RealBufferedSource.request(RealBufferedSource.java:71)
                      at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:225)
                      at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:187)
                      at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:160)
                      at okhttp3.RealCall.execute(RealCall.java:57)
                      at com.test.api.ApiService.get(ApiService.java:145)
                      at com.test.UserApi$1.subscribe(UserApi.java:63)
```

### Old 1.x Code
```java
[...]

Issue #4807, opened 2016-11-04
 FATAL EXCEPTION: RxCachedThreadScheduler-4 or -1 or -2 or -3 ...


yubaokang:
E/AndroidRuntime: FATAL EXCEPTION: RxCachedThreadScheduler-1
                                                                  Process: goujiawang.gjstore, PID: 5960
                                                                  java.io.InterruptedIOException: thread interrupted
                                                                      at okio.Timeout.throwIfReached(Timeout.java:146)
                                                                      at okio.Okio$2.read(Okio.java:135)
                                                                      at okio.Buffer.writeAll(Buffer.java:993)
                                                                      at okhttp3.RequestBody$3.writeTo(RequestBody.java:118)
                                                                      at okhttp3.MultipartBody.writeOrCountBytes(MultipartBody.java:171)
                                                                      at okhttp3.MultipartBody.writeTo(MultipartBody.java:113)
                                                                      at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:189)
                                                                      at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
                                                                      at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
                                                                      at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:170)
                                                                      at okhttp3.RealCall.execute(RealCall.java:60)
                                                                      at retrofit2.OkHttpCall.execute(OkHttpCall.java:174)
                                                                      at com.jakewharton.retrofit2.adapter.rxjava2.CallObservable.subscribeActual(CallObservable.java:41)
                                                                      at io.reactivex.Observable.subscribe(Observable.java:10151)
                                                                      at com.jakewharton.retrofit2.adapter.rxjava2.BodyObservable.subscribeActual(BodyObservable.java:34)
                                                                      at io.reactivex.Observable.subscribe(Observable.java:10151)
                                                                      at io.reactivex.internal.operators.flowable.FlowableFromObservable.subscribeActual(FlowableFromObservable.java:31)
                                                                      at io.reactivex.Flowable.subscribe(Flowable.java:12172)
                                                                      at io.reactivex.internal.operators.flowable.FlowableOnBackpressureLatest.subscribeActual(FlowableOnBackpressureLatest.java:31)
                                                                      at io.reactivex.Flowable.subscribe(Flowable.java:12172)
[...]

PR #4895, opened 2016-11-29
2.x: sys-property to disable calling uncaught exception handler

akarnokd:
This PR adds a new system property `rx2.onerror-nouncaught` to suppress calling the current thread's uncaught exception handler with the default `RxJavaPlugins.onError` as it crashes Android apps (on desktop, it either does nothing or prints the stacktrace).

Related: #4894, #4871, #4863, #4807

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -26,6 +26,11 @@ import io.reactivex.observables.ConnectableObservable;
 
 /**
  * Utility class to inject handlers to certain standard RxJava operations.
+ * <p>System parameters:
+ * <ul>
+ * <li>{@code rx2.onerror-nouncaught}: boolean, default false; if true, the default onError() handler
+ * won't call the current thread's uncaught exception handler (which crashes Android apps).</li>
+ * </ul>
  */
 public final class RxJavaPlugins {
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
Not sure how comfortable I am with the behavior of this, or the library in general, around the distinction between undeliverable errors due to violating the observable contract and undeliverable errors due to disposal. In the latter case, I feel like sending these to the plugins isn't the behavior I would expect. I rely on my app to crash when a notification to onError throws. How are users supposed to distinguish between errors that simply couldn't be delivered because the observer left or serious problems in error handling code?

artem-zinnatullin:
Agreed.

> I feel like sending these to the plugins isn't the behavior I would expect.

+1, I don't see much difference between errors and values being sent to /dev/null in case when observer stopped listening for the stream.

What about reporting error that have no observers to the Plugin only if it's `java.lang.Error`, such as `OutOfMemory` or `VirtualMachineError`?

akarnokd:
> errors due to disposal. In the latter case, I feel like sending these to the plugins isn't the behavior I would expect.

It is up to the source to decide if the error it is about to send should be considered significant or not. For example, `InterruptedIOException` that is thrown by the Retrofit adapter may check for `isDisposed` and not emit the error.

Generally, the end application developer should set up a handler via `RxJavaPlugins.setErrorHandler` and not let the error reach the current thread's uncaught error handler.

JakeWharton:
> What about reporting error that have no observers to the Plugin only if it's java.lang.Error, such as OutOfMemory or VirtualMachineError?

These should always automatically be going to the thread's uncaught handler due to the `throwIfFatal` checks.

> Generally, the end application developer should set up a handler via RxJavaPlugins.setErrorHandler and not let the error reach the current thread's uncaught error handler.

What are we supposed to do with these errors? If we can't differentiate between failures during onError and normal onErrors undeliverable due to dispose then how do we know what's crash worth vs. not? The former should absolutely crash the VM but I don't care about the latter at all.

akarnokd:
RxJava can't decide for the user or for libraries building upon it. All it can do is to either provide a channel for undeliverable exceptions or suppress all of them but if RxJava by default suppressed such errors, there is no "unsuppress" them later.

For the Retrofit case, I think the catch handler that receives the `InterruptedIOException` should check for `isDisposed`/`isCancelled` to detect that the exception is due to cancellation and not some other reason.

Otherwise, I'm not too keen on this PR and one can just simply call `RxJavaPlugins.setErrorHandler(Throwable::printStackTrace)` to bypass the current thread handler.

PR #4899, opened 2016-12-03
2.x: FlowableScanSeed - prevent multiple terminal events

davidmoten:
This PR
* prevents multiple terminal events being emitted when the scan function throws
* prevents processing of a later `onNext` if the previous `onNext` processing resulted in an error emission
* increases coverage to 100% of `FlowableScanSeed`

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableScanSeed.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableScanSeed.java
@@ -51,6 +51,8 @@ public final class FlowableScanSeed<T, R> extends AbstractFlowableWithUpstream<T
         private static final long serialVersionUID = -1776795561228106469L;
 
         final BiFunction<R, ? super T, R> accumulator;
+        
+        boolean done;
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableScanSeed.java
```diff
@@ -51,6 +51,8 @@ protected void subscribeActual(Subscriber<? super R> s) {
         private static final long serialVersionUID = -1776795561228106469L;
 
         final BiFunction<R, ? super T, R> accumulator;
+        
+        boolean done = false;
```
akarnokd: No need to set a default value.

akarnokd (changes requested):

src/main/java/io/reactivex/internal/operators/flowable/FlowableScanSeed.java
davidmoten: righto, fixed

davidmoten:
updated PR with `onNext` protection and test (had forgotten to commit)

akarnokd (approved):

PR #4907, opened 2016-12-07
Use t instead of value to allow for IDE naming

Jawnnypoo:
Naming the value `t` instead of `value` allows the IntelliJ to intelligently name the variable based on the type when autocompleting. 

Before:
```
Single.just(new Dog()).subscribe(new SingleObserver<String>() {
            @Override
            public void onSubscribe(Disposable d) {
                
            }

            @Override
            public void onSuccess(Dog value) {

            }

            @Override

            }
        });
```
After:
```
Single.just(new Dog()).subscribe(new SingleObserver<String>() {
            @Override
            public void onSubscribe(Disposable d) {

            }

            @Override
            public void onSuccess(Dog dog) {

            }

            @Override

            }
        });
```

### DIFF
--- a/src/main/java/io/reactivex/MaybeObserver.java
+++ b/src/main/java/io/reactivex/MaybeObserver.java
@@ -45,10 +45,10 @@ public interface MaybeObserver<T> {
      * <p>
      * The {@link Maybe} will not call this method if it calls {@link #onError}.
      *
-     * @param value
+     * @param t
      *          the item emitted by the Maybe
      */
-    void onSuccess(T value);
+    void onSuccess(T t);
 
     /**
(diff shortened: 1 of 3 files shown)


JakeWharton (approved):
Awesome. Does 2.x need this?

Jawnnypoo:
This is for 2.X. Did I make a mistake? I could add this to 1.X also if desired.

JakeWharton:
Hah, yeah I meant 1.x. Off-by-one error.

Jawnnypoo:
https://github.com/ReactiveX/RxJava/pull/4908 Surprisingly it was already done a lot more in 1.x, so there were only a few cases I found.

akarnokd (approved):

Issue #4909, opened 2016-12-09
2.x: Observable.repeatWhen doesn't call onComplete


otal1105:
I tried below
```java
Observable.range(1, 3).repeatWhen(handler -> handler.take(2))
    .subscribe(
        System.out::println, 
        e -> e.printStackTrace(),
        () -> System.out.println("completed"));
```

It dodin't call onComplete. I got this.
```
1
2
3
1
2
3
1
2
3
```

However, I tried Flowable, it called onComplete.

```java
Flowable.range(1, 3).repeatWhen(handler -> handler.take(2))
    .subscribe(
        System.out::println, 
        e -> e.printStackTrace(),
        () -> System.out.println("completed"));
```

The result was below.
```
1
2
3
1
2
3
completed
```

I tried on 2.0.2.

akarnokd:
Thanks. This is a bug with `Observable.repeat`. I'll post a fix shortly.

akarnokd:
See #4911.

akarnokd:
Closing via #4911

PR #4911, opened 2016-12-09
2.x: fix & update Observable.repeatWhen and retryWhen (again)

akarnokd:
This PR fixes the `Observable.repeatWhen` not completing properly if the handler completes. In addition, the `ObservableRedo` has been split to `ObservableRetryWhen` and `ObservableRepeatWhen` just like their `Flowable` counterpart and are now they use a less convoluted algorithm.

Related: #4909


akarnokd:
I don't know why the travis-ci/push still shows up.

JakeWharton (approved):

Issue #4914, opened 2016-12-11
Add support for Notification for rx.Single


sbcd90:
Hello,

The support for a `Notification` class for `rx.Single` should be added.
In many scenario s, the `Notification` class for `rx.Observable` cannot be re-used.

Version: RxJava 1.x

PR #4915, opened 2016-12-11
#4914 Add support for Notification for rx.Single

sbcd90:
This pull request is created to add support for Notification in rx.Single.

This SingleNotification class provides two kinds of notifications.

- OnSuccess
- OnError

emitted out of rx.Single.

### DIFF
--- a/src/main/java/rx/SingleNotification.java
+++ b/src/main/java/rx/SingleNotification.java
@@ -1,0 +1,182 @@
+/**
+ * Copyright 2014 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


akarnokd:
Why?

sbcd90:
Hello @akarnokd ,

I was trying to write the test cases for supporting `rx.Single` in `Netflix/Hystrix` & saw that the tests for `HystrixObservableCommand` depends on `Notification` for calculating the no. of `onNext`, `onCompleted`, `onError` events emitted as shown [here](https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/test/java/com/netflix/hystrix/TestableExecutionHook.java#L50-L90). 

Now, I can reuse the same Notification class also for writing tests for `rx.Single`. But its not a clean approach.
There could be several other cases like this where the new class `SingleNotification` can be useful.

akarnokd:
You can have this in your own code/library and publish it yourself, I don't see why this should be in RxJava.

PR #4919, opened 2016-12-14
Update Observable.java

daawa:

  - [x] Please include the issue list number(s) or other PR numbers in the description if you are contributing in response to those.

  - [x] Please include a reasonable set of unit tests if you contribute new code or change an existing one. If you contribute an operator, (if applicable) please make sure you have tests for working with an `empty`, `just`, `range` of values as well as an `error` source, with and/or without backpressure and see if unsubscription/cancellation propagates correctly.

change the order of generalized types of Operator to be consistent with Func1

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -199,7 +199,7 @@ public class Observable<T> {
      * @param <T> the upstream's value type (input)
      * @param <R> the downstream's value type (output)
      */
-    public interface Operator<R, T> extends Func1<Subscriber<? super R>, Subscriber<? super T>> {
+    public interface Operator<T, R> extends Func1<Subscriber<? super T>, Subscriber<? super R>> {
         // cover for generics insanity
     }
 


akarnokd:
Such changes violate the established contract of the `Operator` and breaks everybody.

akarnokd:
Correction: The input of the call is the downstream Subscriber which is usually `R` as **r**esult type. The output of the operator is the upstream Subscriber which is usually `T` for inpu**t**.

daawa:
It doesn't matter what the letters are , just keep in line with Func1 interface where T is input, R is output: 

```
public interface Func1<T, R> extends Function {
    R call(T t);
}
```

akarnokd:
Still no good:

```java
Operator<T, R> extends Func1<Subscriber<? super T>, Subscriber<? super R>>
```

The `Func1` receives the subscriber of downstream which is the output. If the `Operator<T, R>` follows the input-output ordering, the `Func1` should be `Func1<Subscriber<? super R>, Subscribe<? super T>>` which would break everybody.

daawa:
ok ,got it . what you mean by _output_ is the emission of observable right?

akarnokd:
Output is what is is emitted towards the downstream. Input is what is received from upstream.

PR #4928, opened 2016-12-19
2.x: don't call the Thread's uncaught handler from RxJavaPlugins.onError

akarnokd:
This PR removes the call to the current thread's uncaught exception handler if there is no `RxJavaPlugins.onError` handler setup. The default behavior of the `UncaughtExceptionHandler` on Android crashes the whole app. Unit tests that expected this call are modified/disabled.

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -12,15 +12,14 @@
  */
 package io.reactivex.plugins;
 
-import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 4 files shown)


JakeWharton:
I'm not a fan of this change either. I still don't understand why there isn't a differentiation between fundamentally undeliverable errors (due to the observable contract) and errors that occur after unsubscription. These are two very different cases that should be handled differently. The former is absolutely crash worthy as it indicates a programming problem in the application whereas the latter is just an event that was explicitly ignored.

akarnokd:
I don't think you can distinguish between the two. What we can do is to make sure workers and futures are cancelled after the upstream's disposables and subscriptions get cancelled. Currently, its almost always in the opposite order and this is why catching an InterruptedException due to Future.cancel the isDisposed still returns false at that moment.

PR #4930, opened 2016-12-20
2.x: add Completable.cache()

akarnokd:
This PR adds the operator `cache()` to `Completable` to cache the terminal event of a `Completable` sequence.

I came across the need for this when [answering a question](http://stackoverflow.com/a/41241793/61158) where the intent of using `Completable` was to indicate an operation has completed but not actually trigger the operation whenever one subscribed.

A small drawback, just like with other `cache()` operators is that once connected, there is no way for cancel it from within the operator itself. Returning `Completable & Disposable` doesn't work because of the the type restriction of `RxJavaPlugins.onAssembly`. There could be a `cache(Consumer<? super Disposable> onDispose)` overload or an example showing `source.ambWith(disposer).cache()`.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -972,6 +972,27 @@ public abstract class Completable implements CompletableSource {
     }
 
     /**
+     * Subscribes to this Completable only once, when the first CompletableObserver
+     * subscribes to the result Completable, caches its terminal event
+     * and relays/replays it to observers.
+     * <p>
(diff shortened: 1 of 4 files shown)


JakeWharton (approved):

PaulWoitaschek:
Great improvement, thanks!

In one of my projects I applied this transformer to a lot of sequences:
```java
private static final CompletableTransformer cache = completable ->
    completable.toObservable()
      .cacheWithInitialCapacity(1)
      .ignoreElements();
```

PR #4934, opened 2016-12-21
2.x: Unify CompositeDisposable error messages

vanniktech:
Unifies the CompositeDisposable error messages and make them a bit more helpful. Just seeing `d is null` reads oddly

### DIFF
--- a/src/main/java/io/reactivex/disposables/CompositeDisposable.java
+++ b/src/main/java/io/reactivex/disposables/CompositeDisposable.java
@@ -40,7 +40,7 @@ public final class CompositeDisposable implements Disposable, DisposableContaine
      * @param resources the array of Disposables to start with
      */
     public CompositeDisposable(Disposable... resources) {
-        ObjectHelper.requireNonNull(resources, "resources is null");
+        ObjectHelper.requireNonNull(resources, "resources are null");
         this.resources = new OpenHashSet<Disposable>(resources.length + 1);
         for (Disposable d : resources) {
             ObjectHelper.requireNonNull(d, "Disposable item is null");
@@ -53,7 +53,7 @@ public final class CompositeDisposable implements Disposable, DisposableContaine
      * @param resources the Iterable sequence of Disposables to start with
      */
     public CompositeDisposable(Iterable<? extends Disposable> resources) {
-        ObjectHelper.requireNonNull(resources, "resources is null");
+        ObjectHelper.requireNonNull(resources, "resources are null");
         this.resources = new OpenHashSet<Disposable>();
         for (Disposable d : resources) {
             ObjectHelper.requireNonNull(d, "Disposable item is null");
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/disposables/CompositeDisposable.java
```diff
@@ -40,7 +40,7 @@ public CompositeDisposable() {
      * @param resources the array of Disposables to start with
      */
     public CompositeDisposable(Disposable... resources) {
-        ObjectHelper.requireNonNull(resources, "resources is null");
+        ObjectHelper.requireNonNull(resources, "resources are null");
```
akarnokd, line 43: It refers to the `resources` parameter so "is" is adequate here.

src/main/java/io/reactivex/disposables/CompositeDisposable.java
```diff
@@ -86,7 +86,7 @@ public boolean isDisposed() {
 
     @Override
     public boolean add(Disposable d) {
-        ObjectHelper.requireNonNull(d, "d is null");
+        ObjectHelper.requireNonNull(d, "Disposable is null");
```
akarnokd, line 89: This is inconsistent as it names the type and not the parameter.

src/main/java/io/reactivex/disposables/CompositeDisposable.java
```diff
@@ -145,7 +145,7 @@ public boolean remove(Disposable d) {
 
     @Override
     public boolean delete(Disposable d) {
-        ObjectHelper.requireNonNull(d, "Disposable item is null");
+        ObjectHelper.requireNonNull(d, "Disposable is null");
```
akarnokd, line 148: It should read `d is null`.

akarnokd (changes requested):
I find only the last one in `delete()` reasonable.

src/main/java/io/reactivex/disposables/CompositeDisposable.java
vanniktech: Are you open to rename the parameter from `d` to `disposable` then?

akarnokd: We want to avoid that in case in some odd JVM language the parameter name change is a breaking change.

vanniktech: Alright but don't you think that the error message `Disposable is null` is better than `d is null` regardless of the parameter name?

akarnokd: `d is null` is totally fine with me.

vanniktech:
If I'm the only one who thinks that those error messages are better and worth the change, I'll leave the decision up to you david. So if you want, you can close it.

PR #4936, opened 2016-12-22
Cleaner integration with CompositeDisposable

stphnd:
Added a convenience method to allow for cleaner, more readable Disposable management;

Adding a disposeWith(Disposable d) method would allow for cleaner, more readable code. Implementation would be simple with something along the lines of:
```
void disposeWith(Disposable d){
    d.add(this);
}
```

Then subscription can be simplified from
```
Disposable d = observable
    .subscribe(() -> foo());
compositeDisposable.add(d);
```
to
```
observable
    .subscribe(()->foo())
    .disposeWith(compositeDisposable);
```

### DIFF
--- a/src/main/java/io/reactivex/disposables/Disposable.java
+++ b/src/main/java/io/reactivex/disposables/Disposable.java
@@ -26,4 +26,9 @@ public interface Disposable {
      * @return true if this resource has been disposed
      */
     boolean isDisposed();
+    
+    /**
+     * Convenience method for adding a Disposable to a CompositeDisposable
+     */
+    void disposeWith(Disposable d);
 }


JakeWharton:
Adding methods to an interface is a breaking change.

stphnd:
Yes, I see that now. Sorry I am new to this. What is the correct way to propose such a change? Is there a place to discuss it first, or am I meant to implement the change fully before having it reviewed here?

JakeWharton:
The earliest it could be incorporated is RxJava 3.x which is likely (and
hopefully) at least two years away.

akarnokd:
I don't see why you'd want this. Each method on Disposable adds a burden to hundreds of internal implementations. Kotlin allows extension methods and you can decorate any RxJava object without breaking anybody else.

PR #4966, opened 2017-01-06
2.x: add strict() operator for strong RS conformance

akarnokd:
This PR adds a new instance operator `strict()`:

Ensures that the event flow between the upstream and downstream follow the Reactive-Streams 1.0 specification by honoring the 3 additional rules (which are omitted in standard operators due to performance reasons).

 * §1.3: onNext should not be called concurrently until onSubscribe returns
 * §2.3: onError or onComplete must not call cancel
 * §3.9: negative requests should emit an onError(IllegalArgumentException)

In addition, if rule §2.12 (onSubscribe must be called at most once) is violated, the sequence is cancelled an onError(IllegalStateException) is emitted. Note that the Reactive-Streams specification doesn't have any sanction for violating §2.12.

The `strict()` operator also replaces the test-only operators that were supporting the Reactive-Streams TCKs already.


src/test/java/io/reactivex/tck/ObserveOnTckTest.java
```diff
@@ -29,9 +29,7 @@ public ObserveOnTckTest() {
     @Override
     public Publisher<Integer> createPublisher(long elements) {
         return FlowableTck.wrap(
-            FlowableAwaitOnSubscribeTck.wrap(
                 Flowable.range(0, (int)elements).observeOn(Schedulers.single())
```
JakeWharton, line 32: nit: indent

JakeWharton (approved):

PR #4967, opened 2017-01-06
2.x: add subjects for Single, Maybe and Completable

akarnokd:
This PR adds 3 new subject types: `SingleSubject`, `MaybeSubject` and `CompletableSubject`. Their purpose is to provide an imperative way to multicast 0-1-error events as well as cache these events for later observers. They are thread-safe by design and there is no need for a serialized wrapper unlike the other `Subject`s.

### DIFF
--- a/src/main/java/io/reactivex/subjects/CompletableSubject.java
+++ b/src/main/java/io/reactivex/subjects/CompletableSubject.java
@@ -1,0 +1,228 @@
+/**
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.subjects;
+
+import java.util.concurrent.atomic.*;
+
+import io.reactivex.*;
+import io.reactivex.annotations.Experimental;
+import io.reactivex.disposables.Disposable;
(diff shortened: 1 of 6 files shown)


src/main/java/io/reactivex/subjects/CompletableSubject.java
```diff
@@ -0,0 +1,228 @@
...
+    static final CompletableDisposable[] TERMINATED = new CompletableDisposable[0];
+
+    final AtomicBoolean once;
+    Throwable error;
+
+    /**
+     * Creates a fresh CompletableSubject.
+     * @return the new CompletableSubject instance
+     */
+    public static CompletableSubject create() {
```
vanniktech, line 51: Should this be annotated with `@CheckReturnValue` ?

src/test/java/io/reactivex/subjects/CompletableSubjectTest.java
```diff
@@ -0,0 +1,226 @@
...
+import io.reactivex.disposables.*;
+import io.reactivex.observers.TestObserver;
+import io.reactivex.plugins.RxJavaPlugins;
+import io.reactivex.schedulers.Schedulers;
+
+public class CompletableSubjectTest {
+
+    @Test
+    public void once() {
+        CompletableSubject ms = CompletableSubject.create();
```
vanniktech: should not all of those be `cs` instead of `ms` for MaybeSubject ?

src/test/java/io/reactivex/subjects/SingleSubjectTest.java
```diff
@@ -0,0 +1,243 @@
...
+import io.reactivex.disposables.*;
+import io.reactivex.observers.TestObserver;
+import io.reactivex.plugins.RxJavaPlugins;
+import io.reactivex.schedulers.Schedulers;
+
+public class SingleSubjectTest {
+
+    @Test
+    public void success() {
+        SingleSubject<Integer> ms = SingleSubject.create();
```
vanniktech: also here should be ss

akarnokd:
@vanniktech updated.

JakeWharton (approved):

PR #4970, opened 2017-01-08
2.x: fix withLatestFrom null checks, lifecycle

akarnokd:
This PR fixes the lack of null check on the result of the `combiner` function as well as fixes the lifecycle of the `FlowableWithLatestFrom` to avoid races with the `onSubscribe`.


hzsweers:
LGTM

ZacSweers (approved):
LGTM

JakeWharton (approved):

PR #4972, opened 2017-01-08
Improve compose() generics

### LINKED ISSUE #4950: 2.x: compose() generics
Migrating some of our code to RxJava 2, we've run into some issues around transformers due to the following signature change:

```java
// 1
public <R> Observable<R> compose(Transformer<? super T, ? extends R> transformer) {     
    return ((Transformer<T, R>) transformer).call(this); 
}

// 2
public final <R> Observable<R> compose(ObservableTransformer<T, R> composer) {     
    return wrap(composer.apply(this));
 }
```

Was this intentionally narrowed? If not, open to a PR to add back the `? super` semantics?

ZacSweers:
Resolves #4950

I tested and built with the added tests on Java 6, 7, and 8. Not sure if you want to keep them in before merging since they're not really functional in nature, just there to make sure they compile.

### DIFF
--- a/src/test/java/io/reactivex/TransformerTest.java
+++ b/src/test/java/io/reactivex/TransformerTest.java
@@ -13,12 +13,11 @@
 
 package io.reactivex;
 
-import static org.junit.Assert.*;
-
+import io.reactivex.exceptions.TestException;
 import org.junit.Test;
 import org.reactivestreams.Publisher;
 
-import io.reactivex.exceptions.TestException;
+import static org.junit.Assert.*;
 
 public class TransformerTest {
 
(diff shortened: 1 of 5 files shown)


ZacSweers:
Side-note: gets even better on Java 8, which removes the need for a type argument at all in those examples!

screen shot 2017-01-08 at 2 58 20 am

src/test/java/io/reactivex/TransformerTest.java
```diff
@@ -83,7 +82,7 @@ public void maybeTransformerThrows() {
     }
 
     @Test
-    public void completabeTransformerThrows() {
+    public void completableTransformerThrows() {
```
ZacSweers, line 85: Took the opportunity to fix a typo in this method name

ZacSweers:
Thanks!

PR #4974, opened 2017-01-08
2.x: add Flowable.parallel() and parallel operators

akarnokd:
This PR adds the `parallel()` method to `Flowable` which opens up a sub-DSL with parallel operations. (Note that only a few operators make sense in a parallel settings.)

This parallel sub-DSL is not limited to computation tasks as it allows specifying the parallelism and the `Scheduler` to run the parallel 'rails'. For example, you can have parallel downloads that block:

```java
Flowable.range(1, 100)
.parallel(10)
.runOn(Schedulers.io())
.map(v -> httpClient.blockingGet("http://server/item/" + v))
.sequential()
.observeOn(AndroidSchedulers.mainThread())
.subscribe(...);
```

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -29,6 +29,7 @@ import io.reactivex.internal.operators.observable.ObservableFromPublisher;
 import io.reactivex.internal.schedulers.ImmediateThinScheduler;
 import io.reactivex.internal.subscribers.*;
 import io.reactivex.internal.util.*;
+import io.reactivex.parallel.ParallelFlowable;
 import io.reactivex.plugins.RxJavaPlugins;
 import io.reactivex.schedulers.*;
 import io.reactivex.subscribers.*;
@@ -10361,6 +10362,100 @@ public abstract class Flowable<T> implements Publisher<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Flowable<T> onTerminateDetach() {
         return RxJavaPlugins.onAssembly(new FlowableDetach<T>(this));
+    }
+
+    /**
+     * Parallelizes the flow by creating multiple 'rails' (equal to the number of CPUs)
+     * and dispatches the upstream items to them in a round-robin fashion.
+     * <p>
+     * Note that the rails don't execute in parallel on their own and one needs to
+     * apply {@link ParallelFlowable#runOn(Scheduler)} to specify the Scheduler where
+     * each rail will execute.
+     * <p>
+     * To merge the parallel 'rails' back into a single sequence, use {@link ParallelFlowable#sequential()}.
+     * <p>
+     * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/flowable.parallel.png" alt="">
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator requires the upstream to honor backpressure and each 'rail' honors backpressure
+     *  as well.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code parallel} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @return the new ParallelFlowable instance
+     * @since 2.0.5 - experimental
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @CheckReturnValue
+    @Experimental
+    public final ParallelFlowable<T> parallel() {
+        return ParallelFlowable.from(this);
+    }
+
+    /**
+     * Parallelizes the flow by creating the specified number of 'rails'
+     * and dispatches the upstream items to them in a round-robin fashion.
+     * <p>
+     * Note that the rails don't execute in parallel on their own and one needs to
+     * apply {@link ParallelFlowable#runOn(Scheduler)} to specify the Scheduler where
+     * each rail will execute.
+     * <p>
+     * To merge the parallel 'rails' back into a single sequence, use {@link ParallelFlowable#sequential()}.
+     * <p>
+     * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/flowable.parallel.png" alt="">
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator requires the upstream to honor backpressure and each 'rail' honors backpressure
+     *  as well.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code parallel} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param parallelism the number of 'rails' to use
+     * @return the new ParallelFlowable instance
+     * @since 2.0.5 - experimental
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @CheckReturnValue
+    @Experimental
+    public final ParallelFlowable<T> parallel(int parallelism) {
+        ObjectHelper.verifyPositive(parallelism, "parallelism");
+        return ParallelFlowable.from(this, parallelism);
+    }
+
+    /**
+     * Parallelizes the flow by creating the specified number of 'rails'
+     * and dispatches the upstream items to them in a round-robin fashion and
+     * uses the defined per-'rail' prefetch amount.
+     * <p>
+     * Note that the rails don't execute in parallel on their own and one needs to
+     * apply {@link ParallelFlowable#runOn(Scheduler)} to specify the Scheduler where
+     * each rail will execute.
+     * <p>
+     * To merge the parallel 'rails' back into a single sequence, use {@link ParallelFlowable#sequential()}.
+     * <p>
+     * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/flowable.parallel.png" alt="">
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator requires the upstream to honor backpressure and each 'rail' honors backpressure
+     *  as well.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code parallel} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param parallelism the number of 'rails' to use
+     * @param prefetch the number of items each 'rail' should prefetch
+     * @return the new ParallelFlowable instance
+     * @since 2.0.5 - experimental
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @CheckReturnValue
+    @Experimental
+    public final ParallelFlowable<T> parallel(int parallelism, int prefetch) {
+        ObjectHelper.verifyPositive(parallelism, "parallelism");
+        ObjectHelper.verifyPositive(prefetch, "prefetch");
+        return ParallelFlowable.from(this, parallelism, prefetch);
     }
 
     /**
--- a/src/main/java/io/reactivex/internal/operators/parallel/ParallelCollect.java
+++ b/src/main/java/io/reactivex/internal/operators/parallel/ParallelCollect.java
@@ -1,0 +1,165 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.internal.operators.parallel;
+
+import java.util.concurrent.Callable;
+
+import org.reactivestreams.*;
+
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.functions.BiConsumer;
+import io.reactivex.internal.subscribers.DeferredScalarSubscriber;
+import io.reactivex.internal.subscriptions.*;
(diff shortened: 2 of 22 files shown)


akarnokd:
I'll restore the +95% coverage in a separate PR.

benjchristensen:
I have a use case that could benefit from this depending on how it is implemented. I don't see the API for `ParallelFlowable.merge(Flowable<Flowable<T>> flowables)` however, which is what I'd need and have to manually do. 

Let me describe the type of parallel processing and see if your goal of `ParallelFlowable` matches it.

100s of network connections, each spread across n event loops (say 16). The semantic behavior is to merge the 100s of connections into a single stream, then do `groupBy` on all of them, and on each `GroupedFlowable` then does a `scan`. With normal `Flowable` this is bad, as it takes the 16 threads and synchronizes them all, even though each source `Flowable` is on one of the 16 threads, and each output `GroupedObservable` can be processed concurrently again on those 16 threads. 

In theory, a `ParallelFlowable.merge(sourceFlowables).groupBy(...).scan(...)` could allow the merge to support concurrent `onNext` and then `ParallelFlowable.groupBy` could re-emit a normal `Flowable` where scan works sequentially again. 

Is this the type of thing you want `ParallelFlowable` to enable?

akarnokd:
@benjchristensen No. ParallelFlowable optimizes for a fixed parallelism level with round-robin dispatch and round-robin join. The closest thing is the `parallelStream()` operator in Java 8 for computation-intensive tasks. Your case has an unknown number of inner sources to merge and an unknown number of groups that could appear.

benjchristensen:
Too bad. Maybe someday I'll get around to making a "ConcurrentFlowable" happen ... but it's been on my todo list for 3 years, so not counting on it :-)

akarnokd:
@JakeWharton Do you want to review this or if not, do you at least willing to accept it into RxJava 2?

src/main/java/io/reactivex/internal/operators/parallel/ParallelCollect.java
```diff
@@ -0,0 +1,165 @@
...
+        this.collector = collector;
+    }
+
+    @Override
+    public void subscribe(Subscriber<? super C>[] subscribers) {
+        if (!validate(subscribers)) {
+            return;
+        }
+
+        int n = subscribers.length;
```
artem-zinnatullin, line 54: `final`

src/main/java/io/reactivex/internal/operators/parallel/ParallelFlatMap.java
```diff
@@ -0,0 +1,92 @@
...
+
+import org.reactivestreams.*;
+
+import io.reactivex.internal.operators.flowable.FlowableFlatMap;
+import io.reactivex.parallel.ParallelFlowable;
+
+/**
+ * Flattens the generated Publishers on each rail.
+ *
+ * @param <T> the input value type
```
artem-zinnatullin, line 26: nit: naming generic type parameters `Input` and `Output` would remove such comments and make code slightly more readable. Or `I` and `O` as a reference to common `I/O` abbr.

src/main/java/io/reactivex/internal/operators/parallel/ParallelFromArray.java
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright 2016 Netflix, Inc.
```
artem-zinnatullin: `(c) 2016-present, RxJava Contributors` here and in all other files.

src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
```diff
@@ -0,0 +1,359 @@
...
+                if (requested.get() != 0) {
+                    actual.onNext(value);
+                    if (requested.get() != Long.MAX_VALUE) {
+                        requested.decrementAndGet();
+                    }
+                    inner.request(1);
+                } else {
+                    SimpleQueue<T> q = inner.getQueue();
+
+                    // FIXME overflow handling
```
artem-zinnatullin: Signal MBE? When do you plan to resolve FIXME? It'll lead to silently dropped values…

src/main/java/io/reactivex/internal/operators/parallel/ParallelPeek.java
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        @Override
+        public void request(long n) {
+            try {
+                parent.onRequest.accept(n);
+            } catch (Throwable ex) {
+                Exceptions.throwIfFatal(ex);
+                RxJavaPlugins.onError(ex);
+            }
```
artem-zinnatullin, line 109: add `return` to avoid request in case of error?

src/main/java/io/reactivex/parallel/ParallelFlowable.java
```diff
@@ -0,0 +1,701 @@
...
+     * Specifies where each 'rail' will observe its incoming values with
+     * no work-stealing and default prefetch amount.
+     * <p>
+     * This operator uses the default prefetch size returned by {@code Flowable.bufferSize()}.
+     * <p>
+     * The operator will call {@code Scheduler.createWorker()} as many
+     * times as this ParallelFlowable's parallelism level is.
+     * <p>
+     * No assumptions are made about the Scheduler's parallelism level,
+     * if the Scheduler's parallelism level is lwer than the ParallelFlowable's,
```
artem-zinnatullin: lower

src/main/java/io/reactivex/parallel/ParallelFlowable.java
```diff
@@ -0,0 +1,701 @@
...
+     * Specifies where each 'rail' will observe its incoming values with
+     * possibly work-stealing and a given prefetch amount.
+     * <p>
+     * This operator uses the default prefetch size returned by {@code Flowable.bufferSize()}.
+     * <p>
+     * The operator will call {@code Scheduler.createWorker()} as many
+     * times as this ParallelFlowable's parallelism level is.
+     * <p>
+     * No assumptions are made about the Scheduler's parallelism level,
+     * if the Scheduler's parallelism level is lwer than the ParallelFlowable's,
```
artem-zinnatullin: lower

src/main/java/io/reactivex/parallel/ParallelFlowable.java
```diff
@@ -0,0 +1,701 @@
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code sequential} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @return the new Px instance
+     * @see ParallelFlowable#sequential(int)
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @CheckReturnValue
+    public final Flowable<T> sequential() {
```
artem-zinnatullin, line 253: I think this should be a verb: `sequentize()`/etc to be consistent with other operators (which are verbs mostly).

Btw, reading chains like:

```java
Flowable.range(1, 100)
.parallel(10)
.runOn(Schedulers.io())
.map(v -> httpClient.blockingGet("http://server/item/" + v))
.sequential()
```

feels strange because `sequential` after `parallel`  looks like an operator that **disables** parallelization of the chain (of course it can't, but I dunno, it just reads strange to me).

src/main/java/io/reactivex/parallel/ParallelFlowable.java
```diff
@@ -0,0 +1,701 @@
...
+     * Sorts the 'rails' of this ParallelFlowable and returns a Publisher that sequentially
+     * picks the smallest next value from the rails.
+     * <p>
+     * This operator requires a finite source ParallelFlowable.
+     *
+     * @param comparator the comparator to use
+     * @return the new Px instance
+     */
+    @CheckReturnValue
+    public final Flowable<T> sorted(Comparator<? super T> comparator) {
```
artem-zinnatullin, line 290: `sort`?

src/main/java/io/reactivex/internal/operators/parallel/ParallelCollect.java
artem-zinnatullin: Same in other similar places would be good.

artem-zinnatullin (changes requested):
Few nits.

artem-zinnatullin:
What about tests and benchmark comparisons with parallelization that you can achieve at the moment, using existing RxJava apis?

src/main/java/io/reactivex/Flowable.java
```diff
@@ -10288,6 +10289,100 @@ public final Completable ignoreElements() {
     }
 
     /**
+     * Parallelizes the flow by creating multiple 'rails' (equal to the number of CPUs)
+     * and dispatches the upstream items to them in a round-robin fashion.
+     * <p>
+     * Note that the rails don't execute in parallel on their own and one needs to
+     * apply {@link ParallelFlowable#runOn(Scheduler)} to specify the Scheduler where
```
artem-zinnatullin, line 10372: What about remove `runOn` and add `Scheduler` as a parameter to `parallel()`?

src/main/java/io/reactivex/internal/operators/parallel/ParallelCollect.java
akarnokd: I don't do those unless the variable has to be accessed from an inner class.

src/main/java/io/reactivex/internal/operators/parallel/ParallelFlatMap.java
akarnokd: This is an established naming pattern with other generic types of RxJava.

src/main/java/io/reactivex/internal/operators/parallel/ParallelCollect.java
artem-zinnatullin: In long methods reader has to spend extra time to check that it's not modified anywhere, but ok

src/main/java/io/reactivex/internal/operators/parallel/ParallelFlatMap.java
artem-zinnatullin: Sure, "just saying"

src/main/java/io/reactivex/internal/operators/parallel/ParallelFromArray.java
akarnokd: I'll update the PR.

src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
akarnokd: Adding it right now.

src/main/java/io/reactivex/internal/operators/parallel/ParallelPeek.java
akarnokd: `FlowableDoOnLifecycle` doesn't return either. There is no good way to report an error and not inject a lot of overhead. See `strict()`.

src/main/java/io/reactivex/parallel/ParallelFlowable.java
akarnokd: Fixing.

src/main/java/io/reactivex/parallel/ParallelFlowable.java
akarnokd: Tis naming matches Java 8 Stream's `parallel()` and `sequential()` operators.

src/main/java/io/reactivex/parallel/ParallelFlowable.java
akarnokd: Matches the naming of `Flowable.sorted()`.

src/main/java/io/reactivex/Flowable.java
akarnokd: It is the same logic as with regular factory methods such as `just`, `range`, `fromIterable` don't take a `Scheduler`, plus you can apply multiple `runOn`'s on a sequence at different stages. For example create a pipeline with stages of parallelism=2 and 3 stages in total.

artem-zinnatullin: >plus you can apply multiple `runOn's` on a sequence at different stages

Ah, that's nice, got it.

src/main/java/io/reactivex/parallel/ParallelFlowable.java
artem-zinnatullin: Makes sense, though JDK is not the best example of good naming.

akarnokd:
I've added a benchmark and here are the results (i7 4770K, Windows 7 x64, Java 8u112):

image

[Raw data](https://gist.github.com/akarnokd/000cf0f0c86c046b44414a34426353d5)

Clearly, parallel has lower overhead than `flatMap`-based, 1 element parallelism.

image

Comparing against `groupBy`, the benefits manifest with longer per-item computation but `groupBy` looks odd: in each compute/parallelism setup the numbers are really close to each other as if there wasn't actual parallel execution with `groupBy`. I have to investigate that further.

src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
```diff
@@ -0,0 +1,368 @@
...
+                        a.onComplete();
+                        return;
+                    }
+                }
+
+                if (e != 0 && r != Long.MAX_VALUE) {
+                    requested.addAndGet(-e);
+                }
+
+                int w = get();
```
davidmoten, line 275: I haven't seen this optimization before with `missed` (calling `get` before `addAndGet`). Is this  particular to the  `ParallelJoin` use case or do you expect to start applying it elsewhere too?

davidmoten (approved):
LGTM, I like it.

src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
akarnokd: There are a couple of places which uses this pattern: range, observeOn, fromArray.

akarnokd: I'm not going to apply them eagerly because it is another local variable/register to worry about when there are lots of other locals in deeper/user code.

akarnokd:
Updated the `groupBy` benchmark. I forgot that `v` was constant and thus the group expression didn't create 1..4 groups. New results (i7 4790, Windows 7 x64, Java 8u112):

image

For smaller computation, parallel has less overhead. For longer computation, they are roughly next to each other. Parallel uses round-robin collection whereas flatMap collects from a source as long as it can.

src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
davidmoten: ok thanks

Issue #4981, opened 2017-01-11
2.x Observable#concatMapEager doesn't always preserve order


ansman:
Sample code (using RxJava 2.0.4):
```java
Observable.just(1, 2, 3, 4, 5)
        .concatMapEager(i -> {
            System.out.println("Processing " + i);
            return i == 3 ? Observable.just(i) : Observable
                    .just(i)
                    .delay(1, TimeUnit.MILLISECONDS, Schedulers.io());
        })
        .observeOn(Schedulers.io())
        .subscribe(i -> System.out.println("Value: " + i));
```

Expected output:
```
Processing 1
Processing 2
Processing 3
Processing 4
Processing 5
Value: 1
Value: 2
Value: 3
Value: 4
Value: 5
```

Actual output:
```
Processing 1
Processing 2
Processing 3
Processing 4
Processing 5
Value: 3
Value: 1
Value: 2
Value: 4
Value: 5
```

As you can see the order is incorrect. This is due to the fact that the item `3` is mapped to a `Callable` source which won't wait for the previous observables.

akarnokd:
Thanks for the feedback. This is indeed a bug with `Observable.concatMapEager()`. `Flowable` works properly.

I'll post a fix shortly. A workaround is to use `just(i).hide()` to break the bad/false internal optimization.

akarnokd:
Closing via #4982

PR #4982, opened 2017-01-11
2.x: fix Observable.concatMapEager bad logic for immediate scalars

akarnokd:
The operator `Observable.concatMapEager` had a bad optimization targeting scalar and callable sources and emitted their values immediately even if it wasn't that particular source's turn for it.

The `Flowable` is not affected, added unit tests for both.

Reported in #4981.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEagerTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEagerTest.java
@@ -1182,7 +1182,6 @@ public class FlowableConcatMapEagerTest {
         .concatMapEager(new Function<Integer, Flowable<Integer>>() {
             @Override
             public Flowable<Integer> apply(Integer i) throws Exception {
-                System.out.println("Processing " + i);
                 return i == 3 ? Flowable.just(i) : Flowable
                         .just(i)
                         .delay(1, TimeUnit.MILLISECONDS, Schedulers.io());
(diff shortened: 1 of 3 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEagerTest.java
```diff
@@ -1175,4 +1175,23 @@ public void innerLong() {
...
         .assertNoErrors();
     }
+
+    @Test
+    public void oneDelayed() {
+        Flowable.just(1, 2, 3, 4, 5)
+        .concatMapEager(new Function<Integer, Flowable<Integer>>() {
+            @Override
+            public Flowable<Integer> apply(Integer i) throws Exception {
+                System.out.println("Processing " + i);
```
vanniktech: Just cluttering it up and can be removed?

src/test/java/io/reactivex/internal/operators/observable/ObservableConcatMapEagerTest.java
```diff
@@ -983,4 +983,23 @@ public void doubleOnSubscribe() {
...
         });
     }
+
+    @Test
+    public void oneDelayed() {
+        Observable.just(1, 2, 3, 4, 5)
+        .concatMapEager(new Function<Integer, ObservableSource<Integer>>() {
+            @Override
+            public ObservableSource<Integer> apply(Integer i) throws Exception {
+                System.out.println("Processing " + i);
```
vanniktech: Same

akarnokd: Updated.

JakeWharton (approved):

Issue #4986, opened 2017-01-12
2.x Observable.combineLatestDelayError delivers Error after completion


StanislavChumarin:
Observable.combineLatestDelayError sends error event after complete event happened and treated as unhandled.
Checked on ```rx.Observable```, ```io.reactivex.Observable``` and ```Flowable```

``` Java
    @Test
    public void testCombine() {
        rx.observers.TestSubscriber<Integer> testSubscriber = rx.observers.TestSubscriber.create();

        rx.Observable<Long> emptyObservable = rx.Observable.empty();
        rx.Observable<Object> errorObservable = rx.Observable.error(new Exception());

        rx.Observable.combineLatestDelayError(
                Arrays.asList(
                        emptyObservable
                                .doOnEach(integerNotification -> System.out.println("emptyObservable: " + integerNotification))
                                .doOnTerminate(() -> System.out.println("emptyObservable: doFinally")),
                        errorObservable
                                .doOnEach(integerNotification -> System.out.println("errorObservable: " + integerNotification))
                                .doOnTerminate(() -> System.out.println("errorObservable: doFinally"))),
                objects -> 0
        )
                .doOnEach(integerNotification -> System.out.println("combineLatestDelayError: " + integerNotification))
                .doOnTerminate(() -> System.out.println("combineLatestDelayError: doFinally"))
                .subscribe(testSubscriber);

        testSubscriber.awaitTerminalEvent();
    }

    @Test
    public void testCombine2() {
        TestObserver<Integer> testObserver = TestObserver.create();

        Observable<Long> emptyObservable = Observable.empty();
        Observable<Object> errorObservable = Observable.error(new Exception());

        Observable.combineLatestDelayError(
                Arrays.asList(
                        emptyObservable
                                .doOnEach(integerNotification -> System.out.println("emptyObservable: " + integerNotification))
                                .doFinally(() -> System.out.println("emptyObservable: doFinally")),
                        errorObservable
                                .doOnEach(integerNotification -> System.out.println("errorObservable: " + integerNotification))
                                .doFinally(() -> System.out.println("errorObservable: doFinally"))),
                objects -> 0
        )
                .doOnEach(integerNotification -> System.out.println("combineLatestDelayError: " + integerNotification))
                .doFinally(() -> System.out.println("combineLatestDelayError: doFinally"))
                .subscribe(testObserver);

        testObserver.awaitTerminalEvent();
    }

    @Test
    public void testCombine2Flowable() {
[...]

Issue #4414, opened 2016-08-23
CombineLatestDelayError emits error before all inner observables emit result


UMFsimke:
Recently I came across following issue while using `combineLatestDealyError` operator. This code has been tested on Android, Nexus 5x running OS 6.0.1.

```
Observable<Boolean> errorObservable = Observable.create(subscriber -> {
    subscriber.onError(new NullPointerException());
});

Observable timeoutObservable = Observable.create(subscriber -> {
    subscriber.onNext(true);
    subscriber.onCompleted();
})
.delay(5, TimeUnit.SECONDS, AndroidSchedulers.mainThread())
.doOnNext(result -> Log.d("Combine", "Delay emitted"));

Observable<Object> zippedObservable = Observable.zip(
        Observable.just(true), timeoutObs,
        (flag1, flag2) -> {
            Log.d("Combine", "On Zip emitted");
            return new Object();
        });

return Observable.combineLatestDelayError(
        Arrays.asList(errorObservable, zippedObservable),
        result -> {
            Boolean flag1 = (Boolean) result[0];
            Object object = result[1];
            Log.d("Combine", "Result of errorObservable: " + flag1);
            Log.d("Combine", "Result of zippedObservable: " + object);
            return null;
        }).subscribeOn(Schedulers.io())
        .observeOn(AndroidSchedulers.mainThread())
        .subscribe(new Subscriber<Object>() {
            @Override
            public void onCompleted() {
                Log.d("Combine", "OnCompleted");
            }

            @Override
            public void onError(Throwable e) {
                Log.d("Combine", "OnError");
            }

            @Override
            public void onNext(Object o) {
                Log.d("Combine", "OnNext");
            }
        });
```

This piece of code is trying to simulate callbacks using delays - for example Android Pay availability callback.

Problem here arises once this code is run. The log will be:

```
OnError
Delay emitted
On Zip emitted
```

This is quite contradictory to the docs that state that expected behavior should be:

```
Delay emitted
On Zip emitted
OnError
```

The expected result does happens if you add `subscriber.onNext(true)`  to the `errorObservable` before `subscriber.onError...`

akarnokd:
Your `errorObservable` doesn't emit any value and since `combineLatestDelayError` won't ever be able to combine values, it terminates with the error immediately.

UMFsimke:
Docs state:

[...]

PR #4987, opened 2017-01-12
2.x: make Obs.combineLatest consistent with Flowable + doc cornercase

akarnokd:
This PR fixes `Observable.combineLatest` to be consistent with `Flowable.combineLatest` by not subscribing to additional input sources if the operator reached a terminal state due to a valueless source (that completes or errors). In addition, such early termination didn't properly cancel the other sources when `delayErrors == true`.

I've also extended the documentation on the overloads to warn about empty sources that will terminate the operator, even with `combineLatestDelayError`, and thus subscription side-effects may not happen.

There is a related issue #4414 where the operator should fully consume each input source no matter what and terminate when all terminate. I'm still considering what would be the best way to introduce this.

Reported in #4986


JakeWharton (approved):

ferhatparmak:
ya

PR #4988, opened 2017-01-12
1.x: Merging an observable of singles.

abersnaze:
The addition of a `Observable<Single<T>> -> Observable<T>` to round out the basic API of `rx.Single`. I need this for doing a flat scan of sorts.

```
Single.merge(events.scan(Single.just(seed), (stateSingle, event) -> {
    return stateSingle.flatMap((state) -> {
        return state.write(event);
    }).cache();
}));
```

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -931,6 +931,110 @@ public class Single<T> {
     }
 
     /**
+     * Flattens any number of Singles into a single Observable, without any transformation.
+     * <p>
+     * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.merge.png" alt="">
+     * <p>
+     * You can combine items emitted by multiple Singles so that they appear as a single Observable, by using
+     * the {@code merge} method.
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code merge} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> merge(Observable<? extends Single<T>> source) {
+        return source.lift(OperatorMergeSingle.<T> instance(false));
+    }
+
+    /**
+     * Flattens any number of Singles into a single Observable, without any transformation.
+     * <p>
+     * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.merge.png" alt="">
+     * <p>
+     * You can combine items emitted by multiple Singles so that they appear as a single Observable, by using
+     * the {@code merge} method.
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code merge} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @param maxConcurrent
+     *            the maximum number of Singles that may be subscribed to concurrently
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> merge(Observable<? extends Single<T>> source, int maxConcurrent) {
+        return source.lift(OperatorMergeSingle.<T> instance(false, maxConcurrent));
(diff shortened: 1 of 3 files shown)


davidmoten:
Looks like a big copy and paste from an existing merge operator. Can you summarize the differences from the original tried and true operator to help review? I also wonder if we can share a decent chunk of code between operators.

abersnaze:
It was a copy for the most part but for changing the class names. It could have be done by mapping the `Observable<Single<T>>` to an `Observable<Observable<T>>`.

src/main/java/rx/Single.java
```diff
@@ -931,6 +931,110 @@ public void onError(Throwable error) {
...
+     * </dl>
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> merge(Observable<? extends Single<T>> source) {
+        return source.lift(OperatorMergeSingle.<T> instance(false));
```
akarnokd, line 952: This could be

```java
return source.map(new Func1<Single<T>, Observable<T>() {
    @Override public Observable<T> call(Single<T> t) {
        return t.toObservable();
    }
}).lift(OperatorMerge.<T>instance(false));
```

src/main/java/rx/Single.java
```diff
@@ -931,6 +931,110 @@ public void onError(Throwable error) {
...
+     * <dd>{@code merge} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> merge(Observable<? extends Single<T>> source) {
```
akarnokd, line 951: Experimental

src/main/java/rx/Single.java
```diff
@@ -931,6 +931,110 @@ public void onError(Throwable error) {
...
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @param maxConcurrent
+     *            the maximum number of Singles that may be subscribed to concurrently
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> merge(Observable<? extends Single<T>> source, int maxConcurrent) {
```
akarnokd, line 975: Experimental

src/main/java/rx/Single.java
```diff
@@ -931,6 +931,110 @@ public void onError(Throwable error) {
...
+     * <dd>{@code merge} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> mergeDelayError(Observable<? extends Single<T>> source) {
```
akarnokd, line 1003: Experimental

src/main/java/rx/Single.java
```diff
@@ -931,6 +931,110 @@ public void onError(Throwable error) {
...
+     *
+     * @param <T> the common value type
+     * @param source
+     *            an Observable of Singles to be merged
+     * @param maxConcurrent
+     *            the maximum number of Singles that may be subscribed to concurrently
+     * @return an Observable that emits all of the items emitted by the source Singles
+     * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
+     */
+    public static <T> Observable<T> mergeDelayError(Observable<? extends Single<T>> source, int maxConcurrent) {
```
akarnokd, line 1033: Experimental

akarnokd (changes requested):
Please reuse the existing merge operator.

abersnaze:
@akarnokd actually do you think that the back pressure could be streamlined because of the guarantee that they'll be one onNext for each subscription? Does it even need a custom Producer if we got rid of the maxConcurrent overloads?

akarnokd:
The [v2 version](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMapSingle.java) is optimized.

akarnokd:
> if we got rid of the maxConcurrent overloads

You may not want to have a million outstanding `Single`-based network request. Limiting flatMap's concurrency is expected here where backpressure helps naturally.

akarnokd:
Any progress on this?

akarnokd:
Closing via #5092.

PR #4992, opened 2017-01-13
2.x: A.flatMapB to eagerly check for cancellations before subscribing

akarnokd:
The `flatMap` implementations always subscribed to the generated inner source even if there was an asynchronous cancel while the function was running. With typical functions, there is only a really tiny window inside function but some users tend to block/sleep in the function and when that returns, the `flatMap` operator is already cancelled. 

If the generated inner source emitted an error disregarding its own cancellation signal (the `error()` operators do this), those errors end up in the `RxJavaPlugins.onError` and crash the app (on Android).

This PR adjusts the `flatMap` implementations to check for the disposed/cancelled state before subscribing to the inner source. For `Observable` and `Flowable`, this has practically no extra overhead as the add/remove already checks for the terminal state and can return a boolean for it. The rest require an explicit `isDisposed()` check.


JakeWharton:
:+1:

PR #5002, opened 2017-01-18
2.x: Add scheduler creation factories

### LINKED ISSUE #4993: 2.x: Should schedulers have RxThreadFactory constructor parameters?
RxJava 1 allowed for specifying a custom thread factory, but it looks like this feature did not make it to RxJava 2. Was this intentional or something that a PR would be welcome for?

ZacSweers:
Resolves #4993

This is a pretty vanilla copy from RxJava 1's implementation. Note that I had to tune NewThread scheduler to not be a singleton to support this.

We had talked about borrowing from project reactor's APIs for different overloads, let me know if you think we should add more fine-grained controls through these.

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
@@ -16,12 +16,12 @@
 
 package io.reactivex.internal.schedulers;
 
-import java.util.concurrent.*;
-import java.util.concurrent.atomic.*;
-
 import io.reactivex.Scheduler;
 import io.reactivex.disposables.*;
 import io.reactivex.internal.disposables.EmptyDisposable;
+
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 
 /**
  * Scheduler that creates and caches a set of thread pools and reuses them if possible.
@@ -37,6 +37,7 @@ public final class IoScheduler extends Scheduler {
     private static final TimeUnit KEEP_ALIVE_UNIT = TimeUnit.SECONDS;
 
     static final ThreadWorker SHUTDOWN_THREAD_WORKER;
+    final ThreadFactory threadFactory;
     final AtomicReference<CachedWorkerPool> pool;
 
     /** The name of the system property for setting the thread priority for this Scheduler. */
@@ -44,9 +45,6 @@ public final class IoScheduler extends Scheduler {
 
     static final CachedWorkerPool NONE;
     static {
-        NONE = new CachedWorkerPool(0, null);
-        NONE.shutdown();
-
         SHUTDOWN_THREAD_WORKER = new ThreadWorker(new RxThreadFactory("RxCachedThreadSchedulerShutdown"));
         SHUTDOWN_THREAD_WORKER.dispose();
 
@@ -56,6 +54,9 @@ public final class IoScheduler extends Scheduler {
         WORKER_THREAD_FACTORY = new RxThreadFactory(WORKER_THREAD_NAME_PREFIX, priority);
 
         EVICTOR_THREAD_FACTORY = new RxThreadFactory(EVICTOR_THREAD_NAME_PREFIX, priority);
+
+        NONE = new CachedWorkerPool(0, null, WORKER_THREAD_FACTORY);
+        NONE.shutdown();
     }
 
     static final class CachedWorkerPool implements Runnable {
@@ -64,11 +65,13 @@ public final class IoScheduler extends Scheduler {
         final CompositeDisposable allWorkers;
         private final ScheduledExecutorService evictorService;
         private final Future<?> evictorTask;
+        private final ThreadFactory threadFactory;
 
-        CachedWorkerPool(long keepAliveTime, TimeUnit unit) {
+        CachedWorkerPool(long keepAliveTime, TimeUnit unit, ThreadFactory threadFactory) {
             this.keepAliveTime = unit != null ? unit.toNanos(keepAliveTime) : 0L;
             this.expiringWorkerQueue = new ConcurrentLinkedQueue<ThreadWorker>();
             this.allWorkers = new CompositeDisposable();
+            this.threadFactory = threadFactory;
 
             ScheduledExecutorService evictor = null;
             Future<?> task = null;
@@ -97,7 +100,7 @@ public final class IoScheduler extends Scheduler {
             }
 
             // No cached worker found, so create a new one.
-            ThreadWorker w = new ThreadWorker(WORKER_THREAD_FACTORY);
+            ThreadWorker w = new ThreadWorker(threadFactory);
             allWorkers.add(w);
             return w;
         }
@@ -143,13 +146,22 @@ public final class IoScheduler extends Scheduler {
     }
 
     public IoScheduler() {
+        this(WORKER_THREAD_FACTORY);
+    }
+
+    /**
+     * @param threadFactory thread factory to use for creating worker threads. Note that this takes precedence over any
+     *                      system properties for configuring new thread creation. Cannot be null.
+     */
+    public IoScheduler(ThreadFactory threadFactory) {
+        this.threadFactory = threadFactory;
         this.pool = new AtomicReference<CachedWorkerPool>(NONE);
         start();
     }
 
     @Override
     public void start() {
-        CachedWorkerPool update = new CachedWorkerPool(KEEP_ALIVE_TIME, KEEP_ALIVE_UNIT);
+        CachedWorkerPool update = new CachedWorkerPool(KEEP_ALIVE_TIME, KEEP_ALIVE_UNIT, threadFactory);
         if (!pool.compareAndSet(NONE, update)) {
             update.shutdown();
         }
--- a/src/main/java/io/reactivex/internal/schedulers/SingleScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/SingleScheduler.java
@@ -12,13 +12,14 @@
  */
 package io.reactivex.internal.schedulers;
 
-import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicReference;
-
 import io.reactivex.Scheduler;
 import io.reactivex.disposables.*;
 import io.reactivex.internal.disposables.EmptyDisposable;
+import io.reactivex.internal.functions.ObjectHelper;
 import io.reactivex.plugins.RxJavaPlugins;
+
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * A scheduler with a shared, single threaded underlying ScheduledExecutorService.
@@ -26,6 +27,7 @@ import io.reactivex.plugins.RxJavaPlugins;
  */
 public final class SingleScheduler extends Scheduler {
 
+    final ThreadFactory threadFactory;
     final AtomicReference<ScheduledExecutorService> executor = new AtomicReference<ScheduledExecutorService>();
 
     /** The name of the system property for setting the thread priority for this Scheduler. */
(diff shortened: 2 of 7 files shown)


ZacSweers:
One other thing I'm worried about - is `Schedulers` the best place for this? RxJava 1 had them in its `RxJavaHooks` class. My concern is that having these alongside the normal `io()`/`computation()`/etc factories will make them look overly inviting for use and accidental abuse.

src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
```diff
@@ -56,6 +54,9 @@
         WORKER_THREAD_FACTORY = new RxThreadFactory(WORKER_THREAD_NAME_PREFIX, priority);
 
         EVICTOR_THREAD_FACTORY = new RxThreadFactory(EVICTOR_THREAD_NAME_PREFIX, priority);
+
+        NONE = new CachedWorkerPool(0, null, WORKER_THREAD_FACTORY);
```
akarnokd, line 58: Any particular reason why this was moved to the end?

src/main/java/io/reactivex/internal/schedulers/SingleScheduler.java
```diff
@@ -47,11 +49,20 @@
...
-        executor.lazySet(createExecutor());
+        this(SINGLE_THREAD_FACTORY);
+    }
+
+    /**
+     * @param threadFactory thread factory to use for creating worker threads. Note that this takes precedence over any
+     *                      system properties for configuring new thread creation. Cannot be null.
+     */
+    public SingleScheduler(ThreadFactory threadFactory) {
+        this.threadFactory = ObjectHelper.requireNonNull(threadFactory, "threadFactory was null");
```
akarnokd: Either this is unnecessary or you forgot to do this with the other new constructors.

src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -180,6 +181,90 @@ public static Scheduler from(Executor executor) {
     }
 
     /**
+     * Create an instance of the default {@link Scheduler} used for {@link Schedulers#computation()}.
+     * @return the created Scheduler instance
+     */
```
akarnokd: please add `@since 2.0.5 - experimental` as well for every new method.

src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -180,6 +181,90 @@ public static Scheduler from(Executor executor) {
...
+    @Experimental
+    public static Scheduler newComputation() {
```
akarnokd: I'm not sure we'd want to allow non-parameterized instances to be created.

src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -180,6 +181,90 @@ public static Scheduler from(Executor executor) {
...
+    /**
+     * Create an instance of the default {@link Scheduler} used for {@link Schedulers#computation()}
+     * except using {@code threadFactory} for thread creation.
+     * @param threadFactory thread factory to use for creating worker threads. Note that this takes precedence over any
+     *                      system properties for configuring new thread creation. Cannot be null.
+     * @return the created Scheduler instance
+     */
+    @Experimental
+    public static Scheduler newComputation(ThreadFactory threadFactory) {
+        return new ComputationScheduler(ObjectHelper.requireNonNull(threadFactory, "threadFactory == null"));
```
akarnokd: The usual NPE message is `threadFactory is null`.

akarnokd (changes requested):

src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
ZacSweers: It requires WORKER_THREAD_FACTORY to be initialized now since it's a parameter

src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -180,6 +181,90 @@ public static Scheduler from(Executor executor) {
...
+     * Create an instance of the default {@link Scheduler} used for {@link Schedulers#computation()}.
+     * @return the created Scheduler instance
+     */
+    @Experimental
+    public static Scheduler newComputation() {
+        return new ComputationScheduler();
+    }
+
+    /**
+     * Create an instance of the default {@link Scheduler} used for {@link Schedulers#computation()}
```
akarnokd: These need some rewording emphasizing that the there is always a new Scheduler instance, unlike with the non-new method call.

src/main/java/io/reactivex/internal/schedulers/SingleScheduler.java
ZacSweers: Oops, meant to remove it from all these constructors and just do it in the factory methods, accidentally missed this one

src/main/java/io/reactivex/schedulers/Schedulers.java
ZacSweers: Will do

src/main/java/io/reactivex/schedulers/Schedulers.java
ZacSweers: Gotcha, will update

src/main/java/io/reactivex/schedulers/Schedulers.java
ZacSweers: Could you expand on that a bit?

akarnokd:
> One other thing I'm worried about - is Schedulers the best place for this? RxJava 1 had them in its RxJavaHooks class. My concern is that having these alongside the normal io()/computation()/etc factories will make them look overly inviting for use and accidental abuse.

Another problem is that people will try to override the scheduler-init with these method calls for `RxJavaPlugins` but referencing `Schedulers` will init the default schedulers regardless.

Indeed, `RxJavaPlugins` sounds like a better place for the methods, but keep only the `ThreadFactory` variants.

ZacSweers:
Heading to bed right now but will update the PR in the morning with requested changes. Also let me know if there's anywhere I should put tests, as I wasn't sure what the right place would be.

src/main/java/io/reactivex/internal/schedulers/IoScheduler.java
akarnokd: Right.

ZacSweers:
> Indeed, RxJavaPlugins sounds like a better place for the methods, but keep only the ThreadFactory variants.

Sounds good to me, will do that in my updates tomorrow

src/main/java/io/reactivex/schedulers/Schedulers.java
akarnokd: What is the use case for creating yet another computation/io/newThread scheduler with the default settings? I thought you wanted to override the `ThreadFactory`.

akarnokd:
Yes, tests that verify the custom thread factory actually worked by checking a custom thread name for each case.

src/main/java/io/reactivex/internal/schedulers/SingleScheduler.java
ZacSweers: e5c4a9e

src/main/java/io/reactivex/schedulers/Schedulers.java
ZacSweers: 262f669

src/main/java/io/reactivex/schedulers/Schedulers.java
ZacSweers: 681b977

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -1890,4 +1888,106 @@ public void accept(final Throwable throwable) throws Exception {
...
+            w.dispose();
+        }
+    }
+
+    @Test
+    public void testCreateComputationScheduler() {
+        ThreadFactory factory = new ThreadFactory() {
+            @Override
+            public Thread newThread(Runnable r) {
+                return new Thread("Test");
```
akarnokd: I think this doesn't allow executing tasks because you ignore `r`.

ZacSweers:
> Indeed, RxJavaPlugins sounds like a better place for the methods

8b4d461

> but keep only the ThreadFactory variants.

8009333

Also tweaked the naming a bit. `newNewThread` was a little weird (went with `create`), and added `Scheduler` suffix since we're not in the `Schedulers` class anymore.

> Yes, tests that verify the custom thread factory actually worked by checking a custom thread name for each case.

I tried setting this up matching some of the cdl-based approaches in the plugin tests but wan't able to get it working (just hangs). Pushed what I had in a1029b4, any insight?

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -1890,4 +1888,106 @@ public void accept(final Throwable throwable) throws Exception {
...
+        };
+
+        Scheduler scheduler = RxJavaPlugins.createComputationScheduler(factory);
+        verifyThread(scheduler.createWorker(), new Predicate<Thread>() {
+            @Override
+            public boolean test(Thread thread) throws Exception {
+                return "Test".equals(thread.getName());
+            }
+        });
+    }
```
akarnokd, line 1944: You have to shut down the scheduler (in a finally block) to avoid leaking threads.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -927,6 +929,58 @@ public static Completable onAssembly(Completable source) {
     }
 
     /**
+     * Create an instance of the default {@link Scheduler} used for {@link Schedulers#computation()}
+     * except using {@code threadFactory} for thread creation.
+     * @param threadFactory thread factory to use for creating worker threads. Note that this takes precedence over any
```
akarnokd, line 934: Please add notes to all of these methods something like this:

> Note that the returned `Scheduler` must be shut down manually if the `ThreadFactory` doesn't  create a daemon thread, otherwise the JVM may not quit.

akarnokd: The "Note that this takes precedence..." may be misleading. The number of threads is still based on the system configuration. I'd remove this sentence entirely

ZacSweers:
Heading on vacation for a couple weeks, but let me know what you think of getting the tests to work. I should have some time here and there to update the PR

akarnokd:
I've already told you: use the runnable in the factory method on the Thread constructor.

ZacSweers:
> use the runnable in the factory method on the Thread constructor.

Aren't I just testing the test ThreadFactory implementations then, and not that the schedulers are hooked up properly to power the scheduler?

akarnokd:
https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/schedulers/RxThreadFactory.java#L66

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -1890,4 +1888,106 @@ public void accept(final Throwable throwable) throws Exception {
             RxJavaPlugins.reset();
         }
     }
+
+    private static void verifyThread(Worker w, Predicate<Thread> threadPredicate) {
+        try {
+            try {
```
JakeWharton: remove double try

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -1890,4 +1888,106 @@ public void accept(final Throwable throwable) throws Exception {
...
+                final AtomicReference<Thread> value = new AtomicReference<>();
```
JakeWharton: Maybe change to `AtomicBoolean` and just run the predicate inside the worker?

JakeWharton: Then you don't even need to supply the `Thread` to the `Predicate` actually. It can verify whatever it wants.

JakeWharton (approved):

JakeWharton:
What's the problem with the tests? They seem to be verifying that the supplied factory was used.

JakeWharton:
Oh you didn't pass the supplied `Runnable` when calling `new Thread()` inside the factory.

ZacSweers:
> Oh you didn't pass the supplied Runnable when calling new Thread() inside the factory.

Ah! This is what I was missing, thanks for pointing that out. I've updated tests (which should now all be passing), and also made them a bit more robust (full integration with schedulers) to simulate a more real world use case.

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
ZacSweers: 90ae664

ZacSweers:
Failing test looks like a flake...

ZacSweers:
Wait no it's not. Repro'd locally, will fix

ZacSweers:
Fixed, I think. I added manual shutdowns of schedulers after they're done. Only IO had this lifecycle issue over tests, but I did the shutdown in all to be safe.

ZacSweers:
Ok this time it seems actually flaky, as the same commit passed in a different travis job against my fork - https://travis-ci.org/hzsweers/RxJava/builds/195074173

akarnokd (approved):

ZacSweers:
Thanks!

PR #5003, opened 2017-01-18
2.x: Improve Flowable tests

coreynicholson:
Some tests were missing the following assertions:
* Assert onComplete is called exactly once.
* Assert onError is not called at all for any throwable.
* Assert the order in which the observable's methods are called.

If you don't think this pull request is adding any value feel free to reject it.

### DIFF
--- a/src/test/java/io/reactivex/flowable/FlowableTests.java
+++ b/src/test/java/io/reactivex/flowable/FlowableTests.java
@@ -224,7 +224,7 @@ public class FlowableTests {
         Flowable<Integer> observable = Flowable.empty();
         observable.firstElement().toFlowable().subscribe(w);
         verify(w, never()).onNext(anyInt());
-        verify(w).onComplete();
+        verify(w, times(1)).onComplete();
         verify(w, never()).onError(any(Throwable.class));
     }
 
@@ -233,7 +233,7 @@ public class FlowableTests {
         Flowable<Integer> observable = Flowable.just(1, 3, 5, 7, 9, 7, 5, 3, 1);
         observable.filter(IS_EVEN).firstElement().toFlowable().subscribe(w);
         verify(w, never()).onNext(anyInt());
-        verify(w).onComplete();
+        verify(w, times(1)).onComplete();
         verify(w, never()).onError(any(Throwable.class));
     }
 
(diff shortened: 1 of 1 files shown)


src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -224,7 +224,7 @@ public void testFirstOfNoneFlowable() {
         Flowable<Integer> observable = Flowable.empty();
         observable.firstElement().toFlowable().subscribe(w);
         verify(w, never()).onNext(anyInt());
-        verify(w).onComplete();
+        verify(w, times(1)).onComplete();
```
akarnokd, line 227: `verify(object)` is the same as `verify(object, times(1))`.

src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -252,7 +252,7 @@ public void testFirstWithPredicateOfNoneMatchingThePredicate() {
         observable.filter(IS_EVEN).firstElement().subscribe(wm);
         verify(wm, never()).onSuccess(anyInt());
         verify(wm, times(1)).onComplete();
-        verify(wm, never()).onError(isA(NoSuchElementException.class));
+        verify(wm, never()).onError(any(Throwable.class));
```
akarnokd, line 255: I think `isA` comes from the upgrade to Mockito 2.1.

src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -345,9 +345,10 @@ public void testMaterializeDematerializeChaining() {
 
         chained.subscribe(observer);
 
-        verify(observer, times(1)).onNext(1);
-        verify(observer, times(1)).onComplete();
-        verify(observer, times(0)).onError(any(Throwable.class));
+        InOrder inOrder = inOrder(observer);
```
akarnokd, line 348: There  is no particular reason to make this in-order.

src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -796,12 +797,11 @@ public void testOfTypeWithPolymorphism() {
 
         observable.subscribe(observer);
 
-        verify(observer, times(1)).onNext(l1);
-        verify(observer, times(1)).onNext(l2);
-        verify(observer, never()).onNext("123");
-        verify(observer, never()).onError(
-                any(Throwable.class));
-        verify(observer, times(1)).onComplete();
+        InOrder inOrder = inOrder(observer);
```
akarnokd, line 800: Why in order?

akarnokd (changes requested):
I don't think these changes are necessary.

src/test/java/io/reactivex/flowable/FlowableTests.java
coreynicholson: It would be incorrect if onNext happened after onComplete, right?

coreynicholson:
Ok, I will close the pull request.

Issue #4952, opened 2017-01-03
Bug with "Sample" operator in 1.2.4


kidinov:
Library version 1.2.4. 

Bug reporting inspired by that [SO question](http://stackoverflow.com/questions/41441240/sample-does-not-stop-on-completion-event/41441944#41441944).

Code sample to reproduce:

```
Observable<String> names = Observable
        .just("Mary", "Patricia", "Linda", "Barbara",
                "Elizabeth", "Jennifer", "Maria", "Susan",
                "Margaret", "Dorothy");
Observable<Long> absoluteDelayMillis = Observable
        .just(0.1, 0.6, 0.9, 1.1,
                3.3, 3.4, 3.5, 3.6,
                4.4, 4.8)
        .map(d -> (long) (d * 1_000));
Observable<String> delayedNames = names
        .zipWith(absoluteDelayMillis,
                (n, d) -> Observable
                        .just(n)
                        .delay(d, TimeUnit.MILLISECONDS))
        .flatMap(o -> o);
delayedNames
        .sample(1, SECONDS)
        .subscribe(System.out::println);
```

Output:

> Linda
> Barbara
> Susan
> Dorothy

But, as far I understand that operator, it should be:

> Linda
> Barbara
> Susan

Because "tick" on 5000 should not happen after `onComplete` called.

akarnokd:
This was a requested feature that made it into [1.1.3](https://github.com/ReactiveX/RxJava/releases/tag/v1.1.3): [Discussion](https://github.com/ReactiveX/RxJava/issues/3657), [PR](https://github.com/ReactiveX/RxJava/pull/3757).

The common documentation of [sample](http://reactivex.io/documentation/operators/sample.html) also indicates emission of the very last event in case of the termination. The RxJava diagram hasn't been updated it seems.

In 2.x, the last element in the internal buffer is not emitted in any of the `sample` variants.

akarnokd:
I can add an overload `scan(long, TimeUnit, Scheduler, boolean)` to emit the very last unsampled value before completing, but it won't get into 2.0.4 as I need to update 2 x 2 operators, plus there is a peculiar race between onComplete() and the sampling run() that needs to be addressed.

akarnokd:
See #4955 for 2.x and #5007 for 1.x.

akarnokd:
Closing via #5007.

PR #5007, opened 2017-01-19
1.x: update sample(time) diagram to indicate emission of last

akarnokd:
This PR updates the 1.x `sample(time)` diagram to indicate the last value is emitted when the sequence completes. It is the same (style) as for 2.x:

Default scheduler:
image

Custom scheduler:
image

Related: #4952


ZacSweers (approved):
LGTM

Issue #5014, opened 2017-01-24
Front page HelloWorld example does not work out of the box


dims12:
Library version is follows:

```
         <dependency>
            <groupId>io.reactivex.rxjava2</groupId>
            <artifactId>rxjava</artifactId>
            <version>2.0.4</version>
        </dependency>
```
Code is follows:

```
import io.reactivex.*;
public class HelloWorld {


   public static void main(String[] args) {
      Flowable.just("Hello world").subscribe(System.out::println);
   }

}
```

error messge is follows:

> Error:(12, 15) java: cannot access org.reactivestreams.Publisher
>   class file for org.reactivestreams.Publisher not found

akarnokd:
Are you using just plain maven? You have to manually reference `org.reactivestreams:reactive-streams:1.0.0' as dependency in maven or switch to Gradle.

dims12:
There is a maven snippet on the page and no any word about `org.reactivestreams:reactive-streams`

akarnokd:
This will be addressed with the release of 2.0.5.

akarnokd:
Closing via #5015.

renderit:
@dims12 @akarnokd Hi, did you solve this? Having the same issue. My pom.xml:
```
<dependency>
      <groupId>org.reactivestreams</groupId>
      <artifactId>reactive-streams</artifactId>
      <version>1.0.0.final</version>
    </dependency>

    <dependency>
      <groupId>io.reactivex.rxjava2</groupId>
      <artifactId>rxjava</artifactId>
      <version>2.0.7</version>
    </dependency>
```
I'm using IDEA btw. When I try:
```
Flowable.just("Hello world").subscribe(System.out::println);
```
I get error: `Method reference are not supported at this language level` around the println statement

xxLestadxx:
Hi, I started a Gradle project in intellij, but I am having the same problem, any thoughts?

akarnokd:
Works for me:

![image](https://user-images.githubusercontent.com/1269832/42451102-83a57f4a-8386-11e8-9363-05a8aaf41239.png)

xxLestadxx:
Thank you very much! ... had a typo...

Issue #4813, opened 2016-11-06
reactive-streams should be scoped as compile dependency


davidmoten:
Something is wrong in the build process that produces the released pom.xml. The pom.xml for 2.0.0 has 

```xml
<dependency>
    <groupId>org.reactivestreams</groupId>
    <artifactId>reactive-streams</artifactId>
    <version>1.0.0</version>
    <scope>runtime</scope>
</dependency>
```

Scope for this dependency should be `compile`.  As it stands maven builds of projects that use RxJava 2 will fail with compile errors. The workaround for anyone encountering this error is to add the `reactive-streams` dependency explicitly (scoped `compile`) above the `rxjava` dependency.

akarnokd:
Found [this](https://discuss.gradle.org/t/maven-publish-plugin-generated-pom-making-dependency-scope-runtime/7494/14).

akarnokd:
See #5015. It works locally for me:

![image](https://cloud.githubusercontent.com/assets/1269832/22243465/bdf6ac8e-e227-11e6-8b3f-c7d8f1abf4b9.png)

akarnokd:
https://oss.jfrog.org/libs-snapshot/io/reactivex/rxjava2/rxjava/2.0.0-DP0-SNAPSHOT/rxjava-2.0.0-DP0-20170124.111838-392.pom

Closing via #5015

PR #5015, opened 2017-01-24
Override RS dependency to compile, replace owner info

akarnokd:
This PR overrides the dependency of Reactive-Streams in the generated POM.xml from `runtime` to `compile` to avoid trouble with plain maven users.

In addition, I took the liberty of overriding the responsible owner references which seem to be hard-coded in the `rxjava-nebula` plugin.

Related: #4813, #5014

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -198,3 +198,26 @@ findbugsMain {
     }
 }
 */
+
+// from https://discuss.gradle.org/t/maven-publish-plugin-generated-pom-making-dependency-scope-runtime/7494/10
+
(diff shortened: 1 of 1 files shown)


davidmoten:
Great, thanks for sorting that out @akarnokd

akarnokd:
Once this passes, I'll merge it so the snapshot gets published and we can verify it [here](https://oss.jfrog.org/libs-snapshot/io/reactivex/rxjava2/rxjava/2.0.0-DP0-SNAPSHOT/).

PR #5020, opened 2017-01-25
2.x: option to fail for using blockingX on the computation scheduler

akarnokd:
This PR adds an `RxJavaPlugins` option `failOnNonBlockingScheduler` that triggers an `IllegalStateException` when the user tries to run a blocking method while the execution is on the `computation()` or `single()` `Scheduler`: 

```java
Flowable.just(1)
.subscribeOn(Schedulers.computation())
.map(v -> Flowable.just("query").subscribeOn(Schedulers.io()).blockingFirst())
.doOnNext(v -> someAPI(v).subscribeOn(Schedulers.newThread()).blockingSubscribe());
.blockingFirst();
```

It is an optional setting, default off.

The check is done before going into an await method (and a few other types of blocking). Most blocking operators usually poll the status and try to avoid the actual blocking thus this shouldn't affect synchronous sequences that one extracts a value from.

Detection of a blocking-sensitive scheduler's thread is done by checking the current thread's class for implementing the `NonBlockingThread` marker interface (currently `internal`).

The `RxThreadFactory` has been updated to allow picking a default `Thread` implementation or a custom one for the `newThread()`. Note that since #5002 you can create custom schedulers by providing a `ThreadFactory`.

This works for RxJava's default schedulers but not for `AndroidSchedulers.mainThread()` where similar blocking should be avoided as well. For them, a plugin-callback action would be more suitable. ~~Question is how that callback should behave (throw, return false, should it be always executed or only when the flag is true).~~

My proposed solution is to have a plugin callback `RxJavaPlugins.setOnBeforeBlocking(BooleanSupplier)` that Android users can define the callback for:

```java
RxJavaPlugins.setOnBeforeBlocking(() -> Looper.myLooper() == Looper.getMainLooper())
RxJavaPlugins.setFailOnNonBlockingScheduler(true);
```

This callback is only executed if the `failOnNonBlockingScheduler` is set to true.

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -13,7 +13,7 @@
 package io.reactivex.plugins;
 
 import io.reactivex.*;
-import io.reactivex.annotations.Experimental;
+import io.reactivex.annotations.*;
 import io.reactivex.flowables.ConnectableFlowable;
 import io.reactivex.functions.*;
 import io.reactivex.internal.functions.ObjectHelper;
@@ -85,6 +85,8 @@ public final class RxJavaPlugins {
 
     static volatile BiFunction<Completable, CompletableObserver, CompletableObserver> onCompletableSubscribe;
 
+    static volatile BooleanSupplier onBeforeBlocking;
+
     /** Prevents changing the plugins. */
     static volatile boolean lockdown;
 
@@ -112,7 +114,8 @@ public final class RxJavaPlugins {
     }
 
     /**
-     * Enables or disables the blockingX operators to fail on a non-blocking
+     * Enables or disables the blockingX operators to fail
+     * with an IllegalStateException on a non-blocking
      * scheduler such as computation or single.
      * @param enable enable or disable the feature
      * @since 2.0.5 - experimental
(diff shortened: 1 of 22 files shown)


src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -929,6 +967,56 @@ public static Completable onAssembly(Completable source) {
...
+     * Set the handler that is called when an operator attemts a blocking
+     * await; the handler should return true to prevent the blocking
+     * and to signal an IllegalStateException instead.
+     * @param handler the handler to set, null resets to the default handler
+     * that always returns false
+     * @see #onBeforeBlocking()
+     * @since 2.0.5 - experimental
+     */
+    @Experimental
+    public static void setOnBeforeBlocking(BooleanSupplier handler) {
```
JakeWharton, line 1001: Do you have a use case in mind for exposing this API in addition to just the boolean?

JakeWharton (approved):

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
akarnokd: See the updated PR description.

Issue #4876, opened 2016-11-23
2.x: Add @NonNull annotation for Function/Predicate/etc.


dlew:
One of the larger tasks when transitioning from 1 to 2 is removing the use of null. Right now there's no automated way to do it, but if we had `@NonNull` then we'd be able to use static analysis to find where we're returning null when we shouldn't be.

I'm not sure the best way to go about it - if we should include jsr305, use Java 8's NonNull, create our own, or something else.

akarnokd:
We can't do external dependencies like jsr305 and I can't find `@NonNull` or `@NotNull` in Java 8's public API. We could introduce our own if your tools support custom annotations for this purpose.

dlew:
It looks like Java 8 doesn't actually include it - it's part of a checker framework. More details [here](https://blogs.oracle.com/java-platform-group/entry/java_8_s_new_type).

Regardless, I'm specifically targeting IntelliJ's "Constant Conditions & Expressions" inspection and it looks like you can tell IntelliJ about which annotations should be considered non-null / nullable. So it would work if RxJava had its own `@NonNull` annotation included with the library (as long as developers are willing to do a little configuration on their end to accept it).

The only downside to not using jsr305 is that we can't use something like `@ParametersAreNonnullByDefault` which would be less verbose.

If people are on board with this, I could start implementing this (but it's a lot of busywork so I don't want to bother unless people want it).

akarnokd:
Looks like there is an interest for new annotations, #4878, so it this has to be coordinated with that. I'm not sure about adding `@NonNull` to the functional interfaces though as it may add internal noise to RxJava. We could, however, annotate parameters and return types of the base reactive classes.

dlew:
Hmm, I'm most concerned about the functional interfaces, actually. Those are the ones where the app may accidentally crash during execution unexpectedly.

Here's a contrived example:

```java
void main() {
    Observable.just("Dan")
        .map(new Function<String, String>() {
            @Override
            public String apply(String s) throws Exception {
                return usuallyDependableMethod(s);
            }
        })
        .subscribe();
}

String usuallyDependableMethod(String name) {
    return (Math.random() < .95) ? "Hello " + name : null;
}
```

I can imagine a lot of circumstances where coders don't even realize that `usuallyDependableMethod()` could sometimes return null (when it's not such an obvious problem). That's where static analysis would really help out.

Jawnnypoo:
[...]

PR #5023, opened 2017-01-27
Add nullability annotations to io.reactivex.annotations interfaces.

tonycosentini:
(Addresses https://github.com/ReactiveX/RxJava/issues/4876)

This pull request annotates everything in the `io.reactivex.annotations` package with a new `@NonNull` annotation defined in RxJava. This doesn’t annotate everything, but I think it’s a good start.

In particular, when migrating a large codebase of RxJava 1.x code to 2.x, catching null returns in these interfaces was very difficult. At best, it’s noticed while migrating, otherwise it’s up to unit tests or a production crash to catch these.

With these new annotations, you get much better IDE support (once you tell IntelliJ about the new annotation):
screen shot 2017-01-27 at 2 30 18 pm
^ in this example, `getPhoneNumber()` is marked as `@Nullable`

In addition, support for this annotation can easily be added to static analysis tools like Infer, checker, or any other popular tool.

Open Questions
* Does this even make sense to add? There seem to be some concerns in the issue.
* The `@NonNull` annotation needs JavaDoc - I’m planning on following up with whatever is in existing NonNull annotations floating around unless anyone thinks otherwise.
* There are no tests since this is essentially just metadata. Should there be something that enforces this? I'm not sure if there is any kind of lint tool that runs on this project, but one approach would be to write a check that enforces everything in specific packages in annotated. This would also help keep future changes annotated.
* Is it worth adding this elsewhere? I find these function interfaces to be the easiest place to run into nullability problems, but I'm curious what others think. A simple, but kind of crazy brute force approach would be to require annotations on every public API.

### DIFF
--- a/src/main/java/io/reactivex/annotations/NonNull.java
+++ b/src/main/java/io/reactivex/annotations/NonNull.java
@@ -11,20 +11,20 @@
  * the License for the specific language governing permissions and limitations under the License.
  */
 
-package io.reactivex.functions;
+package io.reactivex.annotations;
 
-import io.reactivex.annotations.NonNull;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
 
-/**
- * A functional interface (callback) that returns true or false for the given input value.
- * @param <T> the first value
- */
-public interface Predicate<T> {
-    /**
-     * Test the given input value and return a boolean.
-     * @param t the value
-     * @return the boolean result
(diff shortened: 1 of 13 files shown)


vanniktech (approved):
Looks good. I'd say this is for a first try just enough. Would not annotate more.

dlew:
This looks fine to me as a start.

akarnokd:
Great!

PR #5033, opened 2017-01-30
2.x stop distinctUntilChanged from updating its value unless changed

ansman:
This pull request will make the `FlowableDistinctUntilChanged` and `ObservableDistinctUntilChanged` operators keep the first of equal values.

The reasoning for this is the following code:
```
observable = someSource
    .distinctUntilChanged()
    .replay(1)
    .autoConnect();
```

Currently this code will will cache two values if the last two values are equal. Replay will cache the first of them and `distinctUntilChanged` will cache the second since it will update the internal reference even if the items are equal.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChanged.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChanged.java
@@ -84,10 +84,10 @@ public final class FlowableDistinctUntilChanged<T, K> extends AbstractFlowableWi
                 key = keySelector.apply(t);
                 if (hasValue) {
                     boolean equal = comparer.test(last, key);
-                    last = key;
                     if (equal) {
                         return false;
                     }
+                    last = key;
                 } else {
                     hasValue = true;
                     last = key;
@@ -173,10 +173,10 @@ public final class FlowableDistinctUntilChanged<T, K> extends AbstractFlowableWi
                 key = keySelector.apply(t);
                 if (hasValue) {
                     boolean equal = comparer.test(last, key);
-                    last = key;
                     if (equal) {
                         return false;
                     }
+                    last = key;
                 } else {
                     hasValue = true;
                     last = key;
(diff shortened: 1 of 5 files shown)


akarnokd:
This is an odd change about an internal behavior which shouldn't affect user code. What is your use case?

ansman:
I have a singleton observable that looks like this:
```
this.observable = Observable.concat(loadFromCache(), loadFromNetwork())
    .distinctUntilChanged()
    .replay(1)
    .autoConnect();
```

The data loaded is quite large and I only want to cache a single value (the latest).
Most of the time the cached value is the same as the one loaded from network and since `distinctUntilChanged` updates its internal value even if it doesn't emit it I'll end up caching two of these large objects. One which will never be used.

Another solution to this problem is to simply clear the reference in `onError` and `onComplete`.

akarnokd:
Changing an existing and established operator is rarely the answer to one's individual problem, especially if not discussed beforehand.

> Another solution to this problem is to simply clear the reference in onError and onComplete.

Yes, this would be a better option for a PR changing `distinctUntilChanged()`.

However, it doesn't help `replay(1)` because if both sources emit distinct values, `replay` will still retain the first one for continuity reasons.

Other alternative is to use `onTerminateDetach()` to drop any reference to `distinctUntilChanged()`

akarnokd:
> Yes, this would be a better option for a PR changing `distinctUntilChanged()`.

Note that this requires working out the fusion behavior on this operator which requires non-trivial changes.

ansman:
Ok, `onTerminateDetach` seems like a good option then. 

Regarding not caching multiple values. Is the only way around this to store the value elsewhere and remove the observable when it's completed?

akarnokd:
It's unclear what you do if the cache is overwritten because of the network source returns something different.

A guaranteed single element store is `BehaviorSubject` but you have to suppress `onComplete()`:

```java
BehaviorSubject<Data> subject = BehaviorSubject.create();

Observable.concat(loadFromCache(), loadFromNetwork())
    .distinctUntilChanged()
    .onTerminateDetach()
    .subscribe(v -> subject.onNext(v), e -> subject.onError(e));

// use subject

subject.take(1).subscribe(...)
```

ansman:
That's actually a lot better. I've been meaning to find a way to "restart" the fetch later without any of the observers having to do anything. Thanks

PR #5036, opened 2017-01-31
2.x: reintroduce OnErrorNotImplementedException for 0-1 argument subscribe()

akarnokd:
This PR reintroduces the `OnErrorNotImplementedException` wrapper from 1.x and applies it to the `subscribe()` methods that don't define an `onError` handler.

The errors are still routed to the `RxJavaPlugins.onError` handler but now wrapped with `OnErrorNotImplementedException`.

This should help with cases where the developer forgot to add the handler and distinguish such unhandled errors from other undeliverable errors due to lifecycle limitations.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -1702,6 +1702,10 @@ public abstract class Completable implements CompletableSource {
      * Subscribes to this Completable and calls the given Action when this Completable
      * completes normally.
      * <p>
+     * If the Completable emits an error, it is wrapped into an
+     * {@link io.reactivex.exceptions.OnErrorNotImplementedException OnErrorNotImplementedException}
+     * and routed to the RxJavaPlugins.onError handler.
+     * <p>
      * If this Completable emits an error, it is sent to RxJavaPlugins.onError and gets swallowed.
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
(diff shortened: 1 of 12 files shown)


akarnokd:
# Coverage: 95.51%

https://codecov.io/gh/ReactiveX/RxJava/pull/5036

akarnokd:
(Messing with codecov settings, had to reopen this to trigger a webhook of theirs.)

davidmoten (approved):

akarnokd:
I couldn't get Travis re-execute the job, likely due to the problems they are facing right now (even though they claim its only MacOSX builds).

JakeWharton (commented):
:+1:

PaulWoitaschek:
So for the application usage it makes sense to re-throw everything of instance of `OnErrorNotImplementedException` and to log everything else?

My app became quite unstable with RxJava 2 because I'm silently swallowing errors and can't distinguish between my errors and errors thrown after I unsubscribed.

Really looking forward for this change if it does what I think it does!

akarnokd:
If you are using `subscribe()` or `subscribe(Consumer)` and the sequence runs into an error, the error handler will receive an `OnErrorNotImplementedException(originalError)`.

src/main/java/io/reactivex/Completable.java
```diff
@@ -1702,6 +1702,10 @@ public final Disposable subscribe(final Action onComplete, final Consumer<? supe
      * Subscribes to this Completable and calls the given Action when this Completable
      * completes normally.
      * <p>
+     * If the Completable emits an error, it is wrapped into an
```
PaulWoitaschek, line 1705: Now it's twice.

> If the completable emits an error...

PaulWoitaschek: Once with the note about OnErrorNotImplemented..., once as swallowed

akarnokd: Great opportunity to post a PR!

Issue #5044, opened 2017-02-02
Declare subscription order for amb and ambWith in documentation.


JakeWharton:
`amb()` should declare that it subscribes to the supplied observables in the order of the iterable/array. `ambWith()` should declare that it subscribes to the host observable before the supplied 'other' observable.

This allows consumers to explicitly rely on this ordering and a deviation from it to be a bug.

JakeWharton:
I can send a PR for this, I just wanted to make sure that the subscription order is something that we want to put into the public API. In theory a perfectly acceptable implementation of `ambWith` _could_ be `amb(other, this)` (but that'd be bad!).

akarnokd:
This ordered subscribe is implied, what makes you want this to be explicitly defined?

JakeWharton:
We have code whose behavior depends on it. It basically does
behaviorSubject.ambWith(just(sentinel)) to ensure a value is always
synchronously emitted downstream upon subscribe. Would the subscription
order of this (or amb) change we'd never see the subject value when it was
set.

akarnokd:
This looks like you need `BehaviorSubject.createDefault(sentinel)` (or `create(sentinel)` to make sure you have something to start with when subscribing.

Generally, I don't think such ordering guarantee should explicitly be declared on `amb`.

JakeWharton:
That was a contrived example so your suggestion won't work in practice. I
see no reason to omit ordering guarantees around subscription since the
behavior in the synchronous case relies on it. Also it would be completely
illogical to have any ordering but the obvious one that already exists, so
this really isn't a commitment of any challenge.

akarnokd:
Okay then, but don't forget to add unit tests that verify the order is actually held by the implementation (in case of another complete rewrite for 3.x). 5 base types x 3 `amb` types == 15 tests.

valeriyo:
Essentially, docs should (more) explicitly state that this:

`assertThat(amb(just(1), just(2)).toBlocking().single()).isEqualTo(1);`

akarnokd:
Closing via #5047.

PR #5047, opened 2017-02-02
2.x: Document and test amb subscription ordering.

JakeWharton:
Refs #5044.


JakeWharton:
Once wording and tests are approved I'll port to 1.x as well.

akarnokd (approved):
Nice!

(Forgot about testing all signal types as well, so 5 x 3 x 3, but I don't require it.)

JakeWharton:
Hmm I seem to be fighting the cool documentation checker which is being a bit too clever in this instance. I'm using the word "Subscription" at the beginning of a sentence which it thinks means `Subscription` the type. I'll figure out a wording tweak to make it end up as a lowercase "subscription".

JakeWharton:
Also with regard to signal types, I figure the other tests cover those already so in these cases we're just guarding against ordering changes.

akarnokd:
Yes, lowercase it as `Subscription` is interpreted as the reference to the type.

PR #5051, opened 2017-02-02
Feature/#4876 nonnull annotations

jschneider:
Issue: #4876

Starting to add @NonNull and @Nullable annotations.

I decided to add a compileOnly dependency to findbugs:jsr305 artifact.

We could also use "io.reactivex.annotations.NonNull" which already exists. But there is no "io.reactivex.annotations.Nullable" annotation which is at least as important in my experience.

Since these are only annotations, it is no problem when the class files are missing at compile time (JLS 9.6.1.2 Retention).

I have started to add annotations in the Scheduler and RxJavaPlugins.
The test RxJavaPlugins contains some invalid checks using null (lines 1353 and following).

This pull request is work in progress and should be discussed.

Questions so far:
- use jsr305 (compileOnly) or existing annotation in io.reactivex?
- (if not using jsr305): Use own @Nullable annotation or try to get one added to "io.reactivex"? Or skip these completely (which misses the point of the static code analysis)
- Why has there been tests calling the RxJavaPlugins.on*Scheduler with null arguments? Can these be removed securely?

### DIFF
--- a/src/main/java/io/reactivex/Scheduler.java
+++ b/src/main/java/io/reactivex/Scheduler.java
@@ -16,6 +16,7 @@ package io.reactivex;
 import java.util.concurrent.TimeUnit;
 
 import io.reactivex.annotations.Experimental;
+import io.reactivex.annotations.NonNull;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.Exceptions;
 import io.reactivex.functions.Function;
@@ -61,6 +62,7 @@ public abstract class Scheduler {
      *
      * @return a Worker representing a serial queue of actions to be executed
      */
+    @NonNull
     public abstract Worker createWorker();
 
     /**
@@ -69,7 +71,7 @@ public abstract class Scheduler {
      * @return the 'current time'
      * @since 2.0
      */
-    public long now(TimeUnit unit) {
+    public long now(@NonNull TimeUnit unit) {
         return unit.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
     }
 
@@ -105,7 +107,8 @@ public abstract class Scheduler {
      * @return the Disposable instance that let's one cancel this particular task.
      * @since 2.0
      */
-    public Disposable scheduleDirect(Runnable run) {
+    @NonNull
+    public Disposable scheduleDirect(@NonNull Runnable run) {
         return scheduleDirect(run, 0L, TimeUnit.NANOSECONDS);
     }
 
@@ -122,7 +125,8 @@ public abstract class Scheduler {
      * @return the Disposable that let's one cancel this particular delayed task.
      * @since 2.0
      */
-    public Disposable scheduleDirect(Runnable run, long delay, TimeUnit unit) {
+    @NonNull
+    public Disposable scheduleDirect(@NonNull Runnable run, long delay, @NonNull TimeUnit unit) {
         final Worker w = createWorker();
 
         final Runnable decoratedRun = RxJavaPlugins.onSchedule(run);
@@ -159,7 +163,8 @@ public abstract class Scheduler {
      * @return the Disposable that let's one cancel this particular delayed task.
      * @since 2.0
      */
-    public Disposable schedulePeriodicallyDirect(Runnable run, long initialDelay, long period, TimeUnit unit) {
+    @NonNull
+    public Disposable schedulePeriodicallyDirect(@NonNull Runnable run, long initialDelay, long period, @NonNull TimeUnit unit) {
         final Worker w = createWorker();
 
         final Runnable decoratedRun = RxJavaPlugins.onSchedule(run);
@@ -249,7 +254,8 @@ public abstract class Scheduler {
      */
     @SuppressWarnings("unchecked")
     @Experimental
-    public <S extends Scheduler & Disposable> S when(Function<Flowable<Flowable<Completable>>, Completable> combine) {
+    @NonNull
+    public <S extends Scheduler & Disposable> S when(@NonNull Function<Flowable<Flowable<Completable>>, Completable> combine) {
         return (S) new SchedulerWhen(combine, this);
     }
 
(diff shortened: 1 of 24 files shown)


jschneider:
Related issue: #4876

akarnokd:
Maybe it wasn't well articulated, but based on #5023 I thought we try with just the functional interfaces and come back later if it was not enough.

> I decided to add a compileOnly dependency to findbugs:jsr305 artifact.

We rather use our own annotations as most tools allow customizing what annotations to consider.

Expect futher comments inline.

src/main/java/io/reactivex/Scheduler.java
```diff
@@ -122,7 +126,8 @@ public Disposable scheduleDirect(Runnable run) {
      * @return the Disposable that let's one cancel this particular delayed task.
      * @since 2.0
      */
-    public Disposable scheduleDirect(Runnable run, long delay, TimeUnit unit) {
+    @Nonnull
+    public Disposable scheduleDirect(@Nonnull Runnable run, long delay, TimeUnit unit) {
```
akarnokd: `TimeUnit`s should also receive our own non-null annotation.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -28,68 +29,84 @@
...
     static volatile Function<ConnectableObservable, ConnectableObservable> onConnectableObservableAssembly;
 
     @SuppressWarnings("rawtypes")
+    @Nullable
     static volatile Function<Maybe, Maybe> onMaybeAssembly;
 
     @SuppressWarnings("rawtypes")
+    @Nullable
     static volatile Function<Single, Single> onSingleAssembly;
-
```
akarnokd: Please don't remove these empty lines.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -664,7 +712,7 @@ public static void setOnCompletableAssembly(Function<Completable, Completable> o
      * @param onCompletableSubscribe the hook function to set, null allowed
      */
     public static void setOnCompletableSubscribe(
-            BiFunction<Completable, CompletableObserver, CompletableObserver> onCompletableSubscribe) {
+      @Nullable BiFunction<Completable, CompletableObserver, CompletableObserver> onCompletableSubscribe) {
```
akarnokd: Indentation

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -761,7 +809,7 @@ public static void setOnConnectableObservableAssembly(Function<ConnectableObserv
      */
     @SuppressWarnings("rawtypes")
     public static void setOnObservableSubscribe(
-            BiFunction<Observable, Observer, Observer> onObservableSubscribe) {
+      @Nullable BiFunction<Observable, Observer, Observer> onObservableSubscribe) {
```
akarnokd: Please adjust indentation.

src/test/java/io/reactivex/internal/operators/flowable/FlowableReplayTest.java
```diff
@@ -708,14 +708,16 @@ public static Worker workerSpy(final Disposable mockDisposable) {
             this.mockDisposable = mockDisposable;
         }
 
+        @javax.annotation.Nonnull
```
akarnokd: I don't think it is worth updating the unit tests.

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -1349,13 +1349,13 @@ public void onComplete() {
...
-            assertNull(RxJavaPlugins.onComputationScheduler(null));
-
-            assertNull(RxJavaPlugins.onIoScheduler(null));
-
-            assertNull(RxJavaPlugins.onNewThreadScheduler(null));
-
-            assertNull(RxJavaPlugins.onSingleScheduler(null));
+            //this is not valid, because io.reactivex.functions.Function.apply() does not allow null parameters
+            if (false) {
+                assertNull(RxJavaPlugins.onComputationScheduler(null));
```
akarnokd: Instead of "disabling" this, change it to use the thin scheduler below and `assertSame` instead of null.

akarnokd (changes requested):
All annotations should be RxJava's own.

jschneider:
Thanks for the feedback. I will improve the pull request further.

src/main/java/io/reactivex/annotations/Nullable.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.LOCAL_VARIABLE;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+@Documented
```
akarnokd: Could you add a short javadoc to this and `@NonNull` while you are at it?

akarnokd (approved):

src/main/java/io/reactivex/Scheduler.java
jschneider: done

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
jschneider: done

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
jschneider: done

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
jschneider: done

src/test/java/io/reactivex/internal/operators/flowable/FlowableReplayTest.java
jschneider: Idea did it on itself ;-)
If it is ok, I'd just leave it.

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
jschneider: I removed the null checks completely. As you said, below are the assertSame checks. This should be good enough.

PR #5055, opened 2017-02-03
2.x: Feature/#4876 more null checks

jschneider:
I have added some more null checks to help the static code analysis.
I think I have also fixed one possible NPE in NewThreadWorker

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableCombineLatest.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableCombineLatest.java
@@ -16,6 +16,7 @@ package io.reactivex.internal.operators.flowable;
 import java.util.Iterator;
 import java.util.concurrent.atomic.*;
 
+import io.reactivex.annotations.NonNull;
 import io.reactivex.annotations.Nullable;
 import org.reactivestreams.*;
 
@@ -37,8 +38,10 @@ import io.reactivex.plugins.RxJavaPlugins;
 public final class FlowableCombineLatest<T, R>
 extends Flowable<R> {
 
+    @Nullable
     final Publisher<? extends T>[] array;
 
+    @Nullable
     final Iterable<? extends Publisher<? extends T>> iterable;
 
     final Function<? super Object[], ? extends R> combiner;
@@ -47,8 +50,8 @@ extends Flowable<R> {
 
     final boolean delayErrors;
 
-    public FlowableCombineLatest(Publisher<? extends T>[] array,
-            Function<? super Object[], ? extends R> combiner,
+    public FlowableCombineLatest(@NonNull Publisher<? extends T>[] array,
+                    @NonNull Function<? super Object[], ? extends R> combiner,
                     int bufferSize, boolean delayErrors) {
         this.array = array;
         this.iterable = null;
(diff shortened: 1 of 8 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableCombineLatest.java
```diff
@@ -73,6 +76,7 @@ public void subscribeActual(Subscriber<? super R> s) {
         Publisher<? extends T>[] a = array;
         int n;
         if (a == null) {
+            assert iterable != null; //either array or iterable are initialized with non null values
```
akarnokd: Asserts are off by default but they increase the bytecode size; besides we don't use `assert` it anyway.

src/main/java/io/reactivex/internal/operators/flowable/FlowableWindow.java
```diff
@@ -246,6 +246,7 @@ public void onNext(T t) {
 
             if (i == size) {
                 window = null;
+                assert w != null;
```
akarnokd: Please remove these asserts.

src/main/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromMany.java
```diff
@@ -33,21 +35,22 @@
...
-    public FlowableWithLatestFromMany(Publisher<T> source, Publisher<?>[] otherArray, Function<? super Object[], R> combiner) {
+    public FlowableWithLatestFromMany(@NonNull Publisher<T> source, @NonNull Publisher<?>[] otherArray, Function<? super Object[], R> combiner) {
         super(source);
         this.otherArray = otherArray;
         this.otherIterable = null;
         this.combiner = combiner;
     }
 
-    public FlowableWithLatestFromMany(Publisher<T> source, Iterable<? extends Publisher<?>> otherIterable, Function<? super Object[], R> combiner) {
+    public FlowableWithLatestFromMany(@NonNull Publisher<T> source, @NonNull Iterable<? extends Publisher<?>> otherIterable, Function<? super Object[], R> combiner) {
```
akarnokd: `combiner` needs annotation

src/test/java/io/reactivex/schedulers/NewThreadSchedulerTest.java
```diff
@@ -115,4 +115,18 @@ public void run() {
 
         assertEquals(0, calls[0]);
     }
+
+    @Test
+    public void npe() throws Exception {
```
akarnokd: This could be named a bit more informatively.

akarnokd (changes requested):

src/main/java/io/reactivex/internal/operators/flowable/FlowableCombineLatest.java
jschneider: ok. removed it.
Do you have a standard way to suppress false positives?

src/main/java/io/reactivex/internal/operators/flowable/FlowableWindow.java
jschneider: done

src/main/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromMany.java
jschneider: done

src/test/java/io/reactivex/schedulers/NewThreadSchedulerTest.java
jschneider: yep. added some comments

akarnokd (approved):

PR #5059, opened 2017-02-03
When declaring Functions/BiFunctions as parameter, ? super / ? extends is used

jschneider:
I think a function should never be declared without "super/extends" as method parameter.
I have found a few instances where this has been the case.

In the most common cases this shouldn't make a difference, but sometimes it helps a lot...

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -10650,7 +10650,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Maybe<T> reduce(BiFunction<T, T, T> reducer) {
+    public final Maybe<T> reduce(BiFunction<? super T, ? super  T, ? extends T> reducer) {
         ObjectHelper.requireNonNull(reducer, "reducer is null");
         return RxJavaPlugins.onAssembly(new FlowableReduceMaybe<T>(this, reducer));
     }
@@ -11916,7 +11916,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.FULL)
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Flowable<T> scan(BiFunction<T, T, T> accumulator) {
+    public final Flowable<T> scan(BiFunction<? super T, ? super T, ? extends T> accumulator) {
         ObjectHelper.requireNonNull(accumulator, "accumulator is null");
         return RxJavaPlugins.onAssembly(new FlowableScan<T>(this, accumulator));
     }
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -14668,7 +14668,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <K> Single<Map<K, Collection<T>>> toMultimap(Function<? super T, ? extends K> keySelector) {
-        Function<T, T> valueSelector = Functions.identity();
+        Function<? super T, ? extends T> valueSelector = Functions.identity();
         Callable<Map<K, Collection<T>>> mapSupplier = HashMapSupplier.asCallable();
         Function<K, List<T>> collectionFactory = ArrayListSupplier.asFunction();
         return toMultimap(keySelector, valueSelector, mapSupplier, collectionFactory);
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -52,7 +52,7 @@ public final class RxJavaPlugins {
     static volatile Consumer<? super Throwable> errorHandler;
 
     @Nullable
-    static volatile Function<Runnable, Runnable> onScheduleHandler;
+    static volatile Function<? super Runnable, ? extends Runnable> onScheduleHandler;
 
     @Nullable
     static volatile Function<Callable<Scheduler>, Scheduler> onInitComputationHandler;
@@ -260,7 +260,7 @@ public final class RxJavaPlugins {
      * @return the hook function, may be null
      */
     @Nullable
-    public static Function<Runnable, Runnable> getScheduleHandler() {
+    public static Function<? super Runnable, ? extends Runnable> getScheduleHandler() {
         return onScheduleHandler;
     }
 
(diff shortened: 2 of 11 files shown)


akarnokd:
This also needs rebasing. In addition, please check locally that:

  - if you manually change the compiler settings to Java 8 in your IDE, there is no compilation error
  - if you temporarily change the build.gradle to target java 8 and there is no compilation error when building.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -10650,7 +10650,7 @@ public final Completable ignoreElements() {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Maybe<T> reduce(BiFunction<T, T, T> reducer) {
+    public final Maybe<T> reduce(BiFunction<? super T, ? super  T, ? extends T> reducer) {
```
davidmoten, line 10653: This sort of change is a good idea for me but problem is that it is probably a breaking change in that user source that used to compile may no longer compile:

This 
```java
Maybe<Integer> result = flowable.reduce(f);
```
will need to be
```java
Maybe<Integer> result = flowable.<Integer>reduce(f);
```

I haven't checked this yet, can you confirm?

I can live with it. @akarnokd?

akarnokd: `reduce(T, T, T)` was discussed quite a while back and it has to stay non-variant. One of the reasons is the limits of javac and what @davidmoten showed.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -11916,7 +11916,7 @@ public final void safeSubscribe(Subscriber<? super T> s) {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.FULL)
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Flowable<T> scan(BiFunction<T, T, T> accumulator) {
+    public final Flowable<T> scan(BiFunction<? super T, ? super T, ? extends T> accumulator) {
```
akarnokd, line 11919: No good, same situation as with `reduce(T, T, T)`.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -14668,7 +14668,7 @@ public final void subscribe(Subscriber<? super T> s) {
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <K> Single<Map<K, Collection<T>>> toMultimap(Function<? super T, ? extends K> keySelector) {
-        Function<T, T> valueSelector = Functions.identity();
+        Function<? super T, ? extends T> valueSelector = Functions.identity();
```
akarnokd, line 14671: Such changes in our own source is a good indication there will be generics issues for the users of the library.

src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -16,14 +16,26 @@
 
 package io.reactivex.plugins;
 
-import io.reactivex.*;
+import io.reactivex.Completable;
```
akarnokd, line 19: Please don't unroll * imports

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -49,83 +49,83 @@
  */
 public final class RxJavaPlugins {
     @Nullable
-    static volatile Consumer<Throwable> errorHandler;
+    static volatile Consumer<? super Throwable> errorHandler;
 
     @Nullable
-    static volatile Function<Runnable, Runnable> onScheduleHandler;
+    static volatile Function<? super Runnable, ? extends Runnable> onScheduleHandler;
```
akarnokd, line 55: I'm not too keen on these variance changes unless you can show that there is an inference problem with lambdas under 8 without them.

akarnokd (changes requested):

jschneider:
Ok. So I unterstand that we leave the source code as it is.
Will close the pull request. Thanks for the feedback.

src/main/java/io/reactivex/Flowable.java
jschneider: I don't want to keep pushing a finished topic. But just for your interest: This code compiles (have tried with 1.6 and 1.8):

I have added this test case to this pull request, if anybody wants to take a look.

````
@Test
    public void basics() throws Exception {
        Flowable<Integer> flowable = Flowable.just(1, 2, 3, 4);

        BiFunction<Integer, Integer, Integer> sumReducer = new BiFunction<Integer, Integer, Integer>() {
            @Override
            public Integer apply(Integer integer, Integer integer2) throws Exception {
                return integer + integer2;
            }
        };

        Maybe<Integer> reduced = flowable.reduce(sumReducer);

        reduced.subscribe(new Consumer<Integer>() {
            @Override
            public void accept(Integer integer) throws Exception {
                assertEquals(10, integer, 0);
            }
        });
    }

    @Test
    public void reduceWithSuperType() throws Exception {
        Flowable<Integer> flowable = Flowable.just(1, 2, 3, 4);

        BiFunction<Number, Number, Integer> sumReducer = new BiFunction<Number, Number, Integer>() {
            @Override
            public Integer apply(Number integer, Number integer2) throws Exception {
                int value = integer.intValue() + integer2.intValue();
                return value;
            }
        };

        Maybe<Integer> reduced = flowable.reduce(sumReducer);
        reduced.subscribe(new Consumer<Integer>() {
            @Override
            public void accept(Integer integer) throws Exception {
                assertEquals(10, integer, 0);
            }
        });
    }

    @Test
    public void reduceWithSuperTypeNumberFlowable() throws Exception {
        Flowable<Number> flowable = Flowable.<Number>just(1, 2, 3, 4);

        BiFunction<Number, Number, Integer> sumReducer = new BiFunction<Number, Number, Integer>() {
            @Override
            public Integer apply(Number integer, Number integer2) throws Exception {
                int value = integer.intValue() + integer2.intValue();
                return value;
            }
        };

        Maybe<Number> reduced = flowable.reduce(sumReducer);
        reduced.subscribe(new Consumer<Number>() {
            @Override
            public void accept(Number integer) throws Exception {
                assertEquals(10, integer.doubleValue(), 0);
            }
        });
    }

    @Test
    public void reduceWithSuperTypeNumberFlowable2() throws Exception {
        Flowable<Number> flowable = Flowable.<Number>just(1, 2, 3, 4);

        BiFunction<Number, Number, Number> sumReducer = new BiFunction<Number, Number, Number>() {
            @Override
            public Number apply(Number integer, Number integer2) throws Exception {
                int value = integer.intValue() + integer2.intValue();
                return value;
            }
        };

        Maybe<Number> reduced = flowable.reduce(sumReducer);
        reduced.subscribe(new Consumer<Number>() {
            @Override
            public void accept(Number integer) throws Exception {
                assertEquals(10, integer.doubleValue(), 0);
            }
        });
    }

PR #5066, opened 2017-02-03
Update codecov.yml

stevepeak:
Fix codecov.yml

- `target` is used a "coverage must be greater or equal to the target"
- `threshold` is the acceptable amount of loss to still consider the status as successful.


akarnokd:
Thanks!

PR #5068, opened 2017-02-04
Update README.md

soulkeykim:
I tried to use RxJava2 from the example and I get some error message.

Error:(13, 38) java: variable v is already defined in method main(java.lang.String[])

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Flowable.range(1, 10)
   .flatMap(v ->
       Flowable.just(v)
         .subscribeOn(Schedulers.computation())
-        .map(v -> v * v)
+        .map(w -> v * v)
   )
 .blockingSubscribe(System.out::println);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -128,7 +128,7 @@ Flowable.range(1, 10)
   .flatMap(v ->
       Flowable.just(v)
         .subscribeOn(Schedulers.computation())
-        .map(v -> v * v)
+        .map(w -> v * v)
```
JakeWharton: All three of these vars need changed to `w` then.

JakeWharton (changes requested):

soulkeykim (commented):
Renamed all three of vars

akarnokd (approved):

akarnokd:
Thanks for the correction!

PR #5080, opened 2017-02-07
2.x: wrap undeliverable errors

akarnokd:
This PR adds the `UndeliverableException` and `ProtocolViolationException`. The former wraps exceptions that happen beyond the lifecycle of a flow and the latter is added to distinguish validation bugs.

The `RxJavaPlugins.onError` wraps errors into `UndeliverableException` unless `RxJavaPlugins.isBug()` returns true when the `Throwable` is relayed to the (default) handler.

Having the `UndeliverableException` should help with understanding the source of a crash and by having its own stack trace, help locating the offending positions (for example, a missed `isDisposed()` check in a 3rd party library).

### DIFF
--- a/src/main/java/io/reactivex/exceptions/ProtocolViolationException.java
+++ b/src/main/java/io/reactivex/exceptions/ProtocolViolationException.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.exceptions;
+
+import io.reactivex.annotations.Experimental;
+
+/**
+ * Explicitly named exception to indicate a Reactive-Streams
+ * protocol violation.
+ * @since 2.0.6 - experimental
+ */
+@Experimental
+public final class ProtocolViolationException extends IllegalStateException {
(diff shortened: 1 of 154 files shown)


src/main/java/io/reactivex/exceptions/ProtocolViolationException.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
```
JakeWharton, line 2: nit: 2017

src/main/java/io/reactivex/exceptions/UndeliverableException.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
```
JakeWharton, line 2: nit: 2017

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -377,6 +382,42 @@ public static void onError(@NonNull Throwable error) {
...
 
+    /**
+     * Checks if the given error is one of the already named
+     * bug cases that should pass through {@link #onError(Throwable)}
+     * as is.
+     * @param error the error to check
+     * @return true if the error should pass throug, false if
+     * it may be wrapped into an UndeliverableException
+     */
+    static boolean isBug(Throwable error) {
```
JakeWharton, line 393: Shouldn't `CompositeException` be in here as well? While technically they're _always_ undeliverable, they're very crash-worthy exceptions since they indicate an error while processing an error.

JakeWharton (approved):

src/main/java/io/reactivex/exceptions/ProtocolViolationException.java
akarnokd: All our headers are like this, no current year.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
akarnokd: Updated. A few tests have been reverted from assertUndeliverable to assertError. I've also added an isBug unit test.

PR #5093, opened 2017-02-12
2.x: Add doAfterTerminate callback to the Single type.

VeskoI:
As mentioned in [THIS StackOverflow question](http://stackoverflow.com/questions/42168043/why-is-doafterterminate-missing-from-single-in-rxjava-2), doAfterTerminate() is currently missing in RxJava2. There's no strong reason for that, so this PR adds support for it.

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1814,6 +1814,7 @@ public abstract class Single<T> implements SingleSource<T> {
      * @return a Single that emits the same items as the source Single, then invokes the
      *         {@link Action}
      * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX operators documentation: Do</a>
+     * @since 2.0.6 - experimental
      */
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1799,6 +1799,31 @@ public abstract class Single<T> implements SingleSource<T> {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -1799,6 +1799,31 @@
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code doAfterTerminate} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param onAfterTerminate
+     *            an {@link Action} to be invoked when the source Single finishes
+     * @return a Single that emits the same items as the source Single, then invokes the
+     *         {@link Action}
+     * @see <a href="http://reactivex.io/documentation/operators/do.html">ReactiveX operators documentation: Do</a>
+     */
```
akarnokd: Looks good, please add `@since 2.0.6 - experimental` to the javadoc.

src/main/java/io/reactivex/internal/operators/single/SingleDoAfterTerminate.java
```diff
@@ -0,0 +1,100 @@
...
+import io.reactivex.disposables.Disposable;
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.functions.Action;
+import io.reactivex.internal.disposables.DisposableHelper;
+import io.reactivex.plugins.RxJavaPlugins;
+
+/**
+ * Calls an action after pushing the current item or an error to the downstream.
+ * @param <T> the value type
+ */
```
akarnokd: Please add `@since 2.0.6 - experimental` here as well. No need for the `@Experimental` annotation.

akarnokd (approved):

PR #5098, opened 2017-02-16
2.x: Observable.first* with predicate

maxost:
Recently I discovered that there is no first-with-predicate operator in `2.x.` So I've implemented `Observable.firstElement(predicate)`, `Observable.first(defaultItem, predicate)` and `Observable.firstOrError(predicate)` overloads using simple constructs like `filter(predicate).firstElement()`.

Also, I've rewritten all first* tests because seems it was just copied from 1.x with do-it-later intention.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -7197,6 +7197,8 @@ public abstract class Observable<T> implements ObservableSource<T> {
      *         the {@code predicate}, or a default item if the source Observable completes without emitting any items.
      * @see <a href="http://reactivex.io/documentation/operators/first.html">ReactiveX operators documentation: First</a>
      */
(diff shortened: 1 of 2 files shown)


akarnokd:
Hi and thanks for the contribution but I wish you opened a discussion first before spending time on this. The predicate variants were deliberately left out from 2.x as they were just filter().first() etc. with low bang for a buck.

maxost:
I can form rearranged tests into separate pull request, also I spotted that some basic operators have no unit-tests, I can do that. Is there need for that and will it be accepted?

akarnokd:
Don't change existing unit tests unless you are fixing a bug in them (such as flakiness or not testing what it is supposed to). New unit tests are welcome especially if they increase the code coverage.

Issue #5110, opened 2017-02-17
2.x: Reactive-Streams compliance


akarnokd:
There is an [ongoing effort](https://github.com/reactive-streams/reactive-streams-jvm/pull/339) to clarify the reasons behind the [Reactive-Streams](https://github.com/reactive-streams/reactive-streams-jvm/) rules. In the hopes that the rules could be relaxed without too much compatibility problems, I brought up the case that we had to introduce the `strict()` operator to pass the TCK tests. The problematic rules are §1.3 and §3.9 that impose such requirements that add overhead to sequences.

This sparked a [separate discussion](https://github.com/reactive-streams/reactive-streams-jvm/issues/341) about the it was a good idea RxJava 2 exposes itself with the Reactive-Streams interfaces instead of hiding it behind a converter / exposure layer which decoupling allows - in theory - violating the exact rules as long as such violations are not visible by the consumer.

I doubt I could convince the maintainers of the specification (other than @smaldini) it is worth taking the relaxations even at the expense of temporary incompatibility in case vendor's don't relax their particular implementations.

RxJava 2 works with or without the "Reactive-Streams compliant" badge yet as a "brand", it is still advisable to have that badge.

### Suggested API changes

Ensuring conformance is a matter of applying the `StrictSubscriber` wrapper to the incoming `Subscriber` in `Flowable.subscribe`. Unfortunately, this slows down everything as there is only one way to consume a `Flowable` through `subscribe(Subscriber)` (the overloads all delegate to it).

To get the speed back, a new instance method would be introduced that all internal operators would use then on instead of `subscribe`. RxJava 1 has an extra entry point `unsafeSubscribe` to sidestep `SafeSubscriber` thus there is precedence for such alternate entry point.

The method could be called `subscribe` as well that takes a `FlowableSubscriber extends Subscriber`. `FlowableSubscriber`s textual definition can then weaken rules §1.3 and §3.9 inherited. Existing operators would then implement `FlowableSubscriber` in place of `Subscriber`.

I think all of this would be binary compatible change since only `Flowable` would get an extra method.

Suggestions welcome.

JakeWharton:
Is `FlowableSubscriber` a public type just like the non-RS observers?

I've been following the discussion too and I think this is a reasonable action to take.

akarnokd:
Yes, `io.reactivex.FlowableSubscriber`. I don't think hiding it in `internal` is necessary. I'm currently preparing a PR to demonstrate the changes.

[...]

PR #5112, opened 2017-02-18
2.x: Flowable as a Publisher to be fully RS compliant

akarnokd:
This PR performs the changes suggested in #5110.

 - Introduce `FlowableSubscriber` with extra textual specification on its relaxed nature
 - `Flowable.subscribe(Subscriber)` checks for `FlowableSubscriber` and if not found, it wraps the incoming RS `Subscriber` into a `StrictSubscriber` that follows the RS spec to the letter at any cost.
 - Introduce `Flowable.subscribe(FlowableSubscribe)` that most internal operators will use
 - Change `AbstractFlowableWithUpstream` to accept `Flowable` as a source, update operators
 - Some operators were useful with raw `Publisher` input, these were duplicated on their outer containing type but use the same internal `FlowableSubscriber`
 - Removed "cheat" from the TCK tests, adjusted timeout on `delay`
 - Replaced most `implements Subscriber` with `implements FlowableSubscriber`
 - Replaced most `new Subscriber` with `new FlowableSubscriber` in tests, the rest is required for testing the strictness itself.
 - `strict()` is now an identity operator with suggested scheduled removal.

Performance impact estimation

- Most primary use of a `Flowable` should go through `subscribe(FlowableSubscribe)` and thus no overhead change.
- Where the API mandated `Publisher` as input, providing a `Flowable` will have an `instanceof` check at subscription time and routed to `subscribe(FlowableSubscriber) if the consumer is part of RxJava 2 itself.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -23,8 +23,9 @@ import io.reactivex.exceptions.Exceptions;
 import io.reactivex.flowables.*;
 import io.reactivex.functions.*;
 import io.reactivex.internal.functions.*;
-import io.reactivex.internal.fuseable.ScalarCallable;
+import io.reactivex.internal.fuseable.*;
 import io.reactivex.internal.operators.flowable.*;
+import io.reactivex.internal.operators.flowable.FlowableStrict.StrictSubscriber;
 import io.reactivex.internal.operators.observable.ObservableFromPublisher;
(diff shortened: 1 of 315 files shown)


JakeWharton (approved):
Looks good!

akarnokd:
Okay, let's go from here. In the unlikely case the spec becomes more forgiving, the undo is just to remove cast/wrapping from Flowable.subscribe and everything works the same as `FlowableSubscriber` is an interface.

PR #5115, opened 2017-02-18
2.x: benchmark the new strict/interop mode

akarnokd:
Benchmark the overhead of the strict/interop mode.

i5 6440HQ, Windows 10 x64, Java 8u121

image

The numbers are consistent with my expectations; this mobile processor is roughly equivalent to i7 4770 desktop where the cost model is: 1 atomic increment per item equals to roughly 130 Mops/s upper limit, 2 atomic increment per item is roughly 60 Mops/s upper limit. Since the interop mode requires at minimum two atomic increments, 54 Mops/s is a reasonable value to get.

### DIFF
--- a/src/perf/java/io/reactivex/PerfInteropConsumer.java
+++ b/src/perf/java/io/reactivex/PerfInteropConsumer.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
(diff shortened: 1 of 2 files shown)


akarnokd:
Results with the updated benchmark, adding extra overhead by consuming CPU to see where the raw overhead decreases to a reasonably low value:

image

JakeWharton (approved):

PR #5119, opened 2017-02-19
2.x: improve coverage, remove unused code

akarnokd:
This PR adds coverage to some low-coverage classes and removes unused code. 

In addition, the `XFlatMapTest` now retries each test up to 3 times with exponential backoff since they are the most sensitive to Travis CI fluctuations.


JakeWharton (commented):
:+1:

PR #5125, opened 2017-02-21
1.x: reduce stack depth with switchIfEmpty

akarnokd:
Stack depths may get deep if `switchIfEmpty` has to switch to the alternative source at the end of an already long chain leading to the `onCompleted`. Using multiple `switchIfEmpty` amplifies the impact on the stack depth. 

This PR introduces trampolined subscribing to both sources (similar to how `concat` works), thus given synchronous sources, the stack depth rewinds to the level it was when the `switchIfEmpty` was subscribed to. In addition, to reduce the depth even further, the operator has been turned into an `OnSubscribe`-based implementation to avoid the extra depth due to `lift`.

Related: [rxjava StackOverflowError exception in long chain
](http://stackoverflow.com/questions/42363452/rxjava-stackoverflowerror-exception-in-long-chain)

### DIFF
--- a/src/main/java/rx/internal/operators/OnSubscribeSwitchIfEmpty.java
+++ b/src/main/java/rx/internal/operators/OnSubscribeSwitchIfEmpty.java
@@ -16,6 +16,8 @@
 package rx.internal.operators;
 
 
+import java.util.concurrent.atomic.AtomicInteger;
+
 import rx.*;
 import rx.internal.producers.ProducerArbiter;
 import rx.subscriptions.SerialSubscription;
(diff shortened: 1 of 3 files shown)


src/main/java/rx/internal/operators/OnSubscribeSwitchIfEmpty.java
```diff
@@ -69,14 +82,33 @@ public void onCompleted() {
...
-            alternate.unsafeSubscribe(as);
+        void subscribe(Observable<? extends T> source) {
+            if (wip.getAndIncrement() == 0) {
+                do {
+                    if (child.isUnsubscribed()) {
+                        break;
+                    }
+
+                    if (!active) {
+                        if (secondary) {
```
JakeWharton: Is this variable needed? Seems like `source == null` would accomplish the same thing.

JakeWharton (approved):

src/main/java/rx/internal/operators/OnSubscribeSwitchIfEmpty.java
akarnokd: Unfortunately, no. Here is a scenario:

1) Thread 1 subscribes and calls `subscribe(source)`
2) through subscribeOn, Thread 2 calls `onCompleted()`, and thus `subscribe(null)` but finds Thread 1 still inside the loop (`wip != 0`).
3) Thread 1 detects the completion of Thread 2 and loops back. Since `source != null` still in this method, the main source is resubscribed again instead of the alternative.

The `secondary`, persisted indicator works across threads and makes sure no matter where onComplete comes from, the second time we get into that inner part, the alternative source is subscribed next.

akarnokd: On a second thought, you are right. If I set `source = null` after the `source.unsafeSubscribe()` the next round will properly go into the alternative. Updated the PR.

PR #5137, opened 2017-02-25
2.x: add TestSubscriber.withTag

akarnokd:
This PR adds the method `withTag` to the `TestBaseConsumer` that allows setting a textual tag which is then appended to the assertion failure's message:

```
Failure message (latch = 1, values = 0, errors = 0, completions = 0, tag = x: 5, y: 7)
```

Use case, for example, is to add the parameters of the current flow to help identify what settings caused the failure. This comes up with unit tests that have some form of (nested) loop(s) to verify multiple parameter ranges. (It is often much easier than trying to work out JUnit rules and such.)


vanniktech (approved):
I like

JakeWharton (approved):

PR #5153, opened 2017-03-03
2.x: Java 9 compatibility fixes (March 3)

akarnokd:
Java 9 has changed its overload resolution algorithm and things that resolved unambiguously in Java 8 no longer resolve:

```java
Flowable<? extends T> source = ...

source.subscribe(new FlowableSubscriber<T>() { ... });
```

With the code above, javac 8, Eclipse and IntelliJ 2017 EAP picks `Flowable.subscribe(FlowableSubscriber<? super T> s)` as expected. However, javac 9 finds it ambiguous with `Flowable.subscribe(Subscriber<? super T> s)` despite IntelliJ 2017 EAP not indicating any error and still jumping to the right method via CTRL+Click.

The problem may come from the `? extends T` part of the declaration. Having just `Flowable<T>` compiles properly with javac9.

Luckily, we don't need `? extends T` and this PR changes the internal signatures of the affected components.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/BlockingFlowableIterable.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/BlockingFlowableIterable.java
@@ -27,11 +27,11 @@ import io.reactivex.internal.subscriptions.SubscriptionHelper;
 import io.reactivex.internal.util.*;
 
 public final class BlockingFlowableIterable<T> implements Iterable<T> {
-    final Flowable<? extends T> source;
+    final Flowable<T> source;
 
     final int bufferSize;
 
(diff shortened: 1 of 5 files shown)


JakeWharton (approved):

JakeWharton:
Hopefully Travis adds early JDK 9 builds soon.

akarnokd:
If you need only JDK 9, that's possible with some [experimental setup](https://github.com/akarnokd/RxJava2Jdk9Interop/blob/master/.travis.yml). The bigger problem is that Gradle 3.x doesn't work on Jigsaw enabled Jdk 9 builds such as 154+.

PR #5174, opened 2017-03-11
2.x: WIP removes anonymous inner classes.

SleimanJneidi:
This is a WIP. Since the changes are big, an incremental review and PR was recommended by @akarnokd 

 - Removes them  from flowable/observable/operators.
- Issue [#5150](https://github.com/ReactiveX/RxJava/issues/5150)

PR #5176, opened 2017-03-11
2.x: fix private field access, few generics problems

akarnokd:
This PR fixes some private field accessor problems introduced by #5174 as well as a couple of generics warnings.

Note that in IntelliJ, there is a J2ME inspection for private access checks between parent and inner classes.


JakeWharton (approved):

vanniktech:
PMD also has a check for those synthetic methods since 5.5.4

Issue #5172, opened 2017-03-10
1.2.7: replay(1).refCount() keeps latest value in memory even after everyone unsubscribes


ZuZuK:
Code sample:

```java
public class MainActivity extends AppCompatActivity {

    private Observable<Object> testObservable;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        testObservable = Observable.fromCallable(new Callable<Object>() {
            @Override
            public Object call() throws Exception {
                return new VeryBigObject();
            }
        })
        .replay(1)
        .refCount();

        //first subscription - callable and creation of VeryBigObject will be called
        Subscription subscription1 = testObservable.subscribe();
        //second subscription - cached value in OperatorReplay in onNext
        Subscription subscription2 = testObservable.subscribe();
        //first unsubscribe - cached value still in OperatorReplay as we have at least 1 subscriber for refCount
        subscription1.unsubscribe();
        //second unsubscribe - there are no subscribers so my guess is that there should be no cached value in OperatorReplay
        //as no one needs that object
        subscription2.unsubscribe();
        //but after everyone unsubscribe there is a reference to VeryBigObject
    }

    public static class VeryBigObject {
        //it's very big
    }

}
```

Problem: if I'm storing in field any Observable that contains replay(1).refCount() then this field will store hard reference to that latest value stored in replay buffer even if there are no subscriptions to that replay. This value takes memory and it is not good but maybe it's as-design. Also I think that this bug is actual for any replay(***) of any observables chain that have such operator.

Why am I sad: usually I use such construction to not calculate or get from disc some shared single-instance object so if there are at least one subscriber so that object is calculating on subscribe and other subscribers won't call calculation on subscribe but take already calculated value. 

Reference to VeryBigObject

        from `MainActivity` - this (root) object
        from `testObservable (Observable)` - field of MainActivity object stores Observable with replay
        from `onSubscribe (OnSubscribeRefCount)` - field of testObservable stores refCount()
        from `source (OperatorReplay)` - field of OnSubscribeRefCount stores ConnectableObservable of replay()
         from `current (AtomicReference)` - reference to current subscriber, it is only changing to not-null in connect() of OperatorReplay
         from `value (OperatorReplay$ReplaySubscriber)` - value of AtomicReference
         from `buffer (OperatorReplay$SizeBoundReplayBuffer)` - linked list of replay values with limit=1. after everyone unsubscribe it has size=2 index=2
[...]

PR #5181, opened 2017-03-13
1.x: replay().refCount() avoid leaking items between connections

akarnokd:
This PR updates `replay()` to not leak items between reconnections when run with `.refCount()`. The `replay()` operator is designed to hold onto the buffer even after its completion so late subscribers can still receive the cached data. Only a new `connect()` clears this data which may or may not happen. Since `refCount` ensures that there won't be any latecommers to an already completed connection (because it also drops its current subscribers on termination or when reaching zero), the `replay()` can be reset to an empty state.

Reported in #5172

### DIFF
--- a/src/test/java/rx/internal/operators/OnSubscribeRefCountTest.java
+++ b/src/test/java/rx/internal/operators/OnSubscribeRefCountTest.java
@@ -32,9 +32,10 @@ import rx.Observable;
 import rx.Observable.OnSubscribe;
 import rx.Observer;
 import rx.functions.*;
+import rx.observables.ConnectableObservable;
 import rx.observers.*;
 import rx.schedulers.*;
-import rx.subjects.*;
+import rx.subjects.ReplaySubject;
 import rx.subscriptions.Subscriptions;
 
 public class OnSubscribeRefCountTest {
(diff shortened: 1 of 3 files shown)


src/test/java/rx/internal/operators/OnSubscribeRefCountTest.java
```diff
@@ -611,4 +612,139 @@ public void call(Throwable t) {
...
+    public void replayNoLeak() throws Exception {
+        System.gc();
+        Thread.sleep(100);
+
+        long start = ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getUsed();
+
+        source = Observable.fromCallable(new Callable<Object>() {
+            @Override
+            public Object call() throws Exception {
+                return new byte[100 * 1000 * 1000];
```
artem-zinnatullin, line 629: Wouldn't `WeakReference` be a more determined way to detect memory leak?

```java
AtomicReference<WeakReference<Object>> ref = new AtomicReference<>();

Observable
  .fromCallable {
    Object o = new Object();
    ref.set(new WeakReference(o));
    return o;
  }
  .replay(1)
  .refCount();

…

for (i = 0; i < 10; i++) { System.gc(); Thread.sleep(1); }

assertThat(ref.get().get()).isNull()
```

artem-zinnatullin (approved):
👍

src/test/java/rx/internal/operators/OnSubscribeRefCountTest.java
akarnokd: the ~100M is almost as apparent.

artem-zinnatullin: Well, yes and no. I wouldn't trust heap of a process in which I do not control Threads and allocations (both IDE and Gradle start separate JVM with own wrappers around your tests to execute them).

Also, I'm afraid that allocation of such a relatively huge piece of memory as an array may result in OOM on constrained environment like Travis CI.

akarnokd: That's why only 100M, the OOMKiller jumps into action around 800 MB. Plus, the scheduler tests already use such large amount of memory and no kills for those.

artem-zinnatullin: Ok then!

PR #5187, opened 2017-03-15
2.x: reflection-based parameter validator & fixes

akarnokd:
This PR adds a reflection-based parameter validator to check the base reactive classes for proper parameter validation: null checks and argument ranges. The problems detected are fixed as well.


vanniktech (approved):
Nice 👍

akarnokd:
Thanks @vanniktech ! I'm merging this because there are a couple of inbound API changes from other contributors and having this in place avoids conflicts.

PR #5188, opened 2017-03-15
2.x: fix flatMap emitting the terminal exception indicator on cancel

akarnokd:
This PR fixes the case when `Flowable.flatMap` and `Observable.flatMap` would emit the `ExceptionHelper.TERMINTED` indicator exception to downstream due to emission-cancellation race.

Reported on [StackOverflow](http://stackoverflow.com/questions/42580195/what-is-the-meaning-of-no-further-exceptions).

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMap.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMap.java
@@ -421,10 +421,12 @@ public final class FlowableFlatMap<T, U> extends AbstractFlowableWithUpstream<T,
 
                 if (d && (svq == null || svq.isEmpty()) && n == 0) {
                     Throwable ex = errs.terminate();
-                    if (ex == null) {
-                        child.onComplete();
-                    } else {
-                        child.onError(ex);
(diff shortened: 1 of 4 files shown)


akarnokd:
`flatMap` is one of the few operators where it is extremely difficult to make sure certain lines get executed and this probabilistic nature may rarely manifest itself on the slow CI.

JadenChoi:
@akarnokd thanks for the fix for FlatMap, but is there a possibility that we have a similar issue with `FlatMapCompletable`?

On the same StackOverflow thread, Nantoka commented:

> Had the strange exception with Rx Java 2.2.8 as well. I had a flatmapComplete operator wrapped into a flatmapComplete operator. After replacing this with a flatmap followed by a flatmapComplete the error seems to have disappeared. Is say "seems" as the error happened sporadically. 

And we are seeing very similar exception for `FlatMapCompletable` on rxjava:2.2.19 but we aren't getting any useful stacktrace to share though

> io.reactivex.internal.util.ExceptionHelper$Termination: No further exceptions
        at unknown method(unknown file)

akarnokd:
It could be. Note though that 2.x is end-of-life so can't do much about it.

Issue #5185, opened 2017-03-14
2.x concatMapEager should accept 0 for prefetch


ansman:
The documentation for `concatMapEager` states:
```java
@param prefetch hints about the number of expected source sequence values
```

Which leads me to assume that the following code should work:
```java
Observable.fromIterable(someCollection)
    .concatMapEager(this::performTask, 5, someCollection.size())
```

but it will fail if the collection is empty. Either the documentation should be updated to reflect that 0 is not a valid value or it should accept 0 (and perhaps set it to a default value internally).

akarnokd:
Looks like there is a missing validation there as well. PR welcome. Plus, the parameter document should read something like:

`hints about the number of expected values from each inner ObservableSource`

akarnokd:
On a second look, this is not a bug in the operator but perhaps the underdocumentation. Prefetch of 0 or negative has no meaning and if you think there won't be any value, use 1 as the prefetch amount.

akarnokd:
Closing via #5189.

PR #5189, opened 2017-03-15
2.x: fix concatMapEager should accept 0 for prefetch

nmorioka:
Fixes a bug that  Flowable.concatMapEager(mapper , maxConcurrency , prefetch) and Observable.concatMapEager(mapper , maxConcurrency , prefetch) operators will not accept 0 and negative numbers.

Reported in #5185.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -6770,7 +6770,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      * @param mapper the function that maps a sequence of values into a sequence of Publishers that will be
      *               eagerly concatenated
      * @param maxConcurrency the maximum number of concurrent subscribed Publishers
-     * @param prefetch hints about the number of expected source sequence values
+     * @param prefetch hints about the number of expected values from each inner Publisher, must be positive
      * @return the new Publisher instance with the specified concatenation behavior
      * @since 2.0
      */
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -5965,7 +5965,7 @@ public abstract class Observable<T> implements ObservableSource<T> {
      * @param mapper the function that maps a sequence of values into a sequence of ObservableSources that will be
      *               eagerly concatenated
      * @param maxConcurrency the maximum number of concurrent subscribed ObservableSources
-     * @param prefetch hints about the number of expected source sequence values
+     * @param prefetch hints about the number of expected values from each inner ObservableSource, must be positive
      * @return the new ObservableSource instance with the specified concatenation behavior
      * @since 2.0
      */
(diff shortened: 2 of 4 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -6780,7 +6780,6 @@ public final void blockingSubscribe(Subscriber<? super T> subscriber) {
     public final <R> Flowable<R> concatMapEager(Function<? super T, ? extends Publisher<? extends R>> mapper,
             int maxConcurrency, int prefetch) {
         ObjectHelper.verifyPositive(maxConcurrency, "maxConcurrency");
-        ObjectHelper.verifyPositive(prefetch, "prefetch");
```
akarnokd: No, the validations were right, you shouldn't call with non-positive value.

src/main/java/io/reactivex/Observable.java
```diff
@@ -5975,7 +5975,6 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
             int maxConcurrency, int prefetch) {
         ObjectHelper.requireNonNull(mapper, "mapper is null");
         ObjectHelper.verifyPositive(maxConcurrency, "maxConcurrency");
-        ObjectHelper.verifyPositive(prefetch, "prefetch");
```
akarnokd: No, the validations were right, you shouldn't call with non-positive value.

akarnokd:
I believe a 0 breaks the internals of the operator and -1 has the completely opposite effect, taking more memory.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -6770,7 +6770,7 @@ public final void blockingSubscribe(Subscriber<? super T> subscriber) {
      * @param mapper the function that maps a sequence of values into a sequence of Publishers that will be
      *               eagerly concatenated
      * @param maxConcurrency the maximum number of concurrent subscribed Publishers
-     * @param prefetch hints about the number of expected source sequence values
+     * @param prefetch hints about the number of expected values from each inner Publisher
```
akarnokd: I think this should be reinforced:

```
hints about the number of expected values from each inner Publisher, must be positive
```

akarnokd (changes requested):
The verifyPositive should be restored.
Optionally, you could add the ", positive" to the javadoc changes.

nmorioka:
I am sorry with the wrong fix.
Is it only necessary to modify the document?

akarnokd:
> Is it only necessary to modify the document?

Yes. Please modify only the javadoc.

nmorioka:
Thank you! 

Also, I think that the test case name is wrong, so I modifed it.

akarnokd (approved):

vanniktech (approved):

akarnokd:
Thanks for the contribution!

PR #5191, opened 2017-03-16
Correct java doc for refCount() return type. v2.0.7

sadegh:
Java doc for `ConnectableObservable.refCount()` says  it returns a `Flowable`, but it should be `Observable`


src/main/java/io/reactivex/observables/ConnectableObservable.java
```diff
@@ -68,7 +68,7 @@ public final Disposable connect() {
      * Returns an {@code Observable} that stays connected to this {@code ConnectableObservable} as long as there
      * is at least one subscription to this {@code ConnectableObservable}.
      *
-     * @return a {@link Flowable}
+     * @return a {@link Observable}
```
akarnokd: `an`

akarnokd (commented):
Could you change it to `an {@link Observable}`?

sadegh:
fixed!

akarnokd (approved):

akarnokd:
Thanks!

Issue #5165, opened 2017-03-08
1.x: UnicastSubject does not replay onNext calls made prior to subscription if onError is also called prior to subscription.


dano:
I'm using RxJava 1.2.5. Consider this code:
```java
    UnicastSubject<String> s = UnicastSubject.create();
    s.onNext("1");
    s.onNext("2");
    s.onError(new Exception("uh oh"));
    s.subscribe(System.out::println, System.out::println);
```
This outputs:
java.lang.Exception: uh oh

However, the same code using RxJava 2.0.7 version of `UnicastSubject` outputs:
1
2
java.lang.Exception: uh oh

Which is the behavior I expected from 1.x.

akarnokd:
Hi. In RxJava 1.x the UnicastSubject is fail-fast and in 2.x it is delay-error. Currently, there is no option for the other mode in either major version.

dano:
@akarnokd Ah, I see. Thanks for the clarification. Is that documented somewhere that I'm missing? The UnicastSubject docs for 1.x gave me the impression that all events will get replayed once it's subscribed to, regardless of whether or not there's an error event. I also didn't find it mentioned in the "What's Different in 2.0" wiki.

Also, I noticed you added the "PR welcome" label - Are you looking for PRs that add a way to optionally (and not by default) switch to the currently unsupported mode for each major version? Meaning, add a way to delay-error on 1.x and/or a way to fail-fast on 2.x.

akarnokd:
> I also didn't find it mentioned in the "What's Different in 2.0" wiki.

This change was likely simply overlooked.

> added the "PR welcome" label

Yes. Keep the current defaults in each version and add two factory methods:

```java
public static <T> UnicastSubject<T> create(boolean delayError) {
    // ...
}
public static <T> UnicastSubject<T> create(int bufferSize, 
        Runnable onTerminate, boolean delayError) {
    // ...
}
```

dano:
@akarnokd Got it, thanks.

akarnokd:
Closing via #5195, #5217 and #5226.

PR #5195, opened 2017-03-18
1.X: UnicastSubject fail-fast and delay-error behavior

mostroverkhov:
This PR adds support for delay-error behavior to `UnicastSubject` with methods `UnicastSubject<T> create(boolean delayError)`, `UnicastSubject<T> create(int capacityHint, Action0 onTerminated, boolean delayError)`. Behavior of existing factory methods was not changed, and is fail-fast.

Relates to #5165

### DIFF
--- a/src/main/java/rx/subjects/UnicastSubject.java
+++ b/src/main/java/rx/subjects/UnicastSubject.java
@@ -303,6 +303,7 @@ public final class UnicastSubject<T> extends Subject<T, T> {
                 emitting = true;
             }
             Queue<Object> q = queue;
+            boolean delayError = this.delayError;
             for (;;) {
                 Subscriber<? super T> s = subscriber.get();
                 boolean unlimited = false;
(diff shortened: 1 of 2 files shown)


src/main/java/rx/subjects/UnicastSubject.java
```diff
@@ -272,8 +309,7 @@ void replay() {
                 if (s != null) {
                     boolean d = done;
                     boolean empty = q.isEmpty();
-
-                    if (checkTerminated(d, empty, s)) {
+                    if (checkTerminated(d, empty, delayError, s)) {
```
akarnokd, line 313: You should have a local delayError field for this, otherwise you are re-reading the flag on each emission which can cause unnecessary false sharing.

```java
Queue<Object> q = queue;
boolean delayError = this.delayError;
// ...
```

mostroverkhov: done

akarnokd (approved):

akarnokd:
Thanks!

PR #5197, opened 2017-03-19
2.x: add ParallelTransformer interface, params-validation

akarnokd:
This PR adds the `ParallelTransformer` interface to match the other `XTransformer` interfaces, adds `@NonNull` annotations to the `ParallelFlowable` operators and adds the `ParallelFlowable` class to the parameter validator test set.

### DIFF
--- a/src/main/java/io/reactivex/parallel/ParallelFlowable.java
+++ b/src/main/java/io/reactivex/parallel/ParallelFlowable.java
@@ -49,7 +49,7 @@ public abstract class ParallelFlowable<T> {
      * of items must be equal to the parallelism level of this ParallelFlowable
      * @see #parallelism()
      */
-    public abstract void subscribe(Subscriber<? super T>[] subscribers);
+    public abstract void subscribe(@NonNull Subscriber<? super T>[] subscribers);
 
     /**
      * Returns the number of expected parallel Subscribers.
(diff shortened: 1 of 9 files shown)


src/main/java/io/reactivex/parallel/ParallelFlowable.java
```diff
@@ -629,8 +629,8 @@ protected final boolean validate(Subscriber<?>[] subscribers) {
      * @return the ParallelFlowable returned by the function
      */
     @CheckReturnValue
-    public final <U> ParallelFlowable<U> compose(Function<? super ParallelFlowable<T>, ParallelFlowable<U>> composer) {
-        return RxJavaPlugins.onAssembly(to(composer));
+    public final <U> ParallelFlowable<U> compose(@NonNull ParallelTransformer<T, U> composer) {
+        return RxJavaPlugins.onAssembly(composer.apply(this));
```
vanniktech: Missing check here for composer being null?

src/main/java/io/reactivex/parallel/ParallelFlowable.java
```diff
@@ -611,7 +611,7 @@ protected final boolean validate(Subscriber<?>[] subscribers) {
      * @return the value returned by the converter function
      */
     @CheckReturnValue
-    public final <U> U to(Function<? super ParallelFlowable<T>, U> converter) {
+    public final <U> U to(@NonNull Function<? super ParallelFlowable<T>, U> converter) {
         try {
             return converter.apply(this);
```
vanniktech: Missing check here for converter being null

vanniktech (approved):

src/main/java/io/reactivex/parallel/ParallelFlowable.java
akarnokd: Dereference will throw NPE automatically.

vanniktech: I'd still check it in order to be consistent and for the better error message knowing what exactly is null.

akarnokd: Updated this and all other methods.

vanniktech (approved):
Awesome 👍

Issue #5128, opened 2017-02-22
2.x: resilience in the parallel() operators


akarnokd:
When the architecture of the `parallel()` subsystem was designed, the goal was to provide means for efficient parallel computations over recovering from failures in any or most of the parallel 'rails' in it. This resulted fail-fast implementation and little to no means (beyond `flatMap`) to prevent the death of the whole parallel computation due to a single failing component.

Adding the usual `onErrorResumeNext` or `retry` doesn't work because the core driver, `parallel()` can't support resubscription on its individual rails and even if, a failure behind a `runOn` almost certainly would drop any unprocessed elements in its queue.

The given toolset, `ParallelFlowable` already gives the opportunity to include resilience at the cost of overhead: `flatMap` and/or `concatMap`. As with the traditional flows, the inner `Flowable`s to these can have all sorts of operators not available on `parallel()` itself and are confined to the particular 'rail', thus `retry()`/`onErrorResumeNext` can be applied on a per item basis:

```java
Flowable.range(1, 100)
.parallel(20)
.runOn(Schedulers.io())
.concatMap(index -> api.getImage(index).onErrorResumeNext(Flowable.just(notAvailableImage)))
.sequential()
.observeOn(guiThread)
.subscribe(...)
```

On the other hand, parallel-`map`/`doOnNext` (among others) don't offer any ignore/resume option out of box and unless the developer applies `try-catch` to prevent crashing the 'rail'.

To enable finer gained control in this situation, I propose the following overloads:

- `enum FailureHandling`: stop, error, skip, retry
- `map(Function<T, R>, FailureHandling)`
- `map(Function<T, R>, Function<Throwable, FailureHandling>)`
- `filter(Predicate<T>, FailureHandling)`
- `filter(Predicate<T>, Function<Throwable, FailureHandling>)`
- `doOnNext(Action<T>, FailureHandling)`
- `doOnNext(Action<T>, Function<Throwable, FailureHandling>)`

They should cover the the cases where the `Function` hosts a blocking call (if there was a need for timed backoff, the `concatMap` approach is still available).

Thoughts?

akarnokd:
Closing via #5202.

PR #5202, opened 2017-03-20
2.x: add resilient versions of parallel map(), filter() & doOnNext()

akarnokd:
This PR adds 2 new overloads to `ParallelFlowable` operators `map`, `filter` and `doOnNext` to enable per item error handling in case the main function fails with some exception.

```java
Flowable.range(0, 2)
.parallel(1)
.map(v -> 1 / v, ParallelFailureHandling.SKIP)
.sequential()
.test()
.assertResult(1);
```

The new `ParallelFailureHandling` has some default enumeration values to handle the common cases. In addition, the `BiFunction` overload allows bounded retries and/or conditional handling of failures.

Related: #5128.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/parallel/ParallelDoOnNextTry.java
+++ b/src/main/java/io/reactivex/internal/operators/parallel/ParallelDoOnNextTry.java
@@ -1,0 +1,294 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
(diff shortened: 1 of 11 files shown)


AarjavP:
Sorry if this is not the correct place for this question, however it's relevant to the above feature.
I would like to have an error handler/on error callback but continue the processing, or specify the number of retries.
Can the `ParallelFailureHandling` enum be made into an interface instead? I see the enum instances are all used in the same way in all the `tryOnNext` methods of the 3 operators (filter, map, and doOnNext). The existing enums can maybe 'built-in' handler instances. Something like:

built-ins:
```
.map(v -> 1 / v, ParallelFailureHandling.SKIP) //static instance
```

some customizable built-ins:
```
.map(v -> 1 / v, ParallelFailureHandling.retry(5)) //retry 5 times before failing
```

user defined:
```
.map(v -> 1 / v, (retries, err) -> { log.error(err); return ParrallelFailureHandling.Result.SKIP; })
```
The return values could be `SKIP`/`IGNORE`, `RETRY`, `STOP`, `default(val)`, or `error(ex)`

Or perhaps this is already do able using existing api?

akarnokd:
Please always check the available operators first before asking for a feature: http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/parallel/ParallelFlowable.html#map-io.reactivex.functions.Function-io.reactivex.functions.BiFunction-

The enums are just convenience implementations of the BiFunction variants of the operators.

Issue #5203, opened 2017-03-20
Is delaySubscription() interrupts the thread immediately as expected?


hkurokawa:
I ran the below test code with RxJava 2.0.7 and got a failure which succeeded with RxJava 1.x.

```java
    final TestObserver<Boolean> observer = TestObserver.create();
    Observable.<Boolean>create(emitter -> {
      emitter.onNext(Thread.interrupted());
      emitter.onComplete();
    })
        .delaySubscription(100, TimeUnit.MICROSECONDS)
        .subscribe(observer);
    observer.awaitTerminalEvent();
    observer.assertValue(false);
  }
```

If I put `subscribeOn()` before `delaySubscription()`, it does not fail.

I am not sure if it is as expected so want to know if it is the expected behaviour.

Here is some background of my issue: I am using RxJava 2.0.7 with Retrofit 2.2.0 and it seems Retrofit (precisely speaking, Okio, which is being used in OkHttp) throws an error when the thread is interrupted while it tries to read the HTTP response. I am trying to delay the HTTP call to achieve something like debounce() with delaySubscription() operator and encountered this issue. I have a workaround so just want to know if it is the expected behaviour.

Thank you for your support in advance.

akarnokd:
This is an unexpected side-effect of cancelling the delay timer inside the `delaySubscription` operator. I'll post a fix for it shortly.

akarnokd:
See #5207.

hkurokawa:
Thank you for your prompt work on this. Waiting for the next release including this fix.

akarnokd:
Closing via #5207.

PR #5205, opened 2017-03-20
2.x: fix timer() to prevent a self-interrupt when signals its only value

akarnokd:
When the `timer()` is firing its `onNext(0L)`, there is no point in allowing a canellation to bubble up to the current `Future` as it just interrupts the current thread. The fix thus makes sure the cancellation chain is cut off before `onNext` and the current parent `Future` should complete normally.

Related: #5203.


vanniktech (approved):

akarnokd:
Unfortunately, this may break a legitimate use case for interrupting a sleep that happens on the `onNext` call of the timer itself:

```java
to = Observable.timer(1, TimeUnit.SECONDS)
.map(v -> Thread.sleep(5000))
.test();

Thread.sleep(2000);

to.cancel();
```

PR #5207, opened 2017-03-20
2.x: prevent tasks to self interrupt on the standard schedulers

akarnokd:
Task wrappers of the various schedulers and modes (direct & worker) were able to get cancelled via interruption from the same thread they were running.

Related #5203.


vanniktech (approved):
👍 didn't quite understand everything but the tests seem good to me

Issue #5130, opened 2017-02-22
2.x: incorrect images


akarnokd:
This is a container for identifying and tracking fixes to outdated or incorrent marble diagrams.

  - [x] [Single.flattenAsFlowable](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Single.html#flattenAsFlowable(io.reactivex.functions.Function)): refers to flatMapIterable, no reordering of colors
  - [x] [Single.flattenAsObservable](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Single.html#flattenAsObservable(io.reactivex.functions.Function)): refers to flatMapIterable, no reordering of colors
  - [x] [Flowable.flatMapIterable](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Flowable.html#flatMapIterable(io.reactivex.functions.Function)) + 4 no reordering of colors
  - [x] [Observable.flatMapIterable](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#flatMapIterable(io.reactivex.functions.Function)) + 2 no reordering of colors
  - [x] [Flowable.firstElement()](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Flowable.html#firstElement()), indicate cancellation after the first element, shift marbles so the 2nd element doesn't imply the completion like in `singleOrError`.
  - [x] [Observable.firstElement()](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#firstElement()), indicate cancellation after the first element, shift marbles so the 2nd element doesn't imply the completion like in `singleOrError`.

akarnokd:
![Image](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/firstElement.m.png)

![Image](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/first.s.png)

akarnokd:
![Image](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/flattenAsFlowable.png)

![Image](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/flattenAsObservable.png)

akarnokd:
![Image](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/flatMapIterable.f.png)

[...]

PR #5208, opened 2017-03-21
2.x: fix images of firstElement, flattenAsX, flatMapIterable

akarnokd:
This PR links in the new images from #5130 for previously misleading marble diagrams of the operators.


akarnokd:
### Plus UnicastX:

Image
Image

vanniktech (approved):
Those height & width resizes are due to the new images, right?

akarnokd:
Yes, the default with is set to 640 and these new images are slightly bigger in their native form.

JakeWharton (commented):
:+1:

akarnokd:
Thanks for the reviews guys!

PR #5212, opened 2017-03-22
1.x: add link rel=canonical to 1.x javadoc files (semi auto)

akarnokd:
This PR adds `<link rel="canonical" href="..."/>` nodes to the 1.x javadoc via the following program:

https://github.com/akarnokd/akarnokd-misc/blob/master/src/main/java/hu/akarnokd/rxjava/javadoc/AddCanonical.java

How to use it:
- Within a Java 8 project, replace the file directories that point to your checked-out `gh-pages` branch.
- Unpack the rxjava-1.2.x-javadoc.jar into the two places
- Run `AddCanonical`
- Push a PR with the changes.


vanniktech (approved):
Really nice :+1:

PR #5222, opened 2017-03-24
2.x: cleanup based on IntelliJ 2017.1 inspections

akarnokd:
This PR is the result of running IntelliJ inspections on the code:

  - replace `unsubscribe` with `cancel` and `dispose` in the main classes to better match the 2.x terminology
  - write out a couple of local variables
  - remove unnecessary `return` and `continue` statements
  - remove unnecessary variable writes
  - remove unnecessary and unused fields
  - add `final` to some fields
  - fix typos

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
+++ b/src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
@@ -110,13 +110,15 @@ public final class ParallelJoin<T> extends Flowable<T> {
         }
 
         void cancelAll() {
-            for (JoinInnerSubscriber<T> s : subscribers) {
+            for (int i = 0; i < subscribers.length; i++) {
+                JoinInnerSubscriber<T> s = subscribers[i];
                 s.cancel();
             }
         }
 
         void cleanup() {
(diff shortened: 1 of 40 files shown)


src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
```diff
@@ -238,9 +238,7 @@ void drainLoop() {
 
                     boolean empty = true;
 
-                    for (int i = 0; i < n; i++) {
-                        JoinInnerSubscriber<T> inner = s[i];
-
+                    for (JoinInnerSubscriber<T> inner : s) {
```
vanniktech: You might want to leave it as it was before since that way you won't allocate an iterator

vanniktech (approved):

src/main/java/io/reactivex/internal/operators/parallel/ParallelJoin.java
akarnokd: Hmm, yeah, I forget that who knows what optimization the Android VM does (the JVM's escape analysis optimizes this nicely). I'll restore the original indexed version.

PR #5217, opened 2017-03-23
[2.x] UnicastSubject fail fast support

mostroverkhov:
This PR adds support for fail-fast behavior to `UnicastSubject` with methods `UnicastSubject<T> create(boolean delayError)`, `UnicastSubject<T> create(int capacityHint, Runnable onTerminated, boolean delayError)`.  Relates to #5165

PR #5226, opened 2017-03-24
2.x UnicastProcessor fail-fast support

mostroverkhov:
This PR adds support for fail-fast behavior to `UnicastProcessor` with methods `UnicastProcessor<T> create(boolean delayError)`, `UnicastProcessor<T> create(int capacityHint, Runnable onTerminated, boolean delayError)`. Relates to #5165, #5217

### DIFF
--- a/src/main/java/io/reactivex/processors/UnicastProcessor.java
+++ b/src/main/java/io/reactivex/processors/UnicastProcessor.java
@@ -16,6 +16,7 @@ package io.reactivex.processors;
 import io.reactivex.annotations.CheckReturnValue;
 import java.util.concurrent.atomic.*;
 
+import io.reactivex.annotations.Experimental;
 import io.reactivex.annotations.Nullable;
 import org.reactivestreams.*;
 
@@ -49,7 +50,10 @@ public final class UnicastProcessor<T> extends FlowableProcessor<T> {
 
     final AtomicReference<Runnable> onTerminate;
 
+    final boolean delayError;
+
     volatile boolean done;
+
     Throwable error;
 
     final AtomicReference<Subscriber<? super T>> actual;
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/processors/UnicastProcessor.java
```diff
@@ -103,13 +120,44 @@
...
+     * @param capacityHint the hint to size the internal unbounded buffer
+     * @param onCancelled the non null callback
+     * @param delayError deliver pending onNext events before onError
+     * @return an UnicastProcessor instance
+     * @since 2.0.8 - experimental
+     */
+    @CheckReturnValue
+    @Experimental
+    public static <T> UnicastProcessor<T> create(int capacityHint, Runnable onCancelled, boolean delayError) {
+        return new UnicastProcessor<T>(capacityHint, onCancelled,delayError);
```
vanniktech: nit: newline between `onCancelled,delayError`

src/main/java/io/reactivex/processors/UnicastProcessor.java
```diff
@@ -103,13 +120,44 @@
...
+    }
+
+    /**
      * Creates an UnicastProcessor with the given capacity hint.
      * @param capacityHint the capacity hint for the internal, unbounded queue
      * @since 2.0
      */
     UnicastProcessor(int capacityHint) {
```
vanniktech, line 149: Is this one really necessary?

src/main/java/io/reactivex/processors/UnicastProcessor.java
akarnokd: Space.

src/main/java/io/reactivex/processors/UnicastProcessor.java
```diff
@@ -103,13 +120,44 @@
...
+        this(capacityHint, true);
+    }
+
+    /**
+     * Creates an UnicastProcessor with the given capacity hint and delay error flag.
+     * @param capacityHint the capacity hint for the internal, unbounded queue
+     * @param delayError deliver pending onNext events before onError
+     * @since 2.0.8 - experimental
+     */
+    UnicastProcessor(int capacityHint, boolean delayError) {
```
akarnokd: This is unnecessary.

akarnokd (changes requested):
One of the new package-private constructors is unnecessary.

src/main/java/io/reactivex/processors/UnicastProcessor.java
mostroverkhov: could you elaborate a bit?

src/main/java/io/reactivex/processors/UnicastProcessor.java
mostroverkhov: 3 args constructor has non-null onTerminate check; should i move this check to factory?

akarnokd: Could you move that check into the appropriate `create` method(s)?

akarnokd (approved):

PR #5230, opened 2017-03-26
2.x: Fix wrong comments in Functions.java “Function3” -> “BiFunction”

ggikko:
I'm using this library in my company(KakaoBank in Korea) and interested in this library.. so I read whole code these days. I feel really thanks for all people who make awesome library and I just wanted to contribute this library.

### DIFF
--- a/src/main/java/io/reactivex/internal/functions/Functions.java
+++ b/src/main/java/io/reactivex/internal/functions/Functions.java
@@ -24,7 +24,7 @@ import io.reactivex.plugins.RxJavaPlugins;
 import io.reactivex.schedulers.Timed;
 
 /**
- * Utility methods to convert the Function3..Function9 instances to Function of Object array.
+ * Utility methods to convert the BiFunction..Function9 instances to Function of Object array.
  */
 public final class Functions {
 
--- a/src/main/java/io/reactivex/internal/functions/Functions.java
+++ b/src/main/java/io/reactivex/internal/functions/Functions.java
@@ -24,7 +24,7 @@ import io.reactivex.plugins.RxJavaPlugins;
 import io.reactivex.schedulers.Timed;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/internal/functions/Functions.java
```diff
@@ -24,7 +24,7 @@
 import io.reactivex.schedulers.Timed;
 
 /**
- * Utility methods to convert the Function3..Function9 instances to Function of Object array.
+ * Utility methods to convert the BiFunction..Function9 instances to Function of Object array.
```
akarnokd: Actually, `BiFunction, Function3..Function9` would be better.

ggikko:
I also added fail test case toFunction about Functions.java

src/test/java/io/reactivex/internal/functions/FunctionsTest.java
```diff
@@ -169,6 +169,54 @@ public Integer apply(Integer t1, Integer t2, Integer t3, Integer t4, Integer t5,
         }).apply(new Object[20]);
     }
 
+    @Test(expected = NullPointerException.class)
+    public void biFunctionFail() throws Exception {
```
akarnokd: Please add `@SuppressWarnings({"unchecked", "rawtypes"})` to all these new methods.

akarnokd (changes requested):

src/main/java/io/reactivex/internal/functions/Functions.java
ggikko: I agree.

src/test/java/io/reactivex/internal/functions/FunctionsTest.java
ggikko: Thank you for comment.

akarnokd (approved):

Issue #5237, opened 2017-03-27
1.x async Singles crashes for Exceptions thrown in onSuccess


passsy:
Since [1.2.2](https://github.com/ReactiveX/RxJava/releases/tag/v1.2.2) tested with `1.2.9`

`Single` doesn't wrap the `SingleSubscriber` automatically in a `SafeSubscriber`. This leads to problems when an `Exceptions` is thrown on `onSuccess`. It will not be propagated to `onError` for async Singles.

From the [changelog 1.2.2](https://github.com/ReactiveX/RxJava/releases/tag/v1.2.2) I can see this has performance reasons. But the problem is that async Singles now behave differently than sync Singles, or Observables sync or async.

```java
// sync (as expected)
Single.just("test").subscribe(
        result -> {
            System.out.println("got text: " + result);
            throw new IllegalStateException("single sync something is wrong");
            // calls on Error
        },
        e -> {
            System.out.println("caught error " + e.getMessage());
            e.printStackTrace();
        });

// async (behaves differently)
Single.just("test")
        .delay(1, TimeUnit.MILLISECONDS)
        .subscribe(
        result -> {
            System.out.println("got text: " + result);
            throw new IllegalStateException("single async something is wrong");
            // onError will not be called, hard crash
        },
        e -> {
            // not called
            System.out.println("caught error " + e.getMessage());
            e.printStackTrace();
        });

// sync (as expected)
Observable.just("test").subscribe(
        result -> {
            System.out.println("got text: " + result);
            throw new IllegalStateException("observable sync something is wrong");
            // calls on Error
        },
        e -> {
            System.out.println("caught error " + e.getMessage());
            e.printStackTrace();
        });


// async (as expected)
Observable.just("test")
        .delay(1, TimeUnit.MILLISECONDS)
        .subscribe(
                result -> {
                    System.out.println("got text: " + result);
                    throw new IllegalStateException("observable async something is wrong");
                    // calls on Error
                },
                e -> {
                    System.out.println("caught error " + e.getMessage());
                    e.printStackTrace();
                });
```

---

After thinking about this case 1 is the one which is buggy. The synchronous `Single` should crash like the async `Single` case. Analog to the implementation in rx2

akarnokd:
Looks like the lambda-subscriber on Single.java:1777 needs some catch clauses. Would you like to submit a PR?

passsy:
[...]

PR #5241, opened 2017-03-27
1.x fix async single error report (bug #5237)

passsy:
Targets issue #5237 where an `Exception` in `onSuccess` of an async `Single` wasn't forwarded to `onError`. For a sync `Single` it already gets forwarded correctly.

Example:
```java
Single.just("test")
        .delay(1, TimeUnit.MILLISECONDS)
        .subscribe(
        result -> {
            System.out.println("got text: " + result);
            throw new IllegalStateException("single async something is wrong");
            // onError will not be called, hard crash
        },
        e -> {
            // not called
            System.out.println("caught error " + e.getMessage());
            e.printStackTrace();
        });
```

I introduced `SafeSingleSubscriber` analog to `SafeSubscriber` of `Observable`. It makes sure `onError` gets called when `onSuccess` throws.

I had to introduce `private Subscription unsafeSubscribe(SingleSubscriber)` for internal usage of `Single#subscribeOn()` and `Single#unsubscribeOn()` which should *not* be wrapped by `SafeSingleSubscriber`.

Furthermore I added the `subscriber.isUnsubscribed()` case in all three `subscribe` methods analog to `Observable#subscribe`
```java
    // ...
    if (subscriber.isUnsubscribed()) {
        RxJavaHooks.onError(RxJavaHooks.onSingleError(e));
    } else {
        // try forwarding error to subscriber
    }
    // ....
```
as well as `Single#subscribe` now throws `OnErrorFailedException` instead of `RuntimeException` when calling `subscriber.onError(e)` fails.

`Single.using` is now working correctly, I uncommented the missing assertions

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1819,18 +1819,23 @@ public class Single<T> {
         } catch (Throwable e) {
             // special handling for certain Throwable/Error/Exception types
             Exceptions.throwIfFatal(e);
-            // if an unhandled error occurs executing the onSubscribe we will propagate it
-            try {
-                subscriber.onError(RxJavaHooks.onSingleError(e));
-            } catch (Throwable e2) {
-                Exceptions.throwIfFatal(e2);
-                // if this happens it means the onError itself failed (perhaps an invalid function implementation)
-                // so we are unable to propagate the error correctly and will just throw
-                RuntimeException r = new RuntimeException("Error occurred attempting to subscribe [" + e.getMessage() + "] and then again while trying to pass to onError.", e2);
-                // TODO could the hook be the cause of the error in the on error handling.
-                RxJavaHooks.onSingleError(r);
-                // TODO why aren't we throwing the hook's return value.
-                throw r; // NOPMD
+            // in case the subscriber can't listen to exceptions anymore
+            if (subscriber.isUnsubscribed()) {
+                RxJavaHooks.onError(RxJavaHooks.onSingleError(e));
+            } else {
+                // if an unhandled error occurs executing the onSubscribe we will propagate it
+                try {
+                    subscriber.onError(RxJavaHooks.onSingleError(e));
+                } catch (Throwable e2) {
+                    Exceptions.throwIfFatal(e2);
+                    // if this happens it means the onError itself failed (perhaps an invalid function implementation)
+                    // so we are unable to propagate the error correctly and will just throw
+                    RuntimeException r = new OnErrorFailedException("Error occurred attempting to subscribe [" + e.getMessage() + "] and then again while trying to pass to onError.", e2);
+                    // TODO could the hook be the cause of the error in the on error handling.
+                    RxJavaHooks.onSingleError(r);
+                    // TODO why aren't we throwing the hook's return value.
+                    throw r; // NOPMD
+                }
             }
             return Subscriptions.unsubscribed();
         }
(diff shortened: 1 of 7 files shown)


src/main/java/rx/Single.java
```diff
@@ -1959,29 +1953,78 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
      *             if the {@link SingleSubscriber}'s {@code onError} method itself threw a {@code Throwable}
      * @see <a href="http://reactivex.io/documentation/operators/subscribe.html">ReactiveX operators documentation: Subscribe</a>
      */
-    public final Subscription subscribe(final SingleSubscriber<? super T> te) {
-        if (te == null) {
-            throw new IllegalArgumentException("te is null");
+    public final Subscription subscribe(SingleSubscriber<? super T> subscriber) {
```
akarnokd, line 1956: There were concerns that some languages do require the parameter names to remain the same, so no public renamings.

src/main/java/rx/Single.java
```diff
@@ -1959,29 +1953,78 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
...
+        if (subscriber == null) {
+            throw new IllegalArgumentException("subscriber is null");
         }
+
+        /*
+         * See https://github.com/ReactiveX/RxJava/issues/216 for discussion on "Guideline 6.4: Protect calls
+         * to user code from within an Observer"
+         */
+        // if not already wrapped
+        if (!(subscriber instanceof SafeSingleSubscriber)) {
```
akarnokd, line 1966: This undoes a lot of the perf optimizations. Especially without a matching accessible `unsafeSubscribe`.

src/main/java/rx/observers/SafeSingleSubscriber.java
```diff
@@ -0,0 +1,151 @@
...
+     * The {@code Single} will not call this method if it calls {@link #onError}.
+     */
+    @Override
+    public void onSuccess(T t) {
+        try {
+            actual.onSuccess(t);
+        } catch (Throwable e) {
+            // we handle here instead of another method so we don't add stacks to the frame
+            // which can prevent it from being able to handle StackOverflow
+            Exceptions.throwOrReport(e, this);
```
akarnokd, line 70: The `Single` protocol is `onSuccess | onError`. You can't call `onError` when `onSuccess` has been called.

src/test/java/rx/internal/operators/SingleFromEmitterTest.java
```diff
@@ -173,6 +176,8 @@ public void call(SingleEmitter<Object> e) {
 
     @Test
     public void onSuccessThrows() {
+        final AtomicInteger onErrorCallsCounter = new AtomicInteger();
```
akarnokd, line 179: Changing existing unit test is a sign of behavior change that may be incompatible with existing uses.

akarnokd (changes requested):
I'm sorry, the changes to Single are unacceptable. You could have added a `safeSubscribe()` method without breaking anyone.

passsy:
What can I say, tests are working on my maschine 😅 Is it normal that rx1 tests are flaky?

```
pascalwelsch$ ./gradlew test
...
BUILD SUCCESSFUL

Total time: 3 mins 54.445 secs
```

passsy:
I tried to revert to the behaviour in `1.1.9`. I have some problems upgrading because of this behaviour change of not calling `onError` introduced in `1.2.2`. Especially frustrating because this is the behaviour for synchronous Singles but changes when adding a `delay`.

What about those `Single.using` tests? They are now working correctly as far as I understand. Is the change of those assertions intended?

I'll try to change the PR without changing the current behaviour.

src/test/java/rx/internal/operators/SingleFromEmitterTest.java
passsy: I intentional changed the behaviour here

src/main/java/rx/observers/SafeSingleSubscriber.java
passsy: Aye

akarnokd:
> I tried to revert to the behaviour in 1.1.9. I have some problems upgrading because of this behaviour change of not calling onError introduced in 1.2.2.

Where do you have this problem? You could have just delegated to your error handler:

```java
Action1<Throwabe> onErrorHandler = ...

Action1<T> onSuccessHandler = value -> {
    try {
          doSomething(value);
    } catch (Throwable ex) {
          onErrorHandler.call(ex);
    }
};

source.subscribe(onSuccessHandler, onErrorHandler);
```

> What about those Single.using tests? They are now working correctly as far as I understand. Is the change of those assertions intended?

The current (unsafe)subscribe doesn't call unsubscribe. The rest of the test is still valuable.

akarnokd:
> I'll try to change the PR without changing the current behaviour.

It's much cleaner if you start over.

passsy:
Was building in the complete wrong direction
~close

PR #5244, opened 2017-03-30
Remove CheckReturnValue on methods returning Disposable

markelliot:
Per #4878.

### DIFF
--- a/src/test/java/io/reactivex/BaseTypeAnnotations.java
+++ b/src/test/java/io/reactivex/BaseTypeAnnotations.java
@@ -21,6 +21,7 @@ import org.junit.Test;
 import org.reactivestreams.Publisher;
 
 import io.reactivex.annotations.*;
+import io.reactivex.disposables.Disposable;
 
 /**
  * Verifies several properties.
--- a/src/test/java/io/reactivex/BaseTypeAnnotations.java
+++ b/src/test/java/io/reactivex/BaseTypeAnnotations.java
@@ -41,12 +41,11 @@ public class BaseTypeAnnotations {
                 continue;
             }
             if (m.getDeclaringClass() == clazz) {
-                boolean isSubscribeMethod = "subscribe".equals(m.getName()) && m.getParameterTypes().length == 0;
                 boolean isAnnotationPresent = m.isAnnotationPresent(CheckReturnValue.class);
 
-                if (isSubscribeMethod) {
+                if (m.getReturnType().equals(Disposable.class)) {
                     if (isAnnotationPresent) {
-                        b.append("subscribe() method has @CheckReturnValue: ").append(m).append("\r\n");
+                        b.append("method with Disposable return type has @CheckReturnValue: ").append(m).append("\r\n");
                     }
                     continue;
                 }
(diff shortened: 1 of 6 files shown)


src/test/java/io/reactivex/BaseTypeAnnotations.java
```diff
@@ -41,12 +41,11 @@ static void checkCheckReturnValueSupport(Class<?> clazz) {
                 continue;
             }
             if (m.getDeclaringClass() == clazz) {
-                boolean isSubscribeMethod = "subscribe".equals(m.getName()) && m.getParameterTypes().length == 0;
                 boolean isAnnotationPresent = m.isAnnotationPresent(CheckReturnValue.class);
 
-                if (isSubscribeMethod) {
+                if (m.getReturnType().equals(Disposable.class)) {
```
akarnokd, line 47: Looks like this needs to import `io.reactivex.disposables.Disposable`. Did you post this from IntelliJ?

akarnokd (commented):
Because it works for me either way, as I don't use tools that require `CheckReturnValue`, I defer to the judgement of @JakeWharton and @vanniktech on this.

vanniktech (changes requested):
I'm against this change. In the vast majority of the cases that this was triggered, I actually forgot to handle the Disposable and this annotation saved me.

markelliot:
Cool, for consistency, then, should the other Disposable returning subscribe method also pick up this annotation?

vanniktech:
David was against this one. Also like Jake said it's an antipattern and shouldn't be used in applications.

akarnokd:
To sum up, the arg-less `subscribe()` is for testing only and I don't think it should be in production flows. Therefore, a `CheckReturnValue` may be distracting on them. For all the other overloads, the dominant use is with `CompositeDisposable.add()` for flows started by the activity and have to be stopped hence the return value should be taken care of.

BenTilbrook:
Since when did the dominant use case dictate that all other use cases have to suppress a warning? I thought RxJava was a general-purpose lib.

Assigning the `Disposable` to a var doesn't even mean you're disposing it correctly, or at all. I don't think that can be determined by static analysis.

`@CheckReturnValue`'s use here is an excessively opinionated decision that damages user code, by forcing warning suppression (and adding unwanted noise) when the code otherwise handles the `Disposable` effectively using another technique such as RxLifecycle or `takeUntil`.

The annotation is better suited to scenarios where failing to handle the result is _guaranteed_ to be bad or pointless, like creating a copy of an immutable collection using its `add(elements)` method, or indeed, the RxJava operator functions.

JakeWharton:
Strongly disagree. The absence of the annotation creates bugs. Its presence
requires suppression when you're doing non-standard management that itself
doesn't even guarantee correctness which sounds perfectly reasonable.

BenTilbrook:
I feel its use here dilutes the value of the annotation, and is in a sense, a stamp of disapproval for RxLifecycle and equivalent solutions.

akarnokd:
You can always implement `Subscriber`/`Observer` and the `subscribe()` for them doesn't return `Disposable`.

BenTilbrook:
I'd be interested to hear your opinions on RxLifecycle & co, since this decision punishes their use.

akarnokd:
Try [AutoDispose](https://github.com/uber/AutoDispose) instead.

BenTilbrook:
Thanks. Looked into this briefly a while back but didn't see a lot of value in migrating to it. Will give it more consideration now.

PR #5250, opened 2017-03-31
Defer creation of the TimeoutException when using the Single.timeout() operator

dhendry:
Use a defer instead of simply `other = Single.error(new TimeoutException())` since instantiating an exception will cause the current stack trace to be inspected and that overhead should only be incurred when a timeout actually happens.

I caught this problem as I was investigating unusually high CPU usage in one of the systems I operate. When measuring the impact of this change in a real world situation that makes only moderate use of this operator, I observed a *11%* reduction in CPU usage.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -2257,11 +2257,9 @@ public class Single<T> {
             // Use a defer instead of simply   other = Single.error(new TimeoutException())
             // since instantiating an exception will cause the current stack trace to be inspected
             // and we only want to incur that overhead when a timeout actually happens.
-            other = Single.<T>defer(new Func0<Single<T>>()
-            {
+            other = Single.<T>defer(new Func0<Single<T>>() {
                 @Override
-                public Single<T> call()
-                {
+                public Single<T> call() {
                     return Single.<T>error(new TimeoutException());
                 }
             });
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -2254,7 +2254,17 @@ public class Single<T> {
      */
     public final Single<T> timeout(long timeout, TimeUnit timeUnit, Single<? extends T> other, Scheduler scheduler) {
         if (other == null) {
-            other = Single.<T> error(new TimeoutException());
+            // Use a defer instead of simply   other = Single.error(new TimeoutException())
+            // since instantiating an exception will cause the current stack trace to be inspected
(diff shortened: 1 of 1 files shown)


dhendry:
Please let me know if there is anything about this PR that needs to change in order to get this fix merged as the performance improvement seems to be reasonably significant.

src/main/java/rx/Single.java
```diff
@@ -2254,7 +2254,17 @@ public final Completable toCompletable() {
      */
     public final Single<T> timeout(long timeout, TimeUnit timeUnit, Single<? extends T> other, Scheduler scheduler) {
         if (other == null) {
-            other = Single.<T> error(new TimeoutException());
+            // Use a defer instead of simply   other = Single.error(new TimeoutException())
+            // since instantiating an exception will cause the current stack trace to be inspected
+            // and we only want to incur that overhead when a timeout actually happens.
+            other = Single.<T>defer(new Func0<Single<T>>()
+            {
```
JakeWharton: This should go on the previous line after a space.

src/main/java/rx/Single.java
```diff
@@ -2254,7 +2254,17 @@ public final Completable toCompletable() {
...
+                @Override
+                public Single<T> call()
+                {
```
JakeWharton: This should go on the previous line after a space.

JakeWharton (approved):

JakeWharton:
Is this also a problem without `timeout()` operators on other stream types? And also what about 2.x?

dhendry:
@JakeWharton I havent done a super exhaustive audit but, it does not look like this is an issue with any of the other timeout operators in the 1.x branch or the single timeout in the 2.x branch

src/main/java/rx/Single.java
```diff
@@ -2254,7 +2254,15 @@ public final Completable toCompletable() {
      */
     public final Single<T> timeout(long timeout, TimeUnit timeUnit, Single<? extends T> other, Scheduler scheduler) {
         if (other == null) {
-            other = Single.<T> error(new TimeoutException());
+            // Use a defer instead of simply   other = Single.error(new TimeoutException())
```
akarnokd, line 2257: I'd inline this into `SingleTimeout`, that way, there are no extra instances unless necessary.

akarnokd (commented):
If you could fix all of 1.x with this problem, that would be great.

PR #5252, opened 2017-04-01
Add NullPointerException comment for doc

ggikko:
nothing special.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -3631,8 +3631,8 @@ public abstract class Observable<T> implements ObservableSource<T> {
      * @param onSubscribe the ObservableSource instance to wrap
      * @return the new Observable instance
      * @throws NullPointerException
-     *             if {@code source} is null, or
-     *             if {@code onSubscribe} is null
+     *             if {@code onSubscribe} is null, or
+     *             if source is null
      */
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -10342,9 +10342,7 @@ public abstract class Observable<T> implements ObservableSource<T> {
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Observable<T> skipLast(int count) {
-        if (count < 0) {
-            throw new IndexOutOfBoundsException("count >= 0 required but it was " + count);
-        }
+        ObjectHelper.verifyIndexPositive(count);
         if (count == 0) {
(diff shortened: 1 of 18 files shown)


src/main/java/io/reactivex/internal/functions/ObjectHelper.java
```diff
@@ -121,6 +121,20 @@ public static long verifyPositive(long value, String paramName) {
...
     }
+    
+    /**
+     * Validate that the given index is greater than equal to zero or report
+     * an IndexOutOfBoundsException with the count.
+     * @param count the value to validate
+     * @return count
+     * @throws IndexOutOfBoundsException if count &lt; 0
+     */
+    public static long verifyIndexPositive(long count) {
```
JakeWharton, line 132: This should be `verifyCountPositive` if the message and argument are named `count`.

JakeWharton (commented):
You'll likely have much better luck with separate pull requests for each change. In general these clean-up changes are made without consulting the project maintainers and thus have to be closed because any positive changes are lost in a sea of ones not agreed with.

src/main/java/io/reactivex/internal/functions/ObjectHelper.java
ggikko: Thanks for ur comment! :) I got it. I'll change it.

akarnokd: Positive means `> 0` but these arguments must be non-negative because they are allowed to have zero.

src/main/java/io/reactivex/Observable.java
```diff
@@ -1420,6 +1426,9 @@ public static int bufferSize() {
      *                       is interpreted as all inner ObservableSources can be active at the same time
      * @param prefetch the number of elements to prefetch from each inner ObservableSource source
      * @return the new ObservableSource instance with the specified concatenation behavior
+     * @throws NullPointerException
+     *             if {@code maxConcurrency} is null, or
+     *             if {@code prefetch} is null
```
akarnokd, line 1431: `prefetch` is a primitive value and can't be null.

src/main/java/io/reactivex/Observable.java
```diff
@@ -1471,6 +1480,7 @@ public static int bufferSize() {
      * @param <T> the element type
      * @param source the emitter that is called when an Observer subscribes to the returned {@code Observable}
      * @return the new Observable instance
+     * @throws NullPointerException if source is null
```
akarnokd, line 1483: I'm not sure if these need to be spelled out, but I guess other places have them already...

src/main/java/io/reactivex/Observable.java
ggikko: Thank you! I take care of it.

PR #5253, opened 2017-04-01
Add some comments & verifyCountPositive method

ggikko:


akarnokd (changes requested):
I'm not convinced about the need for these changes plus they are worded incorrectly.

ggikko:
yes. I agree. I will change it..

Issue #5216, opened 2017-03-23
2.x BiConsumer nullability annotations violated


davidschreiber:
**RxJava 2.0.7**

The `BiConsumer` interface has both parameters of its `accept()` method marked as `@NonNull`. However, when using the `BiConsumer` interface as result callback for a `Single`, the received parameter values are mutually exclusive being a non-null value and the other one being `null`. Although it seems to make sense in use, it violates the interface annotations and makes usage, at least, confusing.

```
Single.just("Some string")
    .subscribe(new BiConsumer<String, Throwable>() {
        @Override
        public void accept(@io.reactivex.annotations.NonNull String s, @io.reactivex.annotations.NonNull Throwable throwable) throws Exception {
            assert s != null;
            assert throwable == null;
        }
    });

Single.<String>error(new RuntimeException())
    .subscribe(new BiConsumer<String, Throwable>() {
        @Override
        public void accept(@io.reactivex.annotations.NonNull String s, @io.reactivex.annotations.NonNull Throwable throwable) throws Exception {
            assert s == null;
            assert throwable != null;
        }
    });
```

akarnokd:
Indeed, those use cases should override the default annotation with `@Nullable`. Would you like to submit a PR?

bloderxd:
@akarnokd I can submit a PR if @davidschreiber can't, but what do you suggest? An exclusive ```BiConsumer``` for ```Single``` ?

akarnokd:
That would be a breaking change. I see 3 options:

- remove annotations from `BiConsumer`,
- change annotations on `BiConsumer` to `@Nullable` - may introduce a lot of warnings, 
- Do nothing and have your IDE suppress/ignore that warning.

bloderxd:
Well I don't think it's a breaking change, it's just an interface extending ```BiConsumer``` example:

```java
  public void biconsumerWithCorrectNullableAnnotationInThrowable() {
        Single.just(1).subscribe(new SingleBiConsumer<Integer, Throwable>() {
            @Override
            public void accept(@Nullable Integer integer, @Nullable Throwable throwable) throws Exception {
                assertNull(throwable);
            }
        });
    }
```
What do you think?

akarnokd:
Yes, you can simply delete the annotations or ignore them in your code.

bloderxd:
Yes I know, but sometimes it's important to let developers know about property nullability, but ok thank you about your suggestions.

PaulWoitaschek:
[...]

PR #5257, opened 2017-04-01
Remove @NonNull annotations in BiConsumer method parameter

bloderxd:
This PR is part of https://github.com/ReactiveX/RxJava/issues/5216

This PR remove @NonNull annotations from ```BiConsumer``` method parameters, there's no sense use these annotations in ```Single``` context because in this case ```BiConsumer``` always brings you a null parameter.

### DIFF
--- a/src/main/java/io/reactivex/functions/BiConsumer.java
+++ b/src/main/java/io/reactivex/functions/BiConsumer.java
@@ -13,8 +13,6 @@
 
 package io.reactivex.functions;
 
-import io.reactivex.annotations.NonNull;
-
 /**
  * A functional interface (callback) that accepts two values (of possibly different types).
  * @param <T1> the first value type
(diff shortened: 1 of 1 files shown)


akarnokd:
/cc @vanniktech 

I belive `BiConsumer` is very often appears with `zip` and `combineLatest` where the `@NonNull` annotation is important.

JakeWharton:
It's not that important. It just means nullability contracts need to be defined where `BiConsumer` is used instead of on the interface itself.

bloderxd:
Well, it's not in all context that @NonNull annotation is correct in ```BiConsumer```, then I guess it's not safe let it there

vanniktech:
Might make more sense to delete the annotations since the interface is used in different places where different parameters are sometimes null and sometimes not.

Having different interfaces with different annotations tailored for each use case might be the best option although it wouldn't be backwards compatible at least not source and binary compatible.

bloderxd:
@vanniktech like this example?

```java
 public void biconsumerWithCorrectNullableAnnotationInThrowable() {
        Single.just(1).subscribe(new SingleBiConsumer<Integer, Throwable>() {
            @Override
            public void accept(@Nullable Integer integer, @Nullable Throwable throwable) throws Exception {
                assertNull(throwable);
            }
        });
    }
```
Create a extesion of ```BiConsumer``` only for ```Single``` Context?

PR #5271, opened 2017-04-07
2.x: enable link to external JDK, fix Schedulers style

akarnokd:
This PR enables the generated Javadoc to `{@link }` to JDK classes properly (current 2.0.8 doc [shows them](http://reactivex.io/RxJava/2.x/javadoc/2.0.8/io/reactivex/schedulers/Schedulers.html#io()) as plain text). I'm linking to the v7 docs because v6 is ugly.

In addition, I've fixed the style of the `Schedulers` documentation by separating the **Supported system properties** properly into a new paragraph.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,7 @@ javadoc {
     options.addStringOption('top').value = ''
     options.addStringOption('doctitle').value = ''
     options.addStringOption('header').value = ''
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
the docs for 8 look even nicer if you want!

akarnokd:
The Travis build runs on Java 7 so the javadoc style is v7 - giving a consistent view when jumping the links.

PR #5281, opened 2017-04-12
2.x: BehaviorProcessor & Subject terminate-subscribe race

akarnokd:
This PR fixes the race condition in `BehaviorProcessor` and `BehaviorSubject` when `onComplete()` or `onError()` is called concurrently with `subscribe` and the consumer throws a `NullPointerException` instead of relaying the terminal event.

The fix involves having a separate `terminalEvent` atomic field, CAS-ing in the actual or marker `Throwable` and reading that field in `subscribe`.

### DIFF
--- a/src/test/java/io/reactivex/processors/BehaviorProcessorTest.java
+++ b/src/test/java/io/reactivex/processors/BehaviorProcessorTest.java
@@ -756,4 +756,60 @@ public class BehaviorProcessorTest extends DelayedFlowableProcessorTest<Object> 
         .awaitDone(5, TimeUnit.SECONDS)
         .assertResult(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
     }
+
+    @Test
+    public void completeSubscribeRace() throws Exception {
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/processors/BehaviorProcessorTest.java
```diff
@@ -756,4 +756,60 @@ public void run() {
...
+            Runnable r2 = new Runnable() {
+                @Override
+                public void run() {
+                    p.onError(ex);
+                }
+            };
+
+            TestHelper.race(r1, r2);
+
+            ts.assertError(TestException.class);
```
artem-zinnatullin: `assertFailure()`? it also checks that no completion events were emitted

artem-zinnatullin (approved):

vanniktech (approved):
UnicastSubject as well as SerializedSubject still have the `boolean` done. Should those be changed for consistency too?

akarnokd:
No need for that.

PR #5282, opened 2017-04-13
BoundedReplayBuffer no longer leaks memory

AttwellBrian:
The Issue: When `BoundedReplayBuffer` evicts an item to make room for a new item it still contains a strong reference to the evicted item.

New Test: This diff adds a new unit test to verify this behavior. Given this unit test needs to look under the hood of `BoundedReplayBuffer` in order to ensure no memory leaks exist, writing this test requires a new package private method.

Change: This diff fixes misbehavior.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
@@ -616,30 +616,21 @@ public final class ObservableReplay<T> extends ConnectableObservable<T> implemen
             size--;
             // can't just move the head because it would retain the very first value
             // can't null out the head's value because of late replayers would see null
-            Node newHead = new Node(null);
-            Node newNext = next.get();
-            newHead.set(newNext);
-            setFirst(newHead);
-            if (newNext == null) {
-                tail = newHead;
-            }
-        }
-        /* test */ final void removeSome(int n) {
-            Node head = get();
-            while (n > 0) {
-                head = head.get();
-                n--;
-                size--;
-            }
+            setFirst(next);
 
-            setFirst(head);
         }
         /**
          * Arranges the given node is the new head from now on.
-         * @param n the Node instance to set as first
+         * @param next the Node instance to set as first
          */
-        final void setFirst(Node n) {
-            set(n);
+        final void setFirst(Node next) {
+            Node newHead = new Node(null);
+            Node newNext = next.get();
+            newHead.set(newNext);
+            set(newHead);
+            if (newNext == null) {
+                tail = newHead;
+            }
         }
 
         @Override
@@ -740,7 +731,31 @@ public final class ObservableReplay<T> extends ConnectableObservable<T> implemen
         void truncateFinal() {
 
         }
-        /* test */ final  void collect(Collection<? super T> output) {
+        /**
+         * Collect all values in memory even if they will be ignored by subsequent calls to replay(). This is useful
+         * for tracking memory leaks.
+         * @param output all values.
+         */
+        /* test */ final void collectValuesInMemory(Collection<? super T> output) {
+            Node next = getHead();
+            for (;;) {
+                if (next != null) {
+                    Object o = next.value;
+                    Object v = leaveTransform(o);
+                    if (NotificationLite.isComplete(v) || NotificationLite.isError(v)) {
+                        break;
+                    }
+                    T value = NotificationLite.<T>getValue(v);
+                    if (value != null) {
+                        output.add(value);
+                    }
+                } else {
+                    break;
+                }
+                next = next.get();
+            }
+        }
+        /* test */ final void collect(Collection<? super T> output) {
             Node n = getHead();
             for (;;) {
                 Node next = n.get();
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
```diff
@@ -751,6 +757,26 @@ void truncateFinal() {
                 }
             }
         }
+        /* test */ final void collectObjectsInMemory(Collection<? super T> output) {
```
akarnokd: There is a `collect(Collection)` method just above this addition, what's wrong with that?

src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
```diff
@@ -616,7 +616,13 @@ final void removeFirst() {
             size--;
             // can't just move the head because it would retain the very first value
             // can't null out the head's value because of late replayers would see null
-            setFirst(next);
+            Node newHead = new Node(null);
+            Node newNext = next.get();
+            newHead.set(newNext);
```
akarnokd: This can be `lazySet`.

akarnokd (commented):
I have to think about this a bit. Also do you want to fix `ReplayProcessor` and the `replay()` operators?

src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
AttwellBrian: `collect(Collection)` intentionally ignores the value in `head`.

src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
AttwellBrian: I'll take a look at that.

AttwellBrian:
> I have to think about this a bit. Also do you want to fix ReplayProcessor and the replay() operators?

Why? Do these have similar problems?

src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
AttwellBrian: Wouldn't `lazySet()` cause bugs here?

src/main/java/io/reactivex/internal/operators/observable/ObservableReplay.java
```diff
@@ -617,23 +617,20 @@ final void removeFirst() {
             // can't just move the head because it would retain the very first value
             // can't null out the head's value because of late replayers would see null
             setFirst(next);
-        }
```
AttwellBrian, line 620: When testing you can just call `removeFirst()` repeatedly. No need to introduce a test-only method. Plus, doing this reduces some memory leak false positives.

AttwellBrian:
Updated my change so that it fixes leaks due to size truncation in both `BoundedReplayBuffer` and `SizeAndTimeBoundReplayBuffer`.

This pull request makes no attempt to fix leaks that occur due to timeouts. Fully fixing such leaks would require scheduling. Therefore, eagerly fixing these leaks is impractical. This is fine since Rx consumers won't expect timeouts to perform eager cleanup.

akarnokd:
I've thought about this and the problem with the change is that now there are two Node allocations happening per item instead of one. I would accept such change if it happened when the terminal events are received; the head no longer references +1 item for an unknown duration.

Why is this extra 1 item causing you trouble?

> Why? Do these have similar problems?

There are a bunch of operators having the same replay like behavior implemented with the same node logic:

- `ReplaySubject`
- `ReplayProcessor`
- `ObservableReplay`
- `FlowableReplay`

AttwellBrian:
> I've thought about this and the problem with the change is that now there are two Node allocations happening per item instead of one. I would accept such change if it happened when the terminal events are received; the head no longer references +1 item for an unknown duration.

Doesn't sound like a significant improvement. I can't think of any improvement ideas for this. Can you? 

> Why is this extra 1 item causing you trouble?

While Rx is designed with immutable value objects in mind, it is useful for asynchronously emitting other types as well. For example, passing around an `Observable<MapUI>` instead of a `MapUI` inside the Uber app makes it easy to continue progressing application startup while waiting for the map to load.

Optimizations like this have a large affect on application startup time. However, if something in Rx leaks `MapUI` we're in trouble! This will leak the entire app's `Activity`.

> There are a bunch of operators having the same replay like behavior implemented with the same node logic

Interesting. I see this was true with Rx1 as well. So I guess this leak isn't a huge deal. It hasn't caused many issues for us in the last two years.

My takeaway is: never put objects you don't want leaked into rx caches because the cache may live longer than expected.

akarnokd:
There was a previous report when the user actually wanted 1 item cached but due to the head reference, there was 2 object retained. The workaround for that case is the [cacheLast](https://github.com/akarnokd/RxJava2Extensions#flowabletransformerscachelast) extension operator.

> However, if something in Rx leaks MapUI we're in trouble!

I think caching Activities is by itself bad practice.

I'm not willing to pay this extra per item allocation in the standard operator, however, the following are acceptable changes:

- Introduce a trim() method on the Subject and Processor that performs this head swap on demand.
- Have this head swap only implemented for onError and onComplete paths, thus no more +1 retention after the subject/processor has terminated.

AttwellBrian:
> There was a previous report when the user actually wanted 1 item cached but due to the head reference, there was 2 object retained. The workaround for that case is the cacheLast extension operator.

Neat. I'll take a look at that. Looks like it only supports `Flowable` for now. But I guess I can transform back and forth to a flowable.

> I'm not willing to pay this extra per item allocation in the standard operator, however, the following are acceptable changes:
>Introduce a trim() method on the Subject and Processor that performs this head swap on demand.
> Have this head swap only implemented for onError and onComplete paths, thus no more +1 retention after the subject/processor has terminated.

Okay. I'll think about this a bit. 

Thanks for the help and promptness!

PR #5294, opened 2017-04-19
Friendly dependency for beginner developers.

idemax:
Just a small change for 101 developers be able to copy and past, and compile easily without need too much about Gradle.


vanniktech (changes requested):
You almost never want to use this approach as you're not able to reproduce builds.

idemax:
Once the current approach doesn't work, what's the point to keep it in the wiki?

akarnokd:
One of the first things beginner developers have to learn is the ability to substitute into patterns. In addition, any fixed number would need to be updated on each release plus this pattern also encourages developers to look at the latest version badge a couple of lines above.

PR #5297, opened 2017-04-20
Fix JavaDoc image for Single#flatMapObservable

veyndan:
[`Single.flatMapObservable.png`](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.flatMapObservable.png) should be used for `Single#flatMapObservable` instead of [`Single.flatMap.png`](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.flatMap.png).


JakeWharton (approved):

akarnokd:
Thanks!

PR #5301, opened 2017-04-20
2.x: more detailed no-multi-subscribe with std consumers error message

akarnokd:
This PR changes the "Disposable already set!" and "Subscription already set!" messages on the standard consumer classes (`DisposableSubscriber`, `DisposableObserver`, etc.) to something more meaningful:

"It is not allowed to subscribe with a(n) `<class name>` multiple times. Please create a fresh instance of `<class name>` and subscribe that to the target source instead."

Where `<class name>` is a placeholder for the `getClass().getName()` of the subclass of those consumer types. It should clearly state to avoid subscribing with them multiple times as well as printing the full class name to indicate the problem is with the use of the implementor class, and not with the abstract RxJava class.

Inspired by [this StackOverflow](http://stackoverflow.com/questions/43482263/rxjava2-protocolviolationexception-disposable-already-set) question, one of many such questions.

For the internal operators, the original error message stays because when they appear, that is still likely due to an implementation bug (or a misbehaving user-created custom implementation).

### DIFF
--- a/src/main/java/io/reactivex/internal/util/EndConsumerHelper.java
+++ b/src/main/java/io/reactivex/internal/util/EndConsumerHelper.java
@@ -1,0 +1,150 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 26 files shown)


JakeWharton (approved):
Reviewed on mobile. I tried my best.

akarnokd:
Thanks Jake. We can refine the message later if necessary.

Issue #5300, opened 2017-04-20
2.x Single unsubscribeOn missing


I60R:
Also, in order to match 2.x naming convertions that operator should be named `disposeOn`

akarnokd:
Indeed that operator is missing from `Single` and available on the other types. Would you like to convert closest `MaybeUnsubscribeOn` over to it in a PR?

It is called `unsubscribeOn` for historical reasons and 
- to pair up with `subscribeOn`
- to have the same name across the types instead of `disposeOn` and `cancelOn`.

`unsubscribeOn` is the name of (concept of) the operation whereas `doOnCancel` and `doOnDispose` represent the signal types and appropriate method names within the API and as such were adjusted.

I60R:
Thanks for fast feedback and explanation.

About PR: yes, I can do it.
However, I'm not too experienced programmer and can't promise anything about code quality and time

akarnokd:
Oh, I understand. A certain code quality is expected here, plus RxJava 2.0.9 is scheduled to be released tomorrow. I'll do the PR then in a few hours.

I60R:
Looks not too complicated. Can you give me 2-3 hours?

akarnokd:
I've done it. For exercise, see if you can do it yourself and compare it against #5302 to verify your approach.

I60R:
Done it same way, but without testing (oh, it's hardest thing to prove that my code works as expected)

PR #5302, opened 2017-04-20
2.x: add Single.unsubscribeOn()

akarnokd:
This PR adds the `Single.unsubscribeOn()` operator.

Related: #5300.

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -16,6 +16,8 @@ package io.reactivex;
 import java.util.NoSuchElementException;
 import java.util.concurrent.*;
 
+import org.reactivestreams.Publisher;
+
 import io.reactivex.annotations.*;
 import io.reactivex.disposables.Disposable;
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -3056,10 +3057,12 @@ public final Completable toCompletable() {
      * </dl>
      *
      * @return an {@link Observable} that emits a single item T or an error.
+     * @since 2.0.9 - experimental
      */
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     @SuppressWarnings("unchecked")
+    @Experimental
```
JakeWharton: Looks like this landed on the wrong method

JakeWharton (approved):

src/main/java/io/reactivex/Single.java
akarnokd: Thanks, fixed.

PR #5304, opened 2017-04-20
Correct documented return type of Single#flatMapObservable Function argument

veyndan:
The argument of `Single#flatMapObservable` clearly states that it is expecting a `Function` which returns an `ObservableSource` but the documentation contradicts that.


akarnokd (approved):

akarnokd:
Thanks!

Issue #5201, opened 2017-03-20
1.3 API promotions


akarnokd:
Planned release date for 1.3: **May 1st, 2017**

### Promote `Experimental` to stable:

- `rx.exceptions.AssemblyStackTraceException`
- `rx.functions.Cancellable`
- `rx.internal.observers.AssertableSubscriberObservable`
- `rx.internal.operators.FlatMapCompletable`
- `rx.internal.operators.FlatMapSingle`
- `rx.internal.schedulers.SchedulerWhen`
- `rx.internal.util.BackpressureDrainManager`
- `rx.internal.util.BlockingUtils` (the class and its single method)
- `rx.Completable.fromEmitter()`
- `rx.Completable.test()`
- `rx.CompletableEmitter`
- `rx.CompletableSubscriber`
- `rx.Emitter`
- `rx.Scheduler.when()`
- `rx.SingleEmitter`
- `rx.Single.fromEmitter`
- `rx.Single.merge(Observable<Single<T>>)` (2 overloads)
- `rx.Single.mergeDelayError(Observable<Single<T>>)` (2 overloads)
- `rx.Single.cache()`
- `rx.Single.to()`
- `rx.Single.doOnEach()`
- `rx.Single.doOnSuccess()`
- `rx.Single.test()`
- `rx.observers.AssertableSubscriber`
- `rx.observers.AsyncCompletableSubscriber`
- `rx.observers.SafeCompletableSubscriber`
- `rx.observers.TestSubscriber.getCompletions()`
- `rx.observers.TestSubscriber.awaitValueCount()`
- `rx.observers.TestSubscriber.assertValuesAndClear()`
- `rx.schedulers.Schedulers.reset()`
- `rx.plugins.RxJavaCompletableExecutionHook`
- `rx.plugins.RxJavaHooks`
- `rx.plugins.RxJavaPlugins.reset()`
- `rx.plugins.RxJavaPlugins.getCompletableExecutionHook()`
- `rx.plugins.RxJavaPlugins.registerCompletableExecutionHook()`
- `rx.plugins.RxJavaSchedulersHook.createComputationScheduler()` (2 overloads)
- `rx.plugins.RxJavaSchedulersHook.createIoScheduler()` (2 overloads)
- `rx.plugins.RxJavaSchedulersHook.createNewThreadScheduler()` (2 overloads)
- `rx.subjects.UnicastSubject`
[...]

PR #5318, opened 2017-04-27
1.x: apply API promotions for 1.3

akarnokd:
This PR applies the API promotions of #5201.

Note that `Single.unsubscribeOn` remainded experimental and `Observable.fromEmitter` was removed.

Where the experimental version was available, it was moved up as `<p>History: 1.x.y - experimental` similar to how 2.x versioning/promotions happen.

Suggested review strategy:

- verify `@Experimental` and `@Beta` annotations are removed from the method and from Javadoc
- verify `@since 1.3` is present

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -400,7 +400,7 @@ public class Observable<T> {
      *         calls onCompleted
      * @see <a href="http://reactivex.io/documentation/completable.html">ReactiveX documentation:
      *      Completable</a>
-     * @since 1
+     * @since 1.3
      */
     public Completable toCompletable() {
         return Completable.fromObservable(this);
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -110,7 +110,7 @@ public class Observable<T> {
      * <p>
      * You should call the Emitter's onNext, onError and onCompleted methods in a serialized fashion. The
      * rest of its methods are thread-safe.
-     *
+     * <p>History: 1.2.7 - experimental
      * @param <T> the element type
      * @param emitter the emitter that is called when a Subscriber subscribes to the returned {@code Observable}
      * @param backpressure the backpressure mode to apply if the downstream Subscriber doesn't request (fast) enough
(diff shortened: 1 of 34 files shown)


src/main/java/rx/Observable.java
```diff
@@ -350,8 +348,8 @@ protected Observable(OnSubscribe<T> f) {
      * @param <R> the resulting object type
      * @param converter the function that receives the current Observable instance and returns a value
      * @return the value returned by the function
+     * @since 1.3
```
JakeWharton, line 351: History?

src/main/java/rx/Observable.java
```diff
@@ -402,10 +400,8 @@ protected Observable(OnSubscribe<T> f) {
      *         calls onCompleted
      * @see <a href="http://reactivex.io/documentation/completable.html">ReactiveX documentation:
      *      Completable</a>
-     * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical
-     *        with the release number)
+     * @since 1
```
JakeWharton: 1.3

src/main/java/rx/Observable.java
JakeWharton: There's a lot of missing history below as well. Not sure if you were wanting it for every operator or not.

src/main/java/rx/SingleEmitter.java
```diff
@@ -24,11 +23,10 @@
  * <p>
  * All methods are thread-safe; calling onSuccess or onError twice or one after the other has
  * no effect.
- * @since 1.2.3 - experimental (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
- * 
+ * <p>History: 1.2.3 -experimental
```
JakeWharton: super nit: missing space after dash.

JakeWharton (approved):

src/main/java/rx/Observable.java
akarnokd: If it was not recorded in place, I didn't bother looking them all up.

PR #5332, opened 2017-05-04
Add the cast operator to Single.

Mikey-Burns:
Adding the cast operator to Single.  This exists in the 2.x branch but not in the 1.x branch, so callers have had to do unnatural operations like `.map(SomeClass.class:cast)` to get around this.  Added tests similar to those for Observable.cast. 

Signed-off-by: Mike Burns <burnsm523@gmail.com>

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -211,6 +211,21 @@ public class Single<T> {
      */
 
     /**
+     * Casts the success value of the current Single into the target type or signals a
+     * ClassCastException if not compatible.
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code cast} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <R> the target type
+     * @param klass the type token to use for casting the success result from the current Single
+     * @return the new Single instance
+     */
+    public final <R> Single<R> cast(final Class<R> klass) {
+        return lift(new OperatorCast<T, R>(klass));
+    }
+
+    /**
      * Returns an Observable that emits the items emitted by two Singles, one after the other.
      * <p>
      * <img width="640" height="380" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.concat.png" alt="">
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -220,6 +220,7 @@ public class Single<T> {
      * @param <R> the target type
      * @param klass the type token to use for casting the success result from the current Single
      * @return the new Single instance
+     * @since 1.3.1 - experimental
      */
     public final <R> Single<R> cast(final Class<R> klass) {
         return map(new SingleOperatorCast<T, R>(klass));
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -222,7 +222,7 @@ public class Single<T> {
      * @return the new Single instance
      */
     public final <R> Single<R> cast(final Class<R> klass) {
-        return lift(new OperatorCast<T, R>(klass));
+        return map(new SingleOperatorCast<T, R>(klass));
     }
 
     /**
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -222,6 +222,7 @@ public class Single<T> {
      * @return the new Single instance
      * @since 1.3.1 - experimental
      */
+    @Experimental
     public final <R> Single<R> cast(final Class<R> klass) {
         return map(new SingleOperatorCast<T, R>(klass));
     }
--- a/src/main/java/rx/internal/operators/SingleOperatorCast.java
+++ b/src/main/java/rx/internal/operators/SingleOperatorCast.java
@@ -1,0 +1,27 @@
+/*
+ * This is the confidential unpublished intellectual property of EMC Corporation,
+ * and includes without limitation exclusive copyright and trade secret rights
+ * of EMC throughout the world.
+ */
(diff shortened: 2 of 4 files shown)


src/main/java/rx/Single.java
```diff
@@ -211,6 +211,21 @@ protected Single(final Observable.OnSubscribe<T> f) {
...
+     * ClassCastException if not compatible.
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code cast} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param <R> the target type
+     * @param klass the type token to use for casting the success result from the current Single
+     * @return the new Single instance
+     */
+    public final <R> Single<R> cast(final Class<R> klass) {
```
JakeWharton: nit: param doesn't need `final`

src/main/java/rx/Single.java
```diff
@@ -211,6 +211,21 @@ protected Single(final Observable.OnSubscribe<T> f) {
...
+        return lift(new OperatorCast<T, R>(klass));
```
JakeWharton: I didn't realize you could reuse Observable operators on Singles. There's probably a bit of additional overhead, but the reuse is really nice.

src/main/java/rx/Single.java
Mikey-Burns: @JakeWharton I was following the example of [Observable.cast()](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observable.java#L5036).  Would you like me to remove `final` from here or shall I leave it to remain consistent?

src/main/java/rx/Single.java
akarnokd: Actually you should not reuse them but develop/port them directly. Especially this type of Map derivative that could be just a Func1 wrapping the cast and delegating to map.

Mikey-Burns: @akarnokd I will update the PR with a simplified implementation.  Thanks for the suggestion.

src/main/java/rx/Single.java
akarnokd: Please add `@Experimental` and `@since 1.3.1 - experimental` to the javadoc

src/main/java/rx/internal/operators/SingleOperatorCast.java
```diff
@@ -0,0 +1,27 @@
...
+import rx.functions.Func1;
+
+/**
+ * Converts the element of a Single to the specified type.
+ * @param <T> the input value type
+ * @param <R> the output value type
+ */
+public class SingleOperatorCast<T, R> implements Func1<T, R> {
+
+  final Class<R> castClass;
```
akarnokd: 4 spaces.

akarnokd (changes requested):

src/main/java/rx/internal/operators/SingleOperatorCast.java
Mikey-Burns: Sure.  Updated to match the code style.

akarnokd (approved):

src/main/java/rx/internal/operators/SingleOperatorCast.java
```diff
@@ -0,0 +1,27 @@
+/*
```
dano: Probably want to fix these copyright notices :)

akarnokd: @dano Thanks!

src/test/java/rx/internal/operators/SingleOperatorCastTest.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * This is the confidential unpublished intellectual property of EMC Corporation,
```
akarnokd: This too.

akarnokd (changes requested):
Please fix the copyright headers.

src/main/java/rx/internal/operators/SingleOperatorCast.java
Mikey-Burns: Thanks @dano.  IntelliJ default settings getting the better of me again.

akarnokd (approved):

src/test/java/rx/internal/operators/SingleOperatorCastTest.java
```diff
@@ -0,0 +1,52 @@
...
+    @Test
+    public void testSingleCast() {
+        Single<?> source = Single.just(1);
+        Single<Integer> single = source.cast(Integer.class);
+
+        @SuppressWarnings("unchecked")
+        Observer<Integer> observer = mock(Observer.class);
+        single.subscribe(observer);
+        verify(observer, times(1)).onNext(1);
+        verify(observer, times(1)).onNext(1);
```
artem-zinnatullin: Should be `onNext(2)`, same mistake in [`OperatorCastTest`](https://github.com/ReactiveX/RxJava/blob/daf1022ce86291791817b1ef0f45445445e9a15e/src/test/java/rx/internal/operators/OperatorCastTest.java#L37) as well.

artem-zinnatullin (changes requested):

src/test/java/rx/internal/operators/SingleOperatorCastTest.java
Mikey-Burns: @artem-zinnatullin I agree that in OperatorCastTest the second invocation should use 2.  However, here I think the second invocation is actually unnecessary because only one item gets emitted from the Single.  I can remove this line since it is not needed.

artem-zinnatullin (approved):
👍

PR #5344, opened 2017-05-16
2.x: add tryOnError to create/XEmitter API

akarnokd:
This PR adds the `tryOnError` method to the various `Emitter` types used in the `create` operators that allows the developer to avoid the `UndeliverableException` in case a cancellation is racing with the emission of an error. The return value indicates a success of the delivery; in case of `false`, the developer can decide to log/drop the specific error if that makes sense for him/her.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableCreate.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableCreate.java
@@ -134,7 +134,7 @@ public final class FlowableCreate<T> extends Flowable<T> {
             }
         }
 
-       @Override
+        @Override
         public boolean tryOnError(Throwable t) {
            if (emitter.isCancelled() || done) {
                return false;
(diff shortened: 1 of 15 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableCreate.java
```diff
@@ -129,21 +129,27 @@ public void onNext(T t) {
...
-            if (error.addThrowable(t)) {
-                done = true;
-                drain();
-            } else {
+            if (!tryOnError(t)) {
                 RxJavaPlugins.onError(t);
             }
         }
 
+       @Override
```
vanniktech: indentation seems off

vanniktech (approved):

JakeWharton (approved):

src/test/java/io/reactivex/internal/operators/flowable/FlowableCreateTest.java
```diff
@@ -877,4 +877,59 @@ public void cancel() throws Exception {
         }
     }
 
+
+    @Test
+    public void tryOnError() {
+        for (BackpressureStrategy strategy : BackpressureStrategy.values()) {
```
artem-zinnatullin, line 883: :+1:

PR #5353, opened 2017-05-18
1.x: remove semicolon that Eclipse thinks is compile error

davidmoten:
as per title, gradle build works fine but Eclipse doesn't like it.


src/test/java/rx/internal/operators/SingleOperatorCastTest.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 package rx.internal.operators;
-;
```
Mikey-Burns, line 17: Sorry about that.  Good catch.

PR #5367, opened 2017-05-25
2.x: make sure interval+trampoline can be stopped

akarnokd:
This PR fixes the case where a `trampoline` scheduler is used with the `interval` or `intervalRange` operator, the periodic emissions can't be cancelled properly. The synchronous and blocking nature of the periodic schedule is that the `Disposable` is never connected with the downstream and the `interval` stays in a drain-sleep loop per emission indefinitely.

This PR changes the operators to use the `Worker` of a trampoline scheduler that is available before the drain-sleep loop thus a downstream cancellation can stop the loop. Any other "async" scheduler will still use the direct periodic scheduling facility.

In addition, the trampoline loop has been fixed to check for the worker `disposed` state in the inner loop.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableInterval.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableInterval.java
@@ -19,9 +19,11 @@ import java.util.concurrent.atomic.*;
 import org.reactivestreams.*;
 
 import io.reactivex.*;
+import io.reactivex.Scheduler.Worker;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.MissingBackpressureException;
 import io.reactivex.internal.disposables.DisposableHelper;
+import io.reactivex.internal.schedulers.TrampolineScheduler;
 import io.reactivex.internal.subscriptions.SubscriptionHelper;
 import io.reactivex.internal.util.BackpressureHelper;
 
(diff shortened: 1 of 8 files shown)


vanniktech (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableInterval.java
```diff
@@ -43,9 +45,16 @@ public void subscribeActual(Subscriber<? super Long> s) {
         IntervalSubscriber is = new IntervalSubscriber(s);
         s.onSubscribe(is);
 
-        Disposable d = scheduler.schedulePeriodicallyDirect(is, initialDelay, period, unit);
+        Scheduler sch = scheduler;
 
-        is.setResource(d);
+        if (sch instanceof TrampolineScheduler) {
```
artem-zinnatullin, line 50: Can't this fix be made in TrampolineScheduler itself?

akarnokd: No, because the `schedulePeriodic` has to return a `Disposable` in order to allow it to be stopped but the first task stays in the trampoline loop and stays there never returning from the call. The disposability has to happen before that. This is a similar case to synchronous cancellation with streams that have to send a `Disposable`/`Subscription` first to allow cancellation before going into some form of loop that would otherwise not return. The classic Rx.NET and early RxJava suffered from this on a grand scale. This issue only affects interval and trampoline scheduler together.

artem-zinnatullin:
Yeah I see, so alternative solution would be to accept Disposable as
parameter instead of returning it in scheduleX() methods right?

I'm on mobile, but I believe Scheduler is still abstract class as in 1.x
and such overloads could probably be added without breaking compatibility?

akarnokd:
> Yeah I see, so alternative solution would be to accept Disposable as
parameter instead of returning it in scheduleX() methods right?

No, it would require a consumer callback to be called, similar to how `autoConnect` or `connect` has one, to get the `Disposable` out before it gets submitted to an executor.

> I'm on mobile, but I believe Scheduler is still abstract class as in 1.x
and such overloads could probably be added without breaking compatibility?

Doesn't work. In order to not break existing custom schedulers, the new non-abstract methods have to call the old ones which return `Disposable` instead of providing it some way before.

artem-zinnatullin (approved):

artem-zinnatullin:
>No, it would require a consumer callback to be called, similar to
how autoConnect or connect has one to get the Disposable out before it gets
submitted to an executor.

Wouldn't adding inner Disposable into outer one do the same thing?

>Doesn't work. In order to not break existing custom schedulers, the new
non-abstract methods have to call the old ones which
return Disposable instead of providing it some way before.

Which seems possible..

Though okay, it sounds like a huge change and if there are no other
possible problematic cases as Intervals, this PR LGTM

Issue #5376, opened 2017-05-30
Rx 2.x SingleMap should not allow null mapper?


kojilin:
RxJava 2.0.8

Found code like below can works
```
Single.just("1234").map(s -> null)
              .subscribe(s -> {
                  System.out.println(s);
              });
```

Because Single#just not allow null, so SingleMap should not allow it?
should like this? https://github.com/ReactiveX/RxJava/compare/2.x...kojilin:FIX-NPE?expand=1

davidmoten:
Your PR looks good to me, I'd say submit it but remember to remove the "Thank you contributing..."  template from the description.

akarnokd:
Closing via #5378

PR #5377, opened 2017-05-30
2.x: Flowable.combineLatest combiner should not return null

davidmoten:
As per #5376 the combiner should not return null when applied.

* added unit test that failed on original code 
* mentioned constraint in javadoc

I'll look for some others as well.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -167,7 +167,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      * @param sources
      *            the collection of source Publishers
      * @param combiner
-     *            the aggregation function used to combine the items emitted by the source Publishers
+     *            the aggregation function used to combine the items emitted by the source Publishers. Should not return null.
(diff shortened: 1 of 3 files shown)


davidmoten:
Hah, I take that back! The unit test did not fail on original code. I'll check this one with @akarnokd because to get test coverage of this one requires fusion.

davidmoten:
More cases in 

* FlowableFlattenIterable, ObservableFlattenIterable (throws NPE but without helpful message),
* FlowableInternalHelper, ObservableInternalHelper
* FlowableWithLatestFromMany
* MaybeFlatMapIterableObserver  (throws NPE but without helpful message),
* MaybeZipArray
* perhaps more, I stopped looking.

I notice there is a `@NonNull` annotation on the `Function` interface so perhaps the javadoc changes are not necessary.   I'll close this PR for the moment to have a more thorough look.

akarnokd:
This particular lack of null check is on the fusion path. I'll compose a PR that fixes it and all the other places I can find.

davidmoten:
Beaut @akarnokd, thanks

PR #5378, opened 2017-05-30
2.x: make SingleMap not allow map function return null

kojilin:
Found code like below can works
```
Single.just("1234").map(s -> null)
              .subscribe(s -> {
                  System.out.println(s);
              });
```
Because Single#just not allow null, so prevent mapper return null.


akarnokd (approved):

davidmoten (approved):
LGTM, thanks!

artem-zinnatullin (approved):

PR #5379, opened 2017-05-30
2.x: add missing null checks on values returned by user functions

akarnokd:
This PR adds null checks to places where it was missing.

In addition, the `XMapNotification` operators now report a composite exception of the original `Throwable` and the error thrown by the function that should return the continuation source.

Also fixed `TestSubscriber`/`TestObserver` not cancelling the sequence if the fused poll crashed.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableInternalHelper.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableInternalHelper.java
@@ -77,7 +77,8 @@ public final class ObservableInternalHelper {
 
         @Override
         public ObservableSource<T> apply(final T v) throws Exception {
-            return new ObservableTake<U>(itemDelay.apply(v), 1).map(Functions.justFunction(v)).defaultIfEmpty(v);
+            ObservableSource<U> o = ObjectHelper.requireNonNull(itemDelay.apply(v), "The itemDelay returned a null ObservableSource");
+            return new ObservableTake<U>(o, 1).map(Functions.justFunction(v)).defaultIfEmpty(v);
         }
     }
 
@@ -165,7 +166,7 @@ public final class ObservableInternalHelper {
         @Override
         public ObservableSource<R> apply(final T t) throws Exception {
             @SuppressWarnings("unchecked")
-            ObservableSource<U> u = (ObservableSource<U>)mapper.apply(t);
+            ObservableSource<U> u = (ObservableSource<U>)ObjectHelper.requireNonNull(mapper.apply(t), "The mapper returned a null ObservableSource");
             return new ObservableMap<U, R>(u, new FlatMapWithCombinerInner<U, R, T>(combiner, t));
         }
     }
@@ -185,7 +186,7 @@ public final class ObservableInternalHelper {
 
         @Override
         public ObservableSource<U> apply(T t) throws Exception {
-            return new ObservableFromIterable<U>(mapper.apply(t));
+            return new ObservableFromIterable<U>(ObjectHelper.requireNonNull(mapper.apply(t), "The mapper returned a null Iterable"));
         }
     }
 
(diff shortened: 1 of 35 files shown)


src/main/java/io/reactivex/internal/operators/observable/ObservableInternalHelper.java
```diff
@@ -319,7 +320,7 @@ public boolean test(Notification<Object> t) throws Exception {
         @Override
         public Observable<R> apply(T t) throws Exception {
             return RxJavaPlugins.onAssembly(new SingleToObservable<R>(
-                ObjectHelper.requireNonNull(mapper.apply(t), "The mapper returned a null value")));
+                ObjectHelper.requireNonNull(mapper.apply(t), "The mapper returned a null SingleSource")));
```
vanniktech, line 323: Isn't it Observable / ObservableSource?

src/main/java/io/reactivex/observers/TestObserver.java
```diff
@@ -142,6 +142,7 @@ public void onNext(T t) {
             } catch (Throwable ex) {
                 // Exceptions.throwIfFatal(e); TODO add fatal exceptions?
```
vanniktech, line 143: what about this one?

src/test/java/io/reactivex/internal/operators/flowable/FlowableFlatMapTest.java
```diff
@@ -264,7 +264,7 @@ public void testFlatMapTransformsOnErrorFuncThrows() {
 
         source.flatMap(just(onNext), funcThrow((Throwable) null, onError), just0(onComplete)).subscribe(o);
 
-        verify(o).onError(any(TestException.class));
+        verify(o).onError(any(CompositeException.class));
```
vanniktech, line 267: verify that the exceptions are in right order within the CompositeException?

vanniktech (approved):

src/main/java/io/reactivex/internal/operators/observable/ObservableInternalHelper.java
akarnokd: Just open the collapsed section and see what mapper is: `final Function<? super T, ? extends SingleSource<? extends R>> mapper;`

src/main/java/io/reactivex/observers/TestObserver.java
akarnokd: I'm still undecided.

src/test/java/io/reactivex/internal/operators/flowable/FlowableFlatMapTest.java
akarnokd: Tested separately in the FlowableMapNotificationTest.java additions.

artem-zinnatullin (approved):

src/test/java/io/reactivex/flowable/FlowableNullTests.java
```diff
@@ -1341,6 +1341,7 @@ public void flatMapNotificationOnErrorNull() {
     }
 
     @Test(expected = NullPointerException.class)
+    @Ignore("No longer crashes with NPE but signals it; tested elsewhere.")
```
artem-zinnatullin, line 1344: Are there reasons to keep `@Ignore`d tests in this PR?

akarnokd: I had two options: I remove it and I get a question on why or ignore it with comments and get a question why...

artem-zinnatullin: Haha :D

But now you have explanation in GitHub history so maybe remove them?

akarnokd: First it has to get into the 2.x branch for that. Otherwise, there could be a purge of ignored unit test in the future.

vanniktech: > I remove it and I get a question on why or ignore it with comments and get a question why...

I like this way of thinking. :D

davidmoten (approved):
Thanks @akarnokd

Issue #5394, opened 2017-06-05
Add Gitter.im to README.md under Communication


Coffee2CodeNL:
Could you add https://gitter.im/ReactiveX/RxJava to the Communication area in the Readme?

There are some people there, but I think adding it to the readme might spark some activity! :)

akarnokd:
You can post a PR that updates README.md if you want.

akarnokd:
Closing via #5395.

PR #5395, opened 2017-06-05
Update README.md

Coffee2CodeNL:
Added Gitter.im chat for support.

As discussed in #5394

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,10 +171,11 @@ For further details, consult the [wiki](https://github.com/ReactiveX/RxJava/wiki
 
 ## Communication
 
+- [Gitter.im](https://gitter.im/ReactiveX/RxJava)
 - Google Group: [RxJava](http://groups.google.com/d/forum/rxjava)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,10 +171,11 @@ For further details, consult the [wiki](https://github.com/ReactiveX/RxJava/wiki
 
 ## Communication
 
+- [Gitter.im](https://gitter.im/ReactiveX/RxJava)
```
akarnokd: Why regrouping? Actually I prefer StackOverflow because questions can be isolated much better. The gitter channels usually invoke chaos by talking over each other and hard to follow who reacted to what.

Coffee2CodeNL: Realtime -> "Slightly Slower" communication -> Social Media was my idea. 
Unless you want it sorted by preferred communication channels, then i'll update it to your liking.

akarnokd: Please add this to the bottom without changing the order of the existing list elements.

PR #5406, opened 2017-06-11
Fix javadoc for Observable.reduce() and Observable.reduceWith()

devisnik:
The sample code given in the javadoc for `Observable.reduce()` isn't compiling, since the function returns a `Single`. 
This PR changes it to a compiling example. 
It also updates the javadoc for `reduceWith()` to better match its actual behaviour.

*Remark*
I'm not sure whether the sample code for `reduce()` is needed at all. The problem that it addresses might be solved better by using `reduceWith()`, imho.

*Update*
- similar changes applied to `Flowable.reduce()` and `Flowable.reduceWith()`

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -10766,7 +10766,7 @@ public abstract class Flowable<T> implements Publisher<T> {
 
     /**
      * Returns a Flowable that applies a specified accumulator function to the first item emitted by a source
-     * Publisher and a specified seed value derived from calling a specified seedSupplier, then feeds the result
+     * Publisher and a seed value derived from calling a specified seedSupplier, then feeds the result
      * of that function along with the second item emitted by a Publisher into the same function, and so on until
      * all items have been emitted by the source Publisher, emitting the final result from the final call to your
      * function as its sole item.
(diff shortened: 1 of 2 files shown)


akarnokd:
Could you also fix Flowable's javadoc?

devisnik:
@akarnokd Done.
What do you think about referring to `reduceWith()` instead of giving the example that uses `defer()`?

akarnokd:
The `defer` example should stay but you can add `@see` and a 3rd example.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -10723,18 +10723,18 @@ public final Completable ignoreElements() {
...
      * Publisher&lt;T> source = ...
-     * Publisher.defer(() -> source.reduce(new ArrayList&lt;>(), (list, item) -> list.add(item)));
+     * Single.defer(() -> source.reduce(new ArrayList&lt;>(), (list, item) -> list.add(item)));
      *
      * // alternatively, by using compose to stay fluent
      *
      * source.compose(o ->
-     *     Publisher.defer(() -> o.reduce(new ArrayList&lt;>(), (list, item) -> list.add(item)))
-     * );
+     *     Publisher.defer(() -> o.reduce(new ArrayList&lt;>(), (list, item) -> list.add(item)).toFlowable())
```
akarnokd: This should be `Flowable.defer`.

akarnokd (commented):
A small original mention of `Publisher` could be fixed.

devisnik:
Updated according to the feedback.

akarnokd:
Great!

PR #5405, opened 2017-06-09
DefaultObserver javadoc example fix

devisnik:
The example in the javadoc of `DefaultObserver` is not compiling, `DefaultObserver` is not a `Disposable`. 
This PR updates the javadoc to provide a compiling example.

PR #5407, opened 2017-06-11
DefaultSubscriber javadoc sample fix

devisnik:
The example in the javadoc of `DefaultSubscriber` is not compiling, `DefaultSubscriber` is not a `Disposable`. 
This PR updates the javadoc to provide a compiling example.

Fix is similar to #5405 for `DefaultObserver`.


src/main/java/io/reactivex/subscribers/DefaultSubscriber.java
```diff
@@ -52,7 +52,7 @@
  * <p>Example<code><pre>
  * Disposable d =
```
akarnokd: This assignment doesn't work anymore.

devisnik: Sorry, fixed.

Issue #5382, opened 2017-05-31
2.x Single.delay documentation and implementation inconsistent


mmallozzi:
The documentation for Single.delay says that success or error signals will be delayed by the specified amount, but in fact errors are not delayed, only successes. I have no opinion on which one should be correct - I ran into it when mocking out the response for an incomplete backend to test a loading/error UI, e.g. `Single.error(...).delay(3, TimeUnit.SECONDS)`, and found that my errors never had a delay. It seems that Observable has both options, with a version of the method that has a boolean flag `delayError`.

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Single.java#L1636
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/single/SingleDelay.java#L66

akarnokd:
Indeed there is no option for that in `Single` and `Maybe`. Would you like to submit a PR?

mithunsasidharan:
@akarnokd : I've submitted my first PR fixing this -> https://github.com/ReactiveX/RxJava/pull/5409 Kindly review and advice. Thanks.

akarnokd:
Closing via #5409.

PR #5409, opened 2017-06-12
#5382 Corrected Single.delay documentation

mithunsasidharan:
Corrected Single.delay documentation that says that success or error signals will be delayed by the specified amount, but in fact errors are not delayed, only successes.

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1633,8 +1633,7 @@ public abstract class Single<T> implements SingleSource<T> {
     }
 
     /**
-     * Delays the emission of the success or error signal from the current Single by
-     * the specified amount.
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -1633,8 +1633,7 @@
     }
 
     /**
-     * Delays the emission of the success or error signal from the current Single by
-     * the specified amount.
+     * Delays the emission of the success signal from the current Single by the specified amount.
```
akarnokd, line 1636: Please update the `@param time` line as well

mithunsasidharan: @akarnokd : Done. Could you review. Thanks

akarnokd (approved):

PR #5413, opened 2017-06-14
2.x: Add empty source clauses to javadocs of combineLatest operators …

### LINKED ISSUE #5412: combineLatest(Iterable<Publisher<>>, Function<>) not emitting on empty iterable
I am using RxJava 2.10 and operator

io.reactivex.Flowable#combineLatest(java.lang.Iterable<? extends org.reactivestreams.Publisher<? extends T>>, io.reactivex.functions.Function<? super java.lang.Object[],? extends R>)

 but I've seen this behavior in 1.x as well. 

Overload of combineLatests (and combineLatestsDelayError) that takes Iterable of Publishers is not emitting any items and 'combiner' function passed to it is not called when iterable is empty. For this operator I would expect it to scale linearly, that is:

- If I pass Iterable of size 2 as a first argument I expect an Array of size 2 to be passed to combiner function
- If I pass Iterable of size 1 as a first argument I expect an Array of size 1 to be passed to combiner function
- If I pass Iterable of size 0 as a first argument I expect an Array of size 0 to be passed to combiner function

Sroka:
…accepting unspecified number of sources. This commit resolves #5412

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -150,6 +150,9 @@ public abstract class Flowable<T> implements Publisher<T> {
      * If any of the sources never produces an item but only terminates (normally or with an error), the
      * resulting sequence terminates immediately (normally or with all the errors accumulated till that point).
      * If that input source is also synchronous, other sources after it will not be subscribed to.
+     * <p>
+     * If provided array of source Publishers is empty, the resulting sequence completes immediately without emitting
+     * any items and without any calls to combiner function.
      *
      * <dl>
      *  <dt><b>Backpressure:</b></dt>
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -150,6 +150,9 @@ public static int bufferSize() {
      * If any of the sources never produces an item but only terminates (normally or with an error), the
      * resulting sequence terminates immediately (normally or with all the errors accumulated till that point).
      * If that input source is also synchronous, other sources after it will not be subscribed to.
+     * <p>
+     * If provided array of source Publishers is empty, resulting sequence completes immediately without emitting
```
akarnokd: "If the provided array of source Publishers is empty, the resulting sequence completes immediately without emitting any items and without any calls to the combiner function."

src/main/java/io/reactivex/Flowable.java
```diff
@@ -191,6 +194,9 @@ public static int bufferSize() {
      * If any of the sources never produces an item but only terminates (normally or with an error), the
      * resulting sequence terminates immediately (normally or with all the errors accumulated till that point).
      * If that input source is also synchronous, other sources after it will not be subscribed to.
+     * <p>
+     * If there are no source Publishers provided, resulting sequence completes immediately without emitting
```
akarnokd: "If there are no source Publishers provided, the resulting sequence completes immediately without emitting any items and without any calls to the combiner function."

akarnokd (commented):
A couple of "the" are missing.

akarnokd:
Are you going to update this PR based on the feedback?

Sroka:
Argh, totally forgot about it, sorry. I will update it

PR #5415, opened 2017-06-14
2.x: fix doOnNext failure not triggering doOnError when fused

akarnokd:
This PR fixes an issue when in a fused chain of `doOnNext` and `doOnError` the `doOnNext` function fails, the `doOnError` consumer is not called.

Originally reported in Reactor-Core: https://github.com/reactor/reactor-core/issues/664

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableDoOnEach.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableDoOnEach.java
@@ -21,6 +21,7 @@ import io.reactivex.exceptions.*;
 import io.reactivex.functions.*;
 import io.reactivex.internal.fuseable.ConditionalSubscriber;
 import io.reactivex.internal.subscribers.*;
+import io.reactivex.internal.util.ExceptionHelper;
 import io.reactivex.plugins.RxJavaPlugins;
 
 public final class FlowableDoOnEach<T> extends AbstractFlowableWithUpstream<T, T> {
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableDoOnEach.java
```diff
@@ -282,11 +305,33 @@ public int requestFusion(int mode) {
...
+            try {
+                v = qs.poll();
+            } catch (Throwable ex) {
+                Exceptions.throwIfFatal(ex);
+                try {
+                    onError.accept(ex);
+                } catch (Throwable exc) {
+                    throw new CompositeException(ex, exc);
+                }
+                throw ExceptionHelper.<Exception>throwIfThrowable(ex);
```
JakeWharton, line 319: Hmm I don't understand why we both signal `onError` and also re-throw. It seems like it should be one or the other, no?

akarnokd: Because this matches how `doOnError` would be called with the upstream error and then that error is forwarded to the next operator.

src/main/java/io/reactivex/internal/util/ExceptionHelper.java
```diff
@@ -106,6 +106,21 @@ public static RuntimeException wrapOrThrow(Throwable error) {
...
 
+    /**
+     * Workaround for Java 6 not supporting throwing a final Throwable from a catch block.
+     * @param <E> the generic exception type
+     * @param e the Throwable error to return or throw
+     * @return the Throwable e if it is a subclass of Exception
+     * @throws E the generic exception thrown
+     */
+    @SuppressWarnings("unchecked")
+    public static <E extends Throwable> Exception throwIfThrowable(Throwable e) throws E {
```
artem-zinnatullin, line 117: `@CheckReturnValue`

artem-zinnatullin, line 117: Maybe `throwIfThrowableOrReturn`?

artem-zinnatullin (approved):

src/main/java/io/reactivex/internal/util/ExceptionHelper.java
akarnokd: Internal API.

vanniktech (approved):

Issue #5429, opened 2017-06-21
1.x - AsyncOnSubscribe may never complete due to premature unsubscribe from requested observables when onCompleted is emitted from the callback.


yurgisbaykshtis:
The problem can be reproduced using a variety of downstream operators such as zip, zipWith, mergeWith, concatEager, but only if the requested observable is delayed (simulating an async source).  Zip and mergeWith use cases require at least 90 items to be emitted (likely related to a buffer size and request-more threshold)

Below is the list of test case with delay on/off, and different item counts.

```java
 private Observable<Integer> createAsyncOnSubscribe(int count, boolean delay, Scheduler scheduler) {
    return Observable.create(AsyncOnSubscribe.<Integer,Integer>createStateful(()->0,(state, requested, emitter)->{
      if (state == 0) {
        Observable<Integer> items = Observable.range(0, count);
        emitter.onNext(delay? items.delay(1, TimeUnit.SECONDS, scheduler) : items);
      } else {
        emitter.onCompleted();
      }
      return state + 1;
    }))
    .timeout(1, TimeUnit.MINUTES, scheduler);
  }
  
  @Test
  public void testConcatEagerSingleNoDelay() {
    System.out.println("\nconcatEager single no delay:");
    TestScheduler sched = Schedulers.test();
    TestSubscriber<Object> ts = new TestSubscriber<>(); 
    Observable.concatEager(Observable.just(createAsyncOnSubscribe(1, false, sched))).subscribe(ts);
    sched.advanceTimeBy(1, TimeUnit.HOURS);
    System.out.println("errors:" + ts.getOnErrorEvents());
    ts.assertCompleted();
  }

  @Test
  public void testConcatEagerSingleDelay() {
    System.out.println("\nconcatEager single with delay:");
    TestScheduler sched = Schedulers.test();
    TestSubscriber<Object> ts = new TestSubscriber<>(); 
    Observable.concatEager(Observable.just(createAsyncOnSubscribe(1, true, sched))).subscribe(ts);
    sched.advanceTimeBy(1, TimeUnit.HOURS);
    System.out.println("errors:" + ts.getOnErrorEvents());
    ts.assertCompleted();
  }

  @Test
  public void testConcatEager100Delay() {
    System.out.println("\nconcatEager 100 with delay:");
    TestScheduler sched = Schedulers.test();
    TestSubscriber<Object> ts = new TestSubscriber<>(); 
[...]

PR #5430, opened 2017-06-21
1.x: fix premature cleanup in AsyncOnSubscribe when the last Observable is still running

yurgisbaykshtis:
The issue is caused by premature cleanup when onCompleted is emitted from the callback function which may happen while a requested async sequence is still not completed.
The unit test is provided which will fail without the fix.
More cases are provided in the issue https://github.com/ReactiveX/RxJava/issues/5429.


akarnokd:
Thanks for looking into this.

akarnokd (approved):
Will take this for now. The whole operator needs an overhaul otherwise.

Issue #5433, opened 2017-06-21
2.x Time bounded replaySubject getValue() unexpected behavior


nathankooij:
Affects 2.1.1.

The Javadoc for ReplaySubject::getValue states the following:
> Returns a single value the Subject currently has or null if no such value exists.

In combination with a time bounded replay subject I would expect `null` if either:
* there has not yet been an emission
* the emitted item's age is too old

Currently this is not the case as demonstrated by the following test: https://github.com/nathankooij/RxJava/commit/6afa40c0493bd90f012a91f2c20ce4cc2695826e
However, getValues() does return `null` (that is, old items are evicted), which is also demonstrated by the aforementioned test. To me it seems that either the implementation of `getValue/getValues` is inconsistent with each other, and the current behavior is not obvious to me from the documentation.

I've added what I think could be the fix in: https://github.com/nathankooij/RxJava/commit/bf1c1895f5fb66fdb38f1ad9d09212b3462a1ffd

akarnokd:
Yes, it looks like an inconsistency. Do you want to post a PR? Please don't forget to verify/fix `ReplayProcessor` if possible.

nathankooij:
Yes, will open a PR.

akarnokd:
Closing via #5434.

PR #5434, opened 2017-06-21
2.x: fix time bounded replaySubject getValue() inconsistency with getValues() on old items

nathankooij:
See https://github.com/ReactiveX/RxJava/issues/5433.


akarnokd (approved):
LGTM. The build may need several restart as currently it either times out or gets killed by the container for some reason.

vanniktech (approved):

PR #5436, opened 2017-06-23
2.x: Gradle to use less memory, missed an export command

akarnokd:
Looks like there was a missing "export" keyword in the build options and thus gradle ate all 3 GB memory dancing on the edge of the OOMKiller.


akarnokd:
Much better:

https://travis-ci.org/ReactiveX/RxJava/builds/246182350#L687

image

PR #5444, opened 2017-06-26
2x: Fix wording in Async and Publish processors javadoc.

serj-lotutovici:
Seams that these are some leftovers from the 1.x -> 2.x refactoring.

### DIFF
--- a/src/main/java/io/reactivex/processors/PublishProcessor.java
+++ b/src/main/java/io/reactivex/processors/PublishProcessor.java
@@ -23,14 +23,14 @@ import io.reactivex.internal.util.BackpressureHelper;
 import io.reactivex.plugins.RxJavaPlugins;
 
 /**
- * A Subject that multicasts events to Subscribers that are currently subscribed to it.
+ * Processor that multicasts all subsequently observed items to its current {@link Subscriber}s.
  *
  * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/processors/PublishProcessor.java
```diff
@@ -23,14 +23,14 @@
 import io.reactivex.plugins.RxJavaPlugins;
 
 /**
- * A Subject that multicasts events to Subscribers that are currently subscribed to it.
+ * Processor that multicasts all subsequently observed items to its current {@link Subscriber}s.
```
akarnokd, line 26: "A Processor"  ?

akarnokd (approved):

src/main/java/io/reactivex/processors/PublishProcessor.java
serj-lotutovici: In Replay and Behaviour processors it's "Processor" \_(ツ)_/¯. Both seam syntactically correct.

akarnokd: Then the `AsyncProcessor` could be adjusted as well.

vanniktech (approved):

src/main/java/io/reactivex/processors/PublishProcessor.java
serj-lotutovici: done

Issue #5151, opened 2017-03-03
Nullability Annotations


ScottPierce:
It would be helpful if RxJava had nullability annotations that might provide warnings to developers in a good IDE. This would be a helpful addition, especially considering RxJava no longer allows streaming of `null` types.

It would also provide helpful indicators for other JVM languages that have Nullability built in. [Kotlin is a good example of this](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations). Currently using RxJava with Kotlin results in most code being interpreted as potentially nullable. Annotating types in the RxJava API would fix this in a way that a side library, like RxKotlin, couldn't.

akarnokd:
There were several enhancements in this regard in the recent past: #5055, #5051, #5023 for example.

ScottPierce:
I see. Now that you mention it I see them in certain places, but missing from others. It'd be helpful to have them on the `Observer` interfaces for instance. The `OnSubscribe`s too, such as `ObservableOnSubscribe`.

I'm sure there are a few others that are missing them as well, considering how often I see this.

akarnokd:
PR welcome.

ScottPierce:
Sounds reasonable. 

I had loaded the RxJava gradle project the other day, and came across several gradle errors. I ended up having to comment out a lot of the `build.gradle` regarding nebula, and animal sniffer to make gradle happy. Any idea why that might have been?

akarnokd:
Use the gradle in the project: 2.14 . The 3.x don't work properly due to the outdated nebula plugin. In theory, the nebula is just an aggregator of other plugins which appear to be 3.x ready. We could use the help with them too.

ScottPierce:
Oh right. Because I loaded the project with Android Studio, it forced the gradle wrapper to upgrade to 3.3. I'll just use IntelliJ next time. Thanks!

akarnokd:
Continuing with #5341.

PR #5447, opened 2017-06-27
Remove @NonNull annotation in Consumer method parameter

### LINKED ISSUE #5442: 2.x: Consumer contract is violated when using doOnEvent
The following code showcases the problem:
```java
Completable.complete()
  .doOnEvent(t -> {})
```
`t` in this case is annotated with `@NonNull` but it will be `null` since the completable completes.

This is a real pain in Kotlin since the following code crashes in runtime:
```kotlin
Completable.complete()
  .doOnEvent {}
```

ansman:
This fixes #5442

### DIFF
--- a/src/main/java/io/reactivex/functions/Consumer.java
+++ b/src/main/java/io/reactivex/functions/Consumer.java
@@ -13,8 +13,6 @@
 
 package io.reactivex.functions;
 
-import io.reactivex.annotations.NonNull;
-
 /**
  * A functional interface (callback) that accepts a single value.
  * @param <T> the value type
@@ -25,5 +23,5 @@ public interface Consumer<T> {
      * @param t the value
      * @throws Exception on error
      */
-    void accept(@NonNull T t) throws Exception;
+    void accept(T t) throws Exception;
 }


ansman:
This should be safe to merge, see the discussion in #5216

ansman:
I'm curious, why did the code coverage decrease because of this?

akarnokd:
There are unit tests that race for a certain code path inside the operators. Travis-CI lately has become 2-3x slower and often these code paths are no longer exercised. It is sometimes +/- 0.2%. If there is a generic drop unrelated to your changes, there is nothing to worry about.

artem-zinnatullin (approved):

akarnokd:
Looks like this nullability thing is hitting many Kotlin users now (see #5448). What do you think about removing all nullability annotations from the functional types in an additional PR?

svenjacobs:
> What do you think about removing all nullability annotations from the functional types

@akarnokd To add my two cents ;) I think this is a good idea. There is nothing wrong with nullability annotations as long as the code adheres to this contract - meaning `null` should never be passed where `@NonNull` is declared.

akarnokd:
Java 6 doesn't enforce this and only Java 8+ allows annotating the type arguments themselves:

```java
public Observable<T> doOnNext(Consumer<@NonNull T> onNext);

public Completable doOnEvent(Consumer<@Nullable Throwable> onEvent);
```

ansman:
@akarnokd I'd be happy to add a PR for removing them if you wish.
Should all annotations in the `io.reactivex.functions` package be removed then?

akarnokd:
Looks like `Function` is the next to be adjusted: #5448 indicates it's return type may be null with `groupBy`. I'd delay removing all annotations in case Kotlin comes to senses because I doubt RxJava is the only library that has inconsistent nullability annotation usage.

svenjacobs:
> Looks like Function is the next to be adjusted: #5448 indicates it's return type may be null with groupBy. I'd delay removing all annotations in case Kotlin comes to senses because I doubt RxJava is the only library that has inconsistent nullability annotation usage.

@akarnokd What do you mean by "Kotlin comes to senses"? This is by language design and I doubt it will change. I don't understand why we have to discuss this and why it's not clear that as a library developer you should not put `null` into an argument that is annotated by `@NonNull`. As I said before this is not a Kotlin-only problem and certainly will affect Java developers too as here clearly the contract is broken.

akarnokd:
RxJava is a Java 6 library and thus we are bound by the capabilities of Java language level 6. We are also trying to be nice with other languages but that has limits. The best we can do is to remove nullability annotations from controversial functional types that come up.

What I mean by "senses" is it would be Kotlin's responsibility to work with native Java libraries and provide workaround since Java's type system doesn't have a primary notion of nullability; I just leave off/replace the annotations to satisfy some static analysis tools if necessary. The fact that just by upgrading to 1.1.3 compatiblity broke indicates there has to be work done on Kotlin's part.

svenjacobs:
I'm aware that Java 6 cannot check nullability constraints during compile time. However this is not a language issue. IDEs like Intellij IDEA will analyze these annotations and give hints. So as a developer you might think that the argument can never be `null` but in fact it can. So what is the point of this annotation then?

akarnokd:
See #5151. If we could support callsite nullability, we'd do it but we can't. So either annotate for the majority uses or not annotate at all through the common functional interfaces. Adding/removing annotations is considered binary compatible thus any other means, such as alternate interfaces has to be added and accompanied by new methods which also need to be named differently to avoid lambda ambiguity.

> IDEs like Intellij IDEA will analyze these annotations and give hints

The keyword is "hints", the annotations were mostly working for Java/Android but Kotlin no longer considers them as hints. If there was a "I take my chances" type/hint or you could disable the RxJava annotations being considered, that would make things simpler.

JakeWharton:
If they're just hints I'd prefer them entirely removed. They should either
be 100% accurate or entirely absent. We have Java tools which take these
annotations as being 100% accurate and having anything less than that in
actuality will cause bugs.

akarnokd:
AFAIK, the problem with the non-annotated interfaces was that IntelliJ/Kotlin inferred all of them as nullable T and caused inconveniences.

JakeWharton:
Non-annotated types will be inferred as a "platform type" which has
ambiguous nullability. This allows the user to specify it as nullable or
non-nullable (with a runtime check in the latter case) based on what they
know of the data and operator.

akarnokd:
So far, these were the cases that require the removal of annotations:

1) `Consumer` may be called with null (i.e.,`doOnEvent`)
2) `BiConsumer` may be called with null (i.e., `subscribe`)
3) `Function` may return null in `groupBy`'s key selector

The other functional types are never called with null and are not allowed to return null. Option 3 has no PR yet.

ansman:
@akarnokd I added a PR for 3 in #5449

akarnokd:
Thanks @ansman !

svenjacobs:
> The fact that just by upgrading to 1.1.3 compatiblity broke indicates there has to be work done on Kotlin's part.

It just indicates that there were "holes" in the type system which have been fixed ;) Anyway thanks to everybody that this issue has been resolved!

PR #5457, opened 2017-06-30
2.x: See if sudo required results in consistent & faster build times

akarnokd:
I asked the Travis support about the slow build times of lately and they suggested I try `sudo:required` to have an isolated build VM instead of the shared container-based one. I'll rerun this PR a couple of times to see the effects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ jdk:
 # prevent travis running gradle assemble; let the build script do it anyway
 install: true
       
-sudo: false
+sudo: required
 # as per http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 
 # script for build and release via Travis to Bintray


artem-zinnatullin:
>Ran for 11 min 4 sec
>Ran for 12 min 14 sec

Can you please restart build couple times and post resulting build times so we could get some average?

akarnokd:
Ran for 11 min 24 sec
Ran for 11 min 3 sec

Looks reasonable to me. As I understood, the container-based build process uses a 2 core virtualized machine like the sudo one here but multiple builds run on the same VM at the same time, taking CPU time from each other. Time insensitive projects build fine, although much slower, but we have a higher chance of failing due timeouts. Strangely, we ran in container-based mode prior to June relatively fine for at least a year; only the last 3 weeks look bad. I'd think there was a configuration change that previously made sure containers run a build one after the other.

I'll wait for the next response from support.

artem-zinnatullin:
Your assumptions seems valid.

I also use `sudo: required` in a few projects for about half a year (because of Docker) and haven't seen hangs of such builds at all. At the same time regular `sudo: false` builds were always prone to freezes in many projects on Travis.

Another options — try other CIs like CircleCI, however I saw CircleCI builds frozen couple of times (maybe because of Android emulator though).

---

Idk what they're doing at Travis, but build times are crazy, used to be ~14 minutes month ago for 2.x:

[crazy-build-times](https://travis-ci.org/ReactiveX/RxJava/builds)


Month ago:

[month-ago](https://travis-ci.org/ReactiveX/RxJava/builds)

cristaloleg:
Adding stages might help somehow: https://docs.travis-ci.com/user/build-stages/warm-cache/

akarnokd:
@cristaloleg Our problem has nothing to do with startup time or caching but the indication that several concurrent builds are running on the same containerized VM sharing the same 2 core CPU thus generally slowing down things by 2-3x compared to a month ago and before. This makes me think of either misconfiguration or cost saving measure.

akarnokd:
The support was delighted `sudo: required` worked so I assume they are fine running in that mode now on.

PR #5467, opened 2017-07-05
2.x: Removing commented code from tests

pwittchen:
Hi,

In this PR I'm removing commented code from tests as it's generally considered as bad practice.
Part of this code is even not available in the current RxJava API. It does not resolve all of the issues like that, but only part of them. I didn't want to create PR with more changes because it will be harder to review. Rest of the similar issues could be resolved in the future PRs.

Regards,
Piotr

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElementsTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElementsTest.java
@@ -65,8 +65,6 @@ public class FlowableIgnoreElementsTest {
         ts.assertNoErrors();
         ts.assertNoValues();
         ts.assertTerminated();
-        // FIXME no longer testable
-//        ts.assertUnsubscribed();
     }
 
     @Test
@@ -76,8 +74,6 @@ public class FlowableIgnoreElementsTest {
         Flowable.error(ex).ignoreElements().toFlowable().subscribe(ts);
         ts.assertNoValues();
         ts.assertTerminated();
-        // FIXME no longer testable
-//        ts.assertUnsubscribed();
         ts.assertError(TestException.class);
         ts.assertErrorMessage("boo");
     }
@@ -183,8 +179,6 @@ public class FlowableIgnoreElementsTest {
         ts.assertNoErrors();
         ts.assertNoValues();
         ts.assertTerminated();
-        // FIXME no longer testable
-//        ts.assertUnsubscribed();
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableMapTest.java
```diff
@@ -256,139 +256,13 @@ public Integer apply(Integer i) {
...
     private static Map<String, String> getMap(String prefix) {
         Map<String, String> m = new HashMap<String, String>();
         m.put("firstName", prefix + "First");
         m.put("lastName", prefix + "Last");
         return m;
     }
 
-    @Test//(expected = OnErrorNotImplementedException.class)
-    @Ignore("RS subscribers can't throw")
-    public void testShouldNotSwallowOnErrorNotImplementedException() {
```
akarnokd: Please don't remove ignored test methods as they fulfill the crucial task of not getting killed on CI due to lack of test output.

pwittchen: Ok, I'll bring them back.

pwittchen:
@akarnokd I've pushed fix with your suggestion from code review (I kept ignored tests with commented body) and squashed my two commits into one for better git log.

artem-zinnatullin:
@pwittchen you can push without squashing (actually that's easier to review), GitHub can squash commits from PR before merge

src/test/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElementsTest.java
```diff
@@ -193,9 +187,7 @@ public void testErrorReceived() {
         TestException ex = new TestException("boo");
         Flowable.error(ex).ignoreElements().subscribe(ts);
         ts.assertNoValues();
-        ts.assertTerminated();
-        // FIXME no longer testable
-//        ts.assertUnsubscribed();
+        ts.assertTerminated();;
```
artem-zinnatullin: double `;`

artem-zinnatullin (changes requested):

src/test/java/io/reactivex/internal/operators/flowable/FlowableIgnoreElementsTest.java
pwittchen: Thanks. I'll fix this.

artem-zinnatullin:
That would be single line to review if you wouldn't squash :)

artem-zinnatullin (approved):

pwittchen:
@artem-zinnatullin Yeah, but it looks better after merge. Git log history is not "polluted" by fixing commits. There's just one commit with the final version of change :).

artem-zinnatullin:
As I mentioned in the initial comment, GitHub can squash commits before
merge if person who merges it selects that option

akarnokd:
image

pwittchen:
@artem-zinnatullin @akarnokd Thanks! I'll be aware of that in the future.

PR #5470, opened 2017-07-06
fix #5468: eager hooks onError call

### LINKED ISSUE #5468: 1.x Single#flatMapCompletable calls hooks onError when mapping to an error completeable
If you flatmap a Single into a Completeable, and the Completeable is an error, you will have the exception passed into the RxJavaHooks.onError global catch function as well as ending up with the Completeable Error you expect to see.

I have managed to get is down to the follow test case highlighting the problem,

```
@Test
    public void flatMapCompletableErrorCallsHooksErrorHandler() {
        AtomicBoolean wasHooksHandlerCalled = new AtomicBoolean();
        RxJavaHooks.setOnError(t -> wasHooksHandlerCalled.set(true));

        Completable c3  = Single.just(1)
                .flatMapCompletable(i -> Completable.error(new RuntimeException("some exception")));

        c3.get();

        assertThat(wasHooksHandlerCalled.get(), is(false)); //fails
    }
```

I did some digging and I think this is happening due to a bug in SafeCompletableSubscriber, where it calls  RxJavaHooks.onError(e) before it attempts to pass the error in to the wrapped subscribers onError method. ie

```
        RxJavaHooks.onError(e);
        if (done) {
            return;
        }
        done = true;
        try {
            actual.onError(e);
        } catch (Throwable ex) {
            Exceptions.throwIfFatal(ex);

            throw new OnErrorFailedException(new CompositeException(e, ex));
        }
    }
```

Looking at SafeSubscriber, I think the correct behaviour is to only have the call to ` RxJavaHooks.onError(e);` happen inside the catch block, as no exception has leaked at the point where it is currently being called.

DeepSpawn:
Move the call to RxJavaHooks into the catch block, so we only report and it if the underlying subscriber throws in its onError method.

This brings its behaviour in line with that of SafeSubscriber

### DIFF
--- a/src/main/java/rx/observers/SafeCompletableSubscriber.java
+++ b/src/main/java/rx/observers/SafeCompletableSubscriber.java
@@ -54,7 +54,6 @@ public final class SafeCompletableSubscriber implements CompletableSubscriber, S
 
     @Override
     public void onError(Throwable e) {
-        RxJavaHooks.onError(e);
         if (done) {
             return;
         }
@@ -63,6 +62,7 @@ public final class SafeCompletableSubscriber implements CompletableSubscriber, S
             actual.onError(e);
         } catch (Throwable ex) {
             Exceptions.throwIfFatal(ex);
+            RxJavaHooks.onError(e);
 
             throw new OnErrorFailedException(new CompositeException(e, ex));
         }
--- a/src/main/java/rx/observers/SafeCompletableSubscriber.java
+++ b/src/main/java/rx/observers/SafeCompletableSubscriber.java
@@ -55,6 +55,7 @@ public final class SafeCompletableSubscriber implements CompletableSubscriber, S
     @Override
     public void onError(Throwable e) {
         if (done) {
+            RxJavaHooks.onError(e);
             return;
         }
         done = true;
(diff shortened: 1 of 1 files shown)


DeepSpawn:
Not sure if this change warrants a unit test, but I am happy to add them if you think it is necessary

src/main/java/rx/observers/SafeCompletableSubscriber.java
```diff
@@ -54,7 +54,6 @@ public void onCompleted() {
 
     @Override
     public void onError(Throwable e) {
-        RxJavaHooks.onError(e);
         if (done) {
             return;
```
akarnokd: You should have moved it here.

akarnokd (changes requested):

src/main/java/rx/observers/SafeCompletableSubscriber.java
```diff
@@ -63,6 +62,7 @@ public void onError(Throwable e) {
             actual.onError(e);
         } catch (Throwable ex) {
             Exceptions.throwIfFatal(ex);
+            RxJavaHooks.onError(e);
```
DeepSpawn: @akarnokd  so this should be in if(done) only?

akarnokd: Yes.

vanniktech (approved):

akarnokd (approved):

PR #5478, opened 2017-07-10
Fixed  doc for F.reduce(seed, reducer), F.reduceWith(seedSupplier, reducer) and O.reduce(seed, reducer)

ganapathi004:
Corrected the doc for Issue No: #5476


akarnokd:
Thanks, I suspected there was more but GitHub is not able to display those files properly.

Issue #5480, opened 2017-07-11
Suggest Maybe#toSingle in Maybe#defaultIfEmpty documentation


alexeyr:
Looking at http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Maybe.html#defaultIfEmpty%28T%29, I was surprised to find it didn't return `Single`. Searching in closed issues found the answer: https://github.com/ReactiveX/RxJava/issues/5417. Perhaps documentation could be improved by explicitly mentioning `toSingle` as an alternative?

akarnokd:
PR welcome.

akarnokd:
Closing via #5486.

Issue #5417, opened 2017-06-16
2.x defaultIfEmpty for Maybe


uzzu:
Hi. I'm glad that I started using 2.x and be able to use Maybe.
I started using it, and wanted to ask a question.

Now `Maybe#defaultIfEmpty` returns `Maybe`, but because this operator guarantees to emit data, I think should be `Maybe#defaultIfEmpty` returns `Single`.

I would like to know if there is a reason to return `Maybe`.

akarnokd:
It was probably overlooked but we can't change the type now.

vanniktech:
`defaultIfEmptySingle` that returns Single?

akarnokd:
There is the `toSingle(T defaultValue)`.

akarnokd:
Also `defaultIfEmpty` is based on `switchIfEmpty` thus it returns `Maybe`.

uzzu:
OK. :+1:
I write unit tests and it seems that there is no problem with substituting `toSingle(T defaultValue)`, so I will use that.

[IMO] I think that `defaultIfEmpty` is more readable than `toSingle` , so I'd like to return `Single` in 3.x or future versions.

Thank you for answering my question 😄

PR #5485, opened 2017-07-12
Update Maybe#defaultIfEmpty javadoc (#5480)

alexeyr:
Adds a link to Javadoc for `Maybe#defaultIfEmpty` method which has a surprising return type (#5480, #5417).


akarnokd:
Forgot about the javadoc wording guards. I posted #5486 that includes your sentences and an updated validator.

PR #5494, opened 2017-07-17
FlowableWithLatestFrom forgets request

matgabriel:
The following scenario did not work with FlowableWithLatestFrom:
Flowable<?> result = source.combineWithLatest(other, someCombiner)
- client subscribe to the result stream
- source stream emits
- other stream emits
- source stream emits

expected result: result stream emits the combined event
actual result: result stream does not emit anything

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromTest.java
@@ -13,6 +13,7 @@
 
 package io.reactivex.internal.operators.flowable;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.*;
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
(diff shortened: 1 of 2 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromTest.java
```diff
@@ -261,7 +264,10 @@ public void testNoDownstreamUnsubscribe() {
 
     @Test
     public void testBackpressure() {
-        Flowable<Integer> source = Flowable.range(1, 10);
+        TestScheduler testScheduler = new TestScheduler();
```
akarnokd: Please don't change existing tests but add new ones instead.

akarnokd:
Nice find and I'm a bit baffled as [FlowableWithLatestFromMany](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromMany.java#L161) does not have this flaw...

akarnokd (changes requested):

src/test/java/io/reactivex/internal/operators/flowable/FlowableWithLatestFromTest.java
matgabriel: ok, I have added a test, but I still had to modify testBackpressure because it was relying on the fact that the source would not get more request if the other stream does not emit data.

akarnokd (approved):

PR #5511, opened 2017-07-21
Remove unnecessary generic type parameters for delaySubscription methods in Single

plackemacher:
I was running into an issue with using the `delaySubscription(long time, TimeUnit unit)` while testing a `Single` and I saw that the `U` generic type parameter was there unnecessarily. I believe this is only a Kotlin specific issue as I tried playing around in Java and there was no inference error upon compilation. To get it to compile for Kotlin, I would need to specify the generic type parameter and it didn't matter what I set it to as `Any` was a valid option: `.delaySubscription<Any>(5, TimeUnit.SECONDS)`. I also went through `Flowable`, `Observable`, and `Maybe` and saw that the generics were missing for the time-specific `delaySubscription` methods so I believe they should be removed from `Single` as well to help keep things consistent and remove this compilation error for future Kotlin users.

kotlin_error

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1752,7 +1752,6 @@ public abstract class Single<T> implements SingleSource<T> {
      * <dd>{@code delaySubscription} does by default subscribe to the current Single
      * on the {@code computation} {@link Scheduler} after the delay.</dd>
      * </dl>
-     * @param <U> the element type of the other source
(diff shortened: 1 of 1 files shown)


akarnokd:
Thanks. The [JLS](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.13) seems to consider such change as binary compatible.

/cc @JakeWharton

JakeWharton (approved):

Issue #5521, opened 2017-07-29
Feature request: onErrorResumeNextFilter and onErrorReturnFilter


artshell:
It would be nice to have two operators like this.

`Observable<T> onErrorResumeNextFilter(Predicate<? super Throwable>,  final ObservableSource<? extends T> next)`

`Observable<T> onErrorReturnFilter(Predicate<? super Throwable>, final T item)`

Some special Throwable able to obtains next Observable source, Other throwable be passed to downstream by onError()

akarnokd:
You can do both fix the existing `onErrorResumeNext` operators and with the help of transformers:

```java
public static <T> ObservableTransformer<T, T> onErrorResumeNextFilter(
        Predicate<? super Throwable> predicate, final ObservableSource<? extends T> next) {
    return o -> o.onErrorResumeNext(e -> {
       if (predicate.test(e)) {
           return next;
       }
       return Observable.error(e);
    });
}

public static <T> ObservableTransformer<T, T> onErrorReturnFilter(
        Predicate<? super Throwable> predicate, final T item) {
    return onErrorResumeNextFilter(predicate, Observable.just(item));
}
```

artshell:
@akarnokd great an idea，thanks

artshell:
oh, incompatible types
`return predicate.test(thr) ? next : Observable.error(thr);`

```java
public <T> ObservableTransformer<T, T> onErrorResumeNextFilter(final Predicate<? super Throwable> predicate, final ObservableSource<? extends T> next) {
        return new ObservableTransformer<T, T>() {
            @Override
            public ObservableSource<T> apply(@NonNull Observable<T> upstream) {
                return upstream.onErrorResumeNext(new Function<Throwable, ObservableSource<? extends T>>() {
                    @Override
                    public ObservableSource<? extends T> apply(@NonNull Throwable thr) throws Exception {
                        return predicate.test(thr) ? next : Observable.error(thr);
                    }
                });
            }
        };
    }
```
@akarnokd

akarnokd:
Please look at how I wrote the code. Ternaries often won't compile in varions java versions due to limited or inconsistent type inference.

artshell:
@akarnokd thanks, correct compile

PR #5527, opened 2017-07-31
2.x: add onErrorResumeNextFilter() and OnErrorReturnFilter() operators

artshell:
@akarnokd 

I try to add these two operators, but i not so good at english.

Please check it, test, add doc.

Thanks


akarnokd:
Thanks but there is no need for these operators because the function can be achieved with a straightforward transformer function:

https://github.com/ReactiveX/RxJava/issues/5521#issuecomment-318809970

Besides, this PR has problems: no documentation, no unit tests and it fails quality safeguard(s).

artem-zinnatullin:
I agree, behavior seems too specific for core API, sorry @artshell!

artshell:
@akarnokd, @artem-zinnatullin 
okay, no matter

PR #5529, opened 2017-07-31
Add assertValueAt(int, value) to TestObserver

NickFirmani:
I found myself frequently writing tests that want to check that an observable emitted a certain value second or third, but I don't care what the first or other values one, so `assertValues(Object...))` is more brittle.

This PR implements an `assertValueAt(index, value)` method, similar to `assertValue(value)`. Tests are also added.

### DIFF
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -17,6 +17,7 @@ import java.util.*;
 import java.util.concurrent.*;
 
 import io.reactivex.Notification;
+import io.reactivex.annotations.Experimental;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.CompositeException;
 import io.reactivex.functions.Predicate;
@@ -335,7 +336,7 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
 
     /**
      * Assert that this TestObserver/TestSubscriber did not receive an onNext value which is equal to
-     * the given value with respect to Objects.equals.
+     * the given value with respect to null-safe Object.equals.
      *
      * <p>History: 2.0.5 - experimental
      * @param value the value to expect not being received
@@ -403,19 +404,21 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
 
     /**
      * Asserts that this TestObserver/TestSubscriber received an onNext value at the given index
-     * which is equal to the given value with respect to Objects.equals.
+     * which is equal to the given value with respect to null-safe Object.equals.
      * @param index the position to assert on
      * @param value the value to expect
      * @return this
+     * @since 2.1.3 - experimental
      */
     @SuppressWarnings("unchecked")
+    @Experimental
     public final U assertValueAt(int index, T value) {
         int s = values.size();
         if (s == 0) {
             throw fail("No values");
         }
 
-        if (index >= values.size()) {
+        if (index >= s) {
             throw fail("Invalid index: " + index);
         }
 
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -403,6 +403,31 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
 
     /**
      * Asserts that this TestObserver/TestSubscriber received an onNext value at the given index
+     * which is equal to the given value with respect to Objects.equals.
+     * @param index the position to assert on
+     * @param value the value to expect
+     * @return this
+     */
+    @SuppressWarnings("unchecked")
+    public final U assertValueAt(int index, T value) {
+        int s = values.size();
+        if (s == 0) {
+            throw fail("No values");
+        }
+
+        if (index >= values.size()) {
+            throw fail("Invalid index: " + index);
+        }
+
+        T v = values.get(index);
+        if (!ObjectHelper.equals(value, v)) {
+            throw fail("Expected: " + valueAndClass(value) + ", Actual: " + valueAndClass(v));
+        }
+        return (U)this;
+    }
+
+    /**
+     * Asserts that this TestObserver/TestSubscriber received an onNext value at the given index
      * for the provided predicate returns true.
      * @param index the position to assert on
      * @param valuePredicate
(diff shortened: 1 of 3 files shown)


akarnokd:
There is an `assertValueAt(int, Predicate)` just below your changes, what's wrong with that? Also there could be an overload problem because of a `T` - `Predicate` conflict in Java 8+.

NickFirmani:
I personally make heavy use of the `assertValue(T)` method, BaseConsumer#324, which also has a similar paired method `assertValue(Predicate<T>)`. Using that method keeps my tests cleaner, when I know the exact value I expect. It seems pretty verbose to me to implement a `Predicate` when the body will just be `value.equals(expected)`.

If this causes an overload problem, we'll have the same problem with `assertValue(T)` and `assertValue(Predicate<T>)` as well

src/test/java/io/reactivex/observers/TestObserverTest.java
```diff
@@ -1383,6 +1383,48 @@ public void assertValueAtInvalidIndex() {
...
+
+    @Test
+    public void assertValueAtIndexNoMatch() {
+        TestObserver<Integer> ts = new TestObserver<Integer>();
+
+        Observable.just(1, 2, 3).subscribe(ts);
+
+        thrown.expect(AssertionError.class);
+        thrown.expectMessage("Expected: 2 (class: Integer), Actual: 3 (class: Integer) (latch = 0, values = 3, errors = 0, completions = 1)");
+        ts.assertValueAt(2, 2);
```
artem-zinnatullin: Can you please compare something else than integers (String will be ok)? 

Silly code change in assert method can break its logic but pass this test because both index and value are `2` :)

artem-zinnatullin (changes requested):
LGTM, I would use that!

vanniktech:
We need to find some kind of policy for adding new API methods or not.

I'd say let's not get this in. Write a Kotlin extension functions 🙃 or use the `assertValueAt` with a Predicate and then define a test class that implements the interface and has a static factory method `Predicate<T> just(T)`.

If this gets in I also would like to have `mergeWith`, `concatWith`, `startWith` variants for `Single`, `Completable` and `Maybe` on `Observable` & `Flowable`.

artem-zinnatullin:
I'm pretty sure this method will be much more popular than `assertValueAt(int, Predicate)` but that's just my guess

Mr ~~Negotiator~~ @vanniktech 😸 

For simple equals comparisons (which I believe are 95+% of asserts on values) this variant is definitely better than `assertValueAt(int, Predicate)`:

- It has nicer message and prints both expected and actual values including their classes (sometimes that saves tons of time)
- It requires less symbols to type `assertValueAt(3, expected)` even comparing with Kotlin `assertValueAt(3) { it == expected }`

akarnokd:
Independent additions are individually evaluated based on the merits, value, quality and maintenance requirements they pose.

This PR is self contained and adds a convenience pair to an existing method. Plus, once merged in, there will be no need for further development on it.

In contrast, all those `mergeWith(Single)` and co are just shorthands for `mergeWith(Single.toObservable())`s and alike and the premise of "optimize later" actually adds burden to the maintainer(s) of RxJava.

vanniktech:
Alright, you got me convinced now. If those shorthands were to be implemented one by one and also in an optimized way, would you consider PRs for that?

akarnokd:
Yes.

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -403,6 +403,31 @@ public final U assertNever(Predicate<? super T> valuePredicate) {
 
     /**
      * Asserts that this TestObserver/TestSubscriber received an onNext value at the given index
+     * which is equal to the given value with respect to Objects.equals.
```
akarnokd: `Objects` is a Java 8 component but `ObjectHelper` is internal, thus maybe let's say "with respect to null-safe Object.equals()".

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -403,6 +403,31 @@ public final U assertNever(Predicate<? super T> valuePredicate) {
...
+     * @param index the position to assert on
+     * @param value the value to expect
+     * @return this
+     */
```
akarnokd: `@since 2.1.3 - experimental`

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -403,6 +403,31 @@ public final U assertNever(Predicate<? super T> valuePredicate) {
...
+    @SuppressWarnings("unchecked")
+    public final U assertValueAt(int index, T value) {
```
akarnokd: `@Experimental`

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -403,6 +403,31 @@ public final U assertNever(Predicate<? super T> valuePredicate) {
...
+        int s = values.size();
+        if (s == 0) {
+            throw fail("No values");
+        }
+
+        if (index >= values.size()) {
```
akarnokd: Use `s` instead of `values.size()` here.

vanniktech (approved):

akarnokd (changes requested):

NickFirmani:
Thanks for the reviews!

src/test/java/io/reactivex/observers/TestObserverTest.java
NickFirmani: Good call.

src/main/java/io/reactivex/observers/BaseTestConsumer.java
NickFirmani: Fixed the other javadoc I copied that from too :)

src/main/java/io/reactivex/observers/BaseTestConsumer.java
NickFirmani: Nice catch.

akarnokd (approved):

NickFirmani:
CI is failing due to 
`ERROR: JAVA_HOME is set to an invalid directory: /usr/lib/jvm/java-7-oracle`

akarnokd:
There has been an image change on Travis CI and JDK 7 is no longer supported. See #5531.

akarnokd:
Could you rebase the PR so the build settings can take effect?

NickFirmani:
Thanks!

PR #5530, opened 2017-08-01
(refactor) some opportunities to use lambda functions

desmorto:

  - [ ] Please give a description about what and why you are contributing, even if it's trivial.


JakeWharton:
This library targets Java 6 so this isn't feasible

ghost (commented):
Nice, but as long as `RxJava` would target Java 6 (this primarily concerns the Android devices with system older than 4.2) this PR wouldn't be usable.

akarnokd:
RxJava 2 has to stay Java 6 compatible.

desmorto:
ok, thx for the feedback :+1:

PR #5531, opened 2017-08-01
2.x: Try fixing Travis CI lack of java

akarnokd:
Travis CI has replaced the previous Precise image with the Trusty image as part of an ongoing rollout. Unfortunately, the Trusty image fails for `oraclejdk7` because probably it is no longer supported (although I couldn't find anything about it).

This PR tries to identify what's available by trying a matrix of build setups.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: java
 jdk:
 - oraclejdk7
+- openjdk7
+- oraclejdk8
+- openjdk8
 
 # force upgrade Java8 as per https://github.com/travis-ci/travis-ci/issues/4042 (fixes compilation issue)
 #addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: java
+# trusty doesn't support jdk 7 apparently
+dist: precise 
 jdk:
 - oraclejdk7
-- openjdk7
-- oraclejdk8
-- openjdk8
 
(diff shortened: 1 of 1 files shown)


akarnokd:
Looks like Trusty only supports Java 8 builds:

image

There is also a test failure of `MaybeFromRunnableTest.noErrorLoss` due to being sensitive to delays.

Forcing `dist: precise` works.

JakeWharton:
Why not move the build to JDK 8?

akarnokd:
It is an option; there was already a switch to oraclejdk7 back when the JDK 6 support dropped.

Should we switch to `oraclejdk8` and run on `trusty`?

JakeWharton:
That's what I use on all our projects which target Java 6 compatibility. We
use JDK 8 and then Animal Sniffer to ensure no Java 8 APIs are referenced.

It'll be nice when Java 9 comes out and its "-release 6" flag can be used
to accomplish this.

artem-zinnatullin:
I'd propose switching build to Docker where we can better control what environment is actually in the build and test against multiple versions of JDK properly (if needed). 

Also with Docker we'll be able to reproduce CI builds very precisely (except hardware limits).

akarnokd:
That's too risky and may not work due to the encrypted credentials we have here for maven upload.

artem-zinnatullin:
I don't see problems with that, both files and environment variables can be easily passed to Docker container so that should solve maven upload.

And I actually think that keeping environment under Travis control (they update things) is more dangerous.

Issue #5341, opened 2017-05-14
Proposal: Switch to a provided JSR305 artifact for nullability + package annotations


ZacSweers:
I'd like to propose switching away from the current custom `@Nullable` and `@NonNull` annotations and using a combination of a provided JSR305 dependency (that is, only `compileOnly` and not actually packaged, but still visible to static analysis tools).

This would provide two main benefits:
- Use of a standard `@Nullable` annotation (Guava uses the same one). Currently anyone that uses other annotations gets burdened with fully-qualified copies of RxJava's due to IDE autocomplete, and it's a nuisance to clean it up.
- Use of package-level annotations to make the default for everything `@NonNull`. You could use `@ParametersAreNonnullByDefault` for just parameters, or also have a separate artifact with a custom `@EverythingIsNonNullByDefault` annotation like [here](https://github.com/uber/AutoDispose/pull/70/files#diff-216f7cb134879b63987893591e646185) that would also cover fields and return types. This would obviate the need for using `@NonNull` entirely, and is respected by most conventional static analysis tools.
- Potential improvements for interoperability with (Rx)Kotlin - https://youtrack.jetbrains.com/issue/KT-10942

I know RxJava prefers to have no dependencies, but I think that having this as a provided dependency is a good middle ground for the value it provides.

Other prior reading:
  - https://medium.com/square-corner-blog/rolling-out-nullable-42dd823fbd89
  - https://medium.com/@swankjesse/non-null-is-the-default-58ffc0bb9111
  - Some other discussion and justification: https://github.com/square/okio/pull/307

vanniktech:
I'd also prefer if RxJava would use the JSR305 `@NonNull` & `@Nullable`, especially with `@ParametersAreNonnullByDefault`. Although I think RxJavas `@NonNull` and `@Nullable` are now considered stable with the 2.1.0 release and hence not possible to remove if we'd stick to the plan, right?

ZacSweers:
stable, but not a breaking API change technically (from a compile perspective) since they're just annotations and the new annotations would be a superset of their functionality. They could be left in if the idea was to allow other RxJava-dependent libraries to use them if they want, while RxJava itself could use JSR305?

akarnokd:
If you specify `@ParametersAreNonnullByDefault`, how do you override it for `BiConsumer` [to be not defined](https://github.com/ReactiveX/RxJava/issues/5216)?

[...]

PR #5538, opened 2017-08-07
Switch to JSR305 nullability annotations.

artem-zinnatullin:
This PR implements proposal #5341.

Since `@ParametersAreNonnullByDefault` only covers function parameters, I've left `@Nonnull` for return results in existing usage places. 

@hzsweers suggested creating `@EverythingIsNonNullByDefault` but I'm not sure how tooling will recognise it.

@akarnokd we can also change RxJava's `@CheckReturnValue` with `javax.annotation.@CheckReturnValue` in a separate PR if this one gets merged. 

P.S. I've tried to minimize imports regrouping, but some files were written in different code styles regarding imports, so you might find few minor regroupings, sorry.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,10 +21,26 @@ apply plugin: 'nebula.rxjava-project'
 sourceCompatibility = JavaVersion.VERSION_1_6
 targetCompatibility = JavaVersion.VERSION_1_6
 
+configurations {
+    provided
+}
+
+sourceSets {
+    main {
+        compileClasspath += configurations.provided
+        runtimeClasspath += configurations.provided
+    }
+    test {
+        compileClasspath += configurations.provided
+        runtimeClasspath += configurations.provided
+    }
+}
+
 dependencies {
     signature 'org.codehaus.mojo.signature:java16:1.1@signature'
 
     compile 'org.reactivestreams:reactive-streams:1.0.0'
+    provided 'com.google.code.findbugs:jsr305:3.0.2'
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.1.0'
--- a/build.gradle
+++ b/build.gradle
@@ -21,27 +21,14 @@ apply plugin: 'nebula.rxjava-project'
 sourceCompatibility = JavaVersion.VERSION_1_6
 targetCompatibility = JavaVersion.VERSION_1_6
 
-configurations {
-    provided
-}
-
-sourceSets {
-    main {
-        compileClasspath += configurations.provided
-        runtimeClasspath += configurations.provided
-    }
-    test {
-        compileClasspath += configurations.provided
-        runtimeClasspath += configurations.provided
-    }
-}
-
 dependencies {
     signature 'org.codehaus.mojo.signature:java16:1.1@signature'
 
     compile 'org.reactivestreams:reactive-streams:1.0.0'
-    provided 'com.google.code.findbugs:jsr305:3.0.2'
 
+    compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
+    testCompileOnly 'com.google.code.findbugs:jsr305:3.0.2'
+    
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.1.0'
 
--- a/src/main/java/io/reactivex/functions/package-info.java
+++ b/src/main/java/io/reactivex/functions/package-info.java
@@ -18,4 +18,7 @@
  * Functional interfaces of functions and actions of arity 0 to 9 and related
  * utility classes.
  */
+@ParametersAreNonnullByDefault
 package io.reactivex.functions;
+
+import javax.annotation.ParametersAreNonnullByDefault;
--- a/src/main/java/io/reactivex/CompletableEmitter.java
+++ b/src/main/java/io/reactivex/CompletableEmitter.java
@@ -36,7 +36,7 @@ public interface CompletableEmitter {
      * Signal an exception.
      * @param t the exception, not null
      */
-    void onError(@NonNull Throwable t);
+    void onError(Throwable t);
 
     /**
      * Sets a Disposable on this emitter; any previous Disposable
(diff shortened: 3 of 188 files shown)


build.gradle
```diff
@@ -21,10 +21,26 @@ apply plugin: 'nebula.rxjava-project'
...
+        compileClasspath += configurations.provided
+        runtimeClasspath += configurations.provided
+    }
+}
+
 dependencies {
     signature 'org.codehaus.mojo.signature:java16:1.1@signature'
 
     compile 'org.reactivestreams:reactive-streams:1.0.0'
+    provided 'com.google.code.findbugs:jsr305:3.0.2'
```
JakeWharton: `compileOnly`

JakeWharton:
You can make a custom package annotation that specifies that return values are non-null by default in addition to parameters.

artem-zinnatullin:
Will it be picked up by tooling? (I wrote about this concert in description of the PR)

src/main/java/io/reactivex/functions/package-info.java
```diff
@@ -18,4 +18,7 @@
  * Functional interfaces of functions and actions of arity 0 to 9 and related
  * utility classes.
  */
+@ParametersAreNonnullByDefault
```
akarnokd, line 21: Did you update the `BiConsumer`'s `@Nullable` parameter annotation?

ZacSweers:
It will be picked up in tooling if you make it a separate provided artifact. This way the annotations are in the bytecode (tooling will see it) but not distributed as a real dependency. See https://github.com/uber/AutoDispose/pull/70 for an example

Doesn't even need to be released, but does need to be separate to have gradle automatically exclude it nicely. I'm sure you could manually finagle this somehow, but that requires gradle-fu I don't have :)

artem-zinnatullin:
Short update:

- Custom `@EverythingNonNullByDefault` worked inside RxJava itself, but  it's not recognized by external project using version of RxJava with this annotation. Hopefully, separate artifact with this annotation will solve that as pointed by @hzsweers
- @JakeWharton, unfortunately `compileOnly` and custom baked `provided` do not create entry in `pom.xml`. We'll probably need to use `compile` (maybe postprocess `pom.xml` to change scope from `compile` to `provided`): 
  - https://github.com/nebula-plugins/nebula-publishing-plugin/issues/59, 
  - https://discuss.gradle.org/t/publishing-plugin-should-respect-compileonly-configuration/22903/2 

- Looks I'll need to restructure RxJava since it uses root of the project as src module and Gradle goes crazy if I try to add more modules in that setup (still figuring out)
- Undefined nullability cases pointed by @akarnokd seems to be resolvable by using `@Nullable` according to its javadoc: 

>This annotation is useful mostly for **overriding** a {@link Nonnull} annotation. Static analysis tools should generally treat the *annotated items as though they had no annotation*, unless they are configured to minimize false negatives.

I've also tried to create custom `@NullabilityUndefined` which is basically copy of `@Nullable` but with more appropriate name, but IntelliJ does not pick it up and still shows inappropriate nullability warnings. Applying `@Nullable` removes those warnings.

I'll continue working on PR.

akarnokd:
The Nebula plugin is a bane and I don't know how it responds to multi-project Gradle. The main problem is the auto-release part because in order to upload to maven, the generated jars have to be signed and each uploaded. I don't know if the upload script will wait for closing the staging repo on Sonatype.

In my other projects, I have the keys locally and works for the multi-project v3 preview nicely (uploads all jars, then I manually close the single Sonatype staging repo with all of them). I don't know how the signing works with RxJava proper when run on travis-ci.

ZacSweers:
What about switching to a conventional mavencentral release in a similar structure to how square libraries are set up? It follows a pretty standard structure and releasing is a sinch, all you need are sonatype credentials (you can likely ask to have your account given permissions to upload for rxjava if you don't already have it). You could even set up snapshots nicely with this too.

Under the alternative project structure, it would be easy to create a multi-module gradle project

ZacSweers:
We actually just made a little utility library with some of these annotations that you could drop in - https://github.com/uber-common/javax-extras. If you use it as a provided/compileOnly dependency, then you don't need to try to host it in a separate module here

zyxist:
I have some objections against including this PR in the codebase, related to the recent JDK9 release and the arrival of Java modules.

The new module system introduces a rule that a package can be owned by exactly one module/JAR - otherwise the application does not compile. JSR-305 JAR defines a lot of annotations in `javax.annotation` package, which is also used by JDK module `java.xml.ws.annotation`, and `jsr250-api` JAR.

I have already run into issues caused by the presence of both JSR-250 and JSR-305 annotations in the module path caused by other libraries, and in the near future, the problem will become even bigger. I do not question the sense of using static analysis annotations, but from the module perspective, `jsr305` has a design flaw. Java community needs a replacement that uses a different, unique package name.

ZacSweers:
@zyxist Considering they're compile-only, does it matter? They're not on the compilation dependency classpath in consuming projects downstream

zyxist:
@hzsweers -> unfortunately this is not a compile-only issue. Package split prevents a modular application both from compiling and from running, and it pollutes all the downstream dependencies. JSR-305 annotations have `RUNTIME` retention policy, so the JAR must be present at runtime, to pass the JVM verification.

Java provides a command-line switch `--patch-module` that tells it to "include" the content of another JAR into an existing module, but you have to add it to every compilation call, unit test call, and JVM launch. Using it is problematic, because you must hardcode a full path to the JAR archive. It was intended as a temporary solution until there are problematic, but popular JAR-s in the wild, such as `jsr305`.

A couple of days ago, I spent three hours trying to compile (not to mention running) a simple "Hello world" modular application that used both Dagger and Guava. Dagger required `jsr250`, Guava required `jsr305`, both containing the same package `javax.annotation`. Eventually, I managed to do so, but the result was extremely messy, and included storing one of the JAR-s extracted from the Maven cache to a special directory, and hard-coding a path to it in the build file.

In addition, there are some legal issues about using non-standard annotations packaged as `javax.annotation`:

 * https://github.com/spotbugs/spotbugs/issues/130

akarnokd:
Thanks for working on this but the potential conflict in Java 9 modules worries me. Therefore, this type of change may be reconsidered in the v3 era.

PR #5544, opened 2017-08-09
Fix grammar error in README.md

ninearif:
grammar fix :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ APIs marked with the `@Deprecated` annotation at the class or method level will 
 
 #### io.reactivex.internal.*
 
-All code inside the `io.reactivex.internal.*` packages is considered private API and should not be relied upon at all. It can change at any time. 
+All code inside the `io.reactivex.internal.*` packages are considered private API and should not be relied upon at all. It can change at any time. 
 
 ## Full Documentation
 


README.md
```diff
@@ -199,7 +199,7 @@ APIs marked with the `@Deprecated` annotation at the class or method level will
 
 #### io.reactivex.internal.*
 
-All code inside the `io.reactivex.internal.*` packages is considered private API and should not be relied upon at all. It can change at any time. 
+All code inside the `io.reactivex.internal.*` packages are considered private API and should not be relied upon at all. It can change at any time.
```
artem-zinnatullin, line 202: "… code … *is* considered", previous version looks correct to me

JakeWharton: packages is plural

ghost (approved):

README.md
artem-zinnatullin: but "code" is not

JakeWharton: Code is both.

ghost: `All code inside the ... packages is` - @artem-zinnatullin is right ;-)

JakeWharton: Replace code with any other noun and read it. Try cars. Then try deer. Them do code again.

artem-zinnatullin: What if I replace it with something more abstract, like "meaning", "value" or "tragedy": 
"All meaning/value/tragedy … is considered" (probably needs "the" before "code" then).

Obviously I'm not a native English speaker (and David is laughing at this) so I'm out :)

JakeWharton: Yes I think a "the" would solidify it as singular. It's really hard to tell otherwise. I lean towards plural though, but I'm not sure if that's based on what's correct or what I've seen used most.

akarnokd: Grammarly: 

image

jamesgorman2: The grammar parse is (approximately using no specific notation, excluding the second phrase):

> [[All code [inside the `io.reactivex.internal.*` packages]] is [considered private API]]

_[inside the `io.reactivex.internal.*` packages]_ is a qualifier describing the location of _All code_. When we remove it we are left with:

> [[All code] is [considered private API]]

which is correct as _code_ is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun) in English.

ninearif: Oh I see it, 
Thanks @jamesgorman2

akarnokd:
Looks like it was correct all along. It wasn't me but [Ben who wrote it originally for 1.x](https://github.com/ReactiveX/RxJava/blame/d477f9c392d90960671dc883679f2ebeaed86acb/README.md#L53) years ago.

Issue #5234, opened 2017-03-27
2.x: Uncaught errors fail silently in junit tests.


ZacSweers:
I was writing a utility that involved decorating observers and catching uncaught errors from the delegate observer's `onError` handling when I noticed in testing that tests always passed. There would be log output, but tests would pass none the less.

Demo of the issue can be found [here](https://github.com/hzsweers/RxJava/blob/6bf9afcf4c6b14dc25b9b5aa766b75750239b365/src/test/java/io/reactivex/JUnitErrorTest.java), but the gist of it is that the following test passes in standard JUnit:

```java
@Test
public void blerg() {
    Observable.error(new RuntimeException("This should fail the test"))
            .subscribe();
}
```

I'm trying to understand why errors are passed directly to the handler rather than just thrown, and also curious to get thoughts on making this behavior configurable (if for no other reason than to disable in junit tests). I can't find any details about why this behavior happens in JUnit either.

I think this has worrisome implications. In our codebase alone, I dropped in an error watching [rule](https://github.com/uber/AutoDispose/blob/master/autodispose/src/test/java/com/uber/autodispose/RxErrorsRule.java) to track these and discovered over 150 tests with uncaught exceptions that are being marked as passing otherwise.

0kai:
you may use TestObserver to subscribe the result

akarnokd:
Unlike 1.x, `OnErrorNotImplementedException` is not thrown. Also please read on the [error handling](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling) section of the wiki.

ZacSweers:
I did read that, but it doesn't really explain why it deals directly with the current thread's `UncaughtExceptionHandler` rather than just throwing if there's no `onError` hook set.

@0kai I'm not trying to figure out how to test errors, rather I'm trying to make sure that uncaught exceptions in tests aren't marked as passing. At least not silently anyway

akarnokd:
2.x `Subscriber`s and `Observer`s must not throw from their `onXXX` methods. Install an error handler before the test and verify there was no undeliverable error. See how we do it in RxJava tests:

https://github.com/ReactiveX/RxJava/blob/3bfc275ca2f8d14a67305bbadb7bbc6315205250/src/test/java/io/reactivex/parallel/ParallelPeekTest.java#L55

ZacSweers:
[...]

PR #5569, opened 2017-08-26
2.x: Strawman proposal for uncaught exception configuration

ZacSweers:
This is a strawman example/proposal for configurable uncaught exception handling in RxJavaPlugins. This is in ref to #5234.

The two options are to defer to current thread's handler (the current implementation and remains the default) and an optional `throw` option that wraps the throwable into an `UncaughtRxJavaException` for developers to watch for separately (and rethrowing, which requires some sort of `RuntimeException` wrapping).

If the proposal is agreeable/after API design is in place, I can add tests.

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -30,8 +30,27 @@ import io.reactivex.parallel.ParallelFlowable;
 import io.reactivex.schedulers.Schedulers;
 /**
  * Utility class to inject handlers to certain standard RxJava operations.
+ * <p>
+ * <strong>Supported system properties ({@code System.getProperty()}):</strong>
+ * <ul>
+ * <li>{@code rx2.uncaught-behavior} (String): sets the behavior of the uncaught exception handling, default is {@link #UNCAUGHT_HANDLER}. This is {@link Experimental}.</li>
+ * </ul>
  */
 public final class RxJavaPlugins {
+
+    /**
+     * System property value for uncaught exceptions that passes the exception directly to the current thread's
+     * {@link UncaughtExceptionHandler}. This is the default behavior.
+     */
+    @Experimental
+    public static final String UNCAUGHT_HANDLER = "handler";
+
+    /**
+     * System property value for uncaught exceptions that just immediately throws the uncaught exception.
+     */
+    @Experimental
+    public static final String UNCAUGHT_THROW = "throw";
+
     @Nullable
     static volatile Consumer<? super Throwable> errorHandler;
 
@@ -122,6 +141,9 @@ public final class RxJavaPlugins {
      * computation or single scheduler will throw an IllegalStateException.
      */
     static volatile boolean failNonBlockingScheduler;
+
+    /** The name of the system property for setting the uncaught exception behavior. */
+    private static final String KEY_UNCAUGHT_BEHAVIOR = "rx2.uncaught-behavior";
 
     /**
      * Prevents changing the plugins from then on.
@@ -406,9 +428,15 @@ public final class RxJavaPlugins {
     }
 
     static void uncaught(@NonNull Throwable error) {
-        Thread currentThread = Thread.currentThread();
-        UncaughtExceptionHandler handler = currentThread.getUncaughtExceptionHandler();
-        handler.uncaughtException(currentThread, error);
+        String uncaughtBehavior = System.getProperty(KEY_UNCAUGHT_BEHAVIOR, UNCAUGHT_HANDLER);
+        if (UNCAUGHT_THROW.equals(uncaughtBehavior)) {
+            throw new UncaughtRxJavaException(error);
+        } else {
+            Thread currentThread = Thread.currentThread();
+            UncaughtExceptionHandler handler = currentThread.getUncaughtExceptionHandler();
+            handler.uncaughtException(currentThread, error);
+        }
+        // TODO Error on invalid inputs somehow?
     }
 
     /**
--- a/src/main/java/io/reactivex/plugins/UncaughtRxJavaException.java
+++ b/src/main/java/io/reactivex/plugins/UncaughtRxJavaException.java
@@ -1,0 +1,16 @@
+package io.reactivex.plugins;
+
+import io.reactivex.annotations.Experimental;
+
+/**
+ * An exception thrown when an uncaught throwable was received and the {@link RxJavaPlugins#UNCAUGHT_THROW} property was
+ * set. This guarantees the existence of an underlying cause retrievable via {@link #getCause()}.
+ */
+@Experimental
+public final class UncaughtRxJavaException extends RuntimeException {
+
+    public UncaughtRxJavaException(Throwable cause) {
+        super(cause);
+    }
+
+}


akarnokd:
There is a way without changing the plugins:

```java
public class UncaughtCrash {
    
    @Before
    public void before() {
        RxJavaPlugins.setErrorHandler(e -> {
            Thread.currentThread().setUncaughtExceptionHandler((t, f) -> {
                Thread.currentThread().setUncaughtExceptionHandler(null);
                throw (InternalError)f;
            });
            throw new InternalError(e);
         });
    }
    
    @After
    public void after() {
        RxJavaPlugins.setErrorHandler(null);
    }
    
    @Test
    public void test1() {
        RxJavaPlugins.onError(new IOException());
    }
    
    @Test
    public void test2() {
        RxJavaPlugins.onError(new IOException());
    }
}
```

This will bypass the `try-catch` around the consumer invocation in `RxJavaPlugins`.

ZacSweers:
That's not a solution for application use though, which could interfere with crash reporting tools

akarnokd:
> interfere with crash reporting tools

How? 

You can restore the previous handler:

```
    RxJavaPlugins.setErrorHandler(e -> {
        UncaughtExceptionHandler saved = Thread.currentThread().getUncaughtExceptionHandler();
        Thread.currentThread().setUncaughtExceptionHandler((t, f) -> {
            Thread.currentThread().setUncaughtExceptionHandler(saved);
            throw (InternalError)f;
        });
        throw new InternalError(e);
     });
```

or

```java
    RxJavaPlugins.setErrorHandler(e -> {
        UncaughtExceptionHandler saved = Thread.currentThread().getUncaughtExceptionHandler();
        Thread.currentThread().setUncaughtExceptionHandler((t, f) -> {
            saved.uncaughtException(t, f);
            throw (InternalError)f;
        });
        throw new InternalError(e);
     });
```

ZacSweers:
The first solution will get stuck in a cycle of try/catching before eventually bombing out in `CompositeException` due to duplicates found in the chain (but not without creating an extremely deep trace). The decorator _does_ get a chance in its catch clause, but not without the collateral above. If it adds any metadata, it is buried within hundreds of lines of stacktraces :/. This would present an immense problem for grouping in crash reporting tools as well

The second doesn't actually avoid the issue for an android environment because sending it to the saved handler first will just exit the process before the decorator has a chance to handle it in a catch clause again.

akarnokd:
Could you provide an example runnable test that demonstrates the way you want to catch these exceptions? 

> stuck in a cycle of try/catching

`UncaughtRxJavaException` would exhibit the same problem. Both need an extra case in the [`throwIfFatal`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/exceptions/Exceptions.java#L66) helper.

/cc @JakeWharton

ZacSweers:
Can do. I'll be away from my computer tomorrow but will circle back here with it Monday

ZacSweers:
Here's a demo. https://github.com/hzsweers/rxuncaughtissue

Just click the button to see, result will be a long trace like this:

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.rxuncaughtissue, PID: 5469
    java.lang.RuntimeException: io.reactivex.exceptions.UndeliverableException: java.lang.RuntimeException: io.reactivex.exceptions.CompositeException: 2 exceptions occurred. 
        at com.example.rxuncaughtissue.MainActivity$2.accept(MainActivity.java:46)
        at com.example.rxuncaughtissue.MainActivity$2.accept(MainActivity.java:32)
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:355)
        at io.reactivex.Observable.subscribe(Observable.java:10910)
        at io.reactivex.Observable.subscribe(Observable.java:10889)
        at io.reactivex.Observable.subscribe(Observable.java:10767)
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:50)
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15)
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26)
        at android.view.View.performClick(View.java:5637)
        at android.view.View$PerformClick.run(View.java:22429)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6119)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
     Caused by: io.reactivex.exceptions.UndeliverableException: java.lang.RuntimeException: io.reactivex.exceptions.CompositeException: 2 exceptions occurred. 
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:349)
        at io.reactivex.Observable.subscribe(Observable.java:10910) 
        at io.reactivex.Observable.subscribe(Observable.java:10889) 
        at io.reactivex.Observable.subscribe(Observable.java:10767) 
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:50) 
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15) 
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26) 
        at android.view.View.performClick(View.java:5637) 
        at android.view.View$PerformClick.run(View.java:22429) 
        at android.os.Handler.handleCallback(Handler.java:751) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:154) 
        at android.app.ActivityThread.main(ActivityThread.java:6119) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
     Caused by: java.lang.RuntimeException: io.reactivex.exceptions.CompositeException: 2 exceptions occurred. 
        at com.example.rxuncaughtissue.MainActivity$2.accept(MainActivity.java:46)
        at com.example.rxuncaughtissue.MainActivity$2.accept(MainActivity.java:32)
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:355)
        at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77)
        at io.reactivex.internal.disposables.EmptyDisposable.error(EmptyDisposable.java:63)
        at io.reactivex.internal.operators.observable.ObservableError.subscribeActual(ObservableError.java:37)
        at io.reactivex.Observable.subscribe(Observable.java:10903)
        at io.reactivex.Observable.subscribe(Observable.java:10889) 
        at io.reactivex.Observable.subscribe(Observable.java:10767) 
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:50) 
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15) 
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26) 
        at android.view.View.performClick(View.java:5637) 
        at android.view.View$PerformClick.run(View.java:22429) 
        at android.os.Handler.handleCallback(Handler.java:751) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:154) 
        at android.app.ActivityThread.main(ActivityThread.java:6119) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
     Caused by: io.reactivex.exceptions.CompositeException: 2 exceptions occurred. 
        at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77) 
        at io.reactivex.internal.disposables.EmptyDisposable.error(EmptyDisposable.java:63) 
        at io.reactivex.internal.operators.observable.ObservableError.subscribeActual(ObservableError.java:37) 
        at io.reactivex.Observable.subscribe(Observable.java:10903) 
        at io.reactivex.Observable.subscribe(Observable.java:10889) 
        at io.reactivex.Observable.subscribe(Observable.java:10767) 
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:50) 
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15) 
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26) 
        at android.view.View.performClick(View.java:5637) 
        at android.view.View$PerformClick.run(View.java:22429) 
        at android.os.Handler.handleCallback(Handler.java:751) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:154) 
        at android.app.ActivityThread.main(ActivityThread.java:6119) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
     Caused by: io.reactivex.exceptions.CompositeException$CompositeExceptionCausalChain: Chain of Causes for CompositeException In Order Received =>
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:359)
        at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77) 
        at io.reactivex.internal.disposables.EmptyDisposable.error(EmptyDisposable.java:63) 
        at io.reactivex.internal.operators.observable.ObservableError.subscribeActual(ObservableError.java:37) 
        at io.reactivex.Observable.subscribe(Observable.java:10903) 
        at io.reactivex.Observable.subscribe(Observable.java:10889) 
        at io.reactivex.Observable.subscribe(Observable.java:10767) 
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:50) 
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15) 
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26) 
        at android.view.View.performClick(View.java:5637) 
        at android.view.View$PerformClick.run(View.java:22429) 
        at android.os.Handler.handleCallback(Handler.java:751) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:154) 
        at android.app.ActivityThread.main(ActivityThread.java:6119) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
     Caused by: java.lang.RuntimeException: This gets buried
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:49)
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15) 
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26) 
        at android.view.View.performClick(View.java:5637) 
        at android.view.View$PerformClick.run(View.java:22429) 
        at android.os.Handler.handleCallback(Handler.java:751) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:154) 
        at android.app.ActivityThread.main(ActivityThread.java:6119) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
     Caused by: java.lang.RuntimeException: Duplicate found in causal chain so cropping to prevent loop ...
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:359)
        at io.reactivex.internal.observers.LambdaObserver.onError(LambdaObserver.java:77)
        at io.reactivex.internal.disposables.EmptyDisposable.error(EmptyDisposable.java:63)
        at io.reactivex.internal.operators.observable.ObservableError.subscribeActual(ObservableError.java:37)
        at io.reactivex.Observable.subscribe(Observable.java:10903)
        at io.reactivex.Observable.subscribe(Observable.java:10889)
        at io.reactivex.Observable.subscribe(Observable.java:10767)
        at com.example.rxuncaughtissue.MainActivity.simulateError(MainActivity.java:50)
        at com.example.rxuncaughtissue.MainActivity.access$000(MainActivity.java:15) 
        at com.example.rxuncaughtissue.MainActivity$1.onClick(MainActivity.java:26) 
        at android.view.View.performClick(View.java:5637) 
        at android.view.View$PerformClick.run(View.java:22429) 
        at android.os.Handler.handleCallback(Handler.java:751) 
        at android.os.Handler.dispatchMessage(Handler.java:95) 
        at android.os.Looper.loop(Looper.java:154) 
        at android.app.ActivityThread.main(ActivityThread.java:6119) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
```

akarnokd:
I see. Unless there is a free pass for the exception thrown, this PR would end up with the same complicated cause chain.

So if I understand correctly, you want the following two:

1. synchronous sequences should crash and a non-reactive try-catch should be able to grab the exception,
2. shim `Observable`s installed via assembly hooks should be able to try-catch around an `onError` call to a downstream and grab an exception bouncing back from the onError handler.

The first case, when inside an unit test, could be simulated via the `TestHelper.trackPluginErrors` previously mentioned. 

Otherwise and in option 2, the question is what would you even do about that undeliverable error?

Either way, you'd still have to get a free pass for that type of crash by modifying [`throwIfFatal`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/exceptions/Exceptions.java#L66) helper.

And finally, even if such behavior override would be added, I'd go for a static volatile flag instead of a system parameter so it can be enabled and disabled when testing the feature itself.

ZacSweers:
The first case can definitely be handled in tests, but I'm basically trying to write a tool to help with crash reporting by decorating the observer with this observer that, in the event of the delegate observer throwing an `OnErrorNotImplementedException`, would throw its own exception with a bunch of metadata up before crashing.

What modifications would need to go into throwIfFatal? Even if it's something as simple as letting OnErrorNotImplemented bubble up, it'd go a long way to detecting unhandled errors and being able to tie them to the point of subscription. 

Volatile flag sounds fine to me

akarnokd:
> What modifications would need to go into throwIfFatal?

Another `if` with the `RuntimeException` or `Error` to be rethrown.

> letting `OnErrorNotImplemented` bubble up

Not a good idea as it is an established exception type and unexpected new behavior of it.

> in the event of the delegate observer throwing an `OnErrorNotImplementedException`

I'd think you are using the wrong hook then; `setObservableOnSubscribe` is called when subscribing where you could analyze the `LambdaObserver` for the default onError handler (it's a constant). For that, though, the `LambdaObserver` needs modification, perhaps by adding a public `isDefaultErrorConsumer`. This, however, would rely on otherwise discouraged access into the internal packages of RxJava.

Other option I see would be a new hook `RxJavaPlugins.setOnErrorNotImplementedConsumer(Function<Consumer<Throwable>, Consumer<Throwable>>)` where you could basically supply an `onError` handler to the `LambdaObserver` when `subscribe()` and `subscribe(Consumer<T>)` is invoked.

ZacSweers:
I think a key part of this is that it probably shouldn't be tied to `LambdaObserver`, since `OnErrorNotImplementedException` is public and developers could set their own observers with implementations that leverage it (and a throwing `onError` is the real case trying to be handled here). Custom observers are what we actually use as well, and what we would watch for in an onSubscribe hook to know when to decorate (as we wouldn't want to do it on all of rxjava's internal subscribes).

Not really sure how to proceed on this though :/

ZacSweers:
Actually, I think just knowing state about LambdaObserver would be suitable. It's probably not an unsafe assumption that if someone is actually implementing onError, they are responsible for their own implementation.

ZacSweers:
Any further thoughts on exposing LambdaObserver state?

akarnokd:
Modifying `LambdaObserver` (and all the other types) sounds like an acceptable tradeoff.

ZacSweers:
Cool. I can start working on a PR. With regards to it being internal currently, would you want to make it public with `isDefaultErrorConsumer` then? Or maybe a higher level interface that it just implements while keeping the implementation internal?

akarnokd:
Implementing an interface simplifies the detection of the feature but otherwise will make a method public on `LambdaObserver` anyway. 

I'd say let's add a `HasDefaultErrorConsumer` interface into the `io.reactivex.observers` package.

ZacSweers:
Sounds good, I'll start on that

ZacSweers:
Continued in #5590

PR #5577, opened 2017-08-30
Remove unnecessary code

ggikko:


akarnokd:
Please leave it as is. On rare occasions I need that in the given form for diagnosing thread leaks. When commented out, it has no performance penalty.

PR #5579, opened 2017-08-30
2.x: add Maybe null test cases

ggikko:
Add Maybe null test cases.

I will add more test cases for Maybe by studying Maybe. 

This code is similar with Observable, Single,, test case.


akarnokd:
Thanks, but null and argument checks on operators is covered by an automatic test: https://github.com/ReactiveX/RxJava/blob/2.x/src/test/java/io/reactivex/ParamValidationCheckerTest.java

Also please consider opening an issue about what you plan to change so you don't waste time.

ggikko:
Got it. Thanks ur comment.

PR #5581, opened 2017-09-01
2.x: Added a .filter(value) operator shorthand

Statuscope:
NOTE: This is my first ever contribution to anything, so i'm sure there will be revising involved here.

There are times when you simply want to filter events by a single value, and implementing a predicate seems ugly.

I think you should be able to do thinks like ```.filter(false)``` or ```.filter(enum.SOME_EVENT)```.

Let me know what you all think!


akarnokd:
Hi and welcome to the project. 

I wish you had made an [issue](https://github.com/ReactiveX/RxJava/issues) first because the desired use case is considered very rare and the standard solution, namely a lambda such `v -> v.equals(constant)`, is hardly any inconvenience in my experience. 

In addition, new features are vetted against a high standard of quality, expected to be (reasonably) value adding and and contain implementations for all relevant reactive types of RxJava.

Therefore, I'm sorry but the proposed idea in this PR does not meet these requirements. Since a transformer is similarly at least as complicated as a lambda, the alternative for you to would be to use Kotlin and it's extension method feature.

PR #5582, opened 2017-09-02
implement Maybe.switchIfEmpty(Single)

### LINKED ISSUE #4544: Maybe.switchIfEmpty(Single) method
It would be nice to see a method like this:

public class Maybe<T> {
  public Single<T> switchIfEmpty(Single<T> single);
}

Semantically, if you switchIfEmpty to a Single from a Maybe, then it becomes a Single, as you are guaranteed either an error or a value.  It would just make it less verbose than doing this:

maybe.switchIfEmpty(single.toMaybe()).toSingle()

bmaslakov:
Adds `Maybe.switchIfEmpty(Single)`, fixes #4544.

### DIFF
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -2213,7 +2213,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Maybe<T> defaultIfEmpty(T defaultItem) {
         ObjectHelper.requireNonNull(defaultItem, "item is null");
-        return switchIfEmpty(Single.just(defaultItem));
+        return switchIfEmpty(just(defaultItem));
     }
 
 
@@ -3817,11 +3817,11 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Maybe<T> switchIfEmpty(MaybeSource<? extends T> other) {
         ObjectHelper.requireNonNull(other, "other is null");
-        return RxJavaPlugins.onAssembly(new MaybeSwitchIfEmptyToMaybe<T>(this, other));
+        return RxJavaPlugins.onAssembly(new MaybeSwitchIfEmpty<T>(this, other));
     }
 
     /**
-     * Returns a Maybe that emits the items emitted by the source Maybe or the item of an alternate
+     * Returns a Single that emits the items emitted by the source Maybe or the item of an alternate
      * SingleSource if the current Maybe is empty.
      * <p>
      * <img width="640" height="445" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/switchifempty.m.png" alt="">
@@ -3833,14 +3833,14 @@ public abstract class Maybe<T> implements MaybeSource<T> {
      *
      * @param other
      *              the alternate SingleSource to subscribe to if the main does not emit any items
-     * @return  a Maybe that emits the items emitted by the source Maybe or the item of an
+     * @return  a Single that emits the items emitted by the source Maybe or the item of an
      *          alternate SingleSource if the source Maybe is empty.
      */
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Maybe<T> switchIfEmpty(SingleSource<? extends T> other) {
+    public final Single<T> switchIfEmpty(SingleSource<? extends T> other) {
         ObjectHelper.requireNonNull(other, "other is null");
-        return RxJavaPlugins.onAssembly(new MaybeSwitchIfEmptyToSingle<T>(this, other));
+        return RxJavaPlugins.onAssembly(new MaybeSwitchIfEmptySingle<T>(this, other));
     }
 
     /**
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -3835,7 +3835,9 @@ public abstract class Maybe<T> implements MaybeSource<T> {
      *              the alternate SingleSource to subscribe to if the main does not emit any items
      * @return  a Single that emits the items emitted by the source Maybe or the item of an
      *          alternate SingleSource if the source Maybe is empty.
+     * @since 2.1.4 - experimental
      */
+    @Experimental
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Single<T> switchIfEmpty(SingleSource<? extends T> other) {
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -2213,7 +2213,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Maybe<T> defaultIfEmpty(T defaultItem) {
         ObjectHelper.requireNonNull(defaultItem, "item is null");
-        return switchIfEmpty(just(defaultItem));
+        return switchIfEmpty(Single.just(defaultItem));
     }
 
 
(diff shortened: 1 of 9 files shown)


src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptyToMaybeTest.java
```diff
@@ -37,16 +37,6 @@ public void empty() {
     }
 
     @Test
-    public void defaultIfEmptyNonEmpty() {
```
akarnokd: Why were these removed?

src/main/java/io/reactivex/Maybe.java
```diff
@@ -3817,7 +3817,30 @@ public final void subscribe(MaybeObserver<? super T> observer) {
...
+     * </dl>
+     *
+     * @param other
+     *              the alternate SingleSource to subscribe to if the main does not emit any items
+     * @return  a Maybe that emits the items emitted by the source Maybe or the item of an
+     *          alternate SingleSource if the source Maybe is empty.
+     */
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Maybe<T> switchIfEmpty(SingleSource<? extends T> other) {
```
akarnokd: The issue mentioned `Single` as return type.

src/main/java/io/reactivex/Maybe.java
```diff
@@ -3817,7 +3817,30 @@ public final void subscribe(MaybeObserver<? super T> observer) {
...
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code switchIfEmpty} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param other
+     *              the alternate SingleSource to subscribe to if the main does not emit any items
+     * @return  a Maybe that emits the items emitted by the source Maybe or the item of an
+     *          alternate SingleSource if the source Maybe is empty.
+     */
```
akarnokd: Please add the appropriate experimental tags and annotations:

`@since 2.1.4 - experimental`

src/main/java/io/reactivex/Maybe.java
```diff
@@ -3817,7 +3817,30 @@ public final void subscribe(MaybeObserver<? super T> observer) {
...
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
```
akarnokd: Please add `@Experimental`

src/main/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptyToMaybe.java
```diff
@@ -24,11 +24,11 @@
  *
  * @param <T> the value type
  */
-public final class MaybeSwitchIfEmpty<T> extends AbstractMaybeWithUpstream<T, T> {
+public final class MaybeSwitchIfEmptyToMaybe<T> extends AbstractMaybeWithUpstream<T, T> {
```
akarnokd: I'd keep the original name, less change per PR, and name the new version `MaybeSwitchIfEmptySingle`.

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptyToMaybeTest.java
```diff
@@ -24,7 +24,7 @@
 import io.reactivex.processors.PublishProcessor;
 import io.reactivex.schedulers.Schedulers;
 
-public class MaybeSwitchIfEmptyTest {
+public class MaybeSwitchIfEmptyToMaybeTest {
```
akarnokd: Same as before, please keep the original name to this test.

src/main/java/io/reactivex/Maybe.java
```diff
@@ -2213,7 +2213,7 @@ public final T blockingGet(T defaultValue) {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Maybe<T> defaultIfEmpty(T defaultItem) {
         ObjectHelper.requireNonNull(defaultItem, "item is null");
-        return switchIfEmpty(just(defaultItem));
+        return switchIfEmpty(Single.just(defaultItem));
```
akarnokd: Why was this changed, the original `just` should still work?!

akarnokd (dismissed):

akarnokd:
I'm only 50% convinced for the need of this overload.

src/main/java/io/reactivex/Maybe.java
bmaslakov: Of course the original `just` still works. Just thought that it would be better to wrap the `T defaultItem` into `Single` instead of `Maybe` because the `defaultItem` cannot be empty.

akarnokd: Please restore the original state. The less change to existing and practically unrelated code regarding the PR the better.

src/main/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingle.java
```diff
@@ -0,0 +1,124 @@
...
+                    other.subscribe(new OtherSingleObserver<T>(actual, this));
+                }
+            }
+        }
+
+        static final class OtherSingleObserver<T> implements SingleObserver<T> {
+
+            final SingleObserver<? super T> actual;
+
+            final AtomicReference<Disposable> parent;
```
artem-zinnatullin, line 104: `OtherSingleObserver` can extend `AtomicReference` to save field, should we do that @akarnokd?

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingleTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class MaybeSwitchIfEmptySingleTest {
+
+    @Test
+    public void nonEmpty() {
+        Maybe.just(1).switchIfEmpty(Single.just(2)).test().assertResult(1);
```
artem-zinnatullin, line 33: `assertValuesOnly()` @vanniktech?

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingleTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    public void empty() {
+        Maybe.<Integer>empty().switchIfEmpty(Single.just(2)).test().assertResult(2);
+    }
+
+    @Test
+    public void error() {
+        Maybe.<Integer>error(new TestException()).switchIfEmpty(Single.just(2))
+        .test().assertFailure(TestException.class);
```
artem-zinnatullin, line 44: Would be great to also check that no values were emitted to check that it didn't switch to fallback Single

artem-zinnatullin (commented):
LGTM, few comments

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingleTest.java
akarnokd: `onSuccess` translates to `onNext` + `onComplete` in `TestObserver` so this requires `assertResult`.

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingleTest.java
bmaslakov: `assertFailure` does check values:
```
public final U assertFailure(Class<? extends Throwable> error, T... values) {
    return assertSubscribed()
            .assertValues(values)
            .assertError(error)
            .assertNotComplete();
}
```

src/main/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingle.java
akarnokd: If you look at L95, you'll see that `parent` is SwitchIfEmptyMaybeObserver which is declared as `AtomicReference<Disposable>` thus this is not a new instace of an `AtomicReference` here.

artem-zinnatullin: Yes, I saw that, but wouldn't it save field anyway? Though it's super nit

akarnokd: The downstream can interact with only one replaceable reference to the original or the alternative disposable, so you'll need an extra field either way.

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingleTest.java
vanniktech: AssertValuesOnly is mostly for Observable & Flowable

src/test/java/io/reactivex/internal/operators/maybe/MaybeSwitchIfEmptySingleTest.java
vanniktech: Yes assertFailure does the job under the hood

vanniktech (approved):
I like it

PR #5586, opened 2017-09-06
2.x: fix Single.timeout unnecessary dispose calls

akarnokd:
This PR fixes the unnecessary `dispose` call towards the main source when the source terminates and reworks the internals to use less allocation and indirection.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/single/SingleTimeout.java
+++ b/src/main/java/io/reactivex/internal/operators/single/SingleTimeout.java
@@ -14,10 +14,11 @@
 package io.reactivex.internal.operators.single;
 
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import io.reactivex.*;
-import io.reactivex.disposables.*;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.internal.disposables.DisposableHelper;
 
 public final class SingleTimeout<T> extends Single<T> {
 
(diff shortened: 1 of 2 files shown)


vanniktech (approved):

src/main/java/io/reactivex/internal/operators/single/SingleTimeout.java
```diff
@@ -43,97 +44,116 @@ public SingleTimeout(SingleSource<T> source, long timeout, TimeUnit unit, Schedu
...
 
         @Override
-        public void onSubscribe(Disposable d) {
-            set.add(d);
+        public void onError(Throwable e) {
+            Disposable d = get();
+            if (d != DisposableHelper.DISPOSED && compareAndSet(d, DisposableHelper.DISPOSED)) {
+                DisposableHelper.dispose(task);
+                actual.onError(e);
+            }
```
artem-zinnatullin: Shouldn't we deliver error to `RxJavaPlugins.onError()` in `else` block?

src/main/java/io/reactivex/internal/operators/single/SingleTimeout.java
```diff
@@ -43,97 +44,116 @@ public SingleTimeout(SingleSource<T> source, long timeout, TimeUnit unit, Schedu
...
+            Disposable d = get();
+            if (d != DisposableHelper.DISPOSED && compareAndSet(d, DisposableHelper.DISPOSED)) {
+                if (d != null) {
+                    d.dispose();
+                }
+                SingleSource<? extends T> other = this.other;
+                if (other == null) {
+                    actual.onError(new TimeoutException());
+                } else {
+                    this.other = null;
```
artem-zinnatullin, line 117: Hmm, are you trying to avoid memory leak here? Reference to other is still held as a final field in `SingleTimeout`

I mean, final reference to `other` in `TimeoutMainObserver` will make code slightly more readable

src/test/java/io/reactivex/internal/operators/single/SingleTimeoutTest.java
```diff
@@ -69,4 +70,133 @@ public void mainError() {
...
+        assertTrue(fallback.hasObservers());
+
+        to.cancel();
+
+        assertFalse(fallback.hasObservers());
+    }
+
+    @Test
+    public void normalSuccessDoesntDisposeMain() {
+        final int[] calls = { 0 };
```
artem-zinnatullin, line 119: nit: I know you often use arrays for that, but `AtomicInteger` kinda much more readable when you need a reference to single "mutable" int :)

artem-zinnatullin (commented):
Mostly LGTM, but few comments

src/main/java/io/reactivex/internal/operators/single/SingleTimeout.java
akarnokd: The outer class has to reference other because of multiple subscribers. The inner observer has no use to the other once it has switched to it. There was an issue that some time ago that other was some cached resource that didn't go away when the user wanted it. This will somewhat help in that situation.

src/main/java/io/reactivex/internal/operators/single/SingleTimeout.java
akarnokd: Indeed. Fixed.

artem-zinnatullin (approved):

PR #5590, opened 2017-09-06
Implement LambdaConsumerIntrospection

ZacSweers:
Followup from #5569, and allows you to introspect if the resulting observer has missing error consumption and subsequently supplies a default (throwing) one.


ZacSweers:
Wasn't sure how you'd want to do the naming. Can work on adding some tests if this looks good, wanted to get something up for API review early

src/main/java/io/reactivex/observers/HasDefaultErrorConsumer.java
```diff
@@ -0,0 +1,18 @@
+package io.reactivex.observers;
+
+import io.reactivex.annotations.Experimental;
+
+/**
+ * An interface that indicates that the implementing type has default implementations for error consumption.
+ */
```
akarnokd: `@since 2.1.4 - experimental`

src/main/java/io/reactivex/internal/observers/LambdaObserver.java
```diff
@@ -101,4 +104,9 @@ public void dispose() {
     public boolean isDisposed() {
         return get() == DisposableHelper.DISPOSED;
     }
+
+    @Override
+    public boolean hasMissingErrorConsumer() {
```
akarnokd: A test would be great.

akarnokd:
The naming is fine with me.

src/main/java/io/reactivex/observers/HasDefaultErrorConsumer.java
ZacSweers: ded2448

src/main/java/io/reactivex/internal/observers/LambdaObserver.java
ZacSweers: 3f15c24

ZacSweers:
Should I add this to other types' observers too? Or do you want to start with this first?

akarnokd:
> Should I add this to other types' observers too?

That would be great!

artem-zinnatullin:
Btw @hzsweers, you can solve this right now by overriding (via reflection) `io.reactivex.internal.functions.Functions#ON_ERROR_MISSING` with `Consumer<Throwable>` implementation that'll throw error instead of delivering it to the `RxJavaPlugins`

Although current WIP looks fine for the most part 👍

ZacSweers:
yeah I'd prefer something in the public API :).

@akarnokd should there be a separate `HasDefaultErrorConsumer` for LambdaSubscriber for symmetry? Or just reuse?

ZacSweers:
Just reused for now, let me know if you want me to make it separate.

I think I've covered the other observer possibilities, let me know if there's any others I should cover

src/main/java/io/reactivex/observers/HasDefaultErrorConsumer.java
```diff
@@ -0,0 +1,20 @@
...
+
+import io.reactivex.annotations.Experimental;
+
+/**
+ * An interface that indicates that the implementing type has default implementations for error consumption.
+ *
+ * @since 2.1.4 - experimental
+ */
+@Experimental
+public interface HasDefaultErrorConsumer {
```
artem-zinnatullin: I think we should find a better name for this interface for several reasons:

1. `HasDefaultErrorConsumer` sounds like marker interface that you need to use with `instanceof` to get boolean result.
1. Interface `HasDefaultErrorConsumer` and method name `hasMissingErrorConsumer()` are inconsistent.
1. Extensibility — do we plan to extend it with other functions to check `onNext`/`onComplete` implementations?

src/test/java/io/reactivex/internal/observers/CallbackCompletableObserverTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test
+    public void hasMissingErrorConsumer() {
+        CallbackCompletableObserver o = new CallbackCompletableObserver(Functions.EMPTY_ACTION);
+
+        assertTrue(o.hasMissingErrorConsumer());
+    }
+
+    @Test
+    public void isNotMissingErrorConsumer() {
```
artem-zinnatullin: nit: `doesNotHaveMissingErrorConsumer`?

src/main/java/io/reactivex/observers/HasDefaultErrorConsumer.java
JakeWharton: I'd definitely get rid of "default" and "missing" from both. The negated method is really confusing. There's already `Has*` types which do similar things so I don't agree that it needs changed.

artem-zinnatullin: Hm, last sentence dropped from my comment for some reason.

Depending on answer to №3 I have following interface names on my mind `ObserverIntrospection`/`SubscriberIntrospection` or  `ErrorConsumerIntrospection`, hope someone has better suggestions :)

>The negated method is really confusing.

Huh, indeed! Didn't notice that until you pointed

akarnokd: I let you people work out the naming.

ZacSweers:
I agree the names could be better, will update tomorrow

src/main/java/io/reactivex/observers/HasDefaultErrorConsumer.java
ZacSweers: for the method, how about just a simple `isNotImplemented()` or `isThrowing()`?

Assuming the interface itself gets named `HasErrorConsumer`?

JakeWharton: There definitely shouldn't be negation. `!isNotImplementing()` is about the worst thing in the world to see.

I'm curious to the answer to Artem's 3rd question as well.

ZacSweers: I'd be ok with making it a more generalized name and starting with this, but we'd have to make a call on whether or not this is something others should actually be able to use or is just considered read-only public, as new additions would be breaking changes

JakeWharton: You can document the interface isn't for implementation by users. The JDK does this a bunch, or it did before 8.

ZacSweers: That sounds good to me. Should there be specific interfaces for each type in this case to avoid API mismatches?

ZacSweers: also for the name: how about `CompositeObserver`? Makes it clear that it's made up of multiple components

ZacSweers: Started with CompositeObserver, will update the tests in a bit but wanted to get this up for a more concrete proposal.

artem-zinnatullin: People will most likely mistreat `CompositeObserver` and think of it as of a combination of observers because that's how `CompositeDisposable` and `CompositeException` work.

`ObserverInstrumentation`/`ObserverIntrospection`? Such a name also more clearly delivers the intention of consume-only public API

akarnokd: `Composite` is misleading and will show up in content assist when the developer types `Composite`

akarnokd: to get to the disposable/exception variant mentioned.

akarnokd: I'd call it `LambdaConsumerIntrospection` and `hasCustomOnError()`

ZacSweers: Good points. The `Introspection` moniker feels off to me, mostly because I consider introspection an act and don't necessarily think of it as a noun/type at first. I don't know what a better word would be though. Updated in bae194c

src/test/java/io/reactivex/internal/observers/CallbackCompletableObserverTest.java
ZacSweers: Made them clearer in 500df33

ZacSweers:
Ok I've updated the tests and the naming based on CR. The only remaining bit I think is whether or not to use a global introspection interface or make one per observer for better parity. Thoughts?

artem-zinnatullin:
>use a global introspection interface or make one per observer for better parity. Thoughts?

Depends on which level of `per observer` you're talking about. Hopefully, not about all internal implementations :)

I'd create an interface per Rx-type, so you could hook into these `RxJavaPlugins` callbacks:

screen shot 2017-09-09 at 18 55 57

ie: `MaybeObserverIntrospection`, `FlowableSubscriberIntrospection`, `SingleObserverIntrospection`, `ObservableObserverIntrospection`, `CompletableObserverIntrospection`.

Because if we'll decide to add introspection API for other callbacks like `onNext` and `onComplete` in future, some of them will not make sense for types like `Completable` and `Single` and naming won't match their callbacks if we go with same interface for all types.

`LambdaConsumerIntrospection` seems too impl-specific name, I realise that the only way RxJava can understand that ie error handler is default one is to rely on `subscribe()` calls that accept lambdas. But name still feels kinda wrong and too specific. Just my 2 satoshis.

ZacSweers:
Yeah I meant per-type. I think the name is fine as a specific thing since it's only applicable to composed observers like lambdaobserver. `Lambda` itself might not be best since that naming convention seems to only exist for `Observable` and `Flowable`

akarnokd:
I think there is no need to have 5 versions of the same interface; there is always the Subscriber/Observer type discoverable:

```java
public static void errorNotImplemented(Object target) {
    if (target instanceof LambdaConsumerIntrospection) {
        if (target instanceof SingleObserver) {
            // ...
        }
        if (target instanceof Subscriber) {
            // ...
        }
    }
}

**Comment by ZacSweers** [comment]
that works for me. I think this PR is complete then

**Review by akarnokd — APPROVED** [review]

File: src/main/java/io/reactivex/observers/LambdaConsumerIntrospection.java [hunk 5]
```diff
@@ -0,0 +1,23 @@
...
+ */
+@Experimental
+public interface LambdaConsumerIntrospection {
+
+    /**
+     * @return {@code true} if a custom onError consumer implementation was supplied. Returns {@code false} if the
+     * implementation is missing an error consumer and thus using a throwing default implementation.
+     */
+    @Experimental
+    boolean hasCustomOnError();
```
artem-zinnatullin, line 21: Shouldn't we provide `hasDefaultOnError()` instead? cc @JakeWharton 

Because if you think about it from RxJava perspective — that's what library can say about itself: is something  has a *default* value/implementation in compare to what library thinks is default. 

`hasCustomOnError()` basically checks for `hasDefaultOnError()` and then negates the result which is kinda confusing. 

TL;TR: "custom" in the API feels wrong, but maybe it's just me :)

artem-zinnatullin:
@akarnokd questionable decision about types, `Single` and `Completable` might be inconsistent with future addition of `hasDefaultOnNext()` and `hasDefaultOnComplete()` changes in the interface.

up2u of course

akarnokd:
If you want to hijack those consumer types, you already know the target type and which signal types it can deliver. Therefore, if they implement the same broad interface, the impossible methods can return false or throw an `UnsupportedOperationException`.

PR #5592, opened 2017-09-07
2.x: Fix wrapping error signaled after unsubscription into UndeliverableException.

artem-zinnatullin:
For now it's just a failing test since I'm not sure if and how we'll fix that.

Explanation:

`RxJavaPlugins.isBug()` treats 3 common JDK exceptions as user bugs and doesn't wrap them into `UndeliverableException` even if they were actually signalled after observer unsubscription:

- `IllegalStateException`
- `NullPointerException`
- `IllegalArgumentException`

Real-life example:

1. User is doing long (IO, etc.) operation in some part of Rx chain ie `fromCallable()`, `map()`, `create()`, `doOnNext()`, etc. 
1. Then at some point user unsubscribes from the chain.
1. Long operation throws one of 3 common JDK exceptions mentioned above.
1. Exception gets delivered to `RxJavaPlugins.onError()` since there is no observer to deliver it to.
1. Even if user has custom `RxJavaPlugins` error handler, he/she won't be able to understand that this exception is actually an `UndeliverableException` and can be ~safely ignored/logged/etc in most situations (RxJava 1.x [just didn't do anything](https://github.com/ReactiveX/RxJava/blob/8e925345814c291e2ab2f10f4986f1ac9b8ee68a/src/main/java/rx/observers/SafeSubscriber.java#L113) with them).

---

Possible solution:

1. Do not rely on JDK exceptions that could be used by other projects, ie create `RxJavaIllegalStateException` and so on.
1. Explicitly wrap exceptions caught after `isDisposed()` in the operators into `UndeliverableException` before signalling them to `RxJavaPlugins.onError()` (I like this better atm).

### DIFF
--- a/src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
+++ b/src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
@@ -2255,4 +2255,57 @@ public class RxJavaPluginsTest {
         assertTrue(RxJavaPlugins.isBug(new CompositeException(new TestException())));
         assertTrue(RxJavaPlugins.isBug(new OnErrorNotImplementedException(new TestException())));
     }
+
+    @Test
+    public void errorSignalledAfterUnsubscriptionShouldBeWrappedIntoUndeliverableException() throws InterruptedException {
+        final CountDownLatch observableExecutionLatch = new CountDownLatch(1);
+        final AtomicBoolean canSignalErrorNow = new AtomicBoolean();
+
+        final Disposable disposable = Observable
+                .fromCallable(new Callable<Object>() {
+                    @Override
+                    public Object call() throws Exception {
+                        observableExecutionLatch.countDown();
+
+                        while (canSignalErrorNow.get() == false) {
+                            Thread.yield();
+                        }
+
+                        throw new IllegalStateException("Wrap me into UndeliverableException, I'm not a user bug!");
+                    }
+                })
+                .subscribeOn(Schedulers.newThread())
+                .subscribe();
+
+        // Make sure Observable.create() got executed.
+        observableExecutionLatch.await();
+
+        // Drop observer.
+        disposable.dispose();
+
+        // Prepare error handler to capture actual error delivered to it.
+        final AtomicReference<Throwable> throwableSeenByErrorHandler = new AtomicReference<Throwable>();
+        final CountDownLatch errorHandlerLatch = new CountDownLatch(1);
+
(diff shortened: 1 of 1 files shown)


akarnokd:
You should have posted this as a plain issue.

akarnokd:
We can introduce the `RxJavaInternalException` and use that instead of the `IllegalStateException` the `ProtocolViolationException` and `UndeliverableException` uses.

artem-zinnatullin:
Hmm, what about `IllegalArgumentException` and `NullPointerException` (this one is tricky since Reactive Streams requires RxJava to throw it in lots of cases and in many cases means actual user bug)?

I still think that second solution is probably better and more straightforward.

>You should have posted this as a plain issue.

That's questionable, anyone can grab this branch and play with it (otherwise I'd use Kotlin in in the issue for simplicity). 

// Some projects are even closing access to issue tracker to only accept actionable feedback in form of pull requests — so it's an existing practice, I was just trying to stay in the trend lol

akarnokd:
IAE and NPE could mean both user and RxJava problems where the stacktrace should usually point to the offending location. With `UndeliverableException` wrapping, the aim is to capture which operator the exception had to get derailed.

We could differentiate between `RxJavaPlugins.onError` calls that were invoked because an exception travelling on the `onError` path can no longer be forwarded to downstream, independent from what the exception is; and calls where the exception has nowhere to go - there is no further down or other internal state complications that make it infeasible to route it to `onError`.


> That's questionable, anyone can grab this branch and play with it

In their own local fork/copy.

> actionable feedback in form of pull requests

This implies PRs that actually pass. We have a couple of valueless "just testing" PRs per year that would be annoying to any project.

artem-zinnatullin:
Sorry, I'm getting conflicting messages from your comment. Should I go with `RxJavaInternalException` or wrap exception to `UndeliverableException` directly in operators? 

I'd prefer 2nd variant so as a user I'll be able to see that NPE/IAE is also undeliverable.

>In their own local fork/copy.

Well you could add me to members of the repo :trollface: ok ok, I won't post such PRs to RxJava in future.

akarnokd:
These are the steps I was proposing:

- introduce `RxJavaInternalException`
- extend `UndeliverableException`, `ProtocolViolationException` from it
- change `isBug` to consider only `RxJavaInternalException`, `MissingBackpressureException`, `OnErrorNotImplementedException` and `CompositeException` as bugs

akarnokd:
Closing this requires further consideration before a PR.

PR #5594, opened 2017-09-10
2.x: fix assertValueSequence reversed error message

akarnokd:
The error message/variable names in `TestBaseConsumer.assertValueSequence` was reversed: if more items were expected, it said "Fewer" and vice versa.

### DIFF
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -589,10 +589,10 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
         }
 
         if (actualNext) {
-            throw fail("More values received than expected (" + i + ")");
+            throw fail("Fewer values received than expected (" + i + ")");
         }
         if (expectedNext) {
-            throw fail("Fewer values received than expected (" + i + ")");
+            throw fail("More values received than expected (" + i + ")");
         }
         return (U)this;
     }
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -579,11 +579,11 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
                 break;
             }
 
-            T v = expectedIterator.next();
-            T u = actualIterator.next();
+            T u = expectedIterator.next();
+            T v = actualIterator.next();
 
             if (!ObjectHelper.equals(u, v)) {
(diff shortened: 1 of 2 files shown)


src/test/java/io/reactivex/observers/TestObserverTest.java
```diff
@@ -958,13 +958,15 @@ public void assertValueSequence() {
             throw new RuntimeException("Should have thrown");
         } catch (AssertionError ex) {
             // expected
```
vanniktech: let's remove this then?

vanniktech (approved):

src/test/java/io/reactivex/observers/TestObserverTest.java
akarnokd: I say let's keep it to indicate catching te AssertionError is not an accident.

artem-zinnatullin: Usually that's done by naming exception as `expected`:

```java
try {
  // …
  fail();
} catch (AssertionError expected) {
  // Assert exception message, etc.
}
```

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -567,23 +567,23 @@ public final U assertValueSet(Collection<? extends T> expected) {
...
             }
 
-            T v = it.next();
-            T u = vit.next();
+            T v = expectedIterator.next();
+            T u = actualIterator.next();
 
             if (!ObjectHelper.equals(u, v)) {
-                throw fail("Values at position " + i + " differ; Expected: " + valueAndClass(u) + ", Actual: " + valueAndClass(v));
+                throw fail("Values at position " + i + " differ; Expected: " + valueAndClass(v) + ", Actual: " + valueAndClass(u));
```
artem-zinnatullin: This change also needs a test

artem-zinnatullin (changes requested):
LGTM, but one more test pls

src/main/java/io/reactivex/observers/BaseTestConsumer.java
akarnokd: There is no behavior change here, just a swapped variable name. I'll restore the original u - expected, v - actual naming

src/test/java/io/reactivex/observers/TestObserverTest.java
akarnokd: Fine.

artem-zinnatullin (approved):

src/main/java/io/reactivex/observers/BaseTestConsumer.java
artem-zinnatullin: Ah. Well at least we have a valid reason to start working on RxJava 3 — kill these one-letter variable names :trollface:

Issue #5595, opened 2017-09-10
2.x: Documentation error in Observable for "any" operator


RobLewis:
In v.2 javadoc for `Observable`, it is stated (in at least 2 places) that the `any()` operator returns an `Observable`. This appears to be wrong: it returns a `Single` (which is stated correctly in other places). The corresponding entry in the `Flowable` docs appears to be correct.

akarnokd:
Yes, there are a couple of such mistakes, i.e. `all` is wrong as well. Would you like to submit a PR?

RobLewis:
TBH I don't know enough about GitHub to feel comfortable with creating/submitting PRs. Sorry.

akarnokd:
Perhaps someone else can pick it up. PR welcome.

akarnokd:
Closing via #5600.

RobLewis:
Got another one for you: in Observable, it's stated that isEmpty() returns an Observable when it's actually a Single. 

On Sep 14, 2017, at 07:24 AM, David Karnok <notifications@github.com> wrote:

Closed #5595.

akarnokd:
Covered by #5600.

PR #5600, opened 2017-09-14
Updating JavaDoc with correct return types

levaja:
A small PR with updated JavaDoc, mentioned in the issue #5595 (thanks to @RobLewis for reporting it).

Updated documentation about return types, where `Observable` was incorrectly mentioned instead of `Single`.


akarnokd (commented):
Great, thanks!

akarnokd (approved):

akarnokd:
The `Flowable` variants [are okay](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Flowable.html#isEmpty--).

PR #5602, opened 2017-09-14
1.x: Workaround for CHM.keySet bad type with Java 8 compiler

akarnokd:
This PR fixes the problem that `ConcurrentHashMap::keySet`, when compiled on JDK 8, adds the `KeySetView` type in the bytecode which is not available in JDK 7 and before. The change forces the offending place to use the the standard `Map::keySet` which returns a standard `Set`.

### DIFF
--- a/src/main/java/rx/internal/schedulers/NewThreadWorker.java
+++ b/src/main/java/rx/internal/schedulers/NewThreadWorker.java
@@ -16,7 +16,7 @@
 package rx.internal.schedulers;
 
 import java.lang.reflect.*;
-import java.util.Iterator;
+import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -110,10 +110,12 @@ public class NewThreadWorker extends Scheduler.Worker implements Subscription {
     }
 
     /** Purges each registered executor and eagerly evicts shutdown executors. */
-    @SuppressAnimalSniffer // CHM.keySet returns KeySetView in Java 8+; false positive here
     static void purgeExecutors() {
         try {
-            Iterator<ScheduledThreadPoolExecutor> it = EXECUTORS.keySet().iterator();
+            // This prevents map.keySet to compile to a Java 8+ KeySetView return type
+            // and cause NoSuchMethodError on Java 6-7 runtimes.
+            Map<ScheduledThreadPoolExecutor, ScheduledThreadPoolExecutor> map = EXECUTORS;
+            Iterator<ScheduledThreadPoolExecutor> it = map.keySet().iterator();
             while (it.hasNext()) {
                 ScheduledThreadPoolExecutor exec = it.next();
                 if (!exec.isShutdown()) {


JakeWharton (approved):
:+1: !

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -110,10 +110,10 @@ public static void deregisterExecutor(ScheduledExecutorService service) {
     }
 
     /** Purges each registered executor and eagerly evicts shutdown executors. */
-    @SuppressAnimalSniffer // CHM.keySet returns KeySetView in Java 8+; false positive here
     static void purgeExecutors() {
         try {
-            Iterator<ScheduledThreadPoolExecutor> it = EXECUTORS.keySet().iterator();
+            Map<ScheduledThreadPoolExecutor, ScheduledThreadPoolExecutor> map = EXECUTORS;
```
artem-zinnatullin: Let's add a comment I guess

artem-zinnatullin:
Please give me some time to run test scenario that failed before as a sanity check

src/main/java/rx/internal/schedulers/NewThreadWorker.java
```diff
@@ -110,10 +110,12 @@ public static void deregisterExecutor(ScheduledExecutorService service) {
     }
 
     /** Purges each registered executor and eagerly evicts shutdown executors. */
-    @SuppressAnimalSniffer // CHM.keySet returns KeySetView in Java 8+; false positive here
     static void purgeExecutors() {
         try {
-            Iterator<ScheduledThreadPoolExecutor> it = EXECUTORS.keySet().iterator();
+            // This prevents map.keySet to compile to a Java 8+ KeySetView return type
+            // and cause NoClassDefFound on Java 6-7 runtimes.
```
artem-zinnatullin: `NoSuchMethodError`

artem-zinnatullin (approved):
Checked with test scenario from https://github.com/artem-zinnatullin/RxJavaProGuardRules/issues/62#issuecomment-329588157

akarnokd:
Added comment.

I assume if the PR is verified by @artem-zinnatullin, we should release 1.3.2 asap.

src/main/java/rx/internal/schedulers/NewThreadWorker.java
akarnokd: Updated.

vanniktech (approved):

artem-zinnatullin:
mkay, looks like Travis [stuck](https://travis-ci.org/ReactiveX/RxJava/builds/275633702?utm_source=github_status&utm_medium=notification), build is in the queue for an hour.

@akarnokd can you stop/start or retrigger it?

akarnokd:
There is quite a backlog piled up on Travis. May not clear before I go to sleep.

artem-zinnatullin:
Ok, you might want to check "Concurrent jobs" related options in build settings on Travis

akarnokd:
It's not our single job but an overall ~400 started by other projects:

image

Issue #5612, opened 2017-09-21
Completable missing lambda-subscibe() w/ onSubscribe parameter to provide Disposable


jeremy303:
`Completable` does not have a lambda `subscribe()` method that provides a `Disposable` via an `onSubscribe` parameter, equivalent to the one implemented in `Observable`:

```
public final Disposable subscribe(
                    Consumer<? super T> onNext,
                    Consumer<? super Throwable> onError,
                    Action onComplete,
                    Consumer<? super Disposable> onSubscribe)
```

akarnokd:
Neither has `Single` or `Maybe`. I'm not convinced we really need it. Do you have an use case where having the `Disposable` returned by `subscribe()` is not enough?

vanniktech:
Never had the need for it in Flowable / Observable either.

ggikko:
Someone can be confused because Wiki describes that Completable provides a Disposable with an onSubscribe,, 

So, I think it need for them to modify wiki or to provide onSubscribe..

https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#completable

akarnokd:
That is on the default consumer type `CompletableObserver` and is part of the protocol. You can implement `CompletableObserver` and gain access to the `Disposable`. We just don't provide that angle via lambdas.

jeremy303:
@akarnokd The `Disposable` returned by `subscribe()` certainly works.  It may be a personal preference, but I prefer the neatness of adding to my CompositeDisposable in a `subscribe()` lambda vs wrapping the entire Rx chain in a `CompositeDisposable.add()`.  And given that `Observable` has such an overload of `subscribe()`, I would think that for consistency a case could be made for adding an equivalent overload to the other Rx types.

akarnokd:
I'm not convinced we need such overloads. @artem-zinnatullin your opinion?

artem-zinnatullin:
Yeah, I'm not a user of this API either.

@HolySamosa maybe you can use Kotlin? Then you can declare operator `+` for `CompositeDisposable` ([RxKotlin does this](https://github.com/ReactiveX/RxKotlin/blob/49b9d369ce942a74e1aa1b7f8e4a405764239f89/src/main/kotlin/io/reactivex/rxkotlin/disposable.kt#L9)) and use it like this:

```kotlin
compositeDisposable += Observable.fromCallable { something() }.subscribe()
```

[...]

PR #5613, opened 2017-09-21
Add onSubscribe in Completable to provide Disposable

ggikko:
#5612 issue

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -1697,7 +1697,29 @@ public abstract class Completable implements CompletableSource {
         ObjectHelper.requireNonNull(onError, "onError is null");
         ObjectHelper.requireNonNull(onComplete, "onComplete is null");
 
-        CallbackCompletableObserver s = new CallbackCompletableObserver(onError, onComplete);
+        return subscribe(onComplete, onError, Functions.emptyConsumer());
+    }
+
+    /**
+     * Subscribes to this Completable and calls back either the onError or onComplete functions.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code subscribe} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param onComplete the runnable that is called if the Completable completes normally
+     * @param onError the consumer that is called if this Completable emits an error
+     * @param onSubscribe the {@code Consumer} that receives the upstream's Disposable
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/Completable.java
```diff
@@ -1697,7 +1697,29 @@ public final Disposable subscribe(final Action onComplete, final Consumer<? supe
         ObjectHelper.requireNonNull(onError, "onError is null");
         ObjectHelper.requireNonNull(onComplete, "onComplete is null");
 
-        CallbackCompletableObserver s = new CallbackCompletableObserver(onError, onComplete);
+        return subscribe(onComplete, onError, Functions.emptyConsumer());
```
akarnokd: No need for null checks as you delegate to a common method.

src/main/java/io/reactivex/Completable.java
```diff
@@ -1721,9 +1743,7 @@ public final Disposable subscribe(final Action onComplete, final Consumer<? supe
     public final Disposable subscribe(final Action onComplete) {
         ObjectHelper.requireNonNull(onComplete, "onComplete is null");
 
-        CallbackCompletableObserver s = new CallbackCompletableObserver(onComplete);
-        subscribe(s);
-        return s;
+        return subscribe(onComplete, Functions.ON_ERROR_MISSING);
```
akarnokd: No need for null checks. Also since that subscribe delegates to a third method, one call indirection could be avoided by call the new method directly.

src/main/java/io/reactivex/internal/observers/CallbackCompletableObserver.java
```diff
@@ -72,7 +65,15 @@ public void onError(Throwable e) {
 
     @Override
     public void onSubscribe(Disposable d) {
-        DisposableHelper.setOnce(this, d);
+        if (DisposableHelper.setOnce(this, d)) {
+            try {
```
akarnokd, line 69: This part needs coverage testing.

akarnokd (changes requested):

src/main/java/io/reactivex/internal/observers/CallbackCompletableObserver.java
ggikko: I agree. I add test code for onSubscribe.

akarnokd:
After feedback from our top contributors, I decided this won't be added to RxJava. Thank you for your work, but next time, please ask before you start working on a PR that has not been decided upon to save on your time.

PR #5623, opened 2017-09-27
2.x: fix incorrect error message at SubscriptionHelper.setOnce

takuaraki:
I found an incorrect error message at `SubscriptionHelper.setOnce`.
The error message should mention that the second argument `Subscription s` is null, but the error message says `"d is null"`.


akarnokd (approved):

akarnokd:
Thanks!

PR #5624, opened 2017-09-27
2.x: add onTerminateDetach to Single and Completable

akarnokd:
This PR adds the `onTerminateDetach` operator to `Single` and `Completable`. It was available on `Flowable`, `Observable` and `Maybe` already.

Plus:

- Fixes the javadoc for `Maybe.onTerminateDetach`
- Adds unit tests for verifying terminal events do release references with `Maybe` as well.

Their non-existence came up on [StackOverflow](https://stackoverflow.com/questions/46397082/disposablecompletableobserver-leaking-activity-even-though-im-clearing-composit?noredirect=1#comment79870549_46397082).

**Update**

Turns out the downstream's `actual` references were not cleared in `Maybe` after all. Updated the PR to ensure that in all 3 of them and improved coverage.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -1381,6 +1381,24 @@ public abstract class Completable implements CompletableSource {
     }
 
     /**
+     * Nulls out references to the upstream producer and downstream CompletableObserver if
+     * the sequence is terminated or downstream calls dispose().
+     * <dl>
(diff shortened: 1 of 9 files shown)


vanniktech (approved):

artem-zinnatullin:
// Wonder how many people use this api…

akarnokd:
After a LeakCanary report, more and more.

PR #5626, opened 2017-09-28
2.x: add RxJavaPlugins.onCallbackCrash hook?

akarnokd:
This PR suggests a new `RxJavaPlugins` hook: `onCallbackCrash` which should be called by operator implementations when the user-provided callback (`Function`, `Action`, `Consumer`, etc.) crashes for some reason but the exception itself may contain not enough information about what that callback was.

As the test demonstrates, one can call `getClass()` and `getClass().getEnclosingMethod()` to help narrow down the search for the callback's source; Java doesn't provide the source code location this way unfortunately. In addition `getEnclosingMethod()` may return null for Java 8 lambda callbacks and `getClass()` may return a cryptic and unhelpful synthetic method identifier.

Perhaps the only reasonable use case could be the help with finding `null`-returning `Function`s (and the other 0, 2..9 argument variants). This alternative could focus on `ObjectHelper` and replace `requireNonNull` with:

```java
public static <T, R> R nonNullInvoke(Function<T, R> f, T t, 
        String name, String returnDesc) throws Exception {
    R r = f.apply(t);
    if (r != null) {
       return r;
    }
    NullPointerException npe = new NullPointerException(
        "The " + name + " returned a null " + returnDesc);
    RxJavaPlugins.onCallbackCrash(npe, f);
    throw npe;
}
```

The drawback in this case is the heavily increased bytecode count involved which may lead to missed JIT optimizations due to JVM default limits.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableMap.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableMap.java
@@ -22,6 +22,7 @@ import io.reactivex.functions.Function;
 import io.reactivex.internal.functions.ObjectHelper;
 import io.reactivex.internal.fuseable.ConditionalSubscriber;
 import io.reactivex.internal.subscribers.*;
+import io.reactivex.plugins.RxJavaPlugins;
 
 public final class FlowableMap<T, U> extends AbstractFlowableWithUpstream<T, U> {
     final Function<? super T, ? extends U> mapper;
@@ -63,7 +64,7 @@ public final class FlowableMap<T, U> extends AbstractFlowableWithUpstream<T, U> 
             try {
                 v = ObjectHelper.requireNonNull(mapper.apply(t), "The mapper function returned a null value.");
             } catch (Throwable ex) {
-                fail(ex);
+                fail(RxJavaPlugins.onCallbackCrash(ex, mapper));
                 return;
             }
             actual.onNext(v);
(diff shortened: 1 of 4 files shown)


artem-zinnatullin:
Maybe wrap exception with a custom one with proper message and throw it instead of going through `RxJavaPlugins`?

Afaik, errors happened in user defined callbacks for operators are out of scope of Reactive Streams Spec, so RxJava should be able to do pretty much anything here.

```java
public final class UserCallbackException extends Exception {

  public UserCallbackExeption(Object callback, Throwable cause) {
    super(
        "User defined callback (ie map function) has thrown exception $callbackToString()", 
        cause
    );
  }

  private static String callbackToString(Object callback) {
    return "callback info: toString = $callback.toString(), 
                class = $callback.getClass(), 
                enclosingMethod = $callback.getClass().enclosingMethod()";
  }
}
```

My concern is that I'm not sure lots of users will go that long route of defining own handler to get pretty much same basic info you've used as an example in tests in this PR.

Most of the callbacks are lambdas/anonymous-classes that only implement interfaces that RxJava asks them to implement. So I don't think lots of users will be able to gather more additional info than RxJava can.

Unless I'm missing something fundamental, I'd rather prefer implementing something helpful for everybody as a default behavior. With custom exception type and detailed message users will start  seeing this helpful information in their logs & crash reporting tools just by updating RxJava version!

akarnokd:
Well, the [function tagging](https://github.com/akarnokd/RxJava2Extensions#function-tagging) support is one way. I'll close this as it will be already on record as a experimental option.

PR #5627, opened 2017-09-29
Add 'onCompleteThrows', 'onErrorThrows' test for CallbackCompletable

ggikko:
This is test code for CallbackCompletable.

### DIFF
--- a/src/test/java/io/reactivex/internal/observers/CallbackCompletableObserverTest.java
+++ b/src/test/java/io/reactivex/internal/observers/CallbackCompletableObserverTest.java
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) 2016-present, RxJava Contributors.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
  * compliance with the License. You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software distributed under the License is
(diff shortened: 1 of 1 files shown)


src/test/java/io/reactivex/internal/observers/CallbackCompletableObserverTest.java
```diff
@@ -35,4 +41,73 @@ public void customOnErrorShouldReportCustomOnError() {
...
+
+        assertTrue(o.isDisposed());
+
+        RxJavaPlugins.reset();
+    }
+
+    @Test
+    public void onErrorCThrows() {
+
+        List<Throwable> errors = TestHelper.trackPluginErrors();
```
akarnokd: Code between this and the `RxJavaPlugins.reset()` should be put in try-finally.

akarnokd (commented):
The class itself is definitely covered to the max already: https://codecov.io/gh/ReactiveX/RxJava/src/2.x/src/main/java/io/reactivex/internal/observers/CallbackCompletableObserver.java

ggikko:
The class itself is definitely covered to the max already:  

Is this means that pr is not needed.. 

I didn't know how much code is covered by test code. 

I always appreciate your comment. thanks.

akarnokd:
No problem. The [coverage](https://github.com/ReactiveX/RxJava/pull/5627#issuecomment-333061122) reports usually list classes with less than 100% coverage. You can attempt to improve a few of those but note that sometimes it is impossible to cover everything due to hidden bytecodes generated by the javac that can never actually run.

PR #5633, opened 2017-10-02
Upgrade to Gradle 4.2.1, remove nebula plugin dependency

akarnokd:
This PR upgrades the Gradle runtime to 4.2 (which supports Java 9) and gets rid of the `rxjava-nebula` plugin. Its two features, publishing a snapshot and publishing a release has been manually implemented in `build.gradle` based on some [online examples] and [RxAndroid's](https://github.com/ReactiveX/RxAndroid/blob/2.x/rxandroid/build.gradle) variant.

The snapshot publication [works](https://oss.jfrog.org/artifactory/libs-snapshot/io/reactivex/rxjava2/rxjava/2.2.0-SNAPSHOT/) and to test the release, we may have to burn a couple of version tags from 2.1.x.

The PR also renamed the `perf` directory into `jmh` because the replacement JMH plugin expects those files there.


build.gradle
```diff
@@ -1,12 +1,34 @@
...
+    }
+  }
   dependencies {
-    classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:4.0.0'
-    classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.1.0'
+    classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.2.0'
+    classpath 'gradle.plugin.nl.javadude.gradle.plugins:license-gradle-plugin:0.13.1'
+    classpath "me.champeau.gradle:jmh-gradle-plugin:0.4.4"
+    classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.3'
+    classpath "org.jfrog.buildinfo:build-info-extractor-gradle:latest.release"
```
artem-zinnatullin: I'd strongly recommend to replace `latest.release` with particular version like `4.5.2`, otherwise build won't be reproducible

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Jun 28 11:19:41 CEST 2016
+#Mon Oct 02 13:25:01 CEST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.2-bin.zip
```
artem-zinnatullin: Can you please run something like `./gradlew --version` on this branch? 

It should update `gradle/wrapper/gradle-wrapper.jar` and `gradlew` and you need to commit them

artem-zinnatullin (changes requested):
Mostly LGTM, I'd like to check Jar Manifest config more closely later when I have time if you don't mind

build.gradle
akarnokd: Yeah, but these are snapshot/release builds which are by definition non-reproducible as there can be only one version.

gradle/wrapper/gradle-wrapper.properties
akarnokd: I did exactly that.

build.gradle
artem-zinnatullin: Well, non-reproducible also means that RxJava release might fail unpredictably because this plugin was updated even if no one changed anything Gradle-related in RxJava itself…

gradle/wrapper/gradle-wrapper.properties
akarnokd: Looks like 4.2.1 came out just now.

artem-zinnatullin: Well yes, but it still should have updated wrapper jar and bash script hm

akarnokd: Sure.

build.gradle
akarnokd: Okay.

build.gradle
```diff
@@ -16,7 +38,12 @@ apply plugin: 'java'
 apply plugin: 'checkstyle'
 apply plugin: 'jacoco'
 apply plugin: 'ru.vyarus.animalsniffer'
-apply plugin: 'nebula.rxjava-project'
+apply plugin: 'maven'
+apply plugin: 'osgi'
+apply plugin: "me.champeau.gradle.jmh"
```
artem-zinnatullin: super nit: can you please use either single or double quotes for consistency? 

(Separate PR was always overkill for that, but since you're modifying the files it feels like a great opportunity to finally do that)

build.gradle
```diff
@@ -72,14 +102,108 @@ animalsniffer {
...
+    archives javadocJar
+}
+
+jar {
+    manifest {
+        name = 'rxjava'
+        instruction 'Bundle-Vendor', 'RxJava Contributors'
+        instruction 'Bundle-DocURL', 'https://github.com/ReactiveX/RxJava'
+        instruction 'Import-Package', '!org.junit,!junit.framework,!org.mockito.*,*'
+        instruction 'Eclipse-ExtensibleAPI', 'true'
```
artem-zinnatullin: I don't see this in manifest file of 2.1.4, do we need it?

build.gradle
```diff
@@ -72,14 +102,108 @@ animalsniffer {
...
+    archives sourcesJar
+    archives javadocJar
+}
+
+jar {
+    manifest {
+        name = 'rxjava'
+        instruction 'Bundle-Vendor', 'RxJava Contributors'
+        instruction 'Bundle-DocURL', 'https://github.com/ReactiveX/RxJava'
+        instruction 'Import-Package', '!org.junit,!junit.framework,!org.mockito.*,*'
```
artem-zinnatullin: What about testNG? 

As an alternative, you can explicitly declare only `org.reactivestreams;version="[1.0,2)"`.
Or we can figure out how to dynamically collect `compile` dependencies to put them here

akarnokd: I think these are needed by the OSGi plugin MANIFEST.MF. I don't know how relevant this is.

artem-zinnatullin (approved):

vanniktech (approved):
Let's then also remove the `findbugs` & `pmd` apply plugins calls that are commented out?

akarnokd:
Removed findbugs and pmd references. Squashed.

akarnokd:
Let's merge it then. Note the [artifactory](https://oss.jfrog.org/libs-snapshot/io/reactivex/rxjava2/rxjava/2.2.0-SNAPSHOT/) repo seems to [deduplicate](https://www.jfrog.com/blog/database-wrong-fud/) files. I referenced the snapshot in a separate project and it worked.

PR #5640, opened 2017-10-04
2.x: additional warnings for fromPublisher()

akarnokd:
This PR adds JavaDoc warnings for `fromPublisher` as people still try to implement a source through it and fail to follow the specification like this:

```java
// DON'T DO THIS!
Flowable.fromPublisher((Subscriber<Integer> s) -> s.onNext(1));
```

Even though `Publisher` is a single abstract method (SAM) type, it is almost never reasonable to write a source with it that doesn't require a per-subscriber state. The example above fails to setup backpressure properly and it didn't call `s.onSubscribe()` for that, causing `NullPointerException` whenever the `Subscription` would be needed (i.e., for cancellation and requesting) by the downstream operator.

`Maybe` doesn't have `fromPublisher`.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -404,7 +404,7 @@ public abstract class Completable implements CompletableSource {
      * <p>
      * Note that even though {@link Publisher} appears to be a functional interface, it
      * is not recommended to implement it through a lambda as the specification requires
-     * state management not achievable with a stateless lambda.
+     * state management that is not achievable with a stateless lambda.
      * <dl>
      *  <dt><b>Backpressure:</b></dt>
      *  <dd>The returned {@code Completable} honors the backpressure of the downstream consumer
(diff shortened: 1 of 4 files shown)


artem-zinnatullin (approved):

src/main/java/io/reactivex/Completable.java
```diff
@@ -394,6 +394,17 @@ public static Completable fromRunnable(final Runnable run) {
...
+     * The {@link Publisher} must follow the
+     * <a href="https://github.com/reactive-streams/reactive-streams-jvm#reactive-streams">Reactive-Streams specification</a>.
+     * Violating the specification may result in undefined behavior.
+     * <p>
+     * If possible, use {@link #create(CompletableOnSubscribe)} to create a
+     * source-like {@code Completable} instead.
+     * <p>
+     * Note that even though {@link Publisher} appears to be a functional interface, it
+     * is not recommended to implement it through a lambda as the specification requires
+     * state management not achievable with a stateless lambda.
```
vanniktech: maybe add `which is` here? state management `which is` not achievable with a stateless lambda.

vanniktech (approved):

src/main/java/io/reactivex/Completable.java
akarnokd: Or `that is`. Grammarly doesn't complain about the PR's text.

vanniktech: That is would also work.

vanniktech (approved):

Issue #5642, opened 2017-10-04
Maybe.subscribeWith() has incorrect documentation


dimsuz:
`Maybe.subscribeWith()` [documentation ](https://github.com/ReactiveX/RxJava/blob/v2.1.5-RC4/src/main/java/io/reactivex/Maybe.java#L3771-L3799) suggests that it returns a `Disposable` and contains this example (translated to Kotlin):

```kotlin
    val source = Maybe.just(1)
    val composite = CompositeDisposable()
    val ms = object : MaybeObserver<Int> {
      // ...
    }
    composite.add(source.subscribeWith(ms))
```

This code does not compile, because `MaybeObserver` does not implement `Disposable`.

Should the documentation suggest to use `DisposableMaybeObserver` here?
If yes, I can submit a PR.

akarnokd:
Yes, `DisposableMaybeObserver` should be there. PR welcome.

dimsuz:
Closing as PR got merged.

PR #5647, opened 2017-10-05
2.x: Fix subscribeWith documentation examples

dimsuz:
This PR improves a documentation of `subscribeWith` functions in `Single`, `Maybe`, `Completable` by making sure examples are compilable (by using classes which implement `Disposable`).

This should fix issue #5642 .

In that issue @akarnokd approved my idea to use `DisposableMaybeObserver` in the example, but as I looked in other classes, I noticed that `Observable.subscribeWith()` documentation uses a `ResourceObserver` (which implements `Disposable`), so I thought that this is the way to go in all other top-classes?

If you think that this is not the case, and instead all of them should be switched to use `Disposable*Observer` in their examples, let me know, I will do it in this way then.

I think that all those example should use similar classes.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -1686,11 +1686,11 @@ public abstract class Completable implements CompletableSource {
      * Completable source = Completable.complete().delay(1, TimeUnit.SECONDS);
      * CompositeDisposable composite = new CompositeDisposable();
      *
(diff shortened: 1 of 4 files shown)


akarnokd:
`ResourceXObserver` is generally not necessary so `DisposableXObserver` should be in the docs.

dimsuz:
Got it, will update soon. Is it ok to force push into PR branches?

akarnokd:
PR branches are in your own repo so yes.

Issue #4853, opened 2016-11-15
2.x: switchMapSingle and switchMapCompletable


tomaszpolanski:
Hey,

the addition of `flatMapSingle`/`flatMapCompletable` made the usage of `Single` and `Completable` way nicer in RxJava 2.
Do you consider adding `switchMapSingle`/`switchMapCompletable` to RxJava 2 as well?

Cheers

akarnokd:
You mean 1.x? I have no plans for that and I don't want to keep 1.x alive for too long. I'd like to stop enhancing 1.x in 6 months and enter it into a bugfix-only mode. Otherwise 3rd party libraries may delay their upgrade way longer.

vanniktech:
He wants `switchMapSingle` and `switchMapCompletable` for RxJava 2 just like there is `flatMapSingle` and `flatMapCompletable`

tomaszpolanski:
@vanniktech exactly

akarnokd:
Ah right. No plans for extending any other xMap operator

vanniktech:
But then `concatMap` also needs the variants in order to be consistent. To be honest I forgot about those when proposing `flatMapSingle / flatMapCompletable / flatMapMaybe` initially - #4667 

And then there's also `Maybe`. Plus `Observable` and `Flowable` are able of mapping so it's easily another 12 methods.
- Observable.switchMapSingle
- Observable.switchMapMaybe
- Observable.switchMapCompletable
- Observable.concatMapSingle
- Observable.concatMapMaybe
- Observable.concatMapCompletable
- Flowable.switchMapSingle
- Flowable.switchMapMaybe
- Flowable.switchMapCompletable
- Flowable.concatMapSingle
- Flowable.concatMapMaybe
- Flowable.concatMapCompletable

tomaszpolanski:
If it is not on the roadmap because you don't want to add any more xMap operators, then I understand.

But if it is due to lack of the time, would you consider accepting contributions on the topic?

akarnokd:
Sure, but these operators are not easy.

tomaszpolanski:
Yes, I know, those are one of the hardest. 
Still, I will at least have a look and see if I can deliver a PR.

Cheers!

akarnokd:
Any progress on this?

davidmoten:
I'm happy to have a look at this one if @tomaszpolanski doesn't have time at the moment.

tomaszpolanski:
[...]

Issue #5319, opened 2017-04-27
2.x: Observable missing or incorrect images (tracking issue)


akarnokd:
- [x] [Observable.singleOrError()](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#singleOrError()): There is no default value and an empty source should result in an error.
- [x] `Observable.combineLatestDelayError(ObservableSource[], Function)`: missing diagram
- [x] `Observable.combineLatestDelayError(Function, int, ObservableSource...)`: missing diagram
- [x] `Observable.concatArrayEager`: missing diagram
- [x] `Observable.error`: should include the X in the operator body indicating it as being the parameter/return value
- [x] `Observable.fromFuture`: the operator name in the box
- [x] `Observable.fromIterable`: the operator name in the box
- [x] `Observable.fromPublisher`: missing diagram
- [x] `Observable.just`: the element should be a parameter inside the box, such outside elements imply a subject type external push.
- [x] `Observable.just(T, T)` +7: optionally tailor the diagram to the actual number of arguments, plus move the constant values inside the box
- [x] `Observable.rangeLong`: the operator name in the box
- [x] `Observable.switchOnNextDelayError`: the operator name in the box, indicator of an error getting delayed
- [x] `Observable.zipArray`: indicate there is a function instead of implying the output is an array containing one element from each source.
- [x] `Observable.zipIterable`: the operator name in the box
- [x] `Observable.all`: example of a false result
- [x] `Observable.ambWith`: the operator name in the box
- [x] `Observable.blockingFirst`: missing diagram
- [x] `Observable.blockingForEach`: the operator name in the box
- [x] `Observable.blockingIterable`: the operator name in the box, indicate that blocking may happen for each element
- [x] `Observable.blockingLast`: the operator name in the box
- [x] `Observable.blockingLast(T)`: the operator name in the box, indicate default in the box, fix labels
- [x] `Observable.blockingLatest`: missing diagram
- [x] `Observable.blockingMostRecent`: the operator name in the box
- [x] `Observable.blockingNext`: the operator name in the box
- [x] `Observable.blockingSingle`: the operator name in the box
- [x] `Observable.blockingSingle(T)`: the operator name in the box, indicate default in the box, fix labels
- [x] `Observable.toFuture()`: the operator name in the box, indicate multi-value, empty and error behavior
- [x] `Observable.blockingSubscribe`: missing diagram
[...]

PR #5649, opened 2017-10-06
2.x Add concatMapCompletable() to Observable

dweebo:
Add `concatMapCompletable()` to `Observable` as discussed in #4853
I didn't think it made sense in other reactive types.

Code is mostly a copy of `ObservableConcatMap`.  Let me know if there is a better style of code to base this off instead.  It also does not have the option to delay errors as concatMap() does, not sure if that is needed.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -6092,7 +6092,7 @@ public abstract class Observable<T> implements ObservableSource<T> {
      * <p>
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>{@code concatMap} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dd>{@code concatMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
      * @param mapper
@@ -6113,7 +6113,7 @@ public abstract class Observable<T> implements ObservableSource<T> {
      * <p>
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>{@code concatMap} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dd>{@code concatMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
      * @param mapper
@@ -6128,6 +6128,7 @@ public abstract class Observable<T> implements ObservableSource<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Completable concatMapCompletable(Function<? super T, ? extends CompletableSource> mapper, int prefetch) {
         ObjectHelper.requireNonNull(mapper, "mapper is null");
+        ObjectHelper.verifyPositive(prefetch, "prefetch");
         return RxJavaPlugins.onAssembly(new ObservableConcatMapCompletable<T>(this, mapper, prefetch));
     }
 
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -6262,7 +6262,9 @@ public abstract class Observable<T> implements ObservableSource<T> {
      * @param mapper
      *            a function that, when applied to an item emitted by the source ObservableSource, returns a CompletableSource
      * @return a Completable that signals {@code onComplete} when the upstream and all CompletableSources complete
+     * @since 2.1.6 - experimental
      */
+    @Experimental
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Completable concatMapCompletable(Function<? super T, ? extends CompletableSource> mapper) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -6252,6 +6252,46 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+     * order and waits until the upstream and all CompletableSources complete.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code concatMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param mapper
+     *            a function that, when applied to an item emitted by the source ObservableSource, returns a CompletableSource
+     * @return a Completable that signals {@code onComplete} when the upstream and all CompletableSources complete
+     */
```
akarnokd: Please add `@since 2.1.6 - experimental`

src/main/java/io/reactivex/Observable.java
```diff
@@ -6252,6 +6252,46 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Completable concatMapCompletable(Function<? super T, ? extends CompletableSource> mapper) {
```
akarnokd: Please add `@Experimental`.

src/main/java/io/reactivex/Observable.java
```diff
@@ -6252,6 +6252,46 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code concatMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param mapper
+     *            a function that, when applied to an item emitted by the source ObservableSource, returns a CompletableSource
+     * @param prefetch
+     *            the number of elements to prefetch from the current Observable
+     * @return a Completable that signals {@code onComplete} when the upstream and all CompletableSources complete
+     */
```
akarnokd: Please add `@since 2.1.6`.

src/main/java/io/reactivex/Observable.java
```diff
@@ -6252,6 +6252,46 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Completable concatMapCompletable(Function<? super T, ? extends CompletableSource> mapper, int prefetch) {
```
akarnokd: Please add `@Experimental`.

src/main/java/io/reactivex/Observable.java
```diff
@@ -6252,6 +6252,46 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+     * Maps each element of the upstream Observable into CompletableSources, subscribes to them one at a time in
+     * order and waits until the upstream and all CompletableSources complete.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code concatMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param mapper
+     *            a function that, when applied to an item emitted by the source ObservableSource, returns a CompletableSource
+     * @param prefetch
```
akarnokd: A `capacityHint` would capture the underlying behavior better: "the number of upstream items expected to be buffered until the current CompletableSource, mapped from the current item, completes."

akarnokd (commented):
Codewise it looks okay. /cc @artem-zinnatullin, @vanniktech & @davidmoten who was also interested in this.

src/main/java/io/reactivex/Observable.java
```diff
@@ -6252,6 +6252,52 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+     * @param mapper
+     *            a function that, when applied to an item emitted by the source ObservableSource, returns a CompletableSource
+     * @return a Completable that signals {@code onComplete} when the upstream and all CompletableSources complete
+     * @since 2.1.6 - experimental
+     */
+    @Experimental
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Completable concatMapCompletable(Function<? super T, ? extends CompletableSource> mapper) {
+        return concatMapCompletable(mapper, 2);
```
vanniktech, line 6271: Should not we pass 8 here since it'll get overridden to 8 in the implementation?

vanniktech (approved):
Should we also include a diagram?

akarnokd:
Added to #5319 for the diagram.

src/main/java/io/reactivex/Observable.java
akarnokd: Leave it for now.

Issue #5641, opened 2017-10-04
2.x: TestNG log details


akarnokd:
The skipped TestNG tests show up in the Travis-CI without the actual method name:

```
Gradle suite > Gradle test > io.reactivex.tck.ZipWithTckTest. SKIPPED
```

whereas the JUnit tests are fine:

```
io.reactivex.subscribers.SafeSubscriberWithPluginTest > testPluginExceptionWhenUnsubscribing2 SKIPPED
```

It would be great TestNG output would include the method name as well. PR welcome.

akarnokd:
Closing via #5641.

PR #5653, opened 2017-10-09
2.x Upgrade testng to fix issue #5641

dweebo:
Fix for #5641 

Method names now show up in gradle console when tests are skipped and in testng html output (previously no test names were showing up in html output for skipped or completed tests).


akarnokd (approved):

akarnokd:
Thanks! I was afraid of 6.10+ because it broke the Reactive-Streams TCK by mandating parameter injection all over the place.

Issue #5077, opened 2017-02-07
Take Doesn't Reduce Large RequestN


benjchristensen:
When a child subscriber submits a large requestN value, such as Long.MAX_VALUE, the 'take' operator does not reduce it as expected. 

For example, in the following where a default subscribe happens and requests Long.MAX_VALUE up, it is expected that the `take(10)` would adjust the requested size to the max value that take will permit through.

```java
someFlowable.take(10).subscribe(s)
```

Here is a unit test:

```java
    @Test
    public void testDoesntRequestMoreThanNeededFromUpstream2() throws InterruptedException {
        final AtomicLong requests = new AtomicLong();
        TestSubscriber<Long> ts = new TestSubscriber<Long>();
        Flowable.interval(100, TimeUnit.MILLISECONDS)
            .doOnRequest(new LongConsumer() {
                @Override
                public void accept(long n) {
                    System.out.println(n);
                    requests.addAndGet(n);
            }})
            .take(2)
            .subscribe(ts);
        
        ts.awaitTerminalEvent();
        ts.assertComplete();
        ts.assertNoErrors();
        assertEquals(2, requests.get());
    }
```

This errors with:

```
java.lang.AssertionError: expected:<2> but was:<9223372036854775807>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at io.reactivex.internal.operators.flowable.FlowableTakeTest.testDoesntRequestMoreThanNeededFromUpstream2(FlowableTakeTest.java:419)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #5655, opened 2017-10-09
2.x: add limit() to limit both item count and request amount

akarnokd:
This PR adds the operator `limit` to limit the number of items as well as the total request amount.

Some asynchronous boundary-like sources, such as network libraries that translate from/to `Flowable`/`Publisher`, may not modulate or limit downstream requests and thus the other side may produce items unnecessarily. 

Alternatives, such as `take(N)` is designed to go unbounded if its downstream requests more than N to improve performance of local flows and `rebatchRequests(M)` will keep requesting once 75% of M has been received.

Note that requests are negotiated on an operator boundary and `limit`'s amount may not be preserved further upstream. For example, `source.observeOn(Schedulers.computation()).limit(5)` will still request the default (128) elements from the given `source`.

Related discussion: #5077.
Blog: [Java 9 Flow API: taking and skipping - Limiting the request amount](http://akarnokd.blogspot.hu/2017/09/java-9-flow-api-taking-and-skipping.html)

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -9677,6 +9677,53 @@ public abstract class Flowable<T> implements Publisher<T> {
     }
 
     /**
+     * Limits both the number of upstream items (after which the sequence completes)
+     * and the total downstream request amount requested from the upstream to
+     * possibly prevent the creation of excess items by the upstream.
+     * <p>
+     * The operator requests at most the given {@code count} of items from upstream even
+     * if the downstream requests more than that. For example, given a {@code limit(5)},
+     * if the downstream requests 1, a request of 1 is submitted to the upstream
+     * and the operator remembers that only 4 items can be requested now on. A request
+     * of 5 at this point will request 4 from the upstream and any subsequent requests will
+     * be ignored.
+     * <p>
+     * Note that requests are negotiated on an operator boundary and {@code limit}'s amount
+     * may not be preserved further upstream. For example,
+     * {@code source.observeOn(Schedulers.computation()).limit(5)} will still request the
+     * default (128) elements from the given {@code source}.
+     * <p>
+     * The main use of this operator is with sources that are async boundaries that
+     * don't interfere with request amounts, such as certain {@code Flowable}-based
+     * network endpoints that relay downstream request amounts unchanged and are, therefore,
+     * prone to trigger excessive item creation/transmission over the network.
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator requests a total of the given {@code count} items from the upstream.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code limit} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+
+     * @param count the maximum number of items and the total request amount, non-negative.
+     *              Zero will immediately cancel the upstream on subscription and complete
+     *              the downstream.
+     * @return the new Flowable instance
+     * @see #take(long)
+     * @see #rebatchRequests(int)
+     * @since 2.1.6 - experimental
+     */
+    @Experimental
+    @BackpressureSupport(BackpressureKind.SPECIAL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @CheckReturnValue
+    public final Flowable<T> limit(long count) {
+        if (count < 0) {
+            throw new IllegalArgumentException("count >= 0 required but it was " + count);
+        }
+        return RxJavaPlugins.onAssembly(new FlowableLimit<T>(this, count));
+    }
+
+    /**
      * Returns a Flowable that applies a specified function to each item emitted by the source Publisher and
      * emits the results of these function applications.
      * <p>
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
@@ -1,0 +1,137 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.internal.operators.flowable;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.reactivestreams.*;
+
+import io.reactivex.*;
+import io.reactivex.annotations.Experimental;
+import io.reactivex.internal.subscriptions.*;
+import io.reactivex.plugins.RxJavaPlugins;
+
+/**
+ * Limits both the total request amount and items received from the upstream.
+ *
+ * @param <T> the source and output value type
+ * @since 2.1.6 - experimental
+ */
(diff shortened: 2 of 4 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -9677,6 +9677,53 @@ public final Completable ignoreElements() {
...
+     * @return the new Flowable instance
+     * @see #take(long)
+     * @see #rebatchRequests(int)
+     * @since 2.1.6 - experimental
+     */
+    @Experimental
+    @BackpressureSupport(BackpressureKind.SPECIAL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @CheckReturnValue
+    public final Flowable<T> limit(long count) {
```
artem-zinnatullin, line 9719: I like this strict version of `take()`, but thinking about API design I'm afraid that this additional operator could add unnecessary confusion for the users.

Maybe an overload of `take()` with boolean flag that controls "strictness" of the upstream requests would work? (or something better if you have it on your mind)

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
```diff
@@ -0,0 +1,137 @@
...
+                } else {
+                    this.upstream = s;
+                    actual.onSubscribe(this);
+                }
+            }
+        }
+
+        @Override
+        public void onNext(T t) {
+            long r = remaining;
```
artem-zinnatullin, line 79: what's the point of local `r`? It looks like you can just do everything directly on `remaining` here

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
```diff
@@ -0,0 +1,137 @@
...
+            if (r > 0L) {
+                remaining = --r;
+                actual.onNext(t);
+                if (r == 0L) {
+                    upstream.cancel();
```
artem-zinnatullin, line 84: What about this:

```java
if (r > 0L) {
  remaining = --r;
  final boolean completed = r == 0L;
  
  if (completed) {
    upstream.cancel();
  }

  actual.onNext(t);

  if (completed) {
    actual.onComplete();
  }
}
```

This way we cancel upstream before delivering `onNext` notification, it could be important in case if `onNext` triggers long running chain of computations thus delaying cancellation of the upstream (in current version).

artem-zinnatullin: Though, current implementations of `take()` also does this so it might be unwanted behavior change, up2u, we can move this change to 3.x

src/test/java/io/reactivex/internal/operators/flowable/FlowableLimitTest.java
```diff
@@ -0,0 +1,208 @@
...
+import io.reactivex.exceptions.*;
+import io.reactivex.functions.*;
+import io.reactivex.internal.subscriptions.BooleanSubscription;
+import io.reactivex.plugins.RxJavaPlugins;
+import io.reactivex.processors.PublishProcessor;
+import io.reactivex.subscribers.TestSubscriber;
+
+public class FlowableLimitTest implements LongConsumer, Action {
+
+    final List<Long> requests = Collections.synchronizedList(new ArrayList<Long>());
```
artem-zinnatullin: nit: I don't think you need synchronization here, I don't see any concurrent access to this list

src/test/java/io/reactivex/internal/operators/flowable/FlowableLimitTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+    @Override
+    public void accept(long t) throws Exception {
+        requests.add(t);
+    }
+
+    @Override
+    public void run() throws Exception {
+        requests.add(-100L);
```
artem-zinnatullin: super nit: `List<Object>`, `private static final Object CANCELLED = new Object()`, `requests.add(CANCELLED)`

```java
assertEquals(5, requests.get(0));
assertEquals(CANCELLED, requests.get(1));
```

src/main/java/io/reactivex/Flowable.java
akarnokd: Possibly.

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
akarnokd: I avoid re-reading fields.

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
akarnokd: Or it could trigger an interrupted thread and fail code in the downstream unexpectedly.

src/test/java/io/reactivex/internal/operators/flowable/FlowableLimitTest.java
akarnokd: Just in case.

src/test/java/io/reactivex/internal/operators/flowable/FlowableLimitTest.java
akarnokd: -100 is fine.

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
artem-zinnatullin: hm, it's not volatile, so it should be normally in CPU register/cache, but yeah local var has slightly more chances to be there

Although local var creates indirection between read/writes, so overall I think it's neither win nor loss in performance :)

There is interesting question on SO, but answers seem to only compare speed of access to the field when it's already initialized, not taking overall overhead of additional variable https://stackoverflow.com/questions/21613098/java-local-vs-instance-variable-access-speed

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
artem-zinnatullin: Good point! 

But doesn't it mean that `actual.onComplete()` should happen before `upstream.cancel()` then? :trollface:

akarnokd: It's way less likely to happen. There were issues around `onNext` before.

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
akarnokd: The JIT may load it into a register but not everything runs on HotSpot.

src/main/java/io/reactivex/internal/operators/flowable/FlowableLimit.java
artem-zinnatullin: ok, although I think probability is the same here, `upstream.cancel()` has to be called in both ways of call ordering and possibly cause problems. But in current order we guarantee delivery of onNext which is probably good :)

src/test/java/io/reactivex/internal/operators/flowable/FlowableLimitTest.java
```diff
@@ -0,0 +1,210 @@
...
+        Flowable.range(1, 5)
+        .doOnRequest(this)
+        .doOnCancel(this)
+        .limit(5)
+        .test()
+        .assertResult(1, 2, 3, 4, 5);
+
+        assertEquals(2, requests.size());
+        assertEquals(5, requests.get(0).intValue());
+        assertEquals(-100, requests.get(1).intValue());
```
vanniktech: let's use the constant here?

vanniktech (approved):

akarnokd:
Updated. I'm not too fond of a `take(long, boolean)` overload. I suggest having `limit` as the name. I think the feature it provides is mainly useful for network/IO library writers and power users.

vanniktech:
I also prefer to have a dedicated method instead of magic Boolean operators.

artem-zinnatullin (approved):

Issue #5657, opened 2017-10-11
RxJava 1.3 -- timeout not unsubscribing from original observable


sfitts:
Version -- 1.3.0 (did not see any release notes in the subsequent 1.3.x releases about this).

Not sure if this is the expected behavior or not, but when I use the operator:

```
timeout(long timeout, TimeUnit timeUnit, Observable<? extends T> other)
```

it seems that if the timeout occurs and we switch to mirroring the `other` observable the operator does not unsubscribe from the original observable.  Based on the documentation I was expecting that this would essentially execute a "switch" like behavior, which would result in the original observable being unsubscribed.

Here is some code that demonstrates what I'm seeing:

```
        Subject<Long, Long> subject = PublishSubject.create();
        Observable<Long> initialObservable = subject.share().map(value -> {
            System.out.println("Received value " + value);
            return value;
        });

        Observable<Long> timeoutObservable = initialObservable.map(value -> {
           System.out.println("Timeout received value " + value);
           return value;
        });

        TestSubscriber<Long> subscriber = new TestSubscriber<>();
        initialObservable.timeout(1, TimeUnit.SECONDS, timeoutObservable).subscribe(subscriber);
        subject.onNext(5L);
        Thread.sleep(1500);
        subject.onNext(10L);
        subject.onCompleted();

        subscriber.awaitTerminalEvent();
        subscriber.assertNoErrors();
        subscriber.assertValues(5L, 10L);
```

Here is the output:

```
Received value 5
Received value 10
Received value 10
Timeout received value 10
```

I was only expecting one `Received value 10`.

sfitts:
Inserting `this.unsubscribe();` before line 200 in `OperatorTimeoutBase` gives me the behavior I was expecting.  Of course I'm not enough of an RxJava internals expert to know what other, less desirable side effects this might have (though naively it seems that once the timeout subscriber marks itself as terminated, it pretty much gets out of the way).

akarnokd:
Thanks for reporting. This seems to be a bug somewhere in that operator which I can't pinpoint, the 1.x architecture is awkward. See #5660 for the proposed fix.

akarnokd:
Closing via #5660.

PR #5660, opened 2017-10-11
1.x: fix timeout (timed, selector) unsubscribe bug

akarnokd:
This PR fixes an unsubscribe bug somewhere in the timed `timeout` operator reported in #5657 by implementing it in a algorithmically fresh manner.


src/main/java/rx/internal/operators/OperatorTimeout.java
```diff
@@ -1,58 +0,0 @@
...
-
-/**
- * Applies a timeout policy for each element in the observable sequence, using
- * the specified scheduler to run timeout timers. If the next element isn't
- * received within the specified timeout duration starting from its predecessor,
- * the other observable sequence is used to produce future messages from that
- * point on.
- * @param <T> the value type
- */
-public final class OperatorTimeout<T> extends OperatorTimeoutBase<T> {
```
vanniktech, line 31: Is `OperatorTimeoutBase` still used somewhere?

vanniktech (approved):

src/main/java/rx/internal/operators/OperatorTimeout.java
akarnokd: In `OperatorTimeoutWithSelector`.

src/test/java/rx/internal/operators/OperatorTimeoutTests.java
```diff
@@ -427,4 +432,71 @@ public void withDefaultSchedulerAndOther() {
...
+            @Override
+            public void call() {
+                list.add("Unsubscribed");
+            }
+        })
+        .timeout(1, TimeUnit.SECONDS, timeoutObservable, sch).subscribe(subscriber);
+
+        subject.onNext(5L);
+
+        sch.advanceTimeBy(2, TimeUnit.SECONDS);
```
sfitts, line 473: Nice -- did not know about TestScheduler, very handy.

src/main/java/rx/internal/operators/OperatorTimeout.java
sfitts: I probably should have mentioned in my original bug, but this variant of timeout (which uses `OperatorTimeoutWithSelector`) also exhibits the same problem:

```
public final <U, V> Observable<T> timeout(Func0<? extends Observable<U>> firstTimeoutSelector, Func1<? super T, ? extends Observable<V>> timeoutSelector, Observable<? extends T> other)
```

Can this fix be adapted to cover that variant as well (which we also use)?

src/main/java/rx/internal/operators/OnSubscribeTimeoutTimedWithFallback.java
```diff
@@ -0,0 +1,230 @@
...
+            Subscription s = task.get();
+            if (s != null) {
+                s.unsubscribe();
+            }
+
+            actual.onNext(t);
+
+            consumed++;
+
+            startTimeout(idx + 1);
```
sfitts: If I'm reading this correctly, I think there may be a slight behavior difference between this implementation and the existing one.  The existing implementation will timeout (and thus switch) if there are 0 calls to `onNext` before the timeout.  Here I think there needs to be at least 1 call before the timer is set.  I'll confirm this and send a test case when I get a chance.

Turns out we actually want both semantics.  We're using the other timeout variant mentioned below to get the "have to have at least one message" semantic by having `firstTimeoutSelector` return `Observable.never()`.  If there were a way to get both behaviors with just this operator, that would great.

sfitts: Never mind -- I see the `startTimeout` call in `call` above.

sfitts: With this operator I can override `call` to make that optional and get the 2 semantics I'm interested in, so this will be perfect.  Thanks.

sfitts: Except that it's `final`, doh!.  I'm assuming that's a standard policy for operators.  In that case would it be possible to add a `boolean` parameter to the constructor where `true` means start the timer on subscribe and `false` means don't?  Understood if that seems too specialized, but it would save us having to use the more complex operator.

akarnokd: Sorry, no more new features for 1.x, only bugfixes.

src/main/java/rx/internal/operators/OperatorTimeout.java
akarnokd: Yes, they were built on the same abstraction. It wasn't part of the initial report and I try to minimize my interaction with 1.x.

sfitts: Understood.  As I mention above, the new operator actually allows us to solve both of our use cases, so I'm fine with leaving the other operator alone.

src/main/java/rx/internal/operators/OnSubscribeTimeoutTimedWithFallback.java
sfitts: Understood.

akarnokd:
Posted fix for the selector version as well.

sfitts:
Thanks -- we're using the new operators in our code base and they are working as expected.

akarnokd:
Great!

PR #5661, opened 2017-10-11
2.x: improve Flowable.timeout()

akarnokd:
This PR improves the internal overhead of the `Flowable.timeout` operator (its 2 timed and 2 selector-based versions) to use the adapted [indexed atomic state transition approach](http://akarnokd.blogspot.hu/2017/09/java-9-flow-api-timing-out-events.html).

In addition, there was a race condition and potential event loss in `TestScheduler`: 
- `peek()` could return `null` if the task was removed from the queue (via disposing it) between the `isEmpty()` check and `peek()` itself. 
- In addition, if the task was disposed between the `peek` and `remove`, the `remove` could remove the new head of the queue which is no longer what `peek` saw and results in tasks being dropped.


src/main/java/io/reactivex/internal/operators/flowable/FlowableTimeout.java
```diff
@@ -48,299 +47,343 @@ public FlowableTimeout(
...
         }
 
         @Override
         public void onNext(T t) {
-            long idx = index + 1;
-            index = idx;
-
-            actual.onNext(t);
+            long idx = get();
+            if (idx == Long.MAX_VALUE || !compareAndSet(idx, idx + 1)) {
```
artem-zinnatullin, line 97: nit: maybe leave a link to your blog post in comment? So readers could understand the pattern used here

src/main/java/io/reactivex/internal/operators/flowable/FlowableTimeout.java
```diff
@@ -48,299 +47,343 @@ public FlowableTimeout(
...
+        public void onSubscribe(Subscription s) {
+            if (SubscriptionHelper.setOnce(this, s)) {
+                s.request(Long.MAX_VALUE);
+            }
+        }
+
+        @Override
+        public void onNext(Object t) {
+            if (get() != SubscriptionHelper.CANCELLED) {
+                get().cancel();
```
artem-zinnatullin: Extract `get()` result to local var here?

artem-zinnatullin (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableTimeout.java
akarnokd: Not really keen on this. The PR will reference the blog.

src/main/java/io/reactivex/internal/operators/flowable/FlowableTimeout.java
akarnokd: Sure.

src/main/java/io/reactivex/internal/operators/flowable/FlowableTimeout.java
artem-zinnatullin: Right, my point is not to promote your blog haha

Just want to leave some bread crumbs in the code so one could understand that `MAX_VALUE` is used as a notion of terminal state here, one might confuse it with something related to unbounded request mode

I'd just put 

```java
@see <a href="https://akarnokd.blogspot.hu/2017/09/java-9-flow-api-timing-out-events.html">Operator Design Principles</a>
```

To the class's javadoc.

It's obviously a nit and up2u :)

akarnokd: Thanks for the suggestion, but I won't put it there.

vanniktech (approved):

PR #5664, opened 2017-10-12
remove duplcated codes

JSpiner:
`CompositeDisposable` has duplicate check logic.
Checking before `Synchronized` does not seem to be a big performance difference.
So I removed the duplicated code from the code.

### DIFF
--- a/src/main/java/io/reactivex/disposables/CompositeDisposable.java
+++ b/src/main/java/io/reactivex/disposables/CompositeDisposable.java
@@ -64,9 +64,6 @@ public final class CompositeDisposable implements Disposable, DisposableContaine
 
     @Override
     public void dispose() {
-        if (disposed) {
-            return;
-        }
         OpenHashSet<Disposable> set;
         synchronized (this) {
             if (disposed) {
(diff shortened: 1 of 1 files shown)


JakeWharton:
> does not seem to be a big performance difference

Do you have JMH benchmarks to back this up?

JSpiner:
@JakeWharton 
Good point. I will test it.

akarnokd:
These are legitimate checks that help avoid locking when the container has already been disposed. Please don't change them.

JSpiner:
@akarnokd 

Can you tell me what the case is?

Even if I delete duplicate checking code, it can avoid locking issue.
Because of it is checked one more time in `synchronized(this)`.

I agree that it may be little faster that check one more time before `synchronization`.
(it may need performance check, I will check it....)
But I do not know if there is a lock problem.

```java
volatile boolean disposed; //it is global variable. not static variable. lock with `synchronized(this)`
....
public int size() {
        if (disposed) {    // check 1
            return 0;
        }
        synchronized (this) {
            if (disposed) { // check 2 <- duplicated
                return 0;
            }
        ...
    }
}
```

akarnokd:
When the composite has been disposed, all subsequent `add` will go lock-free. In that case, the only behavior is to dispose the incoming `Disposable`. Otherwise, concurrent accesses would block out each other just to discover the container has been disposed.

Note that your "smart" IDE may suggest all sorts of cleanup hints but often it can't consider the context (or it only performs string matching and no flow analysis).

JSpiner:
I cross checked `rx.NET`, `rxSwift`, `rxJs`.
- `rx.NET` : just one check with synchronized.
- `rxSwift` : just one check with synchronized.
- `rxJs` : just one check, no synchronized.

But if they are not used, not mean not need. 
So I'll add a test code to see if it's safe.
And, I want to have more discussion. Can you re-open this PR or create new issue?

akarnokd:
> I cross checked

- Rx.NET: they are ancient tech and use locking all over the place. Since .NET doesn't have runtime JIT, biased locking or lock elision, that costs a lot.
- RxSwift: they copied Rx.NET most of the time without consideration, plus the language is hostile to lock-free approaches anyway.
- RxJS: there is no concurrency in JavaScript so there is no need for locking.

> And, I want to have more discussion.

Have you read [Concurrency in Practice](https://www.amazon.com/Java-Concurrency-Practice-Brian-Goetz/dp/0321349601)?

Issue #5317, opened 2017-04-27
Documentation: Observable.singleElement() states wrong Maybe contract


realdadfish:
The javadoc of `Observable.singleElement()` says:

```
* Returns a Maybe that emits the single item emitted by this Observable if this Observable
* emits only a single item, otherwise if this Observable emits more than one item or no items, an
* {@code IllegalArgumentException} or {@code NoSuchElementException} is signalled respectively.
```

This is not the contract of `Maybe` and also not what the `ObservableSingleMaybe` operator implements. It _will_ complete if there is no value, return a single value and error out if there is more than one value.

Version: 2.0.7

akarnokd:
Do you mean the doc doesn't explicitly mention that if the Observable is empty, the Maybe completes?

jschoedt:
I also got confused about this. It states :

> or **no items**, an {@code IllegalArgumentException} or {@code NoSuchElementException} is signalled respectively.

A Maybe should return **no items** without any exception.

akarnokd:
PR welcome but the actual behavior regarding the exceptions should be verified.

jschoedt:
Agree... one could otherwise use .singleOrError() which is documented to have the same semantics.

akarnokd:
Closing via #5668.

PR #5668, opened 2017-10-15
Fix a misleading documentation of Observable.singleElement()

VeskoI:
This PR improves the documentation of `Observable.singleElement()`, as mentioned in  https://github.com/ReactiveX/RxJava/issues/5317.

An appropriate unit test to verify the `empty` behaviour already exist (`ObservableSingleTest. testSingleWithEmpty()`), I just added one more check there to make it very explicit that an error is not thrown.

As a side note, I'm attaching an updated Marble diagram for the method that includes all 3 states of the resulting Maybe (empty, success, error). I believe this is listed in https://github.com/ReactiveX/RxJava/issues/5319.

singleelement

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -10277,8 +10277,8 @@ public abstract class Observable<T> implements ObservableSource<T> {
     }
 
     /**
-     * Returns a Maybe that emits the single item emitted by this Observable or completes if this Observable is empty.
-     * If this Observable emits more than one item, an {@code IllegalArgumentException} is signalled instead.
+     * Returns a Maybe that completes if this Observable is empty or emits the single item emitted by this Observable,
+     * or signals an {@code IllegalArgumentException} if this Observable emits more than one item.
      * <p>
      * <img width="640" height="315" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/singleElement.png" alt="">
      * <dl>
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -10277,9 +10277,8 @@ public final void safeSubscribe(Observer<? super T> s) {
     }
 
     /**
-     * Returns a Maybe that emits the single item emitted by this Observable if this Observable
-     * emits only a single item, otherwise if this Observable emits more than one item or no items, an
-     * {@code IllegalArgumentException} or {@code NoSuchElementException} is signalled respectively.
+     * Returns a Maybe that emits the single item emitted by this Observable or completes if this Observable is empty.
```
akarnokd: The problem with two sentences is that most IDEs and the Javadoc method listing shows only the first sentence. Could you rewrite it that all relevant behavior is in the same sentence?

akarnokd:
Here is the standard link for the marble:

`https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/singleElement.o.png`

VeskoI:
Thanks, updated the JavaDoc to include the Marble diagram :)

akarnokd (approved):

src/test/java/io/reactivex/internal/operators/observable/ObservableSingleTest.java
```diff
@@ -282,6 +282,7 @@ public void testSingleWithEmpty() {
 
         InOrder inOrder = inOrder(observer);
         inOrder.verify(observer).onComplete();
+        inOrder.verify(observer, never()).onError(any(Throwable.class));
```
vanniktech, line 285: This is actually covered by `verifyNoMoreInteractions` but being explicit here doesn't hurt.

vanniktech (approved):

PR #5672, opened 2017-10-16
2.x: coverage, add SingleToFlowableTest

ggikko:
Add "SingleToFlowableTest" to cover toFlowable with test.

### DIFF
--- a/src/test/java/io/reactivex/TestHelper.java
+++ b/src/test/java/io/reactivex/TestHelper.java
@@ -1230,6 +1230,60 @@ public enum TestHelper {
         }
     }
 
+ /**
+     * Check if the given transformed reactive type reports multiple onSubscribe calls to
+     * RxJavaPlugins.
+     * @param <T> the input value type
+     * @param <R> the output value type
+     * @param transform the transform to drive an operator
+     */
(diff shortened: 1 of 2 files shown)


src/test/java/io/reactivex/TestHelper.java
```diff
@@ -1230,6 +1230,60 @@ protected void subscribeActual(SingleObserver<? super T> observer) {
...
     }
 
+ /**
+     * Check if the given transformed reactive type reports multiple onSubscribe calls to
+     * RxJavaPlugins.
+     * @param <T> the input value type
+     * @param <R> the output value type
+     * @param transform the transform to drive an operator
+     */
+    public static <T, R> void checkDoubleOnSubscribeSingleToObservable(Function<Single<T>, ? extends Publisher<R>> transform) {
```
akarnokd, line 1240: This fails to compile due to clashing with a method having the same signature. It should be called `checkDoubleOnSubscribeSingleToFlowable` for example.

ggikko: My bad. Thanks comment. I reopen pull request.

PR #5673, opened 2017-10-16
2.x: coverage, add SingleToFlowableTest

ggikko:
Add "SingleToFlowableTest" to cover toFlowable with test.

### DIFF
--- a/src/test/java/io/reactivex/TestHelper.java
+++ b/src/test/java/io/reactivex/TestHelper.java
@@ -1234,6 +1234,60 @@ public enum TestHelper {
      * Check if the given transformed reactive type reports multiple onSubscribe calls to
      * RxJavaPlugins.
      * @param <T> the input value type
+     * @param <R> the output value type
+     * @param transform the transform to drive an operator
+     */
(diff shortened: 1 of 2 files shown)


akarnokd:
The operator is covered 100% already btw: [SingleToFlowable](https://codecov.io/gh/ReactiveX/RxJava/src/2.x/src/main/java/io/reactivex/internal/operators/single/SingleToFlowable.java)

akarnokd:
You could, hovewer, see if the other classes could be covered better: https://codecov.io/gh/ReactiveX/RxJava/tree/2.x/src/main/java/io/reactivex/internal/operators

akarnokd:
My mistake, looked at the complexity column.

ggikko:
well.. I thought it was not covered "if (DisposableHelper.validate(this.d, d))" in SingleToFlowable

I got it. thanks for link.

ggikko:
I have a question. There are many duplicate codes in Test code. How about zip these codes to one method for cleaning.

akarnokd:
You mean `TestHelper`? The types are different and they can't be reasonably factored out.

ggikko:
I see. :)

PR #5689, opened 2017-10-26
add onError in try-catch

krcho:
I don't know why It isn't pass exception to "onError" in onSuccess try-catch


krcho:
Test Error 


why this consumer doesn't allow to pass exception to onError??

akarnokd:
Please don't. The `Single` protocol is `onSubscribe (onSuccess | onError)?`, you can't call `onError` after `onSuccess` is called.

Issue #5694, opened 2017-10-31
observeOn unsubscribes child observables/interrupts the thread


stannnous:
Hi,

The following test shows the use of `observeOn` in an observable chain interrupts the thread, cancelling child observables.

    import org.junit.Test;
    import rx.Completable;
    import rx.schedulers.Schedulers;

    import java.util.concurrent.CountDownLatch;

    public class InterruptTest {
        @Test
        public void testInterrupt() throws InterruptedException {
           final int count = 1000;
           CountDownLatch latch = new CountDownLatch(count);

            for (int i = 0; i < count; i++) {
                Completable.complete()
                    .subscribeOn(Schedulers.io())
                    .observeOn(Schedulers.io()) // The problem does not occur if you comment out this line
                    .andThen(Completable.fromAction(() -> {
                        try {
                            Thread.sleep(3000);
                        } catch (InterruptedException e) {
                            System.out.println("Interrupted!"); // This is output periodically
                        }
                    }))
                    .subscribe(() -> latch.countDown());
            }

            latch.await();
            testInterrupt();
        }
    }

The sleep is interrupted periodically printing "Interrupted!" while the test is running. The problem does not occur when the `observeOn` line is commented out.

We've tested this in RxJava 1.3.2 and 2.1.6 with the same results.

akarnokd:
`andThen` and `concat` dispose the previous sources - which is unnecessary on its own - and thus a particular timing may cause the worker of the IO `Scheduler` to be disposed while it is serving the second `Completable`.

akarnokd:
Fix in #5695.

stannnous:
Hi @akarnokd thanks for the fast response. Is there any chance we can get this fixed in 1.x as well?

akarnokd:
Yes, but it may take an additional week to have it released. Besides, we would like to encourage people to switch to v2 as it's only 6 months till RxJava 1.x' end-of-life.

akarnokd:
Closing via #5696

PR #5695, opened 2017-10-31
2.x: fix Completable.concat to use replace (don't dispose old)

akarnokd:
`Completable.andThen`, `concat(Iterable)` and `concatArray()` disposed the previous `Disposable` when receiving the new `Disposable` from the next source which could lead to interruption. `concat(Publisher)` already uses `replace` instead of `update`.

There is a peculiar interplay between `subscribeOn`, `observeOn` and the trampoline in `concat` which can trigger such an interruption: the task of the `observeOn` is cancelled with `mayInterruptIfRunning == true` because the `dispose` call chain shuts down the worker of the `observeOn` from the `subscribeOn`'s thread.

Reported in #5694

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
+++ b/src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
@@ -15,7 +15,13 @@ package io.reactivex.internal.operators.completable;
 
 import io.reactivex.Completable;
 import io.reactivex.Maybe;
+import io.reactivex.functions.Action;
+import io.reactivex.schedulers.Schedulers;
+
+import java.util.concurrent.CountDownLatch;
+
 import org.junit.Test;
+import static org.junit.Assert.*;
 
 public class CompletableAndThenTest {
     @Test(expected = NullPointerException.class)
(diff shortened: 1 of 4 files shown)


vanniktech (approved):

src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
```diff
@@ -63,4 +69,39 @@ public void andThenMaybeError() {
...
+    public void andThenNoInterrupt() throws InterruptedException {
+        for (int k = 0; k < 100; k++) {
+            final int count = 10;
+            final CountDownLatch latch = new CountDownLatch(count);
+            final boolean[] interrupted = { false };
+
+            for (int i = 0; i < count; i++) {
+                Completable.complete()
+                .subscribeOn(Schedulers.io())
+                .observeOn(Schedulers.io()) // The problem does not occur if you comment out this line
```
artem-zinnatullin: Can we delete these comments?

src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
```diff
@@ -63,4 +69,39 @@ public void andThenMaybeError() {
...
+                .andThen(Completable.fromAction(new Action() {
+                    @Override
+                    public void run() throws Exception {
+                        try {
+                            Thread.sleep(30);
+                        } catch (InterruptedException e) {
+                            System.out.println("Interrupted! " + Thread.currentThread()); // This is output periodically
```
artem-zinnatullin: I don't think we need it, or at least message should be changed to something like `System.err.println("Interrupted, but was not expected to")`

src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
```diff
@@ -63,4 +69,39 @@ public void andThenMaybeError() {
             .assertError(RuntimeException.class)
             .assertErrorMessage("bla");
     }
+
+    @Test
+    public void andThenNoInterrupt() throws InterruptedException {
+        for (int k = 0; k < 100; k++) {
+            final int count = 10;
+            final CountDownLatch latch = new CountDownLatch(count);
+            final boolean[] interrupted = { false };
```
artem-zinnatullin, line 78: nit: `AtomicBoolean` would be a nicer replacement for mutable Boolean :)

artem-zinnatullin (approved):

PR #5706, opened 2017-11-06
Remove mentions of Main thread from Schedulers.single() javadoc.

artem-zinnatullin:
To avoid confusion. 

One could think that `single()` can be used to execute code on "Main" thread.

### DIFF
--- a/src/main/java/io/reactivex/schedulers/Schedulers.java
+++ b/src/main/java/io/reactivex/schedulers/Schedulers.java
@@ -243,9 +243,9 @@ public final class Schedulers {
      * <p>
      * Uses:
      * <ul>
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/schedulers/Schedulers.java
```diff
@@ -243,9 +243,9 @@ public static Scheduler newThread() {
      * <p>
      * Uses:
```
artem-zinnatullin, line 244: nit: I don't really understand use of "Uses" here… so if you have some suggestions, I could update this as well

JakeWharton (approved):

PR #5710, opened 2017-11-08
1.x: Single.subscribe should not crash onSuccess into onError

akarnokd:
Fix `Single.subscribe(Action1, Action1)` to not call `onError` if the `onSuccess` action crashes, ensuring that the `Single` protocol `onSuccess|onError` is maintained. Crashes will go into the `RxJavaHooks.onError` handler.

Fixes: #5237.

### DIFF
--- a/src/main/java/rx/Single.java
+++ b/src/main/java/rx/Single.java
@@ -1792,6 +1792,9 @@ public class Single<T> {
             public final void onError(Throwable e) {
                 try {
                     onError.call(e);
+                } catch (Throwable ex) {
+                    Exceptions.throwIfFatal(ex);
+                    RxJavaHooks.onError(new CompositeException(e, ex));
                 } finally {
                     unsubscribe();
                 }
@@ -1801,6 +1804,9 @@ public class Single<T> {
             public final void onSuccess(T args) {
                 try {
                     onSuccess.call(args);
+                } catch (Throwable ex) {
+                    Exceptions.throwIfFatal(ex);
+                    RxJavaHooks.onError(ex);
                 } finally {
                     unsubscribe();
                 }
--- a/src/test/java/rx/CompletableTest.java
+++ b/src/test/java/rx/CompletableTest.java
@@ -2827,14 +2827,27 @@ public class CompletableTest {
         });
     }
 
-    @Test(expected = OnErrorNotImplementedException.class)
+    @Test
     public void propagateExceptionSubscribeOneActionThrowFromOnSuccess() {
-        normal.completable.toSingleDefault(1).subscribe(new Action1<Integer>() {
+        final List<Throwable> list = Collections.synchronizedList(new ArrayList<Throwable>());
+        RxJavaHooks.setOnError(new Action1<Throwable>() {
             @Override
-            public void call(Integer integer) {
-                throw new TestException();
+            public void call(Throwable t) {
+                list.add(t);
             }
         });
+        try {
+            normal.completable.toSingleDefault(1).subscribe(new Action1<Integer>() {
+                @Override
+                public void call(Integer integer) {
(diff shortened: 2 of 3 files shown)


vanniktech (approved):

src/main/java/rx/Single.java
```diff
@@ -1801,6 +1804,9 @@ public final void onError(Throwable e) {
             public final void onSuccess(T args) {
                 try {
                     onSuccess.call(args);
+                } catch (Throwable ex) {
+                    Exceptions.throwIfFatal(ex);
+                    RxJavaHooks.onError(ex);
```
artem-zinnatullin, line 1809: Hm, that's not what Observable does in 1.x, I think we should call `Exceptions.throwOrReport()`, in 1.x we didn't call hooks/plugins in such cases

See 

https://github.com/ReactiveX/RxJava/blob/5b2394c9ee91f298661fff5e043744c84b425808/src/main/java/rx/observers/SafeSubscriber.java#L139

https://github.com/ReactiveX/RxJava/blob/5b2394c9ee91f298661fff5e043744c84b425808/src/main/java/rx/exceptions/Exceptions.java#L222

src/test/java/rx/CompletableTest.java
```diff
@@ -2827,14 +2827,27 @@ public void call(Integer integer) {
         });
     }
 
-    @Test(expected = OnErrorNotImplementedException.class)
```
artem-zinnatullin, line 2830: This is a breaking change that can break lots of unit tests and change runtime behavior

artem-zinnatullin (changes requested):
This looks like breaking and inconsistent (with 1.x Observable) change…

src/main/java/rx/Single.java
akarnokd: Calling that would just bring back the problematic behavior, i.e., calling `onError` when `onSuccess` crashes.

src/test/java/rx/CompletableTest.java
akarnokd: The issue reported an inconsistency: synchronous crash would throw synchronously but an asynchronous crash would get swallowed.

artem-zinnatullin: That is expected for 1.x, this is how it always worked

On Android this will crash the app btw, because default Thread.UncaughtExceptionHandler prints exception and exits the app with dialog

User can always hook into that by overriding default Thread.UncaughtExceptionHandler

akarnokd: In this case, there can't be a fix for the issue reported in #5237; I suggest **wont-fix** there.

artem-zinnatullin: We still can try to sync try/catch logic with Observable's `SafeSubscriber.onCompleted`, basically do something like:

```
Exceptions.throwIfFatal(e);
RxJavaHooks.onError(e);
throw new OnCompletedFailedException(e.getMessage(), e); // Use different exception type here
```

akarnokd: We'd need a public `OnSuccessFailedException` but v1 is in feature freeze. Also I'm not sure if throwing a different exception than what the lambda threw would solve the original problem.

artem-zinnatullin: Yeah I see, maybe just add throwIfFatal then?

akarnokd: And swallow the non-fatal exceptions? That would still break this particular test.

artem-zinnatullin: Ah, sorry, of course current behavior includes throwing fatals, no changes required then, I guess.

PR #5715, opened 2017-11-09
2.x: distinguish between sync and async dispose in ScheduledRunnable

### LINKED ISSUE #5711: Premature thread interruption when using observeOn operator
When using RxJava 2.1.3 to 2.1.6, we observed a possible racing condition which caused the observeOn thread to be interrupted prematurely.

A sample code snippet could be the following (not a complete example):
```java
Observable.fromPublisher(publisher)
.firstElement()
.map(operation)
.blockingGet();
```

In the above example, observeOn operator would create a ObservableObserveOn instance. When there is the first item coming (ObservableObserveOn.ObserveOnObserver.onNext() is called), a Runnable is scheduled via the NewThreadWorker.scheduleActual(). Inside NewThreadWorker.scheduleActual(), a ScheduledRunnable is created wrapping the given Runnable, and submit the ScheduledRunnable to the executor for execution. Then the future returned by the executor is set into ScheduledRunnable via setFuture method.

The problem lies when the ScheduledRunnable.dispose() method is called between submitting to the executor and ScheduledRunnable.setFuture method by the underlying Runnable wrapped in ScheduledRunnable (which is executed in parallel by the executor). In this case, the Runnable contains the firstElement() operator would create a ObservableElementAtMaybe instance. ObservableElementAtMaybe.onNext() method would call the dispose() method before calling actual.onSuccess(). So combining these, if ScheduledRunnable.setFuture is called after ScheduledRunnable.dispose() in ObservableElementAtMaybe.onNext(), the actual.onSuccess() would be interrupted, which is not what we want.

**tl;dr** In the sample code, if the execution order is the following, it would create a thread interruption that interfere with the processing of the chained operators.
```
Thread 1: future = executor.submit(ScheduledRunnable)
Thread 2 (executor thread): ScheduledRunnable.dispose()
Thread 1: ScheduledRunnable.setFuture(future)
Thread 2: actual callback (e.g. actual.onSuccess()) <-- Thread interrupted. Throws exception when we call some synchronization methods in a non-blocking way (e.g. Semaphore.tryAcquire(0, TimeUnit.SECONDS)). 
```

A possible solution is to change the order, i.e. call actual.onSuccess() before dispose(). I searched the code base and found out that in general, dispose() is called before the actual callback. I am not sure if there are other concerns with this specific ordering (i.e. dispose before actual callback), but it certainly sounds odd to me.

akarnokd:
This PR adds logic to distinguish between synchronous and asynchronous dispose calls when `setFuture` is executing. It should prevent interrupting the currently running task body if it requested cancellation indirectly before the `setFuture` was executed by the Thread which scheduled the task.

Fixes #5711

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
+++ b/src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
@@ -26,7 +26,12 @@ implements Runnable, Callable<Object>, Disposable {
     private static final long serialVersionUID = -6120223772001106981L;
     final Runnable actual;
 
-    static final Object DISPOSED = new Object();
+    /** Indicates that the parent tracking this task has been notified about its completion. */
+    static final Object PARENT_DISPOSED = new Object();
+    /** Indicates the dispose() was called from within the run/call method. */
+    static final Object SYNC_DISPOSED = new Object();
+    /** Indicates the dispose() was called from another thread. */
+    static final Object ASYNC_DISPOSED = new Object();
 
     static final Object DONE = new Object();
 
@@ -66,13 +71,13 @@ implements Runnable, Callable<Object>, Disposable {
         } finally {
             lazySet(THREAD_INDEX, null);
             Object o = get(PARENT_INDEX);
-            if (o != DISPOSED && o != null && compareAndSet(PARENT_INDEX, o, DONE)) {
+            if (o != PARENT_DISPOSED && o != null && compareAndSet(PARENT_INDEX, o, DONE)) {
                 ((DisposableContainer)o).delete(this);
             }
 
             for (;;) {
                 o = get(FUTURE_INDEX);
-                if (o == DISPOSED || compareAndSet(FUTURE_INDEX, o, DONE)) {
+                if (o == SYNC_DISPOSED || o == ASYNC_DISPOSED || compareAndSet(FUTURE_INDEX, o, DONE)) {
                     break;
                 }
             }
@@ -85,8 +90,12 @@ implements Runnable, Callable<Object>, Disposable {
             if (o == DONE) {
                 return;
             }
-            if (o == DISPOSED) {
-                f.cancel(get(THREAD_INDEX) != Thread.currentThread());
+            if (o == SYNC_DISPOSED) {
+                f.cancel(false);
+                return;
+            }
+            if (o == ASYNC_DISPOSED) {
+                f.cancel(true);
                 return;
             }
             if (compareAndSet(FUTURE_INDEX, o, f)) {
@@ -99,12 +108,13 @@ implements Runnable, Callable<Object>, Disposable {
     public void dispose() {
         for (;;) {
             Object o = get(FUTURE_INDEX);
-            if (o == DONE || o == DISPOSED) {
+            if (o == DONE || o == SYNC_DISPOSED || o == ASYNC_DISPOSED) {
                 break;
             }
-            if (compareAndSet(FUTURE_INDEX, o, DISPOSED)) {
+            boolean async = get(THREAD_INDEX) != Thread.currentThread();
+            if (compareAndSet(FUTURE_INDEX, o, async ? ASYNC_DISPOSED : SYNC_DISPOSED)) {
                 if (o != null) {
-                    ((Future<?>)o).cancel(get(THREAD_INDEX) != Thread.currentThread());
+                    ((Future<?>)o).cancel(async);
                 }
                 break;
             }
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
```diff
@@ -26,7 +26,12 @@
     private static final long serialVersionUID = -6120223772001106981L;
     final Runnable actual;
 
-    static final Object DISPOSED = new Object();
+    /** Indicates that the parent tracking this task has been notified about its completion. */
+    static final Object PARENT_DISPOSED = new Object();
+    /** Indicates the dispose() was called from within the run/call method. */
+    static final Object SYNC_DISPOSED = new Object();
+    /** Indicates the dispose() was called from another thread. */
+    static final Object ASYNC_DISPOSED = new Object();
```
artem-zinnatullin, line 34: supernit: `ASYNC` doesn't mean another thread :)

But for name conciseness I guess it's fine

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
```diff
@@ -66,13 +71,13 @@ public void run() {
         } finally {
             lazySet(THREAD_INDEX, null);
             Object o = get(PARENT_INDEX);
-            if (o != DISPOSED && o != null && compareAndSet(PARENT_INDEX, o, DONE)) {
+            if (o != PARENT_DISPOSED && compareAndSet(PARENT_INDEX, o, DONE) && o != null) {
```
artem-zinnatullin, line 74: Change of `compareAndSet` position in `if` statement seems to be a behavior change…

Though it doesn't affect `ScheduledRunnableTest` on my machine

Do we need that change?

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
```diff
@@ -99,23 +108,24 @@ public void setFuture(Future<?> f) {
     public void dispose() {
         for (;;) {
             Object o = get(FUTURE_INDEX);
-            if (o == DONE || o == DISPOSED) {
+            if (o == DONE || o == SYNC_DISPOSED || o == ASYNC_DISPOSED) {
                 break;
             }
-            if (compareAndSet(FUTURE_INDEX, o, DISPOSED)) {
+            boolean async = get(THREAD_INDEX) != Thread.currentThread();
+            if (compareAndSet(FUTURE_INDEX, o, async ? ASYNC_DISPOSED : SYNC_DISPOSED)) {
```
artem-zinnatullin, line 115: Hm, what worries me here is that we base `boolean async` on `get(THREAD_INDEX) != Thread.currentThread()`, but in `finally` of `run()` we set `Thread` to null which always gonna give `async == false` 

https://github.com/ReactiveX/RxJava/blob/b99efbfb71d3a76befccd4e36324cefadeefd65a/src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java#L67

Yes, it's `lazySet()`, but aren't we creating unwanted race condition here between `run()` and `dispose()`?

The worst scenario is that if `dispose()` indeed was called from another thread but `lazySet()` memory write becomes visible to it, we'll have `async == false` which is actually wrong in this scenario…

artem-zinnatullin (commented):
LGTM, but comments

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
akarnokd: Yes. With the previous order, if `o` was `null` because of lack of a parent, the CAS didn't happen and the task would partially appear to be still active. With the sync-async marker changes, `isDisposed` was switched to check the parent reference for indication of being disposed instead of the future reference, which required now 3 comparisons instead of 2.

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
akarnokd: Cancelling from the current thread could only happen while the `run` callback is executing. Any other cancellation that comes after this can only be asynchronous. It doesn't matter what value is in `THREAD_INDEX` at this time because that will not be the same thread as the caller's. 

Since we are running with single-threaded thread pools, any subsequent task on the same pool having the same Thread will find the references marked as `DONE` and not cancel the previous task. This is guaranteed by the sequential task processing of the thread pool.

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
artem-zinnatullin: Makes sense, can you please add a test to cover that then? 

When I changed order back to original one on your branch, whole `ScheduledRunnableTest` passed

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
artem-zinnatullin: >It doesn't matter what value is in THREAD_INDEX at this time because that will not be the same thread as the caller's.

Aha, I see now, 👍

artem-zinnatullin (approved):

src/main/java/io/reactivex/internal/schedulers/ScheduledRunnable.java
akarnokd: Sure.

akarnokd: Unit tests added.

artem-zinnatullin (approved):

vanniktech (approved):

Issue #5716, opened 2017-11-10
onComplete() is not getting called on the specified Scheduler


sadegh:
when I use `Observable.take(time, TimeUnit.SECONDS)`, if the result Observable stops due to timeout, `onComplete()` callback gets called on `computation` thread,  not the one that I've specified with `observeOn()`.

Here is a sample code:
```
Observable.just("one", "two")
        .observeOn(Schedulers.io())
        .take(0, TimeUnit.SECONDS)
        .subscribe(new Observer<String>() {
            @Override public void onSubscribe(Disposable d) {
                System.out.println("onSubscribe(), thread: " + Thread.currentThread().getName());
            }

            @Override public void onNext(String s) {
                System.out.println("onNext(), thread: " + Thread.currentThread().getName());
            }

            @Override public void onError(Throwable e) {
                System.out.println("onError(), thread: " + Thread.currentThread().getName());
            }

            @Override public void onComplete() {
                System.out.println("onComplete, thread: " + Thread.currentThread().getName());
            }
        });
```
And as the result you'll get `onComplete, thread: RxComputationThreadPool-1`.

If it's required for all callback methods to run on the specified thread per Reactive specifications, this would be a bug, and it happens on both RxJava 1.x and 2.x

akarnokd:
`observeOn` lasts until the next async operator, in this case [take](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#take-long-java.util.concurrent.TimeUnit-). You can [specify](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#take-long-java.util.concurrent.TimeUnit-io.reactivex.Scheduler-) the Scheduler where `take` will time out. Alternatively, you can swap `take` and `observeOn` so events will end up on the desired thread.

sadegh:
Yeah, I looked into the code and figured what's causing that. However I believe this is kinda inconsistent behavior, (and probably wrong), since all `onNext()` calls are running on the specified scheduler, and all of the sudden `onComplete()` runs on another thread.

At least we should have this behavior documented. It's not known that `take()` **might** change thread for some callbacks! (`onError` is also possible to run on `computation`)

sadegh:
I also looked into the source code and figured why this is happening. 
Here is my quick (and probably naive!) fix that I did to just test my idea about the source of this issue.
I added an `if` to `Observable.take()`

```
public final Observable<T> take(long time, TimeUnit unit) {
[...]

PR #5719, opened 2017-11-11
1.x: Add a sentence to documentation of take() operator

sadegh:
#5716

### DIFF
--- a/src/main/java/rx/Observable.java
+++ b/src/main/java/rx/Observable.java
@@ -10535,6 +10535,8 @@ public class Observable<T> {
     /**
      * Returns an Observable that emits those items emitted by source Observable before a specified time runs
      * out.
+     * If time runs out before Observable is finished, termination events would be called on provided {@link Scheduler},
+     + which is {@code computation} by default.
      * <p>
      * <img width="640" height="305" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/take.t.png" alt="">
      * <dl>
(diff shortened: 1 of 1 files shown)


src/main/java/rx/Observable.java
```diff
@@ -10535,6 +10535,8 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
     /**
      * Returns an Observable that emits those items emitted by source Observable before a specified time runs
      * out.
+     * If time runs out before Observable is finished, termination events would be called on provided {@link Scheduler},
```
akarnokd: A `<p>` would cleanly separate the first sentence and the rest of the Javadoc. Also the wording could use some refinement:

```
<p>
If time runs out before the {@code Observable} completes normally, the {@code onComplete} 
event will be signaled on the default {@code computation} {@link Scheduler}.
```

src/main/java/rx/Observable.java
```diff
@@ -10559,6 +10561,8 @@ public final Subscription subscribe(Subscriber<? super T> subscriber) {
     /**
      * Returns an Observable that emits those items emitted by source Observable before a specified time (on a
      * specified Scheduler) runs out.
+     * If time runs out before Observable is finished, termination events would be called on provided {@link Scheduler},
```
akarnokd:
```
<p>
If time runs out before the {@code Observable} completes normally, the {@code onComplete} 
event will be signaled on the provided {@link Scheduler}.
```

akarnokd (changes requested):

vanniktech (commented):
Agreeing with David

vanniktech (approved):

akarnokd (approved):

PR #5724, opened 2017-11-15
2.x: Add toMap overload with a merge function for updating existing mappings

philleonard:
Equivalent to the `java.util.stream.Collectors#toMap` merge map functionality. Could compliment the other map collectors nicely. There is an argument that this isn't suitable because we can't delegate to `Map#merge` with a language level lower than 8, but wanted to get your thoughts. Thanks!

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -12545,6 +12545,85 @@ public abstract class Observable<T> implements ObservableSource<T> {
     }
 
     /**
+     * Returns a Single that emits a single HashMap containing values corresponding to items emitted by the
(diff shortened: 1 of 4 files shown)


akarnokd:
As you can see, the existing `toMap` is actually implemented over the `collect` operator and are provided for convenience and historical reasons. I don't think RxJava 2 should be extended in this direction since, given Java 8, the proposed custom collector-combiner operation can be easily implemented with lambdas and `HashMap` functionalities.

To avoid spending time on PRs that will likely be not accepted, please post a regular issue first the next time about the proposed extension or feature.

akarnokd:
Thanks for contributing but it seems this PR hasn't sparked interest from the other regulars, therefore, I'm closing this for now.

PR #5729, opened 2017-11-16
2.x: Implement as()

### LINKED ISSUE #5654: 3.x: Proposal: `ObservableConverter` interface and friends
For the `compose()` operators, `ObservableTransformer` was introduced to allow for classes to implement multiple transformer interfaces for composite implementations. I'd like to propose doing the same for the `to()` operator by introducing `ObservableConverter` (and corresponding ones for others).  This would allow for composite converters as well.

If you're open to this, I can contribute a PR!

ZacSweers:
This implement `as()` support as discussed in #5654

I took the opportunity to try to standardize the docs and tests for it (which vary a little bit across implementations of `to()`)

Related: #5654

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -914,7 +914,7 @@ public abstract class Completable implements CompletableSource {
      * This allows fluent conversion to any other type.
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
-     *  <dd>{@code to} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dd>{@code as} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
      * @param <R> the resulting object type
@@ -925,7 +925,7 @@ public abstract class Completable implements CompletableSource {
     @Experimental
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final <R> R as(CompletableConverter<? extends R> converter) {
+    public final <R> R as(@NonNull CompletableConverter<? extends R> converter) {
         try {
             return ObjectHelper.requireNonNull(converter, "converter is null").apply(this);
         } catch (Throwable ex) {
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -927,12 +927,7 @@ public abstract class Completable implements CompletableSource {
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <R> R as(@NonNull CompletableConverter<? extends R> converter) {
-        try {
-            return ObjectHelper.requireNonNull(converter, "converter is null").apply(this);
-        } catch (Throwable ex) {
-            Exceptions.throwIfFatal(ex);
-            throw ExceptionHelper.wrapOrThrow(ex);
-        }
+        return ObjectHelper.requireNonNull(converter, "converter is null").apply(this);
     }
 
     /**
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -921,6 +921,7 @@ public abstract class Completable implements CompletableSource {
      * @param converter the function that receives the current Completable instance and returns a value
      * @return the converted value
      * @throws NullPointerException if converter is null
+     * @since 2.1.7 - experimental
      */
     @Experimental
     @CheckReturnValue
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -922,6 +922,7 @@ public abstract class Completable implements CompletableSource {
      * @return the converted value
      * @throws NullPointerException if converter is null
      */
+    @Experimental
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <R> R as(CompletableConverter<? extends R> converter) {
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -909,6 +909,31 @@ public abstract class Completable implements CompletableSource {
     }
 
     /**
+     * Calls the specified converter function during assembly time and returns its resulting value.
+     * <p>
+     * This allows fluent conversion to any other type.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code to} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <R> the resulting object type
+     * @param converter the function that receives the current Completable instance and returns a value
+     * @return the converted value
+     * @throws NullPointerException if converter is null
+     */
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final <R> R as(CompletableConverter<? extends R> converter) {
+        try {
+            return ObjectHelper.requireNonNull(converter, "converter is null").apply(this);
+        } catch (Throwable ex) {
+            Exceptions.throwIfFatal(ex);
+            throw ExceptionHelper.wrapOrThrow(ex);
+        }
+    }
+
+    /**
      * Subscribes to and awaits the termination of this Completable instance in a blocking manner and
      * rethrows any exception emitted.
      * <dl>
--- a/src/main/java/io/reactivex/CompletableConverter.java
+++ b/src/main/java/io/reactivex/CompletableConverter.java
@@ -13,7 +13,7 @@
 
 package io.reactivex;
 
-import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.*;
 
 /**
  * Convenience interface and callback used by the {@link Completable#as} operator to turn a Completable into another
@@ -21,6 +21,7 @@ import io.reactivex.annotations.NonNull;
  *
  * @param <R> the output type
  */
+@Experimental
 public interface CompletableConverter<R> {
     /**
      * Applies a function to the upstream Completable and returns a converted value of type <R>.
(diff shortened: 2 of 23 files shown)


ZacSweers:
Let me know if you want to add more tests. Some usages of `to()` were mixed in other tests and wasn't sure if they were specifically targeting testing it.

ZacSweers:
I'm not sure I understand the remaining CI errors, as the parameters have the NonNull annotation

src/main/java/io/reactivex/Completable.java
```diff
@@ -909,6 +909,32 @@ public final Completable andThen(CompletableSource next) {
...
+     *
+     * @param <R> the resulting object type
+     * @param converter the function that receives the current Completable instance and returns a value
+     * @return the converted value
+     * @throws NullPointerException if converter is null
+     */
+    @Experimental
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final <R> R as(@NonNull CompletableConverter<? extends R> converter) {
```
akarnokd, line 929: Please add `@Experimental` and `@since 2.1.7 - experimental` to all of them.

src/main/java/io/reactivex/CompletableConverter.java
```diff
@@ -0,0 +1,35 @@
...
+package io.reactivex;
+
+import io.reactivex.annotations.*;
+
+/**
+ * Convenience interface and callback used by the {@link Completable#as} operator to turn a Completable into another
+ * value fluently.
+ *
+ * @param <R> the output type
+ */
```
akarnokd: Please add `@since 2.1.7 - experimental` to these as well.

akarnokd:
Some style checks failed: https://travis-ci.org/ReactiveX/RxJava/builds/302962901#L974

Please add `as` to `ParallelFlowable`.

Please add a unit test that has a converter class which combines all interfaces and is applied to the 6 types of sources.

ZacSweers:
Could you explain what the style checks fix is? I didn't quite understand the message in the log

src/main/java/io/reactivex/Completable.java
ZacSweers: 1914776

ZacSweers:
ParallelFlowable - 8adb583

Started on more converter tests in `ConverterTest` in c1f26ee, composite converter test is in there. Only thing I'm stuck on is the `GenericsSignatureTest` tests (borrowed from `TransformersTest`). Currently they don't compile, but I'm not sure what the goal is (or if it even makes sense to match the checks `TransformersTest` does). Let me know what you think

akarnokd:
You have to extend the `ParamValidationCheckerTest` with instances of your new converter types. Put something like the following into L562.

```java
defaultValues.put(ObservableConverter.class, new ObservableConverter() {
    @Override public Object apply(Observable o) { return o; }
});
```

akarnokd:
As for [these](https://travis-ci.org/ReactiveX/RxJava/builds/302995534#L495), you are using the wrong generic types.

~~`ObservableConverter<A<T, ?>, B<T>>` is applied to `Observable<Integer>` in the test where `Integer` is unrelated to `A<T, ?>`.~~

I overlooked the test using `A`. I'll check code locally.

akarnokd:
The wildcard definition in those compile errors and the wrong `<Integer>` parameter confuses the compiler. I suggest going raw-types with these converters and using `SuppressWarnings({"unchecked", "rawtypes"})` on the problematic tests.

akarnokd:
Posted a [patch](https://github.com/hzsweers/RxJava/pull/2) for your PR.

akarnokd (approved):

src/main/java/io/reactivex/CompletableConverter.java
```diff
@@ -0,0 +1,36 @@
...
+public interface CompletableConverter<R> {
+    /**
+     * Applies a function to the upstream Completable and returns a converted value of type {@code R}.
+     *
+     * @param upstream the upstream Completable instance
+     * @return the converted value
+     * @throws Exception on error
+     */
+    @NonNull
+    R apply(@NonNull Completable upstream) throws Exception;
```
akarnokd: The transformer variants don't define any `throws Exception`. I'd think we can live without them in these interfaces as well and thus no need to `try-catch` apply in the base classes' `as()` methods.

ZacSweers: 5c0dec5

ZacSweers:
Out of curiosity, would this be a candidate for `@Beta` in 2.2.0 since it's so close? Or would it stay `@Experimental` for a time period after?

akarnokd:
2.2.0 is a bit in the future as I'd like to have #5319 finished before it. An experimental component can be promoted to stable regardless.

ZacSweers:
Cool :+1:

src/test/java/io/reactivex/ParamValidationCheckerTest.java
```diff
@@ -560,6 +560,46 @@ public void checkParallelFlowable() {
 
         defaultValues.put(ParallelFailureHandling.class, ParallelFailureHandling.ERROR);
 
+        @SuppressWarnings("rawtypes")
+        class MixedConverters implements FlowableConverter, ObservableConverter, SingleConverter,
```
vanniktech, line 564: Should we rename this to IdentityConverters?

src/test/java/io/reactivex/completable/CompletableTest.java
```diff
@@ -2796,11 +2796,43 @@ public void accept(Object e) { }
...
+    @Test(timeout = 5000)
+    public void asNormal() {
+        Flowable<Object> flow = normal.completable.as(new CompletableConverter<Flowable<Object>>() {
+            @Override
+            public Flowable<Object> apply(Completable c) {
+                return c.toFlowable();
+            }
+        });
+
+        flow.blockingForEach(new Consumer<Object>() {
```
vanniktech: this does not assertion is this wanted?

src/test/java/io/reactivex/flowable/FlowableTests.java
```diff
@@ -1125,6 +1126,34 @@ public Object apply(Flowable<Object> onSubscribe) {
...
+    @Test
+    public void testAsExtend() {
+        final TestSubscriber<Object> subscriber = new TestSubscriber<Object>();
+        final Object value = new Object();
+        Flowable.just(value).as(new FlowableConverter<Object, Object>() {
+            @Override
+            public Object apply(Flowable<Object> onSubscribe) {
+                    onSubscribe.subscribe(subscriber);
+                    subscriber.assertNoErrors();
+                    subscriber.assertComplete();
```
vanniktech, line 1139: Could not this run through without asserting if the Converter is never called?

src/test/java/io/reactivex/observable/ObservableTest.java
```diff
@@ -1165,6 +1165,34 @@ public Object apply(Observable<Object> onSubscribe) {
...
     }
 
+    @Test
+    public void testAsExtend() {
+        final TestObserver<Object> subscriber = new TestObserver<Object>();
+        final Object value = new Object();
+        Observable.just(value).as(new ObservableConverter<Object, Object>() {
+            @Override
+            public Object apply(Observable<Object> onSubscribe) {
+                onSubscribe.subscribe(subscriber);
```
vanniktech: same

vanniktech (commented):
Should the copy right for the new files contain 2016 or 2017?

src/test/java/io/reactivex/ParamValidationCheckerTest.java
ZacSweers: I don't feel strongly about it since this is just a stub for a test

src/test/java/io/reactivex/completable/CompletableTest.java
ZacSweers: Deferring to @akarnokd on this as I was matching the analogous `to()` test

src/test/java/io/reactivex/flowable/FlowableTests.java
ZacSweers: True... I could add something after this to assert that `subscriber` is complete?

vanniktech: How about an AtomicBoolean that you set to true once it's called and you fail the test when it's set to false

ZacSweers: I think I got a better one - 5c3bdf3

src/test/java/io/reactivex/observable/ObservableTest.java
ZacSweers: 5c3bdf3

src/test/java/io/reactivex/completable/CompletableTest.java
akarnokd: Please change these into proper assertions such as `.test().assertResult()`.

ZacSweers: What would I assert for these since there's no value? Or just rewrite the rest in general?

akarnokd: But there is the `onComplete` event.

ZacSweers: there is but `completable.normal` doesn't complete. I wasn't sure if you wanted me to change it to just use `Completable.complete()` or something similar since the original test didn't seem to be testing completion

akarnokd: `normal.completable` [does complete](https://github.com/hzsweers/RxJava/blob/5c3bdf37cd587df0ed4be769ede29e81f7b17d2f/src/test/java/io/reactivex/completable/CompletableTest.java#L110) otherwise `blockingForEach` would block indefinitely and the test would have failed a long a go. A `.test().assertResult(/* no values */)` should suffice.

ZacSweers: oops, I missed that line 🙄

Done in e0d793f

ZacSweers:
Thanks!!

PR #5741, opened 2017-11-24
2.x: internal API to get distinct Workers from some Schedulers

akarnokd:
This PR adds an internal interface `SchedulerMultiWorkerSupport` that allows retrieving multiple workers from a `Scheduler` that implements this interface.

The standard `Scheduler.getWorker()` can be invoked as many times as necessary, but specific implementations such as the `computation()` `Scheduler` and the [`ParallelScheduler`](https://github.com/akarnokd/RxJava2Extensions#parallelscheduler) is not guaranteed to return workers that are backed by distinct single-threaded thread pools. 

This does not effect other scheduler types because: 
- they are single threaded (`single()`) or don't use threads at all (`trampoline()`) and 
- already hand out distinct workers (`io()`, `newThread()`).

Such worker reuse can happen when in a highly concurrent application, typical tasks are mixed with parallel tasks and both pull out workers from these `Scheduler`s. If this is happens, it is possible there will be duplicate threads used by the parallel operations and thus not utilize the originally intended parallelism level.

By implementing this suggested interface, a batch-retrieval can be supported by the `Scheduler`s and they can make sure the caller gets as many distinct thread-pool as possible. If more workers are requested than the `Scheduler`'s parallelism, the workers are handed out in round-robin fashion similar to the standard `createWorker()`.

Why a callback instead of returning an array? 
- even if both require an allocation to set up, there is no need to have all workers visible at once,
- simply less memory usage, and
- avoids looping twice: once for filling in the array and once for using the elements of the array.

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/ComputationScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/ComputationScheduler.java
@@ -15,19 +15,20 @@
  */
 package io.reactivex.internal.schedulers;
 
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReference;
+
 import io.reactivex.Scheduler;
 import io.reactivex.annotations.NonNull;
 import io.reactivex.disposables.*;
 import io.reactivex.internal.disposables.*;
-
-import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicReference;
+import io.reactivex.internal.functions.ObjectHelper;
 
 /**
  * Holds a fixed pool of worker threads and assigns them
  * to requested Scheduler.Workers in a round-robin fashion.
  */
-public final class ComputationScheduler extends Scheduler {
+public final class ComputationScheduler extends Scheduler implements SchedulerMultiWorkerSupport {
     /** This will indicate no pool is active. */
     static final FixedSchedulerPool NONE;
     /** Manages a fixed number of workers. */
@@ -67,7 +68,7 @@ public final class ComputationScheduler extends Scheduler {
         return paramThreads <= 0 || paramThreads > cpuCount ? cpuCount : paramThreads;
     }
 
-    static final class FixedSchedulerPool {
+    static final class FixedSchedulerPool implements SchedulerMultiWorkerSupport {
         final int cores;
 
         final PoolWorker[] eventLoops;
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/internal/schedulers/ComputationScheduler.java
```diff
@@ -96,6 +97,25 @@ public void shutdown() {
                 w.dispose();
             }
         }
+
+        @Override
+        public void createWorkers(int number, WorkerCallback callback) {
+            int c = cores;
```
artem-zinnatullin, line 103: `cores` is `final` in this context, why copy it to local var?

src/main/java/io/reactivex/internal/schedulers/SchedulerMultiWorkerSupport.java
```diff
@@ -0,0 +1,52 @@
...
+     * @param number the number of workers to create, positive
+     * @param callback the callback to send worker instances to
+     */
+    void createWorkers(int number, @NonNull WorkerCallback callback);
+
+    /**
+     * The callback interface for the {@link SchedulerMultiWorkerSupport#createWorkers(int, WorkerCallback)}
+     * method.
+     */
+    interface WorkerCallback {
```
artem-zinnatullin, line 44: `WorkerCreatedCallback`?

src/main/java/io/reactivex/internal/schedulers/SchedulerMultiWorkerSupport.java
```diff
@@ -0,0 +1,52 @@
...
+        /**
+         * Called with the Worker index and instance.
+         * @param index the worker index, zero-based
+         * @param worker the worker instance
+         */
+        void onWorker(int index, @NonNull Scheduler.Worker worker);
```
artem-zinnatullin, line 50: `onCreated`?

artem-zinnatullin (approved):
Rename PR to "Internal API to …"?

src/main/java/io/reactivex/internal/schedulers/ComputationScheduler.java
akarnokd: With potential volatiles around, this won't re-read the field every time it is needed in the loop below.

src/main/java/io/reactivex/internal/schedulers/SchedulerMultiWorkerSupport.java
akarnokd: While still it is an internal API, I don't think we need a more complicated naming.

artem-zinnatullin:
:+1:

vanniktech (approved):

src/main/java/io/reactivex/internal/schedulers/SchedulerMultiWorkerSupport.java
davidmoten: What's the use case for the index?

davidmoten (commented):
Good idea, will be useful.

src/main/java/io/reactivex/internal/schedulers/SchedulerMultiWorkerSupport.java
akarnokd: In `ParallelRunOn`, it let's index into the `parent` and `subscribers` arrays without the need for additional state, such as counting how many times `onWorker` was invoked.

davidmoten: ok, potentially reduces allocations for the user, thanks

davidmoten:
The PR has already been merged but for the record I approve too

akarnokd:
Thanks for the reviews @artem-zinnatullin @vanniktech & @davidmoten !

PR #5748, opened 2017-11-30
2.x: Check runnable == null in *Scheduler.schedule*().

artem-zinnatullin:
To enforce `@NotNull` guarantee for the PR #5734, see  https://github.com/ReactiveX/RxJava/pull/5734/files#r153951238

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/TrampolineScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/TrampolineScheduler.java
@@ -49,6 +49,8 @@ public final class TrampolineScheduler extends Scheduler {
     @NonNull
     @Override
     public Disposable scheduleDirect(@NonNull Runnable run) {
+        // TODO remove if https://github.com/ReactiveX/RxJava/pull/5747 gets merged.
+        ObjectHelper.requireNonNull(run, "runnable is null");
         run.run();
         return EmptyDisposable.INSTANCE;
     }
@@ -56,6 +58,8 @@ public final class TrampolineScheduler extends Scheduler {
     @NonNull
     @Override
     public Disposable scheduleDirect(@NonNull Runnable run, long delay, TimeUnit unit) {
+        // TODO remove if https://github.com/ReactiveX/RxJava/pull/5747 gets merged.
+        ObjectHelper.requireNonNull(run, "runnable is null");
         try {
             unit.sleep(delay);
             run.run();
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/plugins/RxJavaPluginsTest.java
```diff
@@ -1381,10 +1379,7 @@ public void subscribeActual(MaybeObserver t) {
             assertSame(myb, RxJavaPlugins.onAssembly(myb));
 
 
-            assertNull(RxJavaPlugins.onSchedule(null));
```
artem-zinnatullin, line 1384: Not sure what was the reason to check null twice here

vanniktech (approved):

src/main/java/io/reactivex/internal/schedulers/TrampolineScheduler.java
```diff
@@ -49,13 +49,17 @@ public Worker createWorker() {
     @NonNull
     @Override
     public Disposable scheduleDirect(@NonNull Runnable run) {
+        // TODO remove if https://github.com/ReactiveX/RxJava/pull/5747 gets merged.
+        ObjectHelper.requireNonNull(run, "runnable is null");
```
akarnokd: It's "run is null", but otherwise the `run.run()` below would have crashed anyway.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -446,6 +446,8 @@ public static Scheduler onNewThreadScheduler(@NonNull Scheduler defaultScheduler
      */
     @NonNull
     public static Runnable onSchedule(@NonNull Runnable run) {
+        ObjectHelper.requireNonNull(run, "runnable is null");
```
artem-zinnatullin: Forgot to add explanation to PR why I've added check here and not in `Scheduler.schedule*()` methods:

Reason behind that is abstract `Scheduler` has 3 overloads, one of them reuses another one + any concrete scheduler can override any of those methods and call other overloads, so as a result I had checks duplicated in some cases.

This `RxJavaPlugins` call is done by all schedulers before actual schedule/run of the original `Runnable` so it seems to be a good common point that only executed once and guarantees now that `Runnable` is not null for both user defined `RxJavaPlugins.onSchedule()` function and all RxJava schedulers.

akarnokd (dismissed):

akarnokd:
@artem-zinnatullin Could you rebase this PR and reapply your changes?

artem-zinnatullin:
Done, sorry for delay.

akarnokd (approved):

PR #5753, opened 2017-12-05
2.x: Add retry(times, predicate) to Single & Completable and verify behavior across them and Maybe.

vanniktech:
- Add retry(times, predicate) to Single & Completable (Maybe had it already) which just forwards to Flowable for now
- verify behavior across the three reactive types

Happy to get better testing names

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -1546,6 +1546,28 @@ public abstract class Completable implements CompletableSource {
     }
 
     /**
+     * Returns a Completable that when this Completable emits an error, retries at most times
+     * or until the predicate returns false, whichever happens first and emitting the last error.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code retry} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param times the number of times the returned Completable should retry this Completable
+     * @param predicate the predicate that is called with the latest throwable and should return
+     * true to indicate the returned Completable should resubscribe to this Completable.
+     * @return the new Completable instance
+     * @throws NullPointerException if predicate is null
+     * @throws IllegalArgumentException if times is negative
+     * @since 2.1.8 - experimental
+     */
+    @Experimental
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Completable retry(long times, Predicate<? super Throwable> predicate) {
+        return fromPublisher(toFlowable().retry(times, predicate));
+    }
+
+    /**
      * Returns a Completable that when this Completable emits an error, calls the given predicate with
      * the latest exception to decide whether to resubscribe to this or not.
      * <dl>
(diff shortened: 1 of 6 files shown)


akarnokd (approved):

vanniktech:
Given you've added this to the 2.2 milestone should I update the `@since` to `2.2.0` then?

akarnokd:
There is going to be a couple of 2.1.x till then. I don't want to release 2.2.0 until the marbles get fixed.

akarnokd (approved):

src/main/java/io/reactivex/Completable.java
```diff
@@ -1545,6 +1545,28 @@ public final Completable retry(long times) {
         return fromPublisher(toFlowable().retry(times));
     }
 
+    /**
+     * Returns a Completable that when this Completable emits an error, retries at most times
```
artem-zinnatullin, line 1549: nit: Phrasing seems a little bit off here, something like "Returns a Completable that retries when …" sounds more natural I guess

src/main/java/io/reactivex/Single.java
```diff
@@ -2649,6 +2649,26 @@ public final T blockingGet() {
         return toSingle(toFlowable().retry(predicate));
     }
 
+    /**
+     * Repeatedly re-subscribe at most times or until the predicate returns false, whichever happens first
```
artem-zinnatullin, line 2653: Let's make it more consistent with javadoc you've added for Completable? It didn't mention `Repeatedly` and was emphasizing error condition, while with this javadoc you only realize in the end of it that error is initial trigger for whole operator

src/test/java/io/reactivex/completable/CompletableRetryTest.java
```diff
@@ -0,0 +1,93 @@
...
+                throw new IllegalArgumentException();
+            }
+        })
+            .retry(Integer.MAX_VALUE, new Predicate<Throwable>() {
+                @Override public boolean test(final Throwable throwable) throws Exception {
+                    return !(throwable instanceof IllegalArgumentException);
+                }
+            })
+            .test()
+            .assertFailure(IllegalArgumentException.class);
```
artem-zinnatullin, line 48: hm, I believe you should check atomicInteger value here to verify actual number of action retries

artem-zinnatullin: Same below and in tests for other reactive types

artem-zinnatullin (changes requested):
Pls check actual number of retries in tests, otherwise LGTM :)

src/main/java/io/reactivex/Single.java
vanniktech: It should be yeah but if we only change this method then this one is inconsistent with all of the other Single methods.

src/main/java/io/reactivex/Completable.java
vanniktech: I kept this in sync with all of the other retry* java doc sentences.

akarnokd:
The latest changes broke the tests.

src/main/java/io/reactivex/Completable.java
artem-zinnatullin: uh, yeah all rx types have different phrasing for retry operator, alight

artem-zinnatullin (approved):
👍

PR #5760, opened 2017-12-11
2.x: Subject NPE fixes, add UnicastProcessor TCK

akarnokd:
This PR adds the `Subject.refCount()` and `FlowableProcessor.refCount()` that capture the upstream's `Disposable`/`Subscription` and disposes/cancels them if the number of `Observer`s/`Subscriber`s decreases to zero. The Reactive-Streams TCK and thus other implementations may expect such behavior from a `Processor` implementation and this wrapper is required to pass the TCK tests.

While implementing the TCK tests, it turned out the `null`-handling of the `Subject`s and `FlowableProcessor`s were not following the Reactive-Streams specification. They have to throw a `NullPointerException` immediately and not turn them into NPEs for the downstream. These classes and the tests have been fixed as well. Their error messages have been uniformed too.


vanniktech (approved):

src/main/java/io/reactivex/processors/BehaviorProcessor.java
```diff
@@ -175,10 +175,8 @@ public void onSubscribe(Subscription s) {
 
     @Override
     public void onNext(T t) {
-        if (t == null) {
-            onError(new NullPointerException("onNext called with null. Null values are generally not allowed in 2.x operators and sources."));
```
artem-zinnatullin, line 179: 😰😰😰 well this is a little bit breaking change, but since it was always part of RS spec I guess it falls into patch category for RxJava

artem-zinnatullin:
I like RS spec-related fixes 👍 

But not sure about motivation for having `refCount()` specific for processors/subjects, do you have some real-world use-case scenario on mind?

src/main/java/io/reactivex/processors/BehaviorProcessor.java
akarnokd: You were not supposed to call onNext with null anyway and there is no reason to take this extra long path to get the end subscriber a `NullPointerException` via `onError`.

akarnokd:
Other than such the behavior is needed to pass the TCK, none that I can think of. The original idea behind `Processor` is that it acts as a transformation step and would service at most 1 `Subscriber` by that. If that solo `Subscriber` cancels, it is reasonable that cancellation travels upstream. For us, we don't use `Processors` but basically chains of `Subscriber` -> `Subscriber` -> `Subscriber` where the mentioned cancellation is wired up nonetheless.

I can move the operator into the test section and not extend the `Subject`/`FlowableProcessor` classes.

src/main/java/io/reactivex/processors/BehaviorProcessor.java
artem-zinnatullin: Yep I understand :) 

Just noted that it's a little bit breaking behavior for existing users, but since RS doesn't allow nulls anyway we can make this change as "patch" rather than wait for 3.x

artem-zinnatullin:
>I can move the operator into the test section and not extend the Subject/FlowableProcessor classes

That sounds much better to me tbh, I just don't want us to decline PRs with new operators from users and suggest them contribute to rxjava-extras/etc and at the same time add new ones that no user really asked for

Nothing personal :)

akarnokd:
Okay, I'll have a copy of the processor version in test and move both into RxJava2Extensions.

artem-zinnatullin:
Thanks, David! Appreciate that

akarnokd:
Done.

src/test/java/io/reactivex/processors/FlowableProcessorTest.java
```diff
@@ -21,25 +23,29 @@
...
+            fail("No NullPointerException thrown");
+        } catch (NullPointerException ex) {
+            assertEquals("onError called with null. Null values are generally not allowed in 2.x operators and sources.", ex.getMessage());
+        }
 
-        p.test()
-                .assertNoValues()
-                .assertError(NullPointerException.class)
-                .assertErrorMessage("onError called with null. Null values are generally not allowed in 2.x operators and sources.");
+        p.test().assertEmpty().cancel();;
```
artem-zinnatullin: nit `;;` (and in other places in this PR)

src/test/java/io/reactivex/tck/RefCountProcessor.java
```diff
@@ -0,0 +1,221 @@
...
+import io.reactivex.processors.FlowableProcessor;
+
+/**
+ * A FlowableProcessor wrapper that disposes the Subscription set via
+ * onSubscribe if the number of subscribers reaches zero.
+ *
+ * @param <T> the upstream and downstream value type
+ * @since 2.1.8 - experimental
+ */
+@Experimental
```
artem-zinnatullin: nit: remove?

artem-zinnatullin (approved):
Just few small comments, otherwise lgtm 👍

artem-zinnatullin:
// Now started watching rxjava-extras, feel free to ping me if you need review on a PR there

artem-zinnatullin (approved):

akarnokd:
I put my excess things into [RxJava2Extensions](https://github.com/akarnokd/RxJava2Extensions) btw. [rxjava2-extras](https://github.com/davidmoten/rxjava2-extras) is @davidmoten 's repository.

artem-zinnatullin:
yep, I've mixed up naming but subscribed to yours. Somehow I've never needed to use neither of extensions/extras for all these years 🤔

PR #5761, opened 2017-12-12
2.x: fix timed exact buffer calling cancel unnecessarily

akarnokd:
This PR removes the unnnecessary `cancel`/`dispose` call from the exact-boundary timed `buffer()` operators in `Observable` and `Flowable` when the upstream completes normally.

Originally appeared in a [StackOverflow question](https://stackoverflow.com/questions/47772415/rxjava2-completing-a-cold-source-flowable-explicitly).


artem-zinnatullin (approved):

vanniktech (approved):
Huh this totally slipped on my side.

PR #5777, opened 2017-12-23
2.x: Fix race conditions in observeOn operators.

### LINKED ISSUE #5776: 2.x: Race conditions in (Single|Maybe|Completable).observeOn()
I found race conditions in (Single|Maybe|Completable).observeOn() operators.
This is a code sample to reproduce it:

https://github.com/yamasa/RxJava/commit/88e0d884d014c868347d0de9dd20ef128d1826ae

When I call observeOn() with a single-threaded scheduler (e.g. Schedulers.single(), AndroidSchedulers.mainThread()) and then call its subscription's dispose() method from the same thread, I expect onSuccess() is never called after the dispose().
I believe RxJava should guarantee executions on a single-threaded scheduler are sequentially consistent.

This is a similar problem to https://github.com/ReactiveX/RxJava/issues/5742 , but this occurs only in extremely rare cases.

note: observeOn() operator for Observable and Flowable doesn't have this issue.

yamasa:
We should check isDisposed() just before emitting results.

This PR fixes https://github.com/ReactiveX/RxJava/issues/5776

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/completable/CompletableObserveOn.java
+++ b/src/main/java/io/reactivex/internal/operators/completable/CompletableObserveOn.java
@@ -82,6 +82,10 @@ public final class CompletableObserveOn extends Completable {
 
         @Override
         public void run() {
+            if (isDisposed()) {
+                return;
+            }
+
             Throwable ex = error;
             if (ex != null) {
                 error = null;
--- a/src/main/java/io/reactivex/internal/operators/maybe/MaybeObserveOn.java
+++ b/src/main/java/io/reactivex/internal/operators/maybe/MaybeObserveOn.java
@@ -93,6 +93,10 @@ public final class MaybeObserveOn<T> extends AbstractMaybeWithUpstream<T, T> {
 
         @Override
         public void run() {
+            if (isDisposed()) {
+                return;
+            }
+
             Throwable ex = error;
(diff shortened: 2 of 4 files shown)


akarnokd:
There is only a very small window between the scheduler executing the run() and checking for isDisposed which was not worth doing. Besides, the unit test you provided already passes and is not the way to detect this probabilistic race condition.

davidmoten:
A significant pause before run is called is possible if the scheduler is
busy though. Doesn't that make this operator worthy of the suggested
treatment?

akarnokd:
I read #5776 after this PR. Actually, the OP refers to the small window when the task gets scheduled via scheduleDirect and the returned Disposable is atomically swapped into the tracking structure:

```java
    Disposable d = scheduler.scheduleDirect(this);
    // <-----------------------------------------------------
    DisposableHelper.replace(this, d);
```

If the thread initiating the scheduling is pausing at the specified place, the Disposable gets late and the other thread may execute run() in the meantime. This is a probabilistic race and besides, cancellation is best effort; how could you tell that it wasn't the dispose() call that got delayed and the events slipped through anyway? For harder cuts, there are the `onTerminateDetach` operators that can minimize such windows to an absolute minimum.

davidmoten:
Ah i see, thanks.

PR #5780, opened 2017-12-27
2.x: Improve BehaviorSubject JavaDoc + related clarifications

akarnokd:
This PR adds detailed behavior and feature description to the `BehaviorSubject` JavaDoc.

In addition, small mistakes in the `BehaviorProcessor` JavaDoc have also been fixed and both received an extra section: `Error handling` that points at the global error handler for undeliverable errors. Due to this link, the `RxJavaPlugins.onError` JavaDoc has been also expanded.

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -337,7 +337,23 @@ public final class RxJavaPlugins {
 
     /**
      * Called when an undeliverable error occurs.
+     * <p>
+     * Undeliverable errors are those {@code Observer.onError()} invocations that are not allowed happen to
+     * the given consumer type ({@code Observer}, {@code Subscriber}, etc.) due to protocol restrictions
+     * because the consumer has either disposed/cancelled its {@code Disposable}/{@code Subscription} or
+     * has already terminated with an {@code onError()} or {@code onComplete()} signal.
+     * <p>
+     * By default, this global error handler prints the stacktrace via {@link Throwable#printStackTrace()}
+     * and calls {@link java.lang.Thread.UncaughtExceptionHandler#uncaughtException(Thread, Throwable)}
+     * on the current thread.
+     * <p>
+     * Note that on some platforms, the platform runtime terminates the current application with an error if such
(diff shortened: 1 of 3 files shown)


vanniktech (approved):

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -337,7 +337,23 @@ public static Scheduler onComputationScheduler(@NonNull Scheduler defaultSchedul
...
+     * and calls {@link java.lang.Thread.UncaughtExceptionHandler#uncaughtException(Thread, Throwable)}
+     * on the current thread.
+     * <p>
+     * Note that on some platforms, the platform runtime terminates the current application with an error if such
+     * uncaught exceptions happen. In this case, it is recommended the application installs a global error
+     * handler via the {@link #setErrorHandler(Consumer)} plugin method.
+     *
      * @param error the error to report
+     * @see #getErrorHandler()
+     * @see #setErrorHandler(Consumer)
```
tony-root, line 356: @akarnokd 
Do you think it makes sense to reference 
https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling
here?
Is it acceptable to reference Wiki from javadoc?

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -337,7 +337,23 @@ public static Scheduler onComputationScheduler(@NonNull Scheduler defaultSchedul
 
     /**
      * Called when an undeliverable error occurs.
+     * <p>
+     * Undeliverable errors are those {@code Observer.onError()} invocations that are not allowed happen to
```
tony-root: Nit: 
.. that are not allowed **to** happen to ...
?

akarnokd: Updated.

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
akarnokd: Updated.

tony-root (approved):

Issue #5779, opened 2017-12-27
2.x: Inconsistent behavior on throwing UndeliverableException in Observable.merge


tony-root:
Library version: 2.1.7

As stated in [RxJava2 Error Handling](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling), no `Throwable` should be silently swallowed in 2.x. `Throwable`s that cannot be delivered to a disposed downstream get redirected to `RxJavaPlugins`' errorHandler.

However, `Observable.merge` operator does not behave consistently in this case:
- If two different `PublishSubject`s are merged, and emit `onError` one after the other, the second exception **does not get delivered** to `RxJavaPlugins`' errorHandler.
- If one `PublishSubject` get merged with itself (after being filtered for example, which I expect to be a common scenario), and it emit single onError, the exception emitted **gets delivered** to both onError and `RxJavaPlugins`' errorHandler.

The snippet below

```java
public class MergeErrorsExample {
    public static void main(String[] args) {
        RxJavaPlugins.setErrorHandler((error) ->
                printError("RxJavaPlugins.errorHandler", error)
        );

        mergeTwoDifferentObservablesDoesNotThrowUndeliverable();
        mergeSameObservableThrowsUndeliverable();
        mergeSameObservableWithPublishThrowsUndeliverable();
    }

    static void mergeTwoDifferentObservablesDoesNotThrowUndeliverable() {
        System.out.println("Merge two different PublishSubjects DOES NOT throw UndeliverableException");
        PublishSubject<Object> ps1 = PublishSubject.create();
        PublishSubject<Object> ps2 = PublishSubject.create();

        Observable.merge(ps1, ps2).subscribe(
                (next) -> System.out.println("onNext " + next),
                (error) -> printError("onError", error),
                () -> System.out.println("onComplete")
        );

        ps1.onError(new RuntimeException("ps1 exception"));
        ps2.onError(new RuntimeException("ps2 exception"));
        System.out.println();
    }

    static void mergeSameObservableThrowsUndeliverable() {
        System.out.println("Merge same PublishSubject throws UndeliverableException");
        PublishSubject<Boolean> ps1 = PublishSubject.create();

        Observable.merge(
                ps1.filter((condition) -> condition),
                ps1.filter((condition) -> !condition)
        ).subscribe(
                (next) -> System.out.println("onNext " + next),
                (error) -> printError("onError", error),
                () -> System.out.println("onComplete")
        );

[...]

PR #5781, opened 2017-12-27
2.x: Describe merge() error handling.

akarnokd:
This PR adds some clarifications about how `merge` handles (multiple) errors in a new *Error handling* section in its `<dl>` JavaDoc entry.

There exist several dozen variants of `merge` in the various base classes that could also include such JavaDoc addition. Once the wording has been reviewed, the other places will receive a separate PR.

Related: #5779, #5780

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -2965,6 +2965,19 @@ public abstract class Flowable<T> implements Publisher<T> {
      *  backpressure; if violated, the operator <em>may</em> signal {@code MissingBackpressureException}.</dd>
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code merge} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dt><b>Error handling:</b></dt>
+     *  <dd>If any of the source {@code Publisher}s signal a {@code Throwable} via {@code onError}, the resulting
+     *  {@code Flowable} terminates with that {@code Throwable} and all other source {@code Publisher}s are cancelled.
+     *  If more than one {@code Publisher} signals an error, the resulting {@code Flowable} may terminate with the
+     *  first one's error or, depending on the concurrency of the sources, may terminate with a
+     *  {@code CompositeException} containing two or more of the various error signals.
+     *  {@code Throwable}s that didn't make into the composite will be sent (individually) to the global error handler via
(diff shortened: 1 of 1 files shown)


vanniktech (approved):
Does Observable already have this section?

akarnokd:
No. This acts as a testbed for the wording. I hope @artem-zinnatulin can have a look at it too before I copy the text all over the place.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -2965,6 +2965,19 @@ public static int bufferSize() {
...
+     *  If more than one {@code Publisher} signals an error, the resulting {@code Flowable} may terminate with the
+     *  first one's error or, depending on the concurrency of the sources, may terminate with a
+     *  {@code CompositeException} containing two or more of the various error signals.
+     *  {@code Throwable}s that didn't make into the composite will be sent (individually) to the global error handler via
+     *  {@link RxJavaPlugins#onError(Throwable)} method as <em>undeliverable errors</em>. Similarly, {@code Throwable}s
+     *  signaled by source(s) after the returned {@code Flowable} has been cancelled or terminated with a
+     *  (composite) error will be sent to the same global error handler.
+     *  Use {@link #mergeDelayError(Iterable)} to merge sources and terminate only when all source {@code Publisher}s
+     *  have completed or failed with an error.
+     *  </dd>
```
tony-root, line 2980: Does it make sense to explicitly state that split-merge pattern like
```java
PublishSubject<Boolean> ps = PublishSubject.create();

Observable.merge(
        ps.filter((condition) -> condition),
        ps.filter((condition) -> !condition)
).subscribe(
        (next) -> { },
        (error) -> error.printStackTrace()
);

ps.onError(new RuntimeException("Will cause UndeliverableException"));
```
is guaranteed to produce an UndeliverableException in addition to onError?

Maybe in Wiki instead of javadoc?

akarnokd: The main effect is the merge here, not the split. Also I'm not sure how often people find out about this behavior through this split-merge pattern.

tony-root: Indeed merge is the main reason, however every time people use split-merge and **don't find out** about this behavior they introduce a potential hard-to-find bug.

It still seems to me that `merge` should by default behave like `mergeFirstErrorOnly` from #5779 discussion, but I'm definitely not pushing.

PR #5786, opened 2018-01-03
2.x: add error handling section to merge() operator docs

akarnokd:
This PR adds the **Error handling** JavaDoc section to the other `merge` operators throughout the base reactive classes.

Note though that the following `mergeDelayError` methods don't exist yet:

- `Maybe.mergeDelayError(Publisher, int)`
- `Single.mergeDelayError(*)` (`*` all variants of their respective `merge` methods are missing)

These can be added in a separate PR and don't require implementing new operators but to configure existing ones with `delayError == true`.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -451,10 +451,23 @@ public abstract class Completable implements CompletableSource {
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code mergeArray} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dd>If any of the source {@code CompletableSource}s signal a {@code Throwable} via {@code onError}, the resulting
+     *  {@code Completable} terminates with that {@code Throwable} and all other source {@code CompletableSource}s are cancelled.
+     *  If more than one {@code CompletableSource} signals an error, the resulting {@code Completable} may terminate with the
+     *  first one's error or, depending on the concurrency of the sources, may terminate with a
+     *  {@code CompositeException} containing two or more of the various error signals.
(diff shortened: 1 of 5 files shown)


vanniktech (approved):

src/main/java/io/reactivex/Completable.java
```diff
@@ -451,10 +451,23 @@ public static Completable fromRunnable(final Runnable run) {
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code mergeArray} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dd>If any of the source {@code CompletableSource}s signal a {@code Throwable} via {@code onError}, the resulting
+     *  {@code Completable} terminates with that {@code Throwable} and all other source {@code CompletableSource}s are cancelled.
+     *  If more than one {@code CompletableSource} signals an error, the resulting {@code Completable} may terminate with the
+     *  first one's error or, depending on the concurrency of the sources, may terminate with a
+     *  {@code CompositeException} containing two or more of the various error signals.
+     *  {@code Throwable}s that didn't make into the composite will be sent (individually) to the global error handler via
+     *  {@link RxJavaPlugins#onError(Throwable)} method as <em>undeliverable errors</em>. Similarly, {@code Throwable}s
```
artem-zinnatullin: nit: maybe replace `<em>undeliverable` with `<code>UndeliverableException` similarly to CompositeException

artem-zinnatullin (approved):

src/main/java/io/reactivex/Completable.java
akarnokd: Updated.

PR #5790, opened 2018-01-05
2.x: improve request accounting overhead in retry/repeat

akarnokd:
This PR improves the request accounting overhead in the `retry` and `repeat` operators.

Previously, every individual `onNext` invocation signaled a an item production, which had a lot of overhead since the `SubscriptionArbiter` has to serialize invocations of `setSubscription`, `request` and `produced`. The improved algorithm counts the `onNext` calls in a field and calls `produced` once with the total count before subscribing to the upstream again.

Other small changes:
- rename inner class to `RetrySubscriber` (sloppy copy-paste),
- add `isCancelled()` check to `repeatUntil` subscription loop, which was somehow missing.


vanniktech (approved):

artem-zinnatullin (approved):
Nice!

PR #5792, opened 2018-01-05
2.x: Fix flatMap inner fused poll crash not cancelling the upstream

### LINKED ISSUE #5791: 2.x: part of the stream continues after error in Flowable but not Observable.
Given an `Iterable<T>` that always fails on `next()`:

```
public Iterable<String> iterableThatThrowsException()
    {
        return () -> new Iterator<String>()
        {
            @Override
            public boolean hasNext() { true; }

            @Override
            public String next()
            {
                throw new RuntimeException("kaboom, kaboom");
            }
        };
    }
```

and this stream:

```
Flowable.fromIterable(Arrays.asList(1, 2))
                  .doOnNext(i -> out.println("fromIterable onNext() -> " + i))
                  .doOnComplete(() -> out.println("fromIterable onComplete()"))
                  .map(i -> iterableThatThrowsException())
                  .flatMap(l -> Flowable.fromIterable(l))
                  .doOnError(t -> out.println("flatMap onError()"))
                  .subscribe(
                    i -> out.println("subscriber onNext() -> " + i),
                    t -> out.println(
                        "subscriber onError() -> " + t.getClass().getName() +
                        " " + t.getMessage()),
                    () -> out.println("subscriber onComplete()")
                );
```
the `Flowable.fromIterable(Arrays.asList(1, 2))` continues to emit items after `flatMap(l -> Flowable.fromIterable(l))` signals error.

This is the output:
```
fromIterable onNext() -> 1
flatMap onError()
subscriber onError() -> java.lang.RuntimeException kaboom, kaboom
fromIterable onNext() -> 2
fromIterable onComplete()
```

Please note that:

1. When I substitute Flowable with Observable, then it outputs what **I** think is the correct behaviour
```
fromIterable onNext() -> 1
flatMap onError()
subscriber onError() -> java.lang.RuntimeException kaboom, kaboom
```
2. When I add a "dummy"`doOnError()` to the inner `fromIterable` it also produces "correct" output like in 1
```
.flatMap(l -> Flowable.fromIterable(l).doOnError(t -> {}))
```
3. The "incorrect" behaviour is consistent all the way down to version 2.0.0

akarnokd:
This PR fixes the lack of upstream `cancel()` call when an inner, fused source's `queue.poll()` crashes in a non-delayed error mode.

Unit tests were added to verify `Observable.flatMap`, `Flowable.flatMapIterable` and `Observable.flatMapIterable` as well.
  
Fixes #5791

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMap.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMap.java
@@ -85,7 +85,7 @@ public final class FlowableFlatMap<T, U> extends AbstractFlowableWithUpstream<T,
 
         final AtomicLong requested = new AtomicLong();
 
-        Subscription s;
+        Subscription upstream;
 
         long uniqueId;
         long lastId;
@@ -107,8 +107,8 @@ public final class FlowableFlatMap<T, U> extends AbstractFlowableWithUpstream<T,
 
         @Override
         public void onSubscribe(Subscription s) {
-            if (SubscriptionHelper.validate(this.s, s)) {
-                this.s = s;
+            if (SubscriptionHelper.validate(this.upstream, s)) {
+                this.upstream = s;
                 actual.onSubscribe(this);
                 if (!cancelled) {
                     if (maxConcurrency == Integer.MAX_VALUE) {
@@ -132,7 +132,7 @@ public final class FlowableFlatMap<T, U> extends AbstractFlowableWithUpstream<T,
                 p = ObjectHelper.requireNonNull(mapper.apply(t), "The mapper returned a null Publisher");
             } catch (Throwable e) {
                 Exceptions.throwIfFatal(e);
-                s.cancel();
+                upstream.cancel();
                 onError(e);
                 return;
             }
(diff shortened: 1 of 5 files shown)


vanniktech (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMap.java
```diff
@@ -482,6 +482,9 @@ void drainLoop() {
                                     Exceptions.throwIfFatal(ex);
                                     is.dispose();
                                     errs.addThrowable(ex);
+                                    if (!delayErrors) {
+                                        upstream.cancel();
```
artem-zinnatullin, line 486: So this is bug fix, but what's the reason for renaming `s` to `upstream`?

It looks not important to revert but I guess minimizing fix PRs saves time for those who review them while going through changelog :)

artem-zinnatullin (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableFlatMap.java
akarnokd: It reads nicer now.

PR #5799, opened 2018-01-09
2.x: Add missing {Maybe|Single}.mergeDelayError variants

akarnokd:
This PR adds `mergeDelayError` overloads to `Maybe` and `Single`; the infrastructure was there from the beginning but the `delayErrors == true` settings were not exposed publicly.

- `Maybe.mergeDelayError(Publisher, int)`
- `Single.mergeDelayError(Iterable)`
- `Single.mergeDelayError(Publisher)`
- `Single.mergeDelayError(SingleSource, SingleSource)`
- `Single.mergeDelayError(SingleSource, SingleSource, SingleSource)`
- `Single.mergeDelayError(SingleSource, SingleSource, SingleSource, SingleSource)`

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -787,9 +787,9 @@ public abstract class Single<T> implements SingleSource<T> {
      *
      * @param <T> the common value type
      * @param source1
-     *            a Single to be merged
+     *            a SingleSource to be merged
      * @param source2
-     *            a Single to be merged
+     *            a SingleSource to be merged
      * @return a Flowable that emits all of the items emitted by the source Singles
      * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
      * @see #mergeDelayError(SingleSource, SingleSource)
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -913,6 +913,181 @@
...
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The returned {@code Flowable} honors the backpressure of the downstream consumer.</dd>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code mergeDelayError} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <T> the common value type
+     * @param source1
+     *            a Single to be merged
```
vanniktech: technically it's a `SingleSource`. dunno whether you want to distinguish between them

vanniktech (approved):

src/main/java/io/reactivex/Single.java
akarnokd: I'll update it as `Maybe.merge()` say "a MaybeSource to be merged".

vanniktech (approved):

PR #5820, opened 2018-01-25
1.x: Plugin lookup workaround for System.properties access restrictions

### LINKED ISSUE #5819: RxJavaPlugins.java should not be reading all system properties
Hi,

I am working on 1.3.4 version of Rxjava jar and there are certain methods where RxJavaPlugins.java is trying to read all the system properties while getting hooks. Not all companies give access to their system properties on server. For example, we use tomcat and have catalina policy which restricts to properties mentioned there. Due to this, i am getting access denied exceptions.

I propose 2 solutions:
1. Remove system.getproperties() and it should be able to read only properties which are required to get the hooks
2. Provide an implementation via application property which doesn't look for implementation at all.

I am okay with either solution. Please help

akarnokd:
The PR adds a `try-catch` around the System property lookup inside the `RxJavaPlugins` in case a security manager prevents reading arbitrary property entries.

This mainly affects the `rxjava.plugin.[index].class` lookup which were introduced due to the 31 character key limit on Android.

However, when running in a container such as Tomcat, a security manager may prevent reading these type of prefixed entries (where `[index]` can't be known upfront), crashing the initialization.

**Update**:

The `System.getProperties()` can also fail, therefore, retrieving the properties has been factored out into a separate method that returns an empty properties.

Fixes #5819.

### DIFF
--- a/src/main/java/rx/plugins/RxJavaPlugins.java
+++ b/src/main/java/rx/plugins/RxJavaPlugins.java
@@ -105,7 +105,7 @@ public class RxJavaPlugins {
     public RxJavaErrorHandler getErrorHandler() {
         if (errorHandler.get() == null) {
             // check for an implementation from System.getProperty first
-            Object impl = getPluginImplementationViaProperty(RxJavaErrorHandler.class, System.getProperties());
+            Object impl = getPluginImplementationViaProperty(RxJavaErrorHandler.class, getSystemPropertiesSafe());
             if (impl == null) {
                 // nothing set via properties so initialize with default
                 errorHandler.compareAndSet(null, DEFAULT_ERROR_HANDLER);
@@ -147,7 +147,7 @@ public class RxJavaPlugins {
     public RxJavaObservableExecutionHook getObservableExecutionHook() {
         if (observableExecutionHook.get() == null) {
             // check for an implementation from System.getProperty first
-            Object impl = getPluginImplementationViaProperty(RxJavaObservableExecutionHook.class, System.getProperties());
+            Object impl = getPluginImplementationViaProperty(RxJavaObservableExecutionHook.class, getSystemPropertiesSafe());
             if (impl == null) {
                 // nothing set via properties so initialize with default
                 observableExecutionHook.compareAndSet(null, RxJavaObservableExecutionHookDefault.getInstance());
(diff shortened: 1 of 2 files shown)


ravikancherla (approved):

vanniktech (approved):

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -262,6 +262,19 @@ public void registerCompletableExecutionHook(RxJavaCompletableExecutionHook impl
         }
     }
 
+    /**
+     * A security manager may prevent accessing the System properties entirely,
+     * therefore, the SecurityException is turned into an empty properties.
+     * @return the Properties to use for looking up settings
+     */
+    /* test */ static Properties getSystemPropertiesSafe() {
```
artem-zinnatullin, line 270: Remove "test"?

src/main/java/rx/plugins/RxJavaPlugins.java
```diff
@@ -284,25 +297,32 @@ public void registerCompletableExecutionHook(RxJavaCompletableExecutionHook impl
...
                         }
-
-                        break;
                     }
                 }
+            } catch (SecurityException ex) {
+                // https://github.com/ReactiveX/RxJava/issues/5819
+                // We don't seem to have access to all properties.
+                // At least print the exception to the console.
+                ex.printStackTrace();
```
artem-zinnatullin, line 325: Maybe call RxJavaPlugins.onError() or however that api is called in 1.x?

I believe people won't like stacktrace popping up in their logs without ability to swallow it

artem-zinnatullin (approved):
LGTM but few comments

src/main/java/rx/plugins/RxJavaPlugins.java
akarnokd: It indicates the method is `package private` for testing purposes instead of `private`.

vanniktech: It's consistent with the other methods. ¯\_(ツ)_/¯

src/main/java/rx/plugins/RxJavaPlugins.java
akarnokd: Can't call that because this is part of the initialization process of `RxJavaPlugins`, the error handler would come from the same system properties and the default does nothing.

src/main/java/rx/plugins/RxJavaPlugins.java
artem-zinnatullin: Ah, haven't seen these in a while, my bad

src/main/java/rx/plugins/RxJavaPlugins.java
artem-zinnatullin: mhm, I see, ok

PR #5833, opened 2018-02-01
2.x: Fix Observable.switchMap main onError not disposing the current inner source

### LINKED ISSUE #5832: 2.x Observable onError does not dispose switchMap's inner ObservableSource
Hi, I'm facing an issue with Observable's switchMap/switchMapSingle where the inner source does not get disposed if the upstream emits an error. Happens on both version 2.1.6 and 2.1.9. Is this the expected behavior?

Observable version does not dispose the inner source
```        
        Observable
                .create(e -> {
                    e.onNext(true);
                    e.onError(new Throwable());
                })
                .switchMap(__ -> Observable
                        .timer(500L, TimeUnit.MILLISECONDS)
                        .doOnNext(System.out::println))
                .subscribe(__->{}, __->{});
        Thread.sleep(1000L);
```

While the Flowable version does dispose the inner source.

```        
        Flowable
                .create(e -> {
                    e.onNext(true);
                    e.onError(new Throwable());
                }, BackpressureStrategy.BUFFER)
                .switchMap(__ -> Flowable
                        .timer(500L, TimeUnit.MILLISECONDS)
                        .doOnNext(System.out::println))
                .subscribe(__->{}, __->{});
        Thread.sleep(1000L);
```

akarnokd:
The `Observable.switchMap` had bad logic in its main `onError` handler which didn't dispose the current inner `Observable`.

The `Flowable` version didn't have this logic error. Both variants received an unit test to verify the correct behavior.

Fixes #5832.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java
@@ -131,15 +131,15 @@ public final class ObservableSwitchMap<T, R> extends AbstractObservableWithUpstr
 
         @Override
         public void onError(Throwable t) {
-            if (done || !errors.addThrowable(t)) {
+            if (!done && errors.addThrowable(t)) {
                 if (!delayErrors) {
                     disposeInner();
                 }
+                done = true;
+                drain();
+            } else {
                 RxJavaPlugins.onError(t);
-                return;
             }
-            done = true;
-            drain();
         }
 
         @Override
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java
@@ -1153,4 +1153,25 @@ public class FlowableSwitchTest {
         .test()
         .assertFailure(TestException.class);
     }
+
+    @Test
+    public void innerCancelledOnMainError() {
+        final PublishProcessor<Integer> main = PublishProcessor.create();
+        final PublishProcessor<Integer> inner = PublishProcessor.create();
+
+        TestSubscriber<Integer> to = main.switchMap(Functions.justFunction(inner))
+        .test();
+
+        assertTrue(main.hasSubscribers());
+
+        main.onNext(1);
+
+        assertTrue(inner.hasSubscribers());
+
+        main.onError(new TestException());
+
+        assertFalse(inner.hasSubscribers());
+
+        to.assertFailure(TestException.class);
+    }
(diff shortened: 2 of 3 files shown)


vanniktech (approved):

src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java
```diff
@@ -131,15 +131,15 @@ public void onNext(T t) {
 
         @Override
         public void onError(Throwable t) {
-            if (done || !errors.addThrowable(t)) {
+            if (!done && errors.addThrowable(t)) {
```
artem-zinnatullin, line 134: Codecov indicates that some of possible variants are not tested, do you want to cover them in this PR?

screen shot 2018-02-01 at 4 48 30 pm

artem-zinnatullin: It's not that I think code in the patch is broken, but that if we had test that was trying to go into `onError` when it's already `done == true`, we would have a hint that original code had some problems

artem-zinnatullin (approved):

src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java
akarnokd: That line is tested but parallelism on the CI is flaky and may not produce the necessary interleaving there.

The upstream may onError after a crash in the mapper thus we have to route that into the global error handler.

artem-zinnatullin: Can't we test it just by breaking RS protocol with  `main.onComplete(); main.onError()`?

I might miss something, reviewing on mobile

akarnokd: [This](https://github.com/akarnokd/RxJava/blob/8ec030fc052e20b62da80bc6edf6d084c03094d8/src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java#L870) and [this](https://github.com/akarnokd/RxJava/blob/8ec030fc052e20b62da80bc6edf6d084c03094d8/src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java#L1002) tests should cover that line. Apparently, the test did not involve ps2 so the inner source was not involved. Updated the unit tests accordingly.

vanniktech (approved):

artem-zinnatullin (approved):

PR #5837, opened 2018-02-02
2.x: Detail distinct() & distinctUntilChanged() in JavaDoc

akarnokd:
The PR clarifies the comparison and retention behavior of the `distinct` and `distinctUntilChanged` operators (in both `Flowable` and `Observable`).

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -7674,10 +7674,10 @@ public abstract class Flowable<T> implements Publisher<T> {
      * <p>
      * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/distinct.png" alt="">
      * <p>
-     * It is recommended the elements' class {@code T} in the flow overrides the default {@code Object.equals()} to provide
+     * It is recommended the elements' class {@code T} in the flow overrides the default {@code Object.equals()} and {@link Object#hashCode()} to provide
      * meaningful comparison between items as the default Java implementation only considers reference equivalence.
      * <p>
-     * By default, {@code distinct()} uses an internal, per Subscriber {@link java.util.HashSet} to remember
+     * By default, {@code distinct()} uses an internal {@link java.util.HashSet} per Subscriber to remember
      * previously seen items and uses {@link java.util.Set#add(Object)} returning {@code false} as the
      * indicator for duplicates.
      * <p>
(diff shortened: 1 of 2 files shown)


vanniktech (approved):

src/main/java/io/reactivex/Flowable.java
```diff
@@ -7669,9 +7669,24 @@ public final void blockingSubscribe(Subscriber<? super T> subscriber) {
...
-     * Returns a Flowable that emits all items emitted by the source Publisher that are distinct.
+     * Returns a Flowable that emits all items emitted by the source Publisher that are distinct
+     * based on {@link Object#equals(Object)} comparison.
      * <p>
      * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/distinct.png" alt="">
+     * <p>
+     * It is recommended the elements' class {@code T} in the flow overrides the default {@code Object.equals()} to provide
+     * meaningful comparison between items as the default Java implementation only considers reference equivalence.
+     * <p>
+     * By default, {@code distinct()} uses an internal, per Subscriber {@link java.util.HashSet} to remember
```
artem-zinnatullin: supernit: could be phrased like that `By default, {@code distinct()} uses an internal {@link java.util.HashSet} per Subscriber to remember`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -7669,9 +7669,24 @@ public final void blockingSubscribe(Subscriber<? super T> subscriber) {
     }
 
     /**
-     * Returns a Flowable that emits all items emitted by the source Publisher that are distinct.
+     * Returns a Flowable that emits all items emitted by the source Publisher that are distinct
+     * based on {@link Object#equals(Object)} comparison.
      * <p>
      * <img width="640" height="310" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/distinct.png" alt="">
+     * <p>
+     * It is recommended the elements' class {@code T} in the flow overrides the default {@code Object.equals()} to provide
```
artem-zinnatullin: Since we use `HashSet` internally, we should probably recommend properly overriding `hasCode()` as well

artem-zinnatullin: Same below

src/main/java/io/reactivex/Flowable.java
artem-zinnatullin: Same below

artem-zinnatullin (approved):

vanniktech (approved):

PR #5838, opened 2018-02-03
Added nullability annotation for RxJavaPlugins.onCompletableAssembly

afeozzz:
Just annotation that didn't exist

### DIFF
--- a/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/plugins/RxJavaPlugins.java
@@ -86,6 +86,7 @@ public final class RxJavaPlugins {
     @Nullable
     static volatile Function<? super Single, ? extends Single> onSingleAssembly;
 
+    @Nullable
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/plugins/RxJavaPlugins.java
```diff
@@ -86,6 +86,7 @@
     @Nullable
     static volatile Function<? super Single, ? extends Single> onSingleAssembly;
 
+    @Nullable
```
vanniktech, line 89: Do we also need the `@SuppressWarnings("rawtypes")` here?

vanniktech (approved):

src/main/java/io/reactivex/plugins/RxJavaPlugins.java
akarnokd: Completable does not have a type argument.

PR #5841, opened 2018-02-05
2.x: Improve JavaDoc of XObserver types.

akarnokd:
This PR improves the JavaDoc of the 4 RxJava main consumer types:

- `Observer`
- `SingleObserver`
- `MaybeObserver`
- `CompletableObserver`

### DIFF
--- a/src/main/java/io/reactivex/CompletableObserver.java
+++ b/src/main/java/io/reactivex/CompletableObserver.java
@@ -21,7 +21,7 @@ import io.reactivex.disposables.Disposable;
  * <p>
  * When a {@code CompletableObserver} is subscribed to a {@link CompletableSource} through the {@link CompletableSource#subscribe(CompletableObserver)} method,
  * the {@code CompletableSource} calls {@link #onSubscribe(Disposable)}  with a {@link Disposable} that allows
- * cancelling the sequence at any time. A well-behaved
+ * disposing the sequence at any time. A well-behaved
  * {@code CompletableSource} will call a {@code CompletableObserver}'s {@link #onError(Throwable)}
  * or {@link #onComplete()} method exactly once as they are considered mutually exclusive <strong>terminal signals</strong>.
  * <p>
@@ -35,7 +35,7 @@ import io.reactivex.disposables.Disposable;
  * happens, it is the duty of the {@code CompletableObserver} implementation to be ready to receive multiple calls to
  * its methods and ensure proper concurrent behavior of its business logic.
  * <p>
- * Calling {@link #onSubscribe(Disposable)} or{@link #onError(Throwable)} with a
+ * Calling {@link #onSubscribe(Disposable)} or {@link #onError(Throwable)} with a
  * {@code null} argument is forbidden.
  * <p>
  * The implementations of the {@code onXXX} methods should avoid throwing runtime exceptions other than the following cases:
--- a/src/main/java/io/reactivex/CompletableObserver.java
+++ b/src/main/java/io/reactivex/CompletableObserver.java
@@ -17,7 +17,36 @@ import io.reactivex.annotations.NonNull;
 import io.reactivex.disposables.Disposable;
 
 /**
- * Represents the subscription API callbacks when subscribing to a Completable instance.
+ * Provides a mechanism for receiving push-based notification of a valueless completion or an error.
+ * <p>
+ * When a {@code CompletableObserver} is subscribed to a {@link CompletableSource} through the {@link CompletableSource#subscribe(CompletableObserver)} method,
+ * the {@code CompletableSource} calls {@link #onSubscribe(Disposable)}  with a {@link Disposable} that allows
+ * cancelling the sequence at any time. A well-behaved
+ * {@code CompletableSource} will call a {@code CompletableObserver}'s {@link #onError(Throwable)}
+ * or {@link #onComplete()} method exactly once as they are considered mutually exclusive <strong>terminal signals</strong>.
+ * <p>
+ * Calling the {@code CompletableObserver}'s method must happen in a serialized fashion, that is, they must not
+ * be invoked concurrently by multiple threads in an overlapping fashion and the invocation pattern must
+ * adhere to the following protocol:
+ * <p>
+ * <pre><code>    onSubscribe (onError | onComplete)?</code></pre>
+ * <p>
+ * Subscribing a {@code CompletableObserver} to multiple {@code CompletableSource}s is not recommended. If such reuse
+ * happens, it is the duty of the {@code CompletableObserver} implementation to be ready to receive multiple calls to
+ * its methods and ensure proper concurrent behavior of its business logic.
+ * <p>
+ * Calling {@link #onSubscribe(Disposable)} or{@link #onError(Throwable)} with a
+ * {@code null} argument is forbidden.
+ * <p>
+ * The implementations of the {@code onXXX} methods should avoid throwing runtime exceptions other than the following cases:
+ * <ul>
+ * <li>If the argument is {@code null}, the methods can throw a {@code NullPointerException}.
+ * Note though that RxJava prevents {@code null}s to enter into the flow and thus there is generally no
+ * need to check for nulls in flows assembled from standard sources and intermediate operators.
+ * </li>
+ * <li>If there is a fatal error (such as {@code VirtualMachineError}).</li>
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/Observer.java
```diff
@@ -19,14 +19,35 @@
...
+ * <pre><code>    onSubscribe onNext* (onError | onComplete)?</code></pre>
+ * <p>
+ * Subscribing an {@code Observer} to multiple {@code ObservableSource}s is not recommended. If such reuse
+ * happens, it is the duty of the {@code Observer} implementation to be ready to receive multiple calls to
+ * its methods and ensure proper concurrent behavior of its business logic.
+ * <p>
+ * Calling {@link #onSubscribe(Disposable)}, {@link #onNext(Object)} or {@link #onError(Throwable)} with a
+ * {@code null} argument is forbidden.
+ * <p>
+ * The implementations of the {@code onXXX} methods should avoid throwing runtime exceptions other than the following cases:
```
artem-zinnatullin: Another valid case is fatal business logic violation. Otherwise people might think that swallowing errors is preferred because doc said so

artem-zinnatullin (commented):
Great rewrite! But comment about errors from onX methods

src/main/java/io/reactivex/Observer.java
akarnokd: What do you mean by "fatal business logic violation"?

src/main/java/io/reactivex/CompletableObserver.java
```diff
@@ -17,7 +17,36 @@
 import io.reactivex.disposables.Disposable;
 
 /**
- * Represents the subscription API callbacks when subscribing to a Completable instance.
+ * Provides a mechanism for receiving push-based notification of a valueless completion or an error.
+ * <p>
+ * When a {@code CompletableObserver} is subscribed to a {@link CompletableSource} through the {@link CompletableSource#subscribe(CompletableObserver)} method,
+ * the {@code CompletableSource} calls {@link #onSubscribe(Disposable)}  with a {@link Disposable} that allows
+ * cancelling the sequence at any time. A well-behaved
```
vanniktech: should we use `cancelling` here or `disposing` since the method is called dispose?

src/main/java/io/reactivex/CompletableObserver.java
```diff
@@ -17,7 +17,36 @@
...
+ * {@code CompletableSource} will call a {@code CompletableObserver}'s {@link #onError(Throwable)}
```
vanniktech: I'd use `may` call a since you can also just call onSubscribe and then do nothing.

src/main/java/io/reactivex/CompletableObserver.java
```diff
@@ -17,7 +17,36 @@
...
+ * be invoked concurrently by multiple threads in an overlapping fashion and the invocation pattern must
+ * adhere to the following protocol:
+ * <p>
+ * <pre><code>    onSubscribe (onError | onComplete)?</code></pre>
+ * <p>
+ * Subscribing a {@code CompletableObserver} to multiple {@code CompletableSource}s is not recommended. If such reuse
+ * happens, it is the duty of the {@code CompletableObserver} implementation to be ready to receive multiple calls to
+ * its methods and ensure proper concurrent behavior of its business logic.
+ * <p>
+ * Calling {@link #onSubscribe(Disposable)} or{@link #onError(Throwable)} with a
```
vanniktech: I think there's a space missing between `or` and `{`

vanniktech (approved):
The documentation is getting slicker and slicker.

src/main/java/io/reactivex/CompletableObserver.java
akarnokd: Technically yes, but we don't want to encourage that for end users because they'll just hang.

vanniktech: Alright works for me.

src/main/java/io/reactivex/Observer.java
artem-zinnatullin: When for some reason `Observer` finds itself in inconsistent state during `onXXX()` execution.

In this case exception thrown from `onXXX()` will reach `RxJavaPlugins` as `OnErrorNotImplementedException` and in case of Android will crash the application, on JVM user can hook into that and specify required behavior.

This is fail fast design principle, when instead of swallowing error states and writing defensive code you fail as soon as possible preventing inconsistent states from going deeper into your system.

I'm just afraid that this doc might discourage people from that design principle

akarnokd: > In this case exception thrown from onXXX() will reach RxJavaPlugins as OnErrorNotImplementedException and in case of Android will crash the application, on JVM user can hook into that and specify required behavior.

That is the property of the lambda-subscribe methods and not when using a plain `Observer`.

The sentence is the derivative of the Reactive Streams rule [2.13](https://github.com/reactive-streams/reactive-streams-jvm#2.13) and is intended to discourage throwing from these methods as they may not end up where users expect it. Remember those issues where the sync code propagated the exception on the main thread but as soon as asynchrony was involved, no errors were suddenly thrown by the unit test. So don't throw from `Observer` methods and you'll be fine.

artem-zinnatullin: >That is the property of the lambda-subscribe methods and not when using a plain `Observer`.

Ah right, but exception is still delivered to `RxJavaPlugins`' error handler.

>The sentence is the derivative of the Reactive Streams rule 2.13 and is intended to discourage throwing from these methods as they may not end up where users expect it. Remember those issues where the sync code propagated the exception on the main thread but as soon as asynchrony was involved, no errors were suddenly thrown by the unit test.

Yep, RS 2.13, yep RxJava 2.x broke unit testing. However I still think throwing exception from `onXXX` is okay **if** system finds itself in inconsistent state.

[RS 2.13](https://github.com/reactive-streams/reactive-streams-jvm#2.13) says:

>Calling `onSubscribe`, `onNext`, `onError` or `onComplete` MUST [return normally](#term_return_normally) except when any provided parameter is `null` in which case it MUST throw a `java.lang.NullPointerException` to the caller, for all other situations the only legal way for a `Subscriber` to signal failure is by cancelling its `Subscription`. In the case that this rule is violated, any associated `Subscription` to the `Subscriber` MUST be considered as cancelled, and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.

Specifically:

>In the case that this rule is violated, any associated Subscription to the Subscriber MUST be considered as cancelled, and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.

Which actually sort of allows throwing exceptions from `onXXX()` since RxJava indeed provides a way to `raise this error condition in a fashion that is adequate for the runtime environment`.

---

What I want to say is that there are thousands of projects where exceptions happen in `onXXX()` methods, crash apps and then appear as fatal errors in bug tracking systems. 

Correct solution will be to fix such errors by fixing the code that produces them in the first place  rather than somehow not throwing them.

It's also unclear how one should avoid throwing exceptions from `onXXX()` if Java has unchecked exceptions and some JVM languages like Kotlin treat every exception as unchecked. I hope you don't expect people to wrap code in every `onXXX()` callback into a `try/catch` :)

akarnokd: There is no guarantee where such exceptions will end up:

- if the flow is synchronous, the `subscribe` will end up throwing
- if there is asynchrony, depending on the scheduler, the error may end up in
  - `RxJavaPlugins.onError`,
  - The current thread's uncaught exception handler directly, or
  - get swallowed in a custom scheduler implementation

From the Observable's perspective, an Observer is the end consumer thus it is the `Observer`'s responsibility to "crash further down". Yes, this means unreliable code should be wrapped into `try-catch`, specifically in `onError` or `onComplete`. If there is a crash in `onNext`, as the rule says, dispose the `Disposable` and signal the error by some other means, for example, by calling `this.onError()`.

vanniktech (approved):

src/main/java/io/reactivex/Observer.java
akarnokd: Added this type of clarification to the JavaDoc.

artem-zinnatullin (approved):
I don't really agree with wording about try/catch, but also don't want to block this PR because it's 99% awesome

PR #5842, opened 2018-02-05
prevent encapsulation issues

igor-suhorukov:

---

Member values are subject to change from anywhere in the code and may not meet the programmer's assumptions.


JakeWharton:
These don't appear to be actual issues of encapsulation.

akarnokd:
Please leave those field accesses as they are. There is no reason to introduce those method calls on internal components.

PR #5850, opened 2018-02-14
Fix a race condition that may make OperatorMaterialize emit too many terminal notifications

pkolaczk:
We're using RxJava 1.3.5 and one of our tests occasionally fails (very hard to reproduce). The test asserts that `anObservable.toBlocking.toList` (RxScala) expression throws a particular exception, because we expect the observable to be in error state by emitting a single `onError`. However, in rare cases, we observe that there is no exception thrown, and an empty list is returned instead. We verified that the problem is not the observable we use. The observable properly fires the `doOnError` handler prior to  the test assertion failure. Therefore we suspected something wrong happening in the `toBlocking.toList` conversion. I debugged that in fact this conversion uses `OperatorMaterialize` as a part of `BlockingOperatorToIterator`. 

I took a look at the `OperatorMaterialize` and found an obvious bug. The `busy` variable is never set to true anywhere, despite the comments showing the author of the code intended it to be true when the `drain` loop was running.

The PR fixes this problem by setting `busy` to true at the entry of the `drain` method. 

However I still have some doubts about the original code: 
- Why does `drain` have to be called from `requestMore`? Why isn't it enough to call it from `onCompleted` and `onError`? It looks like the `drain` logic does anything only if the `terminalNotification` is set, and it can be set only by `onCompleted` and `onError`.
- What is this `drain` logic supposed to do? Looks rather complex and the intent of it is not obvious. If we don't call it from `requestMore` then maybe we don't need the `drain` logic at all? Why not emit the terminal notifications directly from `onError` and `onCompleted` ?

Thanks,
Piotr Kołaczkowski

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 jdk:
 - oraclejdk8
-sudo: false
+sudo: required
 # as per http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 
 git:
--- a/src/main/java/rx/internal/operators/OperatorMaterialize.java
+++ b/src/main/java/rx/internal/operators/OperatorMaterialize.java
@@ -134,6 +134,7 @@ public final class OperatorMaterialize<T> implements Operator<Notification<T>, T
                     missed = true;
                     return;
                 }
+                busy = true;
             }
             // drain loop
             final AtomicLong localRequested = this.requested;
--- a/src/test/java/rx/internal/operators/OperatorMaterializeTest.java
+++ b/src/test/java/rx/internal/operators/OperatorMaterializeTest.java
@@ -29,9 +29,12 @@ import org.junit.Test;
 import rx.Notification;
 import rx.Observable;
 import rx.Subscriber;
+import rx.TestUtil;
+import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.observers.TestSubscriber;
 import rx.schedulers.Schedulers;
+import rx.subjects.PublishSubject;
 
 public class OperatorMaterializeTest {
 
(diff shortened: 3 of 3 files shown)


akarnokd:
> Why does drain have to be called from requestMore?
> What is this drain logic supposed to do? 

Let's assume the source generates N item, completes and the consumer requests N notifications. Since the source completion is another Notification, it can't be emitted unless the consumer request yet another notification. Since this can happen after the original `onCompleted`, the request itself has to trigger the emission of that completion notification. Because `onXXX` methods can race with `request` calls and trigger emission, there has to be a serialization of signals towards the consumer. The busy/missed drain loop ensures this. 

[Further reading.](http://akarnokd.blogspot.hu/2015/05/operator-concurrency-primitives.html)

akarnokd (commented):
It would be great if you'd adapted [this v2](https://github.com/ReactiveX/RxJava/blob/2.x/src/test/java/io/reactivex/internal/subscribers/SinglePostCompleteSubscriberTest.java#L25) test to verify this change. The `TestUtil.race()` is there for v1.

pkolaczk:
Thanks! I get it. You're right, somehow I missed the fact that `onCompleted` may be called earlier than the request (still thinking in terms of cold observables we have in our system, that never emit anything until requested). 

BTW: This is not the real cause of the bug I'm hunting for, because I've just found another occurrence that was not invoking iterators / lists at all. So I dug more into it, and I found another race in OperatorMerge. I submitted a PR.

davidmoten (approved):
Great find @pkolaczk, thanks!

akarnokd (approved):

pkolaczk:
Unit test confirms that without the patch `OperatorMaterialize` can emit two onComplete notifications.

pkolaczk:
May I rebase and fix the commit message of the fix? It says "miss the item" while in fact it may emit too many items.

akarnokd:
If you really want to.

pkolaczk:
Misleading commit messages are bad.

akarnokd:
This PR keeps failing due to overcrowded Travis CI. In v2, we solved this with `sudo: required` in [.travis.yml](https://github.com/ReactiveX/RxJava/blob/2.x/.travis.yml#L14). Could you please update the PR to make this happen on 1.x? https://github.com/ReactiveX/RxJava/blob/1.x/.travis.yml#L4

vanniktech (approved):

PR #5851, opened 2018-02-14
Fix a race condition in OperatorMerge.InnerSubscriber#onError

pkolaczk:
Inner subscriber must queue the error first before setting done, so that after `emitLoop` removes the subscriber,`emitLoop` is guaranteed to notice the error. Otherwise it would be possible that inner subscribers count was 0, and at the same time the error queue was empty.

We observed that sometimes a result of flattening two observables doesn't properly mark the resulting observable as failed, despite the inner observable properly reporting onError. Instead, the error condition was ignored and the result observable was completing normally.

### DIFF
--- a/src/main/java/rx/internal/operators/OperatorMerge.java
+++ b/src/main/java/rx/internal/operators/OperatorMerge.java
@@ -847,8 +847,11 @@ public final class OperatorMerge<T> implements Operator<T, Observable<? extends 
         }
         @Override
         public void onError(Throwable e) {
-            done = true;
+            // Need to queue the error first before setting done, so that after emitLoop() removes the subscriber,
+            // it is guaranteed to notice the error. Otherwise it would be possible that inner subscribers count was 0,
+            // and at the same time the error queue was empty.
             parent.getOrCreateErrorQueue().offer(e);
+            done = true;
             parent.emit();
         }
         @Override
--- a/src/test/java/rx/internal/operators/OperatorMergeTest.java
+++ b/src/test/java/rx/internal/operators/OperatorMergeTest.java
@@ -1205,6 +1205,34 @@ public class OperatorMergeTest {
         assertTrue(latch.await(10, TimeUnit.SECONDS));
     }
 
+    @Test
+    public void testConcurrentMergeInnerError() {
+        for (int i = 0; i < 1000; i++) {
+            final TestSubscriber<Integer> ts = TestSubscriber.create();
+            final PublishSubject<Integer> ps1 = PublishSubject.create();
+            final PublishSubject<Integer> ps2 = PublishSubject.create();
+            final Exception error = new NullPointerException();
+            Action0 action1 = new Action0() {
+                @Override
+                public void call() {
+                    ps1.onNext(1);
+                    ps1.onCompleted();
+                }
+            };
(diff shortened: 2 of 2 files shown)


src/main/java/rx/internal/operators/OperatorMerge.java
```diff
@@ -847,8 +847,11 @@ public void onNext(T t) {
         }
         @Override
         public void onError(Throwable e) {
-            done = true;
+            // Need to queue the error first before setting done, so that after emitLoop() removes the subscriber,
+            // it is guaranteed to notice the error. Otherwise it would be possible that inner subscribers count was 0,
+            // and at the same time the error queue was empty.
             parent.getOrCreateErrorQueue().offer(e);
+            done = true;
```
akarnokd, line 854: Please add an unit test that checks this race condition.

pkolaczk: How do I do this? This looks like a very rare / hard to trigger one. A thread would have to stop for a moment after setting done and before queuing the error at the time when emitLoop was running on another thread.

akarnokd: That's what `TestUtil.race` is for. Have one subject `onCompleted()` and another `onError()`, `mergeDelayError` them. Loop this at least 1000 times, check a `TestSubscriber` for error termination. Remember to create the subjects and subscriber inside the loop due to the terminal event nature.

akarnokd (approved):

pkolaczk:
Wow! I wrote a test and I'm surprised it reproduced the issue so easily. When I reverted the patch, the test failed with:
```
java.lang.AssertionError: No errors (1 completion)

	at rx.observers.TestSubscriber.assertionError(TestSubscriber.java:667)
	at rx.observers.TestSubscriber.assertError(TestSubscriber.java:557)
	at rx.internal.operators.OperatorMergeTest.testConcurrentMergeInnerError(OperatorMergeTest.java:1232)
	...
```

pkolaczk:
@akarnokd Thanks, can you shed some light when it will be released to maven?

akarnokd:
Your other PR still incomplete without verifying the fix via a similar unit test. I usually announce the next version 3-5 days before release, however, since this affects a critical operator `merge`, it could be cut sort and release as soon both your PRs clear.

pkolaczk:
I know, working on it. Sorry, different time zone maybe. I've just started the day. ;)

akarnokd:
Thanks for the fixes. These were long-standing bugs you stumbled upon very close to the end-of-life of RxJava 1.x (March 31, 2018), after which no further development and bugfixes will be accepted. Are you in the process of moving to RxJava 2.x by any chance?

pkolaczk:
We'd like to, however the limiting factor at the moment is lack of RxScala for 2.x. A large part of our project is already on RxJava 2.x, so using just one version everywhere would be much better for everyone.

The options are either going to Scala 2.12 and using RxJava 2.x with no wrapper at all, thanks to much better Scala-Java lambda/SAM interop in Scala 2.12, or switching to something totally else. Monix maybe.

davidmoten:
LGTM, really appreciate this fix too @pkolaczk!

Issue #5763, opened 2017-12-14
2.x -- group by, custom map variant


sfitts:
We are using release 2.1.7.

As is noted in the "What's Different" page, the variant of `groupBy` that accepts a custom map factory is not currently part of the 2.x API.  We were relying on this to implement "eviction" of groups that haven't been seen for a "long time" (where "long" is user specified).  Our system maintains Flowable instances for its entire lifetime (which is hopefully weeks/months), so it is important to be able to reclaim resources as we go.

Anyway -- my questions are -- are the plans to bring this back?  Is there a suggested alternative for implementing the disposal of groups after a period of inactivity?  Any guidance would be appreciated.

akarnokd:
It was a community contribution which was not ported to 2.x by the original author. You can timeout groups and switch to an empty flow:

```java
source.groupBy(v -> key(v))
.map(g -> g.timeout(5, TimeUnit.MINUTES, Observable.empty())
.flatMap(g -> /* the usual group processing */)
```

sfitts:
Thanks, that should work nicely.

davidmoten:
I was author of that contribution to 1.x #3931. Using `timeout` was discussed but it had problems. In particular you can lose emissions. If @akarnokd is in favour I can investigate adapting the `groupBy` operator to use a Map factory again.

akarnokd:
Yes, PR welcome.

sfitts:
@davidmoten thanks for the info re: timeout dropping emissions.  We have essentially the same use case that you describe in #3931, so something that avoids that issue for Rx2 would be great.

davidmoten:
Just looking at the PR now. @akarnokd do you have any advice for optimizing the RxJava 2.x compile experience in Eclipse (latest)? I just waited 5 min 40s for the project to compile after modifying some code and this happens frequently. 

PS Just realized I'm not using the Eclipse Gradle plugin for the project in Eclipse but rather running `./gradlew eclipse` and importing as an Eclipse project. What do you do @akarnokd?

akarnokd:
None. The bug has been reported and hopefully we get it backported to 4.7.

davidmoten:
Ok. You still using Eclipse for dev on RxJava?

akarnokd:
Yes. I just do something else for 10 minutes while Eclipse is compiling. After that, as long as you don't touch the main classes, it's okay.

davidmoten:
Eclipse slow compilation bug:

[...]

PR #5860, opened 2018-02-21
2.x: Flowable.groupBy add overload with evicting map factory

davidmoten:
As per discussion in #5763, this PR adds an overload for `Flowable.groupBy` that specifies an `evictingMapFactory`. 

An example of usage taken from the new javadoc:

```java
Function<Consumer<Object>, Map<Integer, Object>> evictingMapFactory = 
   notify ->
       CacheBuilder
         .newBuilder() 
         .maximumSize(3)
         .removalListener(entry -> { 
              try {
                  // emit the value not the key!
                  notify.accept(entry.getValue());
              } catch (Exception e) {
                  throw new RuntimeException(e);
              }
            })
         .<Integer, Object> build()
         .asMap();
         
 // Emit 1000 items but ensure that the
 // internal map never has more than 3 items in it           
 Flowable
   .range(1, 1000)
   // note that number of keys is 10
   .groupBy(x -> x % 10, x-> x, true, 16, evictingMapFactory)
   .flatMap(g -> g)
   .forEach(System.out::println);
```

Note that I based this operator on the 1.x version which I think may have a bug that goes as far as the signature of that method. The eviction consumer should not be given a key but rather the value from the map. I'll pursue the 1.x issue after dealing with this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ apply plugin: "me.champeau.gradle.jmh"
 apply plugin: "com.github.hierynomus.license"
 apply plugin: "com.jfrog.bintray"
 apply plugin: "com.jfrog.artifactory"
+apply plugin: "eclipse"
 
 sourceCompatibility = JavaVersion.VERSION_1_6
 targetCompatibility = JavaVersion.VERSION_1_6
@@ -55,7 +56,7 @@ def reactiveStreamsVersion = "1.0.2"
 def mockitoVersion = "2.1.0"
 def jmhLibVersion = "1.19"
 def testNgVersion = "6.11"
-
+def guavaVersion = "24.0-jre"
 // --------------------------------------
 
 repositories {
@@ -73,6 +74,7 @@ dependencies {
 
     testImplementation "org.reactivestreams:reactive-streams-tck:$reactiveStreamsVersion"
     testImplementation "org.testng:testng:$testNgVersion"
+    testImplementation "com.google.guava:guava:$guavaVersion"
 }
 
 javadoc {
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -9708,9 +9708,116 @@ public abstract class Flowable<T> implements Publisher<T> {
         ObjectHelper.requireNonNull(valueSelector, "valueSelector is null");
         ObjectHelper.verifyPositive(bufferSize, "bufferSize");
 
-        return RxJavaPlugins.onAssembly(new FlowableGroupBy<T, K, V>(this, keySelector, valueSelector, bufferSize, delayError));
+        return RxJavaPlugins.onAssembly(new FlowableGroupBy<T, K, V>(this, keySelector, valueSelector, bufferSize, delayError, null));
     }
 
+    /**
+     * Groups the items emitted by a {@code Publisher} according to a specified criterion, and emits these
+     * grouped items as {@link GroupedFlowable}s. The emitted {@code GroupedFlowable} allows only a single
+     * {@link Subscriber} during its lifetime and if this {@code Subscriber} cancels before the
+     * source terminates, the next emission by the source having the same key will trigger a new
+     * {@code GroupedPublisher} emission. The {@code evictingMapFactory} is used to create a map that will
+     * be used to hold the {@link GroupedFlowable}s by key. The evicting map created by this factory must 
+     * notify the provided {@code Consumer<Object>} with the entry value (not the key!) when an entry in this
+     * map has been evicted. The next source emission will bring about the  completion of the evicted
+     * {@link GroupedFlowable}s and the arrival of an item with the same key as a completed {@link GroupedFlowable}
+     * will prompt the creation and emission of a new {@link GroupedFlowable} with that key.
+     * 
+     * <p>A use case for specifying an {@code evictingMapFactory} is where the source is infinite and fast and
+     * over time the number of keys grows enough to be a concern in terms of the memory footprint of the 
+     * internal hash map containing the {@link GroupedFlowable}s.
+     * 
+     * <p>The map created by an {@code evictingMapFactory} must be thread-safe.
+     * 
+     * <p>An example of an {@code evictingMapFactory} using <a href="https://google.github.io/guava/releases/24.0-jre/api/docs/com/google/common/cache/CacheBuilder.html">CacheBuilder</a> from the Guava library is below:
+     * 
+     * <pre>
+     * Function&lt;Consumer&lt;Object&gt;, Map&lt;Integer, Object&gt;&gt; evictingMapFactory = 
+     *   notify -&gt;
+     *       CacheBuilder
+     *         .newBuilder() 
+     *         .maximumSize(3)
+     *         .removalListener(entry -&gt; {
+     *              try {
+     *                  // emit the value not the key!
+     *                  notify.accept(entry.getValue());
+     *              } catch (Exception e) {
+     *                  throw new RuntimeException(e);
+     *              }
+     *            })
+     *         .&lt;Integer, Object&gt; build()
+     *         .asMap();
+     *
+     * // Emit 1000 items but ensure that the
+     * // internal map never has more than 3 items in it           
+     * Flowable
+     *   .range(1, 1000)
+     *   // note that number of keys is 10
+     *   .groupBy(x -&gt; x % 10, x -&gt; x, true, 16, evictingMapFactory)
+     *   .flatMap(g -&gt; g)
+     *   .forEach(System.out::println);
+     * </pre>
+     * 
+     * <p>
+     * <img width="640" height="360" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/groupBy.png" alt="">
+     * <p>
+     * <em>Note:</em> A {@link GroupedFlowable} will cache the items it is to emit until such time as it
+     * is subscribed to. For this reason, in order to avoid memory leaks, you should not simply ignore those
+     * {@code GroupedFlowable}s that do not concern you. Instead, you can signal to them that they may
+     * discard their buffers by applying an operator like {@link #ignoreElements} to them.
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>Both the returned and its inner {@code GroupedFlowable}s honor backpressure and the source {@code Publisher}
+     *  is consumed in a bounded mode (i.e., requested a fixed amount upfront and replenished based on
+     *  downstream consumption). Note that both the returned and its inner {@code GroupedFlowable}s use
+     *  unbounded internal buffers and if the source {@code Publisher} doesn't honor backpressure, that <em>may</em>
+     *  lead to {@code OutOfMemoryError}.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code groupBy} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param keySelector
+     *            a function that extracts the key for each item
+     * @param valueSelector
+     *            a function that extracts the return element for each item
+     * @param delayError
+     *            if true, the exception from the current Flowable is delayed in each group until that specific group emitted
+     *            the normal values; if false, the exception bypasses values in the groups and is reported immediately.
+     * @param bufferSize
+     *            the hint for how many {@link GroupedFlowable}s and element in each {@link GroupedFlowable} should be buffered
+     * @param evictingMapFactory
+     *            The factory used to create a map that will be used by the implementation to hold the 
+     *            {@link GroupedFlowable}s. The evicting map created by this factory must
+     *            notify the provided {@code Consumer<Object>} with the entry value (not the key!) when
+     *            an entry in this map has been evicted. The next source emission will bring about the
+     *            completion of the evicted {@link GroupedFlowable}s. See example above.
+     * @param <K>
+     *            the key type
+     * @param <V>
+     *            the element type
+     * @return a {@code Publisher} that emits {@link GroupedFlowable}s, each of which corresponds to a
+     *         unique key value and each of which emits those items from the source Publisher that share that
+     *         key value
+     * @see <a href="http://reactivex.io/documentation/operators/groupby.html">ReactiveX operators documentation: GroupBy</a>
+     *
+     * @since 2.1.10
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @Beta
+    public final <K, V> Flowable<GroupedFlowable<K, V>> groupBy(Function<? super T, ? extends K> keySelector,
+            Function<? super T, ? extends V> valueSelector,
+            boolean delayError, int bufferSize, 
+            Function<? super Consumer<Object>, ? extends Map<K, Object>> evictingMapFactory) {
+        ObjectHelper.requireNonNull(keySelector, "keySelector is null");
+        ObjectHelper.requireNonNull(valueSelector, "valueSelector is null");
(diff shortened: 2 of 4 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
         ObjectHelper.requireNonNull(valueSelector, "valueSelector is null");
         ObjectHelper.verifyPositive(bufferSize, "bufferSize");
 
-        return RxJavaPlugins.onAssembly(new FlowableGroupBy<T, K, V>(this, keySelector, valueSelector, bufferSize, delayError));
+        return RxJavaPlugins.onAssembly(new FlowableGroupBy<T, K, V>(this, keySelector, valueSelector, bufferSize, delayError, null));
     }
 
+    /**
+     * Groups the items emitted by a {@code Publisher} according to a specified criterion, and emits these
+     * grouped items as {@link GroupedFlowable}s. The emitted {@code GroupedPublisher} allows only a single
```
akarnokd: `GroupedPublisher` -> `GroupedFlowable`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     * {@link Subscriber} during its lifetime and if this {@code Subscriber} cancels before the
+     * source terminates, the next emission by the source having the same key will trigger a new
+     * {@code GroupedPublisher} emission. The {@code evictingMapFactory} is used to create a map that will
```
akarnokd: `GroupedPublisher` -> `GroupedFlowable`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     * be used to hold the {@link GroupedFlowable}s by key. The evicting map created by this factory must 
+     * notify when a value in this map has been evicted. The next source emission will bring about the 
+     * completion of the evicted {@link GroupedFlowable}s and the arrival of an item with the same key as a
+     * completed {@link GroupedFlowable} will prompt the creation and emission of a new {@link GroupedFlowable}
+     * with that key.
+     * 
+     * <p>A use case for specifying an {@code evictingMapFactory} is where the source is infinite and fast and
+     * over time the number of keys grows enough to be a concern in terms of the memory footprint of the 
+     * internal hash map containing the {@link GroupedFlowables}.
```
akarnokd: `{@link GroupedFlowable}s`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     * 
+     * <p>An example of an {@code evictingMapFactory} using <a href="https://google.github.io/guava/releases/19.0/api/docs/com/google/common/cache/CacheBuilder.html">CacheBuilder</a> from the Guava library is below:
+     * 
+     * <pre>
+     * Function&lt;Consumer&lt;Object&gt;, Map&lt;Integer, Object&gt;&gt; evictingMapFactory = 
+     *   notify -&gt;
+     *       CacheBuilder
+     *         .newBuilder() 
+     *         .maximumSize(3)
+     *         .removalListener(entry -> {
```
akarnokd: `->` -> `-&gt;`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     *   .flatMap(g -&gt; g)
+     *   .forEach(System.out::println);
+     * </pre>
+     * 
+     * <p>
+     * <img width="640" height="360" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/groupBy.png" alt="">
+     * <p>
+     * <em>Note:</em> A {@link GroupedFlowable} will cache the items it is to emit until such time as it
+     * is subscribed to. For this reason, in order to avoid memory leaks, you should not simply ignore those
+     * {@code GroupedPublisher}s that do not concern you. Instead, you can signal to them that they may
```
akarnokd: `GroupedPublisher` -> `GroupedFlowable`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     * discard their buffers by applying an operator like {@link #ignoreElements} to them.
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>Both the returned and its inner {@code Publisher}s honor backpressure and the source {@code Publisher}
```
akarnokd: `Publisher` -> `GroupedFlowable`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     *  is consumed in a bounded mode (i.e., requested a fixed amount upfront and replenished based on
+     *  downstream consumption). Note that both the returned and its inner {@code Publisher}s use
```
akarnokd: `Publisher` -> `GroupedFlowable`

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,112 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     *            a function that extracts the return element for each item
+     * @param delayError
+     *            if true, the exception from the current Flowable is delayed in each group until that specific group emitted
+     *            the normal values; if false, the exception bypasses values in the groups and is reported immediately.
+     * @param bufferSize
+     *            the hint for how many {@link GroupedFlowable}s and element in each {@link GroupedFlowable} should be buffered
+     * @param evictingMapFactory
+     *            The factory used to create a map that will be used by the implementation to hold the 
+     *            {@link GroupedFlowable}s. The evicting map created by this factory must notify when a
+     *            value in this map has been evicted. The next source emission will bring about the
```
akarnokd: How? By calling the provided input `Consumer<Object>` with something?

src/main/java/io/reactivex/internal/operators/flowable/FlowableGroupBy.java
```diff
@@ -78,13 +95,21 @@ protected void subscribeActual(Subscriber<? super GroupedFlowable<K, V>> s) {
...
+                Function<? super T, ? extends V> valueSelector, int bufferSize, boolean delayError, 
+                Function<? super Consumer<Object>, ? extends Map<K, Object>> mapFactory) throws Exception {
             this.actual = actual;
             this.keySelector = keySelector;
             this.valueSelector = valueSelector;
             this.bufferSize = bufferSize;
             this.delayError = delayError;
-            this.groups = new ConcurrentHashMap<Object, GroupedUnicast<K, V>>();
+            if (mapFactory == null) {
+                this.groups = new ConcurrentHashMap<Object, GroupedUnicast<K, V>>();
```
akarnokd: I'm not a fan of this structure. These could be constructor arguments prepared before the `GroupBySubscriber` gets initialized. So the try-catch will only cover the non-null case only and the use of `GroupBySubscriber` can look like before:

```java
source.subscribe(new GroupBySubscriber<T, K, V>(s, keySelector, valueSelector, bufferSize, delayError, groups, evictedGroups));
```

src/main/java/io/reactivex/internal/operators/flowable/FlowableGroupBy.java
```diff
@@ -371,6 +408,27 @@ public boolean isEmpty() {
             return queue.isEmpty();
         }
     }
+    
+    @SuppressWarnings("unchecked")
+    private static <K,V> Map<Object, GroupedUnicast<K, V>> createMap(
```
akarnokd: Not sure why this is a separate method. Also please avoid `private`.

akarnokd (changes requested):

davidmoten:
Thanks @akarnokd. I've addressed your comments and also added `@Beta` and `@since 2.1.11`.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9787,9 +9787,10 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
      *            the hint for how many {@link GroupedFlowable}s and element in each {@link GroupedFlowable} should be buffered
      * @param evictingMapFactory
      *            The factory used to create a map that will be used by the implementation to hold the 
-     *            {@link GroupedFlowable}s. The evicting map created by this factory must notify when a
-     *            value in this map has been evicted. The next source emission will bring about the 
-     *            completion of the evicted {@link GroupedFlowable}s. See example using Guava above.
+     *            {@link GroupedFlowable}s. The evicting map created by this factory must
+     *            notify the provided {@code Consumer<Object>} with the entry value (not the key!) when
+     *            an entry in this map has been evicted.. The next source emission will bring about the
```
akarnokd: Two dots.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9798,10 +9799,13 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
      *         unique key value and each of which emits those items from the source Publisher that share that
      *         key value
      * @see <a href="http://reactivex.io/documentation/operators/groupby.html">ReactiveX operators documentation: GroupBy</a>
+     *
+     * @since 2.1.11
```
akarnokd: This should fit into 2.1.10, assuming @vanniktech or @artem-zinnatullin approves.

akarnokd (approved):

build.gradle
```diff
@@ -43,6 +43,7 @@ apply plugin: "me.champeau.gradle.jmh"
 apply plugin: "com.github.hierynomus.license"
 apply plugin: "com.jfrog.bintray"
 apply plugin: "com.jfrog.artifactory"
+apply plugin: "eclipse"
```
vanniktech, line 46: why do we need the eclipse plugin?

build.gradle
```diff
@@ -55,7 +56,7 @@ def reactiveStreamsVersion = "1.0.2"
 def mockitoVersion = "2.1.0"
 def jmhLibVersion = "1.19"
 def testNgVersion = "6.11"
-
+def guavaVersion = "19.0"
```
vanniktech: Latest version is 24.0

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,116 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     * {@link GroupedFlowable}s and the arrival of an item with the same key as a completed {@link GroupedFlowable}
+     * will prompt the creation and emission of a new {@link GroupedFlowable} with that key.
+     * 
+     * <p>A use case for specifying an {@code evictingMapFactory} is where the source is infinite and fast and
+     * over time the number of keys grows enough to be a concern in terms of the memory footprint of the 
+     * internal hash map containing the {@link GroupedFlowable}s.
+     * 
+     * <p>The map created by an {@code evictingMapFactory} must be thread-safe.
+     * 
+     * <p>An example of an {@code evictingMapFactory} using <a href="https://google.github.io/guava/releases/19.0/api/docs/com/google/common/cache/CacheBuilder.html">CacheBuilder</a> from the Guava library is below:
```
vanniktech: Probably we should reference a newer version link

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,116 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     *            })
+     *         .&lt;Integer, Object&gt; build()
+     *         .asMap();
+     *
+     * // Emit 1000 items but ensure that the
+     * // internal map never has more than 3 items in it           
+     * Flowable
+     *   .range(1, 1000)
+     *   // note that number of keys is 10
+     *   .groupBy(x -&gt; x % 10, x-&gt; x, true, 16, evictingMapFactory)
```
vanniktech: we probably want `x -&gt; x` here to be consistent with the first parameter

src/main/java/io/reactivex/Flowable.java
```diff
@@ -9708,9 +9708,116 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     *   .flatMap(g -&gt; g)
+     *   .forEach(System.out::println);
+     * </pre>
+     * 
+     * <p>
+     * <img width="640" height="360" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/groupBy.png" alt="">
+     * <p>
+     * <em>Note:</em> A {@link GroupedFlowable} will cache the items it is to emit until such time as it
+     * is subscribed to. For this reason, in order to avoid memory leaks, you should not simply ignore those
+     * {@code GroupedFlowable}s that do not concern you. Instead, you can signal to them that they may
```
vanniktech, line 9766: `Instead, you can signal to them that they may` sounds very weird to me. Can we rephrase it somehow?

vanniktech (approved):
Looking good. Let's get this into 2.1.10

build.gradle
davidmoten: This is so I can run `./gradlew eclipse` for import into Eclipse without using gradle plugin. It's not required for the general build obviously but is a convenience for me. Can remove if has any negative effects.

build.gradle
davidmoten: Yes I know but 19.0 will do. I don't plan to keep this test dependency up to date. No point as it is for demonstration purposes only.

src/main/java/io/reactivex/Flowable.java
davidmoten: Not that keen to track this against latest release of guava in the future but can go to v24 if required.

src/main/java/io/reactivex/Flowable.java
davidmoten: don't understand that ??

src/main/java/io/reactivex/Flowable.java
davidmoten: this is a legacy comment that occurs in every `groupBy` overload. Perhaps we could leave rework of that phrase to another PR? I agree that the para is a bit clunky and could probably do with explicit  examples like don't filter on GroupedFlowable keys (`.filter(g -> g.getKey() > 10)`) do blah instead.

davidmoten:
I've 

* updated the guava dependency to 24.0-jre
* updated the CacheBuilder link to point to Guava 24.0 javadoc
* set the since version to 2.1.10

src/main/java/io/reactivex/Flowable.java
davidmoten: have moved to 24.0

build.gradle
vanniktech: Works for me.

src/main/java/io/reactivex/Flowable.java
vanniktech: `x-&gt; x` becomes `x-> x`
`x -&gt; x` becomes `x -> x` which is then more readable and also consistent with the first parameter

src/main/java/io/reactivex/Flowable.java
vanniktech: Then leave as is.

vanniktech (approved):

davidmoten:
Added that missing space to javadoc (`x -> x`) and squashed commits. Thanks for the review @vanniktech.

vanniktech (approved):
:shipit:

akarnokd (approved):

PR #5861, opened 2018-02-22
2.x: Fix Javadoc warnings, links to the JDK types

akarnokd:
The definition of multiple external links was wrong in `build.gradle` where the `options.links()` is actually defined as `String...` and multiple calls are not additional. This made the generated JavaDoc not have links to the JDK types. The fix now properly uses the varargs of the method. In addition, somehow the plain `http://` still wouldn't generate the proper links probably because they are redirected to `https://` for which the javadoc tool is not prepared.

In addition, 4 dangling `<p>` tags were showing up as warnings and have been removed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,8 +91,10 @@ javadoc {
     options.addStringOption("doctitle").value = ""
     options.addStringOption("header").value = ""
 
-    options.links("http://docs.oracle.com/javase/7/docs/api/")
-    options.links("http://www.reactive-streams.org/reactive-streams-${reactiveStreamsVersion}-javadoc/")
+    options.links(
+        "https://docs.oracle.com/javase/7/docs/api/",
+        "http://www.reactive-streams.org/reactive-streams-${reactiveStreamsVersion}-javadoc/"
+    )
 
     if (JavaVersion.current().isJava7()) {
         // "./gradle/stylesheet.css" only supports Java 7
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -91,8 +91,10 @@ javadoc {
     options.addStringOption("doctitle").value = ""
     options.addStringOption("header").value = ""
 
-    options.links("http://docs.oracle.com/javase/7/docs/api/")
-    options.links("http://www.reactive-streams.org/reactive-streams-${reactiveStreamsVersion}-javadoc/")
+    options.links(
+        "https://docs.oracle.com/javase/7/docs/api/",
```
JakeWharton, line 95: Any reason not to use 8 or 9?

akarnokd: No particular reason.

artem-zinnatullin: Since RxJava 2.x is compatible with JDK 6, should we link 6?

artem-zinnatullin (approved):

build.gradle
akarnokd: Well, the Javadoc is built with Java 8 and the JDK 6 docs look ugly to me anyway. Style-wise, 8 would be more seamless, 9 would be nicer due to the search box. Anything 7+ may trigger confusion about the JDK's default methods not available in Android or the misunderstanding that it's only the docs that link to Java 9, not that RxJava now depends on Java 9.

PR #5870, opened 2018-02-28
2.x: Add Flowable.switchMapCompletable{DelayError} operator

akarnokd:
This PR adds the `Flowable.switchMapCompletable` and `Flowable.switchMapCompletableDelayError` operators as requested by #4853.

The associated new marbles are:

switchMapCompletable

switchMapCompletableDelayError

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -14114,6 +14114,98 @@ public abstract class Flowable<T> implements Publisher<T> {
         return switchMap0(mapper, bufferSize, false);
     }
 
+
+    /**
+     * Maps the upstream values into {@link CompletableSource}s, subscribes to the newer one while
+     * disposing the subscription to the previous {@code CompletableSource}, thus keeping at most one
+     * active {@code CompletableSource} running.
+     * <p>
+     * <img width="640" height="521" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/switchMapCompletable.f.png" alt="">
+     * <p>
+     * Since a {@code CompletableSource} doesn't produce any items, the resulting reactive type of
+     * this operator is a {@link Completable} that can only indicate successful completion or
+     * a failure in any of the inner {@code CompletableSource}s or the failure of the current
+     * {@link Flowable}.
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator consumes the current {@link Flowable} in an unbounded manner and otherwise
+     *  does not have backpressure in its return type because no items are ever produced.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code switchMapCompletable} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dt><b>Error handling:</b></dt>
+     *  <dd>If either this {@code Flowable} or the active {@code CompletableSource} signals an {@code onError},
+     *  the resulting {@code Completable} is terminated immediately with that {@code Throwable}.
+     *  Use the {@link #switchMapCompletableDelayError(Function)} to delay such inner failures until
+     *  every inner {@code CompletableSource}s and the main {@code Flowable} terminates in some fashion.
+     *  If they fail concurrently, the operator may combine the {@code Throwable}s into a
+     *  {@link io.reactivex.exceptions.CompositeException CompositeException}
+     *  and signal it to the downstream instead. If any inactivated (switched out) {@code CompletableSource}
+     *  signals an {@code onError} late, the {@code Throwable}s will be signalled to the global error handler via
+     *  {@link RxJavaPlugins#onError(Throwable)} method as {@code UndeliverableException} errors.
+     *  </dd>
+     * </dl>
+     * @param mapper the function called with each upstream item and should return a
+     *               {@link CompletableSource} to be subscribed to and awaited for
+     *               (non blockingly) for its terminal event
+     * @return the new Completable instance
+     * @since 2.1.11 - experimental
+     * @see #switchMapCompletableDelayError(Function)
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -13951,6 +13951,97 @@ public final void subscribe(FlowableSubscriber<? super T> s) {
...
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator consumes the current {@link Flowable} in an unbounded manner and otherwise
+     *  does not have backpressure in its return type because no items are ever produced.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code switchMapCompletableDelayError} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dt><b>Error handling:</b></dt>
+     *  <dd>Errors of this {@code Flowable} and all the {@code CompletableSource}s, who had the chance
+     *  to run to their completion, are delayed until
+     *  all of the terminate in some fashion. At this point, if there was only one failure, the respective
```
artem-zinnatullin: [all of the]**m**

src/main/java/io/reactivex/internal/operators/flowable/FlowableSwitchMapCompletable.java
```diff
@@ -0,0 +1,239 @@
...
+                this.upstream = s;
+                downstream.onSubscribe(this);
+                s.request(Long.MAX_VALUE);
+            }
+        }
+
+        @Override
+        public void onNext(T t) {
+            CompletableSource c;
+
```
artem-zinnatullin: hm, I'm looking at `Flowable.switchMap` and `Observable.switchMap` and they dispose current first:

https://github.com/ReactiveX/RxJava/blob/8068404179b0d2e07da6f0e10ea95110d98e118d/src/main/java/io/reactivex/internal/operators/flowable/FlowableSwitchMap.java#L109

https://github.com/ReactiveX/RxJava/blob/8068404179b0d2e07da6f0e10ea95110d98e118d/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java#L105

While this operator maps first and then disposes current which give current more time to work which makes it more `delayErrors` friendly I guess, however I still have a question about `delayErrors` below

src/main/java/io/reactivex/internal/operators/flowable/FlowableSwitchMapCompletable.java
```diff
@@ -0,0 +1,239 @@
...
+            SwitchMapInnerObserver o = new SwitchMapInnerObserver(this);
+
+            for (;;) {
+                SwitchMapInnerObserver current = inner.get();
+                if (current == INNER_DISPOSED) {
+                    break;
+                }
+                if (inner.compareAndSet(current, o)) {
+                    if (current != null) {
+                        current.dispose();
```
artem-zinnatullin: Why do we have to do it in loop? Do we expect concurrent `onNext` from `upstream`? Or I'm missing something…

src/main/java/io/reactivex/internal/operators/flowable/FlowableSwitchMapCompletable.java
artem-zinnatullin: I've tried to adapt `Flowable.switchMap` behavior to respect `delayErrors` if mapper throws but failed so far, this is the test I wrote, maybe it'll help you:

```java
@Test
public void delayErrorsErrorInMapper() {
    final PublishProcessor<Object> trigger = PublishProcessor.create();

    TestSubscriber<Integer> ts = Flowable
            .just(1, 2)
            .switchMapDelayError(new Function<Integer, Publisher<Integer>>() {
                @Override
                public Publisher<Integer> apply(Integer integer) throws Exception {
                    if (integer.equals(1)) {
                        return Flowable
                                .combineLatest(Flowable.just(1), trigger, new BiFunction<Integer, Object, Integer>() {
                                    @Override
                                    public Integer apply(Integer integer, Object o) {
                                        return integer;
                                    }
                                })
                                .doOnCancel(new Action() {
                                    @Override
                                    public void run() throws Exception {
                                        trigger.toString();
                                    }
                                });
                    } else {
                        // Fail for second item.
                        throw new TestException("Item " + integer + " mapping failure.");
                    }
                }
            })
            .test();

    ts.assertNoValues();
    ts.assertNotTerminated();

    trigger.onNext(new Object());

    ts.assertValue(1);
    ts.assertError(new Predicate<Throwable>() {
        @Override
        public boolean test(Throwable throwable) {
            return throwable instanceof TestException && throwable.getMessage().equals("Item 2 mapping failure.");
        }
    });
}
```

artem-zinnatullin (commented):
Looks good, but comment about mapper

src/main/java/io/reactivex/internal/operators/flowable/FlowableSwitchMapCompletable.java
akarnokd: Yes, essentially it gives a larger window for errors but otherwise has no effect because there are no items to forward and a late completion has no effect anyway.

Mapping errors are terminal errors, as if the upstream ended with a failure. The problem is that the operator can't decide if it is okay to continue. You can always do `try-catch` and return a `Completable.error`.

src/main/java/io/reactivex/internal/operators/flowable/FlowableSwitchMapCompletable.java
akarnokd: `current` could be null or the previous inner observer. When that old inner observer is completing, it CASes itself to null, which fails this CAS and forces a retry. In addition, a `dispose` can also change the `inner` to the `INNER_DISPOSED` concurrently at which point it shouldn't be overridden and the loop should simply end.

vanniktech (approved):

akarnokd:
Rebased and refactored due to `concatMapCompletable` introducing the `internal.mixed` package.

vanniktech (approved):

artem-zinnatullin (approved):

Issue #5622, opened 2017-09-26
3.x: Breaking API changes


akarnokd:
This issue collects the API changes proposed for 3.x: unused operators and overloads, mistakes in signatures, etc.

- **Remove**: `Flowable.subscribe(/* 4 args */)` with `Consumer<? super Subscription>` -> #6517
- **Remove**: `Observable.subscribe(/* 4 args */)` with `Consumer<? super Disposable>` -> #6517
- **Change**: `Completable.blockingGet()` signature -> #6517
- **Change**: `Maybe.defaultIfEmpty` to return `Single`. -> #6517
- **Change**: `X.to()` to use dedicated non-generic functional type which then can be implemented by a single class for multiple base types. (#5654) -> #6514
- **Remove**: `BehaviorSubject.getValues()` -> #6516
- **Remove**: `BehaviorProcessor.getValues()`  -> #6516
- **Change**: `Single.toCompletable` to `Single.ignoreElement` to be consistent with the other types. -> #6517
- **Remove**: `AsyncSubject.getValues()` -> #6516
- **Remove**: `AsyncProcessor.getValues()` -> #6516

ZacSweers:
I'd like to propose making `takeUntil` and similar variants take a `Maybe`. The contract is that it takes until the first emission, which is the same contract as a `Maybe`.

JakeWharton:
But why? A Single seems more appropriate contract-wise, but at that point
I'd generalize it back to Observable which already exists. The need for
those overloads is actually a need for polymorphic stream types instead.

ZacSweers:
I guess it depends. Sometimes you takeUntil() something that may or may not actually emit, like many lifecycle handling patterns do

JakeWharton:
But single isn't required to emit.

ZacSweers:
I suppose this is true, if the source stream completes it just disposes the single?

artem-zinnatullin:
If we decide to keep target version #5620 lower than Java 8 then:

- Would be great to remove `throws Exception` from `io.reactivex.functions.Function#apply()` method.

else

- Would be great to remove own `io.reactivex.functions.Function` and replace it with `java.util.function.Function`

Sounds good?

akarnokd:
@artem-zinnatullin Why? Adding the `throws Exception` was well worth the effort. In fact, the idea was to make it `throws Throwable` in v3. Btw, the topic warrants its own issue/discussion.

artem-zinnatullin:
[...]

PR #5876, opened 2018-03-02
2.x: Add note about NoSuchElementException to Single.zip().

artem-zinnatullin:


src/main/java/io/reactivex/Single.java
```diff
@@ -1294,7 +1296,7 @@
      * </dl>
      * @param <T> the common value type
      * @param <R> the result value type
-     * @param sources the Iterable sequence of SingleSource instances
+     * @param sources the Iterable sequence of SingleSource instances, must not be empty
```
akarnokd: "An empty sequence will result in an `onError` signal of `NoSuchElementException`."

src/main/java/io/reactivex/Single.java
```diff
@@ -1736,7 +1740,7 @@
      * </dl>
      * @param <T> the common value type
      * @param <R> the result value type
-     * @param sources the array of SingleSource instances
+     * @param sources the array of SingleSource instances, must not be empty
```
akarnokd: "An empty array will result in an `onError` signal of `NoSuchElementException`."

akarnokd (changes requested):
There is no such requirement that these should be not empty, only consequence of being empty.

src/main/java/io/reactivex/Single.java
```diff
@@ -1279,6 +1279,8 @@
      * value and calls a zipper function with an array of these values to return a result
      * to be emitted to downstream.
      * <p>
+     * If the {@code Iterable} of {@link SingleSource}s is empty a {@link NoSuchElementException} error is signalled after subscription.
```
artem-zinnatullin, line 1282: I can add explanation if you think it adds value, something like: 

```java
/**
 * Since {@link Single} can't complete without value, there is nothing zip operator
 * can do about zipping 0 {@code Single}s.
 */
```

akarnokd:
This may be a desirable documentation enhancement the other `zip` operators all around. Would you like to do them as well in this or another PR?

src/main/java/io/reactivex/Single.java
akarnokd: This doesn't sound good to me. Simply state the fact that empty sequence results in a NoSuchElementException.

artem-zinnatullin:
Yep, I'll check other rx types to update that, for some types like `Observable` it's fine since they can just signal completion on empty input.

Will post comment with update in any case :)

artem-zinnatullin:
@akarnokd in internal discussion at work with @kxfang we've come to conclusion with that `Single.zip(Iterable)` and `Single.zipArray()` should return `Maybe` to turn this case into normal completion event

If you think it's reasonable let's add it to the list of breaking 3.x changes #5622?

akarnokd:
An empty input sequence to zip sounds more like a mistake to be warned about instead of a feature.

akarnokd (approved):

artem-zinnatullin:
So I briefly checked `zip*(Iterable/Array)` overloads of :

- `Observable`
- `Flowable`
- `Maybe`
- `Completable`

And they emit completion if input is an empty collection of sources.

I agree that if you care about emitted value, "silent" completion is not great. 
And if you don't care about value, you probably should convert result of `zip` to `Completable` to make it explicit.
However to convert result to `Completable` you need to be able to do it naturally (`.toCompletable()`), which is not consistent with current `Single.zip()` behavior because it emits error.

For consistency with other rx types I'd still suggest to return `Maybe` as a result of `Single.zip(Iterable/Array)`

An alternative would be to emit error from other rx types in this scenario, but such change doesn't sound pragmatic

Note:

>`Completable.zip()` doesn't exist because there are no values to zip, but `merge()` basically does `zip()` in case of `Completable`.

vanniktech (approved):

PR #5879, opened 2018-03-02
2.x: Add Flowable.generateAsync to bridge 1-by-1 async APIs

akarnokd:
This PR proposes a new source operator to bridge async APIs that can be repeatedly called to produce the next item (or terminate in some way) asynchronously and only call the API again once the result has been received and delivered to the downstream, while honoring the backpressure of the downstream. This means if the downstream stops requesting, the API won't be called until the latest result has been requested and consumed by the downstream.

The operator was inspired by [this StackOverflow](https://stackoverflow.com/questions/49059458/is-there-an-equivalent-of-project-reactors-flux-create-that-caters-for-push-p) question.

Example APIs could be [AsyncEnumerable](https://github.com/akarnokd/async-enumerable#async-enumerable) style, coroutine style or [async-await](https://github.com/electronicarts/ea-async).

There are no convenience overloads unlike `generate`, because these APIs likely require their interaction points created for each individual subscriber and otherwise should not be run from multiple subscriptions of the same Flowable.

(I'm still thinking about how a batched variant of this could be reasonably implemented. Batched means the downstream request is presented to the async preparation call and multiple items could be generated with a single API invocation. There are, however, complications such as controlling the request channel over multiple intermediate operators in between or how to detect and act on the case when there are fewer items delivered than requested and the API didn't indicate completion.)

The JavaDoc features the following example:

#### Example
Let's assume there is an async API with the following interface definition:

```java
interface AsyncAPI<T> extends AutoCloseable {

    CompletableFuture<Void> nextValue(Consumer<? super T> onValue);

}
```

When the call succeeds, the `onValue` is invoked with it. If there are no more items, the
` CompletableFuture`  returned by the last ` nextValue`  is completed (with ` null` ).
If there is an error, the same ` CompletableFuture`  is completed exceptionally. Each
` nextValue`  invocation creates a fresh ` CompletableFuture`  which can be cancelled
if necesary. ` nextValue`  should not be invoked again until the ` onValue`  callback
has been notified.

An instance of this API can be obtained on demand, thus the state of this operator consists of the
` AsyncAPI`  instance supplied for each individual {@code Subscriber}. The API can be transformed into
a ` Flowable`  as follows:

```java
Flowable<Integer> source = Flowable.<Integer, AsyncAPI<Integer>>generateAsync(

    // create a fresh API instance for each individual Subscriber
    () -> new AsyncAPIImpl<Integer>(),

    // this BiFunction will be called once the operator is ready to receive the next item
    // and will invoke it again only when that item is delivered via emitter.onNext()
    (state, emitter) -> {
        // issue the async API call
        CompletableFuture<Void> f = state.nextValue(

            // handle the value received
            value -> {

                // we have the option to signal that item
                if (value % 2 == 0) {
                    emitter.onNext(value);
                } else if (value == 101) {
                    // or stop altogether, which will also trigger a cleanup
                    emitter.onComplete();
                } else {
                    // or drop it and have the operator start a new call
                    emitter.onNothing();
                }
            }
        );

        // This API call may not produce further items or fail
        f.whenComplete((done, error) -> {
            // As per the CompletableFuture API, error != null is the error outcome,
            // done is always null due to the Void type
            if (error != null) {
                emitter.onError(error);
            } else {
                emitter.onComplete();
            }
        });

        // In case the downstream cancels, the current API call
        // should be cancelled as well
        emitter.replaceCancellable(() -> f.cancel(true));

        // some sources may want to create a fresh state object
        // after each invocation of this generator
        return state;
    },

    // cleanup the state object
    state -> { state.close(); }
);
```

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -2272,6 +2272,139 @@ public abstract class Flowable<T> implements Publisher<T> {
     }
 
     /**
+     * Generates items by invoking a callback, for each downstream request one by one, that sets up an
+     * asynchronous call to some API that eventually responds with an item, an error or termination, while
+     * making sure there is only one such outstanding API call in progress and honoring the
+     * backpressure of the downstream.
+     * <p>
+     * This operator allows the bridging of the asynchronous and backpressurable world with the reactive world,
+     * where backpressure is the emergent effect of making sure there is only one outstanding API call
+     * at a time which responds with at most one item per invocation.
+     * <p>
+     * Note that the implementation may have one outstanding API call even if the downstream hasn't requested more
+     * and as such, the resulting item may get cached until the downstream requests for more.
+     * <p>
+     * During the async response, the invocation protocol of the {@link FlowableAsyncEmitter} should be as follows:
+     * <pre><code>
+     *     (onNext | onNothing)? (onError | onComplete)?
+     * </code></pre>
+     * In words, an {@code onNext} or {@code onNothing} (which indicates this particular API call resulted in no
+     * items and the next API call can proceed) may be followed by a terminal event.
+     * <p>
+     * The methods {@link FlowableAsyncEmitter#onNext(Object)}, {@link FlowableAsyncEmitter#onError(Throwable)},
+     * {@link FlowableAsyncEmitter#onComplete()} and {@link FlowableAsyncEmitter#onNothing()} should not be called
+     * concurrently with each other or outside the context of the generator. The rest of the
+     * {@link FlowableAsyncEmitter} methods are thread-safe.
+     * <p>
+     * <b>Example:</b><br>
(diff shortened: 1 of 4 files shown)


vanniktech (approved):
Do you think that this is something that should live in RxJava itself?

I, personally, don't have any use case for something like that but then again I'm not using Flowable for frontend development.

akarnokd:
Since this is a source operator, it doesn't have to live in RxJava and alternative async API's aren't that widespread at the moment either. It looks like, however, that there is a significant push towards such APIs and even the JDK developers think those are the future - despite all the shortcomings when using them in a multi-value scenario.

I'm currently undecided about this.

akarnokd:
Moved to [RxJava 2 Extensions](https://github.com/akarnokd/RxJava2Extensions#flowablesgenerateasync).

PR #5887, opened 2018-03-04
2.x: Fix window(Observable|Callable) upstream handling

### LINKED ISSUE #5881: Incorrect upstream management in window(Observable) and window(Callable)
While working on coverage improvements, I've noticed `Flowable.window(Publisher<B>)` and `Flowable.window(Callable<Publisher<B>>)` (plus their `Observable` counterpart) manage the upstream incorrectly as they don't cancel it when both the main output and the current window have been cancelled.

akarnokd:
This PR fixes the upstream handling in `Observable.window(ObservableSource)` and `Observable.window(Callable<ObservableSource>)` operators to make sure if both the main output and the inner windows have been disposed, the upstream is also disposed.

Fixes #5881.

*Remark:*

There are a couple of `w != null` checks showing up as partially covered. In theory, with the right interleaving, `w` can become `null`, but unfortunately the usual `TestHelper.race()` testing is unlikely to trigger that case. There would be a higher chance with 3 threads (one disposing, one completing the main and one completing the inner window) but the current CI is effectively 1.5 - 2 cores/threads that tend to not produce a thorough interleaving.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableWindowBoundary.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableWindowBoundary.java
@@ -18,167 +18,199 @@ import java.util.concurrent.atomic.*;
 import io.reactivex.*;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.internal.disposables.DisposableHelper;
-import io.reactivex.internal.observers.QueueDrainObserver;
 import io.reactivex.internal.queue.MpscLinkedQueue;
-import io.reactivex.internal.util.NotificationLite;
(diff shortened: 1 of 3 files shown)


vanniktech (approved):

vanniktech:
The issue will get closed once this PR is merged but I don't see the Flowable implementation here.

akarnokd:
The issue refers to `window(Observable)` thus the `Observable` variants. I'll post a separate PR once I checked the `Flowable` variants.

akarnokd:
Yes, `Flowable.window(Publisher)` is also incorrect.

PR #5890, opened 2018-03-05
Added nullable annotations to subjects

bangarharshit:
Adding nullable annotations to subjects. The `throwable` and `getValue` methods are annotated with Nullable.

Quick question about styling - In some places, javadocs has trailing whitespaces but it is absent in other places. Is there a styling policy around that? Intellij is removing the trailed whitespace automatically.


akarnokd (approved):

vanniktech (approved):
Trailing whitespace was added by accident and should not be there so it's fine to remove it.

PR #5893, opened 2018-03-06
2.x: Fix Flowable.publish(-|Function) subscriber swap possible data loss

akarnokd:
This PR fixes an avoidable dataloss in the following subscriber-swap scenario with the `publish()` and `publish(Function<Flowable, Publisher>)`.

When an `onNext` changes the current array of subscribers (an existing consumer cancelled or a new one arrived), the change detection is delayed and items may get dropped even though a fresh consumer could take those values.

The algorithms were updated to fix this case as well as the emission tracking in `FlowableMulticastProcessor`: a global `emitted` value is not good here after all (subscribers joining late would indicate an incorrect aggregate demand and get overflown).

One of the advanced uses of `publish(Function)` is to implement consumption mode changes by having a mode cancel the subscription to the shared `Flowable` and synchronously subscribe a new consumer with a different behavior. 

The discovery of this shortcoming was due to a special transformation pattern: apply a transformer if the source is not empty, and in this case, deliver all items of it.

```java
FlowableTransformer<T, U> transformer = ...
source.publish(shared ->
   // let's see if there is at least 1 source item
   shared.take(1)
   // if there is one item, stop this phase and keep the item
   .concatMap(first ->
       // given the very first item, apply the transformation
       // on the "original" sequence by reattaching the first
       // item to the rest of the source
       transformer.apply(shared.startWith(first))
   )
)
.subscribe(/* ... */);
```

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
@@ -19,7 +19,7 @@ import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 
-import org.junit.Test;
+import org.junit.*;
 import org.reactivestreams.*;
 
 import io.reactivex.*;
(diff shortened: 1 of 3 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
```diff
@@ -941,4 +941,365 @@ protected void subscribeActual(Subscriber<? super Integer> s) {
         ref.get().add(new InnerSubscriber<Integer>(new TestSubscriber<Integer>()));
         ref.get().remove(null);
     }
+
+    @Test
+    @Ignore("publish() keeps consuming the upstream if there are no subscribers, 3.x should change this")
```
vanniktech, line 946: Should we document this somewhere in one of the 3.x issues?

vanniktech (approved):

src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
akarnokd: Posted as #5899.

PR #5895, opened 2018-03-07
Fix buffer() documentation to correctly describe onError behaviour

### LINKED ISSUE #5894: 2.x: buffer documentation incorrect for onError behaviour
The documentation for most `buffer` operators reads as follows:

> When the source ObservableSource completes or encounters an error, the resulting ObservableSource emits the current buffer and propagates the notification from the source ObservableSource.

This is incorrect; none of the `buffer` operators emit their buffer before propagating an error:

```
@Override
    buffer = null;
    actual.onError(t);
}
```

Note that this documentation was fixed in 1.x in #3561 (this is effectively a duplicate of issue #3560).

adamsp:
Fixes #5894. I just copied the sentence from the fixed docs in 1.x and updated to use ObservableSource/Publisher naming.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -5379,8 +5379,9 @@ public abstract class Observable<T> implements ObservableSource<T> {
     /**
      * Returns an Observable that emits buffers of items it collects from the source ObservableSource. The resulting
      * ObservableSource emits connected, non-overlapping buffers, each containing {@code count} items. When the source
-     * ObservableSource completes or encounters an error, the resulting ObservableSource emits the current buffer and
-     * propagates the notification from the source ObservableSource.
+     * ObservableSource completes, the resulting ObservableSource emits the current buffer and propagates the notification
+     * from the source ObservableSource. Note that if the source ObservableSource issues an onError notification
+     * the event is passed on immediately without first emitting the buffer it is in the process of assembling.
      * <p>
      * <img width="640" height="320" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/buffer3.png" alt="">
      * <dl>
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -5923,18 +5949,20 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
     /**
      * Returns an Observable that emits buffers of items it collects from the source ObservableSource. The resulting
      * ObservableSource emits connected, non-overlapping buffers. It emits the current buffer and replaces it with a
-     * new buffer whenever the ObservableSource produced by the specified {@code closingIndicator} emits an item.
+     * new buffer whenever the ObservableSource produced by the specified {@code boundarySupplier} emits an item.
```
adamsp, line 5952: Updated to match the actual parameter

src/main/java/io/reactivex/Observable.java
```diff
@@ -5923,18 +5949,20 @@ public final void blockingSubscribe(Observer<? super T> subscriber) {
...
+     * is passed on immediately without first emitting the buffer it is in the process of assembling.
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>This version of {@code buffer} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
      *
      * @param <B> the value type of the boundary-providing ObservableSource
      * @param boundarySupplier
      *            a {@link Callable} that produces an ObservableSource that governs the boundary between buffers.
-     *            Whenever the source {@code ObservableSource} emits an item, {@code buffer} emits the current buffer and
+     *            Whenever the supplied {@code ObservableSource} emits an item, {@code buffer} emits the current buffer and
```
adamsp, line 5965: Everywhere else the word "source" is used to indicate the observable in `source.buffer(...)` - here we provide a `boundarySupplier` so this comment is referring to the _supplied_ observable.

akarnokd (approved):

vanniktech (approved):

PR #5902, opened 2018-03-08
2.x: Perf measure of Flowable flatMap & flatMapCompletable

akarnokd:
This PR adds two benchmarks that measure the synchronous and asynchronous behavior of `Flowable.flatMap` and `Flowable.flatMapCompletable` when both are run with a `Completable` source.

Apart from the `item == 1` case, which is due to lack of scalar optimizations with `flatMapCompletable`, the `flatMapCompletable` is faster by a good margin:

image

I'd say the asynchronous use is slightly faster in general with `flatMapCompletable`:

image

-----------

Originally, this PR started out as an attempt to optimize the inner consumer tracking in `flatMapCompletable` when the `maxConcurrency` is non-default by using a special containers for 1..8 instead of the `CompositeDisposable`. Unfortunately, this made it slightly slower in general (PR column):

image

The async benchmark is within +/- 3% most of the time, and generally inconclusive.

image

An additional test method was added to verify the new tracking classes, which is left as an additional mean to verify `flatMapCompletable`'s behavior in general.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ targetCompatibility = JavaVersion.VERSION_1_6
 def junitVersion = "4.12"
 def reactiveStreamsVersion = "1.0.2"
 def mockitoVersion = "2.1.0"
-def jmhLibVersion = "1.19"
+def jmhLibVersion = "1.20"
 def testNgVersion = "6.11"
 def guavaVersion = "24.0-jre"
 def jacocoVersion = "0.8.0"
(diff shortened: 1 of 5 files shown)


davidmoten (approved):
Cool, I'll look out for the increased perf potential here, thanks.

vanniktech (approved):
Should Observable get the new test too?

akarnokd:
I'll get to it in the regular coverage improvements.

PR #5906, opened 2018-03-09
2.x: Add public constructor for TestScheduler that takes the time.

### LINKED ISSUE #5901: 2.x: Pulib constructor for TestScheduler that already takes the time
I'd like to have the ability to have an instantiated TestScheduler which already has a time. Internally it'd just delegate to `advanceTimeTo` and the constructor would also take the same parameters as that method.

What are your thoughts?

vanniktech:
- adds a time constructor to TestScheduler
- should I make it experimental or will it just be stable?

Fixes #5901

### DIFF
--- a/src/main/java/io/reactivex/schedulers/TestScheduler.java
+++ b/src/main/java/io/reactivex/schedulers/TestScheduler.java
@@ -35,6 +35,25 @@ public final class TestScheduler extends Scheduler {
     // Storing time in nanoseconds internally.
     volatile long time;
 
+    /**
+     * Creates a new TestScheduler without a particular time.
+     */
+    public TestScheduler() {
+        // No-op.
+    }
+
+    /**
+     * Moves the Scheduler's clock to a particular moment in time.
+     *
+     * @param delayTime
+     *          the point in time to move the Scheduler's clock to
+     * @param unit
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/schedulers/TestScheduler.java
```diff
@@ -35,6 +35,25 @@
...
 
+    /**
+     * Creates a new TestScheduler without a particular time.
+     */
+    public TestScheduler() {
+        // No-op.
+    }
+
+    /**
+     * Moves the Scheduler's clock to a particular moment in time.
```
akarnokd: "Creates a new TestScheduler with the specified initial virtual time."

src/main/java/io/reactivex/schedulers/TestScheduler.java
```diff
@@ -35,6 +35,25 @@
     // Storing time in nanoseconds internally.
     volatile long time;
 
+    /**
+     * Creates a new TestScheduler without a particular time.
```
akarnokd: "Creates a new TestScheduler with initial virtual time of zero."

akarnokd (changes requested):

akarnokd (approved):

PR #5907, opened 2018-03-09
2.x: Cleanup test local variable names

akarnokd:
Add an unit test that scans the unit test files for common local variable misnaming due to copy-pasting between the types, such as:

- `TestObserver ts` <-> `TestSubscriber to`
- `PublishSubject pp` <-> `PublishProcessor ps`

See the new `CheckLocalVariablesInTests` for other patterns described with a regexp.

### DIFF
--- a/src/test/java/io/reactivex/completable/CompletableTest.java
+++ b/src/test/java/io/reactivex/completable/CompletableTest.java
@@ -2599,24 +2599,24 @@ public class CompletableTest {
 
     @Test(timeout = 5000)
     public void subscribeObserverNormal() {
-        TestObserver<Object> ts = new TestObserver<Object>();
+        TestObserver<Object> to = new TestObserver<Object>();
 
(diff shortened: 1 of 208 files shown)


src/test/java/io/reactivex/completable/CompletableTest.java
```diff
@@ -3404,32 +3404,32 @@ public Object call() throws Exception {
...
-        ts.assertComplete();
-        ts.assertNoErrors();
+        to.assertValue(1);
+        to.assertComplete();
+        to.assertNoErrors();
     }
 
     @Test(timeout = 5000)
     public void startWithObservableError() {
         Observable<Object> c = normal.completable
```
vanniktech: should be `o` and could be added to the new test class

vanniktech (approved):

vanniktech (approved):

PR #5910, opened 2018-03-12
2.x: Improve coverage, fix operator logic 03/12

akarnokd:
Improve coverage of RxJava internal components. See the change explanation as comments below.

### DIFF
--- a/src/main/java/io/reactivex/exceptions/CompositeException.java
+++ b/src/main/java/io/reactivex/exceptions/CompositeException.java
@@ -278,7 +278,7 @@ public final class CompositeException extends RuntimeException {
      * @param e the {@link Throwable} {@code e}.
      * @return The root cause of {@code e}. If {@code e.getCause()} returns {@code null} or {@code e}, just return {@code e} itself.
      */
-    private Throwable getRootCause(Throwable e) {
+    /*private */Throwable getRootCause(Throwable e) {
         Throwable root = e.getCause();
         if (root == null || cause == root) {
             return e;
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferBoundarySupplier.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferBoundarySupplier.java
@@ -196,24 +196,20 @@ extends AbstractFlowableWithUpstream<T, U> {
 
             BufferBoundarySubscriber<T, U, B> bs = new BufferBoundarySubscriber<T, U, B>(this);
 
-            Disposable o = other.get();
-
-            if (!other.compareAndSet(o, bs)) {
-                return;
-            }
-
-            U b;
-            synchronized (this) {
-                b = buffer;
-                if (b == null) {
-                    return;
+            if (DisposableHelper.replace(other, bs)) {
+                U b;
+                synchronized (this) {
+                    b = buffer;
+                    if (b == null) {
+                        return;
+                    }
+                    buffer = next;
                 }
-                buffer = next;
+
+                boundary.subscribe(bs);
+
+                fastPathEmitMax(b, false, this);
             }
-
-            boundary.subscribe(bs);
-
-            fastPathEmitMax(b, false, this);
         }
 
         @Override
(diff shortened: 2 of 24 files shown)


src/main/java/io/reactivex/exceptions/CompositeException.java
```diff
@@ -278,7 +278,7 @@ public int size() {
      * @param e the {@link Throwable} {@code e}.
      * @return The root cause of {@code e}. If {@code e.getCause()} returns {@code null} or {@code e}, just return {@code e} itself.
      */
-    private Throwable getRootCause(Throwable e) {
+    /*private */Throwable getRootCause(Throwable e) {
```
akarnokd, line 281: That `cause == root` case is borderline impossible, unless this method gets called with `this`.

src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferBoundarySupplier.java
```diff
@@ -196,24 +196,20 @@ void next() {
...
-            if (!other.compareAndSet(o, bs)) {
-                return;
-            }
-
-            U b;
-            synchronized (this) {
-                b = buffer;
-                if (b == null) {
-                    return;
+            if (DisposableHelper.replace(other, bs)) {
```
akarnokd, line 199: Use dedicated helper instead of a CAS, plus the old code could unset a terminal indicator.

src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferTimed.java
```diff
@@ -324,9 +320,10 @@ public void request(long n) {
 
         @Override
         public void cancel() {
-            clear();
+            cancelled = true;
```
akarnokd, line 323: `cancelled` is checked at various places but was never actually set.

src/main/java/io/reactivex/internal/operators/flowable/FlowableDebounceTimed.java
```diff
@@ -58,7 +58,7 @@ protected void subscribeActual(Subscriber<? super T> s) {
 
         Subscription s;
 
-        final SequentialDisposable timer = new SequentialDisposable();
+        Disposable timer;
```
akarnokd, line 61: Timer is only manipulated from `onNext`, `onError` or `onComplete` and doesn't have to be atomically safe. The operator disposes the entire worker to stop any outstanding timer activity anyway.

src/main/java/io/reactivex/internal/operators/observable/ObservableBufferBoundarySupplier.java
```diff
@@ -190,24 +190,20 @@ void next() {
...
-            if (!other.compareAndSet(o, bs)) {
-                return;
-            }
-
-            U b;
-            synchronized (this) {
-                b = buffer;
-                if (b == null) {
-                    return;
+            if (DisposableHelper.replace(other, bs)) {
```
akarnokd, line 193: Use the helper method instead of a CAS that could replace a terminal indicator.

src/main/java/io/reactivex/internal/operators/observable/ObservableDebounceTimed.java
```diff
@@ -51,7 +51,7 @@ public void subscribeActual(Observer<? super T> t) {
 
         Disposable s;
 
-        final AtomicReference<Disposable> timer = new AtomicReference<Disposable>();
+        Disposable timer;
```
akarnokd, line 54: `timer` is only manipulated from `onNext`, `onError` and `onComplete`.

src/main/java/io/reactivex/internal/operators/single/SingleTakeUntil.java
```diff
@@ -85,12 +85,9 @@ public void onSubscribe(Disposable d) {
         public void onSuccess(T value) {
             other.dispose();
 
-            Disposable a = get();
+            Disposable a = getAndSet(DisposableHelper.DISPOSED);
```
akarnokd, line 88: Should be called at most once anyway, so biasing towards the `a != DisposableHelper.DISPOSED` being true.

src/main/java/io/reactivex/internal/util/MergerBiFunction.java
```diff
@@ -58,8 +58,7 @@ public MergerBiFunction(Comparator<? super T> comparator) {
             while (at.hasNext()) {
                 both.add(at.next());
             }
-        } else
-        if (s2 != null) {
+        } else {
```
akarnokd, line 61: At this point either `s1` or `s2` can be null.

akarnokd:
Done with the explanations.

vanniktech (approved):

Issue #5868, opened 2018-02-27
1.x: groupBy with evicting map factory fix


davidmoten:
I'm preparing a PR to fix a bug with the operator `groupBy` where an evicting map factory is specified.

The problem with the operator extends as far as the method signature. This is what it is now:
```java
<K, R> Observable<GroupedObservable<K, R>> groupBy(
    Func1<? super T, ? extends K> keySelector,
    Func1<? super T, ? extends R> elementSelector, 
    Func1<Action1<K>, Map<K, Object>> evictingMapFactory) 
```
This is what it should be (the only change is `Action1<K>` becomes `Action1<Object>`:
```java
<K, R> Observable<GroupedObservable<K, R>> groupBy(
    Func1<? super T, ? extends K> keySelector,
    Func1<? super T, ? extends R> elementSelector, 
    Func1<Action1<Object>, Map<K, Object>> evictingMapFactory) 
```
and the javadoc will be changed to indicate that the value from the map should be passed to the `Action1` and not the key.

This is a breaking change. How do we want this to happen?

I've got a few options:

* replace the method (breaking change, move to 1.4)
* replace the method (breaking change, move to 2.0, that will be confusing!)
* deprecate the existing method and add `groupByV2`
* deprecate the existing method and throw a `DontCallMeException` and add `groupByV2`
* remove the existing method and add `groupByV2`

Any ideas/preferences? I'm ready to go with the PR but just need to decide the change in the API.

akarnokd:
Yeah, too bad it is marked stable in 1.x. Did somebody run into a signature problem?

Deprecation but it should remain operational. Also I'd call the new method `groupByEvictable` or something like it, no version in method names.

davidmoten:
>Yeah, too bad it is marked stable in 1.x. Did somebody run into a signature problem?

Nope, I ran into a logic problem (that I created) while reviewing the old code for transfer to the new code. In short the operator uses the map to lookup an evicted GroupedObservable by key but if it has been evicted then you can't get the GroupedObservable to complete. I haven't looked closely at this yet and am not sure how the old tests were working but I'll report back on this.  The new 2.x method uses a map that is contracted to report evictions by value, not key. I imported the new 2.x tests into 1.x and two failed on 1.x till I applied the 2.x technique.

>Deprecation but it should remain operational. Also I'd call the new method groupByEvictable or something like it, no version in method names.

[...]

PR #5917, opened 2018-03-15
1.x: fix and deprecate evicting groupBy and add new overload

davidmoten:
See #5868

There is a problem with the existing 1.x `Observable.groupBy` overload with an evicting map factory where depending on the eviction behaviour of the supplied map factory, groups may not be completed and may leak memory. The problem extends to the method signature in that the eviction action should report the evicted `GroupedObservable` not the corresponding key.

This PR

* fixes the existing method at the expense of doubling up the internal map `groups` ( into `groupsCopy`) so that we can always successfully lookup the evicted group from the `groupsCopy` map. Evictions from `groups` are mirrored in `groupsCopy` after the lookup.
* deprecates the problematic overload indicating that it uses more memory than the preferred new overload
* adds a new method `groupBy` that has the corrected signature and offers `bufferSize` and `delayError` parameters for a bit more flexibility and to differentiate the erased signature from the deprecated method
* adds tests that were ported from the 2.x tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'
-    testCompile 'com.google.guava:guava:19.0'
+    testCompile 'com.google.guava:guava:24.0-jre'
     testCompile 'com.pushtorefresh.java-private-constructor-checker:checker:1.2.0'
 
     perfCompile 'org.openjdk.jmh:jmh-core:1.11.3'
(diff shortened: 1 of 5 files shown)


akarnokd (approved):

davidmoten:
Added a null check because a race with `cancel` could clear `groupsCopy`

davidmoten:
Can we get another 1.x release with this change soonish?

akarnokd:
Done.

davidmoten:
Thank you!

PR #5919, opened 2018-03-15
2.x: Add fusion support to ObservableSwitchMap inner source

akarnokd:
This PR improves the performance of the `Observable.switchMap` operator by adding fusion support as well as upgrading the `Completable` and `Single` conversion to `Observable` to be fuseable. The PR also fixes an error in the `DeferredScalarDisposable` implementation as the protocol requires calling with `null`.

The plain (`Observable.switchMap(Observable)`) conversion has promising performance improvements:

image

The conversion `Observable.switchMap(Maybe.toObservable)` has promising performance improvements:

image

The conversion `Observable.switchMap(Completable.toObservable)` is also promising, but there is a 11% loss in the scalar case for some reason:

image

Finally, the conversion  `Observable.switchMap(Single.toObservable)` got a bigger hit for the scalar case, 21% loss, but the rest are impressive:

image

### DIFF
--- a/src/jmh/java/io/reactivex/xmapz/ObservableSwitchMapCompletablePerf.java
+++ b/src/jmh/java/io/reactivex/xmapz/ObservableSwitchMapCompletablePerf.java
@@ -32,9 +32,9 @@ public class ObservableSwitchMapCompletablePerf {
     @Param({ "1", "10", "100", "1000", "10000", "100000", "1000000" })
     public int count;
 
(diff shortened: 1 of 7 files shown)


akarnokd:
I don't know why those losses appeared. The change to `ObserverCompletableObserver` has only structural change but no code path change. The `SingleToObservableObserver` now supports fusion which means longer code path on a success value.

vanniktech (approved):

PR #5922, opened 2018-03-18
change compile to implementation.

Apsaliya:
Change `compile` to `implementation` in README.


akarnokd:
According to [this](https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#new_configurations) table, the `api` would be a better match. Plus, I believe it requires `java-library` and Gradle 3.4. I'd simply leave the readme as is.

PR #5932, opened 2018-03-24
Adding support for Intellij IDEA project generation

Valloric:
(All in title.)


artem-zinnatullin:
It works just fine as is tho

akarnokd:
I've been working with the code base in both Eclipse and IntelliJ without import problems. I believe the `eclipse` and `idea` plugins are there to generate some project structure outside the IDEs, but I find it much more easy to let the respective IDE's simply import RxJava as Gradle project.

vanniktech:
Same here.

Valloric:
I'll take your word then; I'm used to doing `gradle idea` to generate the project files.

PR #5940, opened 2018-03-31
2.x: Allow @SchedulerSupport on constructors.

vanniktech:
Motivation behind this is having a class which takes multiple constructor parameters where one can be a Scheduler with a default and a custom overload. In both cases I'd want to slam `@SchedulerSupport` on it so it can be handled properly by static analysis as well.


artem-zinnatullin (approved):
lgtm

artem-zinnatullin:
nit: maybe rename PR title to `2.x: Allow @SchedulerSupport on constructors.`?

"Also" just distracts from the subject :)

PR #5942, opened 2018-03-31
Removed TERMINATED check in onNext

kjkrum:
`TERMINATED` is an empty array, so iterating over it would be harmless. The cost of the check is probably minuscule, but it was imposed on the common case where the subject is not terminated. In essence, it was optimizing for misuse of the API.


akarnokd:
Could you also update [PublishProcessor](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/processors/PublishProcessor.java#L192)?

kjkrum:
@akarnokd Done.

akarnokd (approved):

vanniktech (approved):

artem-zinnatullin (approved):

artem-zinnatullin (approved):

PR #5947, opened 2018-04-03
2.x: fix groupBy cancellation with evicting map factory

### LINKED ISSUE #5933: 2.x -- groupBy with evicting map -- seeing inconsistent behavior when eviction occurs
I am using version 2.1.11.  I'm seeing what I believe to be a bug in the groupBy operator when configured to use an evicting map.  Here is a test that demonstrates what I'm seeing:

```java
    private static class TestTicker extends Ticker {
        long tick = 0;

        @Override
        public long read() {
            return tick;
        }
    }
    
    @Test
    public void testGroupByEviction() {
        FlowableProcessor<Integer> source = PublishProcessor.create();
        TestSubscriber<Integer> subscriber1 = new TestSubscriber<>();
        TestTicker testTicker = new TestTicker();

        Function<Consumer<Object>, Map<Object, Object>> mapFactory = action -> {
            return CacheBuilder.newBuilder()
                    .expireAfterAccess(5, TimeUnit.SECONDS)
                    .removalListener(notification -> {
                        try {
                            action.accept(notification.getValue());
                        } catch (Exception ex) {
                            throw new RuntimeException(ex);
                        }
                    }).ticker(testTicker).build().asMap();
        };

        final List<String> list = Collections.synchronizedList(new ArrayList<String>());
        Flowable<Integer> stream = source.doOnCancel(() -> list.add("Source canceled"))
                .groupBy(v -> v, Functions.identity(), false, Flowable.bufferSize(), mapFactory)
                .flatMap(group -> {
            return group
                    .doOnComplete(() -> list.add("Group completed"))
                    .doOnCancel(() -> list.add("Group canceled"));
        });
        stream.doOnCancel(() -> list.add("Outer group by canceled.")).subscribe(subscriber1);
        
        // Send 3 in the same group and wait for them to be seen
        source.onNext(1);
        source.onNext(1);
        source.onNext(1);
        subscriber1.awaitCount(3);
        
        // Advance time far enough to evict the group.
        // NOTE -- Comment this line out to make the test "pass".
        testTicker.tick = TimeUnit.SECONDS.toNanos(6);
        
        // Send more data in the group (triggering eviction and recreation)
        source.onNext(1);
        source.onNext(1);

        // Wait for the last 2 and then cancel the subscription
        subscriber1.awaitCount(5);
        subscriber1.cancel();

        // Observe the result.  Note that right now the result differs depending on whether eviction occurred or
        // not.  The observed sequence in that case is:  Group completed, Outer group by canceled., Group canceled.
        // The addition of the "Group completed" is actually fine, but the fact that the cancel doesn't reach the
        // source seems like a bug.  Commenting out the setting of "tick" above will produce the "expected" sequence.
        assertEquals(Arrays.asList(
                // "Group completed", -- this is here when eviction occurs
                "Outer group by canceled.", 
                "Group canceled",
                "Source canceled"  // This is *not* here when eviction occurs
        ), list);
    }
```

As you can see from this when a group eviction occurs and the root subscription is canceled, the cancel dose not propagate beyond the `groupBy` (it does when there is no eviction).  I think the reason for this is that after the eviction processing (line 188 of FlowableGroupBy) the groupCount is 3 when I believe it should be 2).  This leads the groupBy to conclude that there are 2 active groups when the cancel occurs, when in fact there is only one.  This doesn't happen when there is no eviction (you can see this by commenting out the line that updates "tick").

davidmoten:
Fixes #5933.

When the outer stream is cancelled there may be evicted groups present in the queue to be completed.

I've added code to complete the evicted groups in the `cancel` method as suggested by @akarnokd. To protect against a race condition where the EvictionAction is called just after the evicted groups are completed in `cancel` I've added a `cancelled` check to `EvictionAction`. 

However, I am concerned about a race that I'll place in the comments on the code below.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableGroupBy.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableGroupBy.java
@@ -104,7 +104,8 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
         final AtomicInteger groupCount = new AtomicInteger(1);
 
         Throwable error;
-        volatile boolean done;
+        volatile boolean finished;
+        boolean done;
 
         boolean outputFused;
 
@@ -178,12 +179,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
 
             group.onNext(v);
 
-            if (evictedGroups != null) {
-                GroupedUnicast<K, V> evictedGroup;
-                while ((evictedGroup = evictedGroups.poll()) != null) {
-                    evictedGroup.onComplete();
-                }
-            }
+            completeEvictions();
 
             if (newGroup) {
                 q.offer(group);
@@ -197,6 +193,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                 RxJavaPlugins.onError(t);
                 return;
             }
+            done = true;
             for (GroupedUnicast<K, V> g : groups.values()) {
                 g.onError(t);
             }
@@ -205,7 +202,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                 evictedGroups.clear();
             }
             error = t;
-            done = true;
+            finished = true;
             drain();
         }
 
@@ -220,6 +217,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                     evictedGroups.clear();
                 }
                 done = true;
+                finished = true;
                 drain();
             }
         }
@@ -237,8 +235,23 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
             // cancelling the main source means we don't want any more groups
             // but running groups still require new values
             if (cancelled.compareAndSet(false, true)) {
+                completeEvictions();
                 if (groupCount.decrementAndGet() == 0) {
                     s.cancel();
+                }
+            }
+        }
+
+        private void completeEvictions() {
+            if (evictedGroups != null) {
+                int count = 0;
+                GroupedUnicast<K, V> evictedGroup;
+                while ((evictedGroup = evictedGroups.poll()) != null) {
+                    evictedGroup.onComplete();
+                    count++;
+                }
+                if (count != 0) {
+                    groupCount.addAndGet(-count);
                 }
             }
         }
@@ -279,7 +292,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                     return;
                 }
 
-                boolean d = done;
+                boolean d = finished;
 
                 if (d && !delayError) {
                     Throwable ex = error;
@@ -321,7 +334,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                 long e = 0L;
 
                 while (e != r) {
-                    boolean d = done;
+                    boolean d = finished;
 
                     GroupedFlowable<K, V> t = q.poll();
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableGroupBy.java
```diff
@@ -420,14 +430,22 @@ public boolean isEmpty() {
...
         @Override
-        public void accept(GroupedUnicast<K,V> value) {
-            evictedGroups.offer(value);
+        public void accept(GroupedUnicast<K,V> group) {
+            if (!cancelled.get()) {
+                evictedGroups.offer(group);
+            } else {
+                // note that this call to `onComplete` is safe because 
+                // GroupedUnicast is designed to allow concurrent access
+                group.onComplete();
```
davidmoten: I'm concerned about a race with `group.onNext` where cancelled has just happened and `group.onComplete` gets called above which then drains in `State`. At that point `group.onNext` gets called from another thread and that emission is lost. I'm guessing some significant rework might be needed to protect against this. It is a cancellation scenario so perhaps not important. Thoughts?

src/test/java/io/reactivex/internal/operators/flowable/FlowableGroupByTest.java
```diff
@@ -13,32 +13,66 @@
...
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.Test;
 import org.mockito.Mockito;
-import org.reactivestreams.*;
-
-import com.google.common.cache.*;
-
-import io.reactivex.*;
+import org.reactivestreams.Publisher;
```
akarnokd: Please don't unroll star imports.

davidmoten:
Fixed imports

davidmoten:
I'm thinking now that we need to complete evicted groups in the `onError` and `onComplete` methods as well (and deal with late arrivals to `evictedGroups`).

davidmoten:
I've added a bit more code to complete evicted groups in the `onError` and `onComplete` methods as well. I'll keep pondering it a bit.

src/main/java/io/reactivex/internal/operators/flowable/FlowableGroupBy.java
```diff
@@ -420,14 +438,24 @@ public boolean isEmpty() {
...
             this.evictedGroups = evictedGroups;
+            this.cancelled = cancelled;
+            this.sourceDone = sourceDone;
         }
 
         @Override
-        public void accept(GroupedUnicast<K,V> value) {
-            evictedGroups.offer(value);
+        public void accept(GroupedUnicast<K,V> group) {
+            if (!cancelled.get() && !sourceDone.get()) {
```
davidmoten: insufficient race protection here, I'll fix

davidmoten:
Excuse the delay, busy times. I'll return to this soonish. Can close the PR and reopen later if desired.

davidmoten:
Having a look now

davidmoten:
I've redone this PR. 

The test failure submitted in #5933 arose because `groupCount` needs to be decremented when groups are evicted from the map (I'm doing it when we complete evicted groups which happens before checking `groupCount` is 0 for the purpose of cancelling the source). We also needed to complete evictions when `cancel` is called as @akarnokd pointed out.

I've also included a perf improvement (without stats!) where a volatile boolean `done` served both as post-terminal event protection and as a memory barrier in the `drain` method. Splitting those two functions into `boolean done` and `volatile boolean finished` means particularly that every source `onNext` emission will do one less volatile read for post-terminal event protection (and `onComplete` and `onError` benefit too). There's a slight allocation cost I suppose.

akarnokd (approved):

vanniktech (approved):

PR #5957, opened 2018-04-12
2.x: Add Single.ignoreElement, deprecate toCompletable

akarnokd:
Rename-deprecate `Single.toCompletable` to `ignoreElement` to be in line with the same operator in the other classes and also indicate more clearly in the method name that it ignores the element.

In addition, the marbles have been updated on both `Single` and `Maybe` of this operator.

image
image


vanniktech (commented):
We deprecate a method that's stable and then add a new one that should be used that's experimental.

Does not seem right to me.

akarnokd:
I'd be completely fine with `ignoreElement` being standard right away.

vanniktech:
Then let's do it.

vanniktech (approved):

Issue #4856, opened 2016-11-16
Conflate Operator


mrudangit:
RxJava should have conflate operator. I see we have sample, debounce etc but actually they are not equivalent to conflate . 

Conflate operator should behave as Sample if updates are faster than the given interval. But if updates stops and new update comes after a long period it should push immediately. Instead of waiting for sampling period.

JakeWharton:
Is the behavior your after equivalent to zip(stream, interval())? I too
want such an operator and I was about to file an issue tomorrow. It finally
hit my three separate use cases bar for wanting it to be first party.

mrudangit:
Not sure how would you achieve with zip. 

sample at timeout out interval if source is emitting too fast but if source is slower than the timeout interval emit right away. 

O.O.O.O.O...............O 
--->-->-->-->-->-->-->-->
----X-------X-------------X

abersnaze:
Is `throttleFirst` closer to what you are looking for?

![](https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/throttleFirst.png)

JakeWharton:
throttleFirst drops events. I believe this doesn't want that. It's like "at
most" every X interval.

mrudangit:
throttleLast/Sample is what i want if source is publishing too fast but then publish the item right away don't wait for the sample timeout if it was published slower than then timeout interval. 

took stab at this  

https://github.com/mrudangit/HelloRxJava/blob/master/src/main/java/com/solutionarchitects/Conflation.java

abersnaze:
How about this?
![image](https://cloud.githubusercontent.com/assets/406038/20335040/e0c40332-ab73-11e6-9238-17bd9ff8cbbc.png)

JakeWharton:
For my case I would need green before the blues. If that's not conflate
I'll file a separate issue.

abersnaze:
I was being lazy about the timing of the end of yellow's timeout and cyan's arrival. I was mostly trying to highlight that this is `throttleFirstAndLastButNotLastIfItIsTheSameAsTheFirst`
- red is not emitted twice
- blue is emitted at the end of cyan's interval

mrudangit:
Red should output at the end of first period, Green at the end of 2nd period, dark blue at the end of 3rd period , not at the start of period as shown. How to generate diagram. probably can explain that way better

abersnaze:
Was I misinterpreting this statement?

> publish the item right away don't wait for the sample timeout

[...]

PR #5968, opened 2018-04-22
Adding Flowable.throttleAndSample - Conflate operator

0ximDigital:
This pull request is to add Flowable.throttleAndSample operator as discussed in #4856


src/main/java/io/reactivex/Flowable.java
```diff
@@ -15505,6 +15505,82 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
...
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code throttleAndSample} operates by default on the {@code computation} {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param windowDuration time to wait before emitting another item after emitting the last item
+     * @param unit           the unit of time of {@code windowDuration}
+     * @return a Flowable that performs the throttle and sample operation
+     */
+    @SchedulerSupport(SchedulerSupport.COMPUTATION)
```
akarnokd: These need version tag `@since 2.1.13 - experimental` and `@Experimental` annotation.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,194 @@
...
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+public final class FlowableThrottleAndSample<T> extends AbstractFlowableWithUpstream<T, T> {
+
+    private final long windowDuration;
```
akarnokd: Please no `private` components.

src/test/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSampleTest.java
```diff
@@ -0,0 +1,299 @@
+package io.reactivex.internal.operators.flowable;
```
akarnokd, line 13: Please add the standard file header.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,194 @@
+package io.reactivex.internal.operators.flowable;
```
akarnokd, line 13: Please add the standard file header.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,194 @@
...
+                timer.replace(worker.schedule(this, windowDuration, unit));
+            } else {
+                set(value);
+            }
+        }
+
+        @Override
+        public void run() {
+            final Disposable throttling = timer.get();
+            throttling.dispose();
```
akarnokd: There is a race here. If `run()` gets here and `onNext` is fired, throttling will be disposed and the `onNext` value gets emitted. Here then the cached value gets emitted as well and now there are two tasks delayed for the subsequent interactions.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,194 @@
...
+            super(actual, windowDuration, unit, worker);
+        }
+
+        @Override
+        void complete() {
+            final T value = getAndSet(null);
+            if (value != null) {
+                emit(value);
+            }
+            actual.onComplete();
```
akarnokd: This could race with the emission of `run()`.

akarnokd (changes requested):

src/main/java/io/reactivex/Flowable.java
```diff
@@ -15505,6 +15505,82 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
         return sample(intervalDuration, unit, scheduler);
     }
 
+    /**
+     * Returns a Flowable that emits an item immediately and then no frequently than specified duration,
+     * unless more than specified duration has passed. In which case the next item is emitted immediately.
+     * Optionally emits the very last upstream item when the upstream completes.
+     * <p>
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
```
akarnokd: Backpressure definition is required.

src/main/java/io/reactivex/Flowable.java
akarnokd: Also `@BlackpressureSupport`.

akarnokd (dismissed):

src/test/java/io/reactivex/flowable/FlowableNullTests.java
```diff
@@ -2251,6 +2251,16 @@ public void throttleLastSchedulerNull() {
         just1.throttleLast(1, TimeUnit.SECONDS, null);
     }
 
+    @Test(expected = NullPointerException.class)
```
akarnokd: This is not necessary, null checks are done automatically in `ParamValidationCheckerTest`.

akarnokd:
Please run the unit tests in `io.reactivex` locally first to resolve the style validation issues so you don't have to wait 15 minutes for Travis CI.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -15505,6 +15505,112 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
         return sample(intervalDuration, unit, scheduler);
     }
 
+    /**
+     * Returns a Flowable that emits an item immediately and then no frequently than specified duration,
```
artem-zinnatullin, line 15509: supernit: duration → period or interval?

src/main/java/io/reactivex/Flowable.java
```diff
@@ -15505,6 +15505,112 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
...
+     * unless more than specified duration has passed. In which case the next item is emitted immediately.
+     * Optionally emits the very last upstream item when the upstream completes.
```
artem-zinnatullin, line 15511: Remove "very last" part from javadoc of this overload?

artem-zinnatullin: And other similar overloads

src/main/java/io/reactivex/Flowable.java
0ximDigital: Sounds good 👍

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        @Override
+        public void cancel() {
+            cancelTimer();
+            subscription.cancel();
+        }
+
+        private void cancelTimer() {
+            DisposableHelper.dispose(timer);
```
akarnokd, line 150: Not really necessary, `worker.dispose()` will dispose any outstanding timer action.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        @Override
+        public void run() {
+            emitFromRun();
+        }
+
+        @Override
+        void complete() {
+            actual.onComplete();
```
akarnokd, line 184: the `emitFromRun` may still be running at this point and you have concurrent `onNext`/`onComplete` invocation.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,221 @@
...
+            worker.dispose();
+        }
+
+        void emit(@NonNull final T value) {
+            long r = requested.get();
+            if (r != 0L) {
+                actual.onNext(value);
+                BackpressureHelper.produced(requested, 1);
+            } else {
+                cancel();
```
akarnokd, line 160: `isDone` should be set here so that `onError`/`onComplete` doesn't get called again.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,221 @@
...
+
+        @Override
+        public void onError(final Throwable t) {
+            if (isDone) {
+                RxJavaPlugins.onError(t);
+                return;
+            }
+            isDone = true;
+            cancelTimer();
+            actual.onError(t);
```
akarnokd, line 123: Same problem as with `onComplete`, `emitFromRun` may be still going on.

akarnokd (changes requested):

akarnokd:
Since you have to take into account the concurrency of `emitFromRun` with the termination of the main source, you most likely can't avoid a queue-drain approach. The closest similar operator is the `onBackpressureLatest` except the drain is triggered by the `run()` and not `request`.

src/main/java/io/reactivex/internal/operators/flowable/FlowableThrottleAndSample.java
```diff
@@ -0,0 +1,221 @@
...
+    protected void subscribeActual(final Subscriber<? super T> s) {
+        final SerializedSubscriber<T> serial = new SerializedSubscriber<T>(s);
+        if (emitLast) {
+            source.subscribe(new ThrottleAndSampleEmitLast<T>(serial, windowDuration, unit, scheduler.createWorker()));
+        } else {
+            source.subscribe(new ThrottleAndSampleNoLast<T>(serial, windowDuration, unit, scheduler.createWorker()));
+        }
+    }
+
+    abstract static class ThrottleAndSampleSubscriber<T> extends AtomicReference<T> implements FlowableSubscriber<T>, Subscription, Runnable {
```
akarnokd, line 60: Accessing `timerInProgress` happens more often so `extends AtomicInteger` is better. Have the `AtomicReference` as instance field instead.

akarnokd:
Since no progress was made on this PR, I'm closing this in favor of #5979.

PR #5975, opened 2018-04-28
2.x: Fix refCount() connect/subscribe/cancel deadlock

akarnokd:
This PR fixes a deadlock issue with the `refCount` operator when a subscription leads to a blocking execution while the lock is being held, preventing other subscription or cancellation from executing on other threads.

The bug was discovered as the cause of a reported hang on [the Google groups](https://groups.google.com/forum/#!topic/rxjava/G0axw5PbOF4).

The [code](https://github.com/akarnokd/RxJava2Extensions/blob/master/src/main/java/hu/akarnokd/rxjava2/operators/FlowableRefCountTimeout.java) has been developed in the extensions project where the operator has [more features](https://github.com/akarnokd/RxJava2Extensions#flowabletransformersrefcount). The overloads supporting these features can be added in a separate enhancement PR.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableRefCount.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableRefCount.java
@@ -114,13 +114,18 @@ public final class FlowableRefCount<T> extends Flowable<T> {
         sd.replace(scheduler.scheduleDirect(rc, timeout, unit));
     }
 
+    void clear(RefConnection rc) {
+        connection = null;
+        DisposableHelper.dispose(rc);
+        if (source instanceof Disposable) {
+            ((Disposable)source).dispose();
+        }
+    }
+
     void terminated(RefConnection rc) {
         synchronized (this) {
             if (connection != null) {
-                connection = null;
-                if (source instanceof Disposable) {
-                    ((Disposable)source).dispose();
-                }
+                clear(rc);
             }
         }
     }
@@ -128,11 +133,7 @@ public final class FlowableRefCount<T> extends Flowable<T> {
     void timeout(RefConnection rc) {
         synchronized (this) {
             if (rc.subscriberCount == 0 && rc == connection) {
-                connection = null;
-                DisposableHelper.dispose(rc);
-                if (source instanceof Disposable) {
-                    ((Disposable)source).dispose();
-                }
+                clear(rc);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
```diff
@@ -962,4 +963,38 @@ public void badSourceCompleteDisconnect() {
             assertTrue(ex.getCause() instanceof TestException);
         }
     }
+
+    @Test(timeout = 7500)
+    public void blockingSourceAsnycCancel() throws Exception {
+        BehaviorProcessor<Integer> bp = BehaviorProcessor.createDefault(1);
+
+        Flowable<Integer> o = bp
```
vanniktech: should be `f`

vanniktech (approved):

vanniktech (approved):

vanniktech (approved):

davidmoten:
:+1: Great to see that bit of locking removed. I actually had launched into a non-blocking rewrite of the `refCount` operator a couple of months ago and had it working with current tests but started bringing in tests from 1.x and got a failure or two with some quite unusual tests, then jumped to something else leaving it on my to-do queue.  I'll cast another look at the rewrite soon.

src/main/java/io/reactivex/internal/operators/flowable/FlowableRefCount.java
```diff
@@ -31,199 +34,202 @@
...
+        boolean connect = false;
+        synchronized (this) {
+            conn = connection;
+            if (conn == null) {
+                conn = new RefConnection(this);
+                connection = conn;
+            }
+
+            long c = conn.subscriberCount;
+            if (c == 0L && conn.timer != null) {
```
artem-zinnatullin, line 78: looks like this condition is only actual for `resubscribeBeforeTimeout` test

artem-zinnatullin: It's strange to see logic and tests for something that user cannot use in RxJava

As you mentioned in PR description, let's maybe expose additional time related refCount api or try to shrink unused code?

artem-zinnatullin (commented):
looks good, but comment :)

src/main/java/io/reactivex/internal/operators/flowable/FlowableRefCount.java
akarnokd: Done: #5986.

Issue #5974, opened 2018-04-25
Single.concatEager is not implemented


boris-petrov:
`Single.concatEager` and `Completable.concatEager` are not available. `Observable.concatEager` and `Maybe.concatEager` are. Is there any reason for that or you just haven't gotten to that?

It's kind of a pain as I currently cast all my Singles to Maybes, use `Maybe.concatEager` and then cast back... :)

akarnokd:
How would `Completable.concatEager` work?

boris-petrov:
@akarnokd - well, I guess subscribe to all Completables at the same time and complete when all of them complete. I think the current `Completable.concat` subscribes to the Completables one after the other?

akarnokd:
There is already an operator for that specific behavior with `Completable`s.

boris-petrov:
Ah, OK, sorry for the confusion then. But the `Single.concatEager` is missing indeed.

akarnokd:
PR welcome.

Desislav-Petrov:
What should be the behaviour of the Single.concatEager - subscribe to all and push  just the first value received ?

akarnokd:
It should match `Maybe.concatEager`.

Desislav-Petrov:
Ok.. i think i know what's going on - will try to submit a pull req for initial review tonight (please bear in mind it will be my first contrib :))

akarnokd:
Closing via #5976.

PR #5976, opened 2018-04-28
Adding eager concats to Single

Desislav-Petrov:
- Adding concatEager operator for Singles covering:
   concatEager(Publisher<? extends SingleSource<? extends T>> sources)
   concatEager(Iterable<? extends SingleSource<? extends T>> sources)
   concatArrayEager(SingleSource<? extends T>... sources)
  
- Issue: https://github.com/ReactiveX/RxJava/issues/5974

 - Added tests for both methods, I wasn't sure if I should add tests covering just a single element in the vararg list and respectively an iterable with a single element.. any thoughts on this?


src/main/java/io/reactivex/Single.java
```diff
@@ -320,6 +320,54 @@
...
+     * @return the new Flowable instance with the specified concatenation behavior
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    @BackpressureSupport(BackpressureKind.FULL)
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public static <T> Flowable<T> concatEager(Iterable<? extends SingleSource<? extends T>> sources) {
+        return Flowable.fromIterable(sources).concatMapEager(SingleInternalHelper.<T>toFlowable());
+    }
+
```
akarnokd, line 396: There is another overload you missed:

`public static <T> Flowable<T> concatEager(Publisher<? extends SingleSource<? extends T>> sources)`

Desislav-Petrov: Oops - thanks for the review, will update shortly

akarnokd:
First of all, there is no reason keep the welcome text in the description.

> Added tests for both methods, I wasn't sure if I should add tests covering just a single element in the vararg list and respectively an iterable with a single element.. any thoughts on this?

Since this operator relies on `Flowable.concatMapEager`, there is no need for too much testing, only ensuring that the `Single.concatEager` is implemented and `Single` remains 100% covered.

src/main/java/io/reactivex/Single.java
```diff
@@ -344,6 +344,32 @@
         return Flowable.fromArray(sources).concatMapEager(SingleInternalHelper.<T>toFlowable());
     }
 
+    /**
+     * Concatenates a Publisher sequence of Publishers eagerly into a single stream of values.
```
akarnokd: SingleSources

akarnokd: Could you also fix the JavaDoc of the `Maybe` version?

src/main/java/io/reactivex/Single.java
```diff
@@ -345,7 +345,7 @@
     }
 
     /**
-     * Concatenates a Publisher sequence of Publishers eagerly into a single stream of values.
+     * Concatenates a Publisher sequence of Single sources eagerly into a single stream of values.
```
akarnokd: `SingleSource`s

Desislav-Petrov: hm.. i can see this "Single sources" instead of SingleSources in a few other places within the Javadoc of the Single class - guess we should fix all?

src/main/java/io/reactivex/Single.java
Desislav-Petrov: good idea, will do

src/main/java/io/reactivex/Single.java
akarnokd: Sure. I think they are remnants from the time when there was no `SingleSource` interface yet.

Desislav-Petrov: oh i see - fixed in all places within the Single class.

src/main/java/io/reactivex/Maybe.java
```diff
@@ -412,7 +412,7 @@
     }
 
     /**
-     * Concatenates a Publisher sequence of Publishers eagerly into a single stream of values.
+     * Concatenates a Publisher sequence of Maybe sources eagerly into a single stream of values.
```
akarnokd: Please fix these too.

Desislav-Petrov: 3 of them found and fixed.

akarnokd (approved):

akarnokd:
Great! Thanks for contributing. Once one of the regular reviewers gives an approval, I'll merge this. (It's midnight over here so that could happen 8-10 hours from now.)

vanniktech (approved):

PR #5978, opened 2018-04-29
2.x: Fix Flowable.take to route post-cancel errors to RxJavaPlugins.onError

akarnokd:
This PR makes sure errors emitted by the upstream just after the `Flowable.take` has received the specified amount of items get routed to `RxJavaPlugins.onError`.

`Observable.take()` and `Flowable.limit()` were already doing this. Added unit tests to verify the behavior across the board.


vanniktech (approved):
ObservableLimitTest already has this unit test?

akarnokd:
Not this specific one.

vanniktech:
If this is tested in a different test then fine, otherwise it might make sense to also add it.

akarnokd:
There is no `ObservableLimit`.

vanniktech:
Alright. Then merge.

PR #5979, opened 2018-04-29
2.x: Add throttleLatest operator

akarnokd:
This PR adds the `throttleLatest` operator to `Observable` and `Flowable`, also known as `conflate` (#4856): it is a combination of `throttleFirst` and `sample` whereby frequent items are sampled but the first item outside the sampling window will be emitted immediately:

throttleLatest

throttleLatest

Resolves: #4856
Replaces: #5968


vanniktech (approved):

mrudangit:
Thank You @akarnokd

PR #5982, opened 2018-04-29
2.x: deprecate getValues() in Subjects/Processors

akarnokd:
This PR deprecates the `getValues()` and `getValues(T[])` methods in `AsyncSubject`, `BehaviorSubject`, `AsyncProcessor` and `BehaviorProcessor`.

These methods were remnants of an early 2.x API design where `Subject` itself had these methods for the sake of `ReplaySubject`  and `SerializedSubject` and thus every other subject type had to implement them. The listed subjects and processors above have at most one value accessible via `getValue()` which can then be assigned to an array index manually if needed.


vanniktech (approved):

artem-zinnatullin (approved):
noice

PR #5986, opened 2018-04-30
2.x: Add refCount with count & disconnect timeout

akarnokd:
This PR exposes the additional `refCount` operation modes from #5975:

- Connect only when the specified number of `Subscriber`s/`Observer`s have subscribed
- Disconnect when the given amount of time elapsed since the very last `Subscriber`/`Observer` cancelled/disposed.
- The combination of both.

In addition, the original `refCount` received extra JavaDocs details.


vanniktech (approved):
Nice

vanniktech (approved):
Nice

Issue #2631, opened 2015-02-06
Automate Publishing of Javadocs


benjchristensen:
Javadocs should be auto-generated for snapshot and final release builds.

If it needs to be a shell script it can be incorporated into the Travis build process here: https://github.com/ReactiveX/RxJava/blob/1.x/gradle/buildViaTravis.sh

Javadocs get published to gh-pages here: https://github.com/ReactiveX/RxJava/tree/gh-pages/javadoc

Each version should go in it's own directory. Whatever the latest version is should also be in the root. There should also be a /snapshot/ folder for the latest snapshot.

benjchristensen:
Here is the hack of a script I use to generate Javadocs:

```sh
#!/bin/sh

[...]

PR #5993, opened 2018-05-05
Proposing solution for: Automate Publishing of Javadocs #2631

rpmcdougall:
This is referencing issue https://github.com/ReactiveX/RxJava/issues/2631 to automate publishing of JavaDocs for 2.x

I've added in a Gradle task using https://github.com/ajoberstar/gradle-git-publish and provided a block to pick up a Github token for auth if present. You can publish a doc with this method by generating the JavaDocs with the already included Gradle task javadocJar and then running the gitPublishPush Gradle task. I have this working here: https://rpmcdougall.github.io/RxJava/

Thanks!


rpmcdougall:
Looks like I will need to do some work to get it to fit the current javadoc folder structure and properly generate for specific versions in 2.x to match what is currently there.

build.gradle
```diff
@@ -103,6 +107,30 @@ javadoc {
...
+    // branch will be created if it doesn't exist
+    branch = 'gh-pages'
+
+    contents {
+        from 'build/docs/javadoc'
+        from(javadoc) {
+            into 'api'
+        }
+    }
+    commitMessage = 'Publishing a new page'
```
akarnokd: Including the version message would be great.

build.gradle
```diff
@@ -103,6 +107,30 @@ javadoc {
     }
 }
 
+
+
+if (project.hasProperty('githubToken')) {
+    System.setProperty('org.ajoberstar.grgit.auth.username', project.githubToken)
+}
+
+gitPublish {
```
akarnokd: Can this be conditional on the build mode? We'd like to publish only on release builds.

build.gradle
```diff
@@ -103,6 +107,30 @@ javadoc {
     }
 }
 
+
+
+if (project.hasProperty('githubToken')) {
```
akarnokd: How do I create this token value? `-PgithubToken={$javadocGithubToken}` in [buildViaTravis](https://github.com/ReactiveX/RxJava/blob/2.x/gradle/buildViaTravis.sh#L21)?

akarnokd:
Please also make sure it doesn't delete the v1 javadocs by first clearing all the target directories.

rpmcdougall:
I've noted all of your suggestions, and I'll try to get those in order. As far as the token, I was thinking it was more of an optional thing depending on how one would authenticate to the repo. If its just travis that will be doing the publishing, I don't know that its completely necessary.

build.gradle
```diff
@@ -103,6 +107,30 @@ javadoc {
...
+        branch = 'gh-pages-test'
+
+        contents {
+            from 'build/docs'
+        }
+
+        preserve {
+            include '1.x/**'
+            include '2.x/**'
+            include 'images/**'
```
akarnokd: There is an `rx` directory here to be preserved as well: https://github.com/ReactiveX/RxJava/tree/gh-pages/javadoc

build.gradle
```diff
@@ -103,6 +107,30 @@ javadoc {
     }
 }
 
+
+if (rootProject.hasProperty("releaseMode")) {
```
akarnokd: releaseMode==full

build.gradle
```diff
@@ -103,6 +107,30 @@ javadoc {
...
+    gitPublish {
+
+        repoUri = 'https://github.com/rpmcdougall/RxJava.git'
```
akarnokd: I guess this is for testing it in your own repo, right?

rpmcdougall: Indeed, it will be changed to the proper repo when everything else is worked out.

rpmcdougall:
So far it appears to be matching up with the current gh-pages branch as far as maintaining what needs to be there. However, I'm not sure the appropriate way to test making sure the release conditional is working as I'm super green when it comes to releases. Hopefully what I've got so far is mostly in line with what you are looking for. 

Here is the latest publish on my repo:
https://github.com/rpmcdougall/RxJava/tree/gh-pages
https://rpmcdougall.github.io/RxJava/javadoc/

I'm also curious how the project is migrating previous release documentation to the 2.x directory since I'm wondering how that might fit into what I am working on here if at all.

akarnokd:
To be clear, in the manual process I copy the files to 3 locations:

- under `/javadoc`, preserving the `rx` and `resources` directory
- under `2.x/javadoc`
- under `2.x/javadoc/2.x.y` release version specific

akarnokd:
I've cobbled together a script in one of my support libraries and it seems to work. I guess custom copying patterns can be made via [these kinds of cps](https://github.com/akarnokd/RxJava2Jdk8Interop/blob/master/push.sh#L41).

rpmcdougall:
I was also just looking into making a couple more Gradle copy specs, but I could also do something liek that as well.
https://docs.gradle.org/current/javadoc/org/gradle/api/file/CopySpec.html

rpmcdougall:
Latest push seems to do the trick except for the root of 2.x/javadoc which doesnt seem to be caught up with 2.19. Are some of those previous releases in that directory not full releases? I believe I was caught between my preserve statements and my copyspec colliding a bit, so I'll have to work on that.

https://github.com/rpmcdougall/RxJava/tree/cpstest/2.x/javadoc

akarnokd:
I started working on a bash script version: https://github.com/ReactiveX/RxJava/tree/JavadocAutoPublish and https://github.com/ReactiveX/RxJava/blob/JavadocAutoPublish/gradle/push.sh . I'm not proficient with shell scripts so it may take some tries to get it working with RxJava as well.

akarnokd:
Thanks for your efforts, but for now, we'll go with the shell script variant, which should give more control over what happens.

PR #5995, opened 2018-05-05
Implement 'toString' method for some Emitters

strekha:
When use `.create` method it's unclear why `emitter` is null (if call `toString` or observe object via debugger).

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableCreateTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableCreateTest.java
@@ -950,6 +950,7 @@ public class FlowableCreateTest {
                 @Override
                 public void subscribe(FlowableEmitter<Object> emitter) throws Exception {
                     assertTrue(emitter.toString().contains(entry.getValue().getSimpleName()));
+                    assertTrue(emitter.serialize().toString().contains(entry.getValue().getSimpleName()));
                 }
             }, entry.getKey()).test().assertEmpty();
         }
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableCreateTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableCreateTest.java
@@ -942,7 +942,7 @@ public class FlowableCreateTest {
         emitterMap.put(BackpressureStrategy.MISSING, FlowableCreate.MissingEmitter.class);
         emitterMap.put(BackpressureStrategy.ERROR, FlowableCreate.ErrorAsyncEmitter.class);
         emitterMap.put(BackpressureStrategy.DROP, FlowableCreate.DropAsyncEmitter.class);
-        emitterMap.put(BackpressureStrategy.LATEST, FlowableCreate.DropAsyncEmitter.class);
+        emitterMap.put(BackpressureStrategy.LATEST, FlowableCreate.LatestAsyncEmitter.class);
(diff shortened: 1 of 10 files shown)


akarnokd:
Please add unit tests that check these `toString()`s to some extent.

akarnokd (changes requested):
Please add unit tests.

src/test/java/io/reactivex/internal/operators/flowable/FlowableCreateTest.java
```diff
@@ -931,4 +933,25 @@ public void subscribe(FlowableEmitter<Object> e) throws Exception {
...
+        emitterMap.put(BackpressureStrategy.ERROR, FlowableCreate.ErrorAsyncEmitter.class);
+        emitterMap.put(BackpressureStrategy.DROP, FlowableCreate.DropAsyncEmitter.class);
+        emitterMap.put(BackpressureStrategy.LATEST, FlowableCreate.DropAsyncEmitter.class);
+        emitterMap.put(BackpressureStrategy.BUFFER, FlowableCreate.BufferAsyncEmitter.class);
+
+        for (final Map.Entry<BackpressureStrategy, Class<? extends FlowableEmitter>> entry : emitterMap.entrySet()) {
+            Flowable.create(new FlowableOnSubscribe<Object>() {
+                @Override
+                public void subscribe(FlowableEmitter<Object> emitter) throws Exception {
+                    assertTrue(emitter.toString().contains(entry.getValue().getSimpleName()));
```
akarnokd, line 952: If this fails, the test still passes because the error is turned into `onError`. Please add the `.assertEmpty()` after test(). In addition, please do the test for the serialized emitter as well.

akarnokd:
The `Maybe` and `Completable` variants are missing.

src/test/java/io/reactivex/internal/operators/observable/ObservableCreateTest.java
```diff
@@ -643,4 +643,14 @@ public void subscribe(ObservableEmitter<Object> e) throws Exception {
             RxJavaPlugins.reset();
         }
     }
+
+    @Test
+    public void emitterHasToString() {
+        Observable.create(new ObservableOnSubscribe<Object>() {
+            @Override
+            public void subscribe(ObservableEmitter<Object> emitter) throws Exception {
+                assertTrue(emitter.toString().contains(ObservableCreate.CreateEmitter.class.getSimpleName()));
```
akarnokd, line 652: Please add the serialized test here as well.

akarnokd (approved):

vanniktech (commented):
Aren't there more cases where we should do this then?

I'm still unsure how much the benefit is worth it.

akarnokd:
> Aren't there more cases where we should do this then?

Maybe on the emitters of generators. Any user-facing API object could be a candidate.

> I'm still unsure how much the benefit is worth it.

There is occasionally a question where the OP states the emitter is null and he is getting an `NPE`. Then I had to explain that when debugging, the `toString` value returns "null" and not the emitter reference is null.

vanniktech:
Should we cover the generators too?

strekha:
Think we should. So I will add this soon.
Are there any other places except the generators? Because I don't know the API so well.

akarnokd:
The `generate` methods. Please put them into a separate PR.

vanniktech (approved):
Alright. Separate PR it is.

PR #5996, opened 2018-05-05
2.x: Automatically publish the generated JavaDocs from CI

akarnokd:
This PR adds a script that publishes the generated JavaDocs from CI back to the `gh-pages` branch upon successful snapshot and release builds (this latter needs testing though via -RC releases).

The snapshot JavaDocs will go into the same directory: http://reactivex.io/RxJava/2.x/javadoc/snapshot/

The release builds update
- http://reactivex.io/RxJava/javadoc/ while preserving the old `rx` links,
- http://reactivex.io/RxJava/2.x/javadoc/ for the latest 2.x and
- http://reactivex.io/RxJava/2.x/javadoc/2.1.13/ for the specific release version.

The `push.sh` script uses a Travis-CI secured token `GITHUB_TOKEN` to talk to the GitHub API. This token was created for my account via https://github.com/settings/tokens (requires repo direct push access), and provided to Travis via "Environment variables" on the page https://travis-ci.org/ReactiveX/RxJava/settings .

If, for some reason the publishing fails, the build still counts as successful. The reason for this is that restarting the otherwise successful build will try to reupload the same library version into maven which fails. In this case, the JavaDocs can be updated via the old manual method.


akarnokd:
PRs don't trigger: https://travis-ci.org/ReactiveX/RxJava/builds/375403314#L4010

Branches do trigger: https://travis-ci.org/ReactiveX/RxJava/builds/375403308#L4018

.travis.yml
```diff
@@ -20,6 +20,7 @@ script: gradle/buildViaTravis.sh
 # Code coverage
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  - bash gradle/push.sh
```
vanniktech: push_javadoc?

gradle/push.sh
```diff
@@ -0,0 +1,120 @@
...
+git add *.js
+git add *package-list*
+
+# remove tracked but deleted files
+echo -e "Removing deleted files"
+git add -u
+
+# commit all
+echo -e "commit Travis build: $TRAVIS_BUILD_NUMBER for $buildTag"
+git commit --message "Travis build: $TRAVIS_BUILD_NUMBER for $buildTag"
```
vanniktech: Should we include in the message that I'll update the documentation?

vanniktech (approved):

.travis.yml
akarnokd: Okay.

gradle/push.sh
akarnokd: I don't think there is much ambiguity. These go into the `gh-pages` tab: https://github.com/ReactiveX/RxJava/tree/gh-pages

vanniktech (approved):

akarnokd:
Looks like the release mode also works:

- http://reactivex.io/RxJava/javadoc/
- http://reactivex.io/RxJava/2.x/javadoc/
- http://reactivex.io/RxJava/2.x/javadoc/2.1.14-RC1/

PR #5998, opened 2018-05-07
Export path from RxJava to InputStream

jonathanlb:
The submitted pattern allows exporting RxJava computations to a non-reactive system, for example for use in publishing computations to REST using Scalatra or Jackson frameworks as endpoints that can accept InputStream instances whose data can be streamed to the client, as opposed to blocking for a final value from an Observable or Flowable.  Rx operations can enrich such REST frameworks, for example merging with interval() sequences to provide heartbeats, or subscribing to other merged sources to aggregate long-lived connections.  Outside of REST, other applications include serializing RX streams for archival, debugging, and monitoring.

The unit tests demonstrate example operation, coverage, and error behavior.  While the marshalling in unit tests uses toString().getBytes(), the marshalling is more general. I've demonstrated to myself at home using basic types like Integer and Long, but the examples seemed contrived.  I would expect most users to serialize to JSON or XML strings, but the framework leaves open a general hook for non-text solutions.

Back pressure is addressed for flowables by requesting values when pulled by the InputStream.read() operations.  Buffering can be added by wrapping the ReactiveInputStream in a BufferedReader/etc and keeping the the ReactiveInputStream code simple.

Internally buffering is limited using the rx2.buffer-size property used in other areas of the codebase to limit the size of queue of items to be passed along to a reader.

Apologetically, I first thought that the submission might receive warmer reception in RxJavaStrings. 
- Marshalling performed in ReactiveInputStream is not limited to Strings; 
- the additional code is in ReactiveInputStream is minimal;
- RxJavaStrings doesn't look to have momentum moving to Rx2. 
I've hacked around with ReactiveInputStream in conjunction ObjectInput/OutputStreams and Java serialization, though I would not promote serialization that way, but one could plug in a serialization library and go to the races....

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -12,6 +12,7 @@
  */
 package io.reactivex;
 
+import java.io.InputStream;
 import java.util.*;
 import java.util.concurrent.*;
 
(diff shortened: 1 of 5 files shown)


akarnokd:
This is out of scope of RxJava; we don't do IO on our own. One reason is that the choice of serialization/marshalling method can vary quite a bit. The second reason is how the binary format is established; NIO-based `ByteBuffer`s are generally better performing but harder to work with (i.e., pooling).

You can, however, have your own separate library for this and provide transformers/converters for fluent interop.

For future reference, please open an issue first before you try to add non-trivial code so we can discuss your ideas and avoid spending time on PRs that will be rejected.

jonathanlb:
Thanks, @akarnokd. One class might not justify a library, but I've archived the work at [RxJavaIO](https://github.com/jonathanlb/RxJavaIO).  I'd be thrilled if ReactiveX would host it, for better visibility in case others want to expand, improve, contribute a NIO path, etc.

davidmoten:
Note that there are a bunch of RxJava IO, String and byte related operators already at [rxjava2-extras](https://github.com/davidmoten/rxjava2-extras) and at https://github.com/akarnokd/RxJava2Extensions.

PR #6010, opened 2018-05-17
2.x: Add assertValueSetOnly and assertValueSequenceOnly to TestObserver + TestSubscriber

vanniktech:
This basically copies the behavior of `assertValuesOnly` to the `assertValueSet` and `assertValueSequence` equivalent methods.

Coped the tests exactly from `assertValuesOnly` and adjusted them with the new added methods.

### DIFF
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -591,7 +591,7 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
      * Assert that the TestObserver/TestSubscriber received only the specified values in any order without terminating.
      * @param expected the collection of values expected in any order
      * @return this;
-     * @since 2.1.14
+     * @since 2.1.14 - Experimental
      */
     @SuppressWarnings("unchecked")
     @Experimental
@@ -644,7 +644,7 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
      * Assert that the TestObserver/TestSubscriber received only the specified values in the specified order without terminating.
      * @param sequence the sequence of expected values in order
      * @return this;
-     * @since 2.1.14
+     * @since 2.1.14 - Experimental
      */
     @SuppressWarnings("unchecked")
     @Experimental
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -588,6 +588,21 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -587,6 +587,21 @@ public final U assertValueSet(Collection<? extends T> expected) {
         return (U)this;
     }
 
+    /**
+     * Assert that the TestObserver/TestSubscriber received only the specified values in any order without terminating.
+     * @param values the values expected
+     * @return this;
+     * @since 2.1.14
```
akarnokd: Please have these like this: `2.1.14 - experimental`

src/test/java/io/reactivex/observers/TestObserverTest.java
```diff
@@ -1501,4 +1501,128 @@ public void assertValuesOnlyThrowsWhenErrored() {
...
+        to.assertValueSetOnly(Collections.<Integer>emptySet());
+
+        to.onNext(5);
+        to.assertValueSetOnly(Collections.singleton(5));
+
+        to.onNext(-1);
+
+        try {
+            to.assertValueSetOnly(Collections.singleton(5));
+            fail();
```
akarnokd: `fail` throws `AssertionError` so this test always passes. Perhaps throw a `RuntimeException` or add a fail message and verify the `catch` wasn't that message. Also this is overlooked in many other test using `fail` and `catch (AssertionError)` so it would be great if this could be fixed as well.

akarnokd (commented):
Some changes would be great.

vanniktech:
Addressed your feedback.

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -587,6 +587,21 @@ public final U assertValueSet(Collection<? extends T> expected) {
         return (U)this;
     }
 
+    /**
+     * Assert that the TestObserver/TestSubscriber received only the specified values in any order without terminating.
+     * @param expected the collection of values expected in any order
+     * @return this;
+     * @since 2.1.14
```
akarnokd: Please add ` - experimental`.

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -625,6 +640,21 @@ public final U assertValueSequence(Iterable<? extends T> sequence) {
         return (U)this;
     }
 
+    /**
+     * Assert that the TestObserver/TestSubscriber received only the specified values in the specified order without terminating.
+     * @param sequence the sequence of expected values in order
+     * @return this;
+     * @since 2.1.14
```
akarnokd: Please add ` - experimental`.

vanniktech:
Whops. Forgot to commit that one, thanks.

akarnokd (approved):

PR #6012, opened 2018-05-22
Stops Integer boxing via null checks in Observable.concatEager

jnlopar:
Replaces with verifyPositive as in other similar methods.

This seems to be a trivial change and makes this more consistent with other areas of the code. Let me know if it needs its own issue or any specific tests. Given the methods only accept primitive ints, and passing a null Integer would throw a runtime exception, I'm not sure there's anything needed outside of the already present testing.


artem-zinnatullin (approved):
Noice 

also means that there were no unit tests that'd check boundaries of the parameters but it's not that critical tbh

akarnokd:
There is already a reflection based range check for the base classes so these null checks can be simply removed; the delegated calls should perform the verifyPositive calls already too.

akarnokd:
Thanks for contribution. I've posted a broader solution to this problem in #6014.

PR #6014, opened 2018-05-23
2.x: Fix & prevent null checks on primitives

### LINKED ISSUE #6013: Should we have a code check for ObjectHelper.requireNonNull on a primitive?
I found a couple small instances of `ObjectHelper.requireNonNull` being used on primitive ints, which I patched in #6012 . Does the Rx codebase have a code pattern checking framework that we could augment to find other potential instances of this that could be causing unnecessary boxing? Errorprone, for example, has this check: https://errorprone.info/bugpattern/PreconditionsCheckNotNullPrimitive for a similar pattern on Guava's `Preconditions.checkNotNull`, so having one for `ObjectHelper` could be potentially useful here.

akarnokd:
This PR removes the accidental null checks on primitives in two `Observable` methods and adds a trap method to `ObjectHelper` to both highlight and fail the tests in case the null checks are still attempted. The method is marked as deprecated so it will also show up in (Eclipse's) problems window.

Replaces #6012 
Resolves #6013

### DIFF
--- a/src/main/java/io/reactivex/internal/functions/ObjectHelper.java
+++ b/src/main/java/io/reactivex/internal/functions/ObjectHelper.java
@@ -128,4 +128,17 @@ public final class ObjectHelper {
             return ObjectHelper.equals(o1, o2);
         }
     }
+
+    /**
+     * Trap null-check attempts on primitives.
+     * @param value the value to check
+     * @param message the message to print
+     * @return the value
+     * @deprecated this method should not be used as there is no need
+     * to check primitives for nullness.
+     */
+    @Deprecated
+    public static long requireNonNull(long value, String message) {
+        throw new InternalError("Null check on a primitive: " + message);
+    }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/internal/functions/ObjectHelper.java
```diff
@@ -128,4 +128,17 @@ public boolean test(Object o1, Object o2) {
...
+     * Trap null-check attempts on primitives.
+     * @param value the value to check
+     * @param message the message to print
+     * @return the value
+     * @deprecated this method should not be used as there is no need
+     * to check primitives for nullness.
+     */
+    @Deprecated
+    public static long requireNonNull(long value, String message) {
+        throw new InternalError("Null check on a primitive: " + message);
```
artem-zinnatullin, line 142: Probably not the best choice of error type, `InternalError` indicates JVM error, combined with "Null check on a primitive" message might mislead the developer into thinking that something is wrong with bytecode/VM hehe

akarnokd: Especially if they run a value types enabled vm ;)

Issue #6015, opened 2018-05-23
Single.flatMapPublisher full implementation


davidmoten:
I noticed `Single.flatMapPublisher` is implemented as `toFlowable().flatMapPublisher(...)` which subjects the one flowable to batched requests. I'd like to have requests go straight through to the one flowable. Can I contribute this operator?

The operator at the moment is at https://github.com/davidmoten/rxjava2-http/blob/master/src/main/java/org/davidmoten/rx2/io/internal/FlowableSingleFlatMapPublisher.java without fusion. I can add more tests, fusion (transitive as per `doOnEach`) and follow the projects's naming and visibility conventions.

akarnokd:
Sure.

akarnokd:
In case you haven't started yet, you may want to wait for #6017 so the changes can be benchmarked.

davidmoten:
That's brilliant, thanks

davidmoten:
A question about fusion. I notice that `timeout(Long, TimeUnit)` doesn't support fusion and I suspect the operator I'm writing is similar. The problem is that a subscription is handed to downstream while the Single emission is being calculated and the resultant mapped Flowable is unknown. Just after the subscription is handed to downstream (via `onSubscribe`) `requestFusion` is called but of course at that point the fusion characteristics of the mapped Flowable is unknown and `QueueFuseable.NONE` is returned. 

Should I chase fusion for this operator? Any suggestions?

akarnokd:
Fusion won't work here due to the delayed subscription to the Flowable. Don't bother with it.

davidmoten:
Righto, ta.

PR #6021, opened 2018-05-26
2.x: Single.flatMapPublisher full implementation

davidmoten:
See discussion in #6015.

Resolves: #6015

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
+++ b/src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
@@ -1,0 +1,137 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.internal.operators.single;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import io.reactivex.Flowable;
+import io.reactivex.FlowableSubscriber;
+import io.reactivex.Scheduler;
+import io.reactivex.SingleObserver;
+import io.reactivex.SingleSource;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.functions.Function;
+import io.reactivex.internal.functions.ObjectHelper;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
+
+/**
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+import io.reactivex.Flowable;
+import io.reactivex.SingleObserver;
+import io.reactivex.SingleSource;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.functions.Function;
+import io.reactivex.internal.functions.ObjectHelper;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
+
+public final class SingleFlatMapPublisher<S, T> extends Flowable<T> {
```
akarnokd: Please add a JavaDoc here (you can copy the docs from the operator) that includes `@since 2.1.15` (no need for experimental). Also it would be great of the type arguments would be `T, R`.

src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+        this.mapper = mapper;
+    }
+
+    @Override
+    protected void subscribeActual(Subscriber<? super T> actual) {
+        source.subscribe(new SingleFlatMapPublisherObserver<S, T>(actual, mapper));
+    }
+
+    static final class SingleFlatMapPublisherObserver<S, T> extends AtomicLong
+            implements SingleObserver<S>, Subscriber<T>, Subscription {
```
akarnokd: Please `FlowableSubscriber`, otherwise the safety layer is added as a mapped `Flowable` will think it talks to an arbitrary consumer.

src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+
+        private static final long serialVersionUID = 7759721921468635667L;
+
+        final Subscriber<? super T> actual;
+        final Function<? super S, ? extends Publisher<? extends T>> mapper;
+        final AtomicReference<Subscription> parent = new AtomicReference<Subscription>();
```
akarnokd: Nit: initialize in constructor?

src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+        public void onSubscribe(Disposable d) {
+            this.disposable = d;
+            actual.onSubscribe(this);
+        }
+
+        @Override
+        public void onSuccess(S value) {
+            Publisher<? extends T> f;
+            try {
+                f = ObjectHelper.requireNonNull(mapper.apply(value), "mapper returns null");
```
akarnokd: Usually it says `The mapper returned a null Publisher`.

src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+            } catch (Exception e) {
```
akarnokd: `Throwable e`

akarnokd (changes requested):

src/main/java/io/reactivex/internal/operators/single/SingleFlatMapPublisher.java
davidmoten: Thanks, I'll go and check my other libraries for that one.

davidmoten:
I've made those changes.

akarnokd (approved):

vanniktech (approved):

PR #6024, opened 2018-05-27
2.x: Dedicated {Single|Maybe}.flatMap{Publisher|Observable} & andThen(Observable|Publisher) implementations

akarnokd:
This PR implements the following operators directly instead of conversions between base types:

- `Single.flatMapObservable`
- `Maybe.flatMapObservable`
- `Maybe.flatMapPublisher`
- `Completable.andThen(Observable)`
- `Completable.andThen(Publisher)`

The `Single.flatMapPublisher` was reimplemented in #6021 already.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/mixed/CompletableAndThenObservable.java
+++ b/src/main/java/io/reactivex/internal/operators/mixed/CompletableAndThenObservable.java
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.internal.operators.mixed;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.reactivex.*;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.internal.disposables.DisposableHelper;
+
+/**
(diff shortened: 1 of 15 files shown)


src/main/java/io/reactivex/internal/operators/mixed/CompletableAndThenObservable.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        @Override
+        public void onError(Throwable t) {
+            downstream.onError(t);
+        }
+
+        @Override
+        public void onComplete() {
+            ObservableSource<? extends R> p = other;
```
vanniktech: `o`

src/main/java/io/reactivex/internal/operators/mixed/MaybeFlatMapObservable.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        @Override
+        public void onSubscribe(Disposable d) {
+            DisposableHelper.replace(this, d);
+        }
+
+        @Override
+        public void onSuccess(T t) {
+            ObservableSource<? extends R> p;
```
vanniktech: `o`

src/main/java/io/reactivex/internal/operators/mixed/SingleFlatMapObservable.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        @Override
+        public void onSubscribe(Disposable d) {
+            DisposableHelper.replace(this, d);
+        }
+
+        @Override
+        public void onSuccess(T t) {
+            ObservableSource<? extends R> p;
```
vanniktech: `o`

vanniktech (approved):

akarnokd:
Benchmark comparison with baseline: (i7 4770K, Windows 7 x64, Java 8u172, JMH 1.20):

image

Other that 3 cases with slightly degraded throughput (probably Windows related), these are nice improvements.

Comparing `Maybe` and `Single` flattening:

image

These should be practically the same. There is also a slight performance drop for `count=1 flatMapHide publisher` which is likely due to the backpressure inside `Flowable.just()`.

vanniktech (approved):

davidmoten (approved):

Issue #5988, opened 2018-04-30
2.x: Flowable.blockingSubscribe(lambdas) is unbounded and can lead to OOME


akarnokd:
I can't remember why, but `Flowable.blockingSubscribe(Consumer,...)` uses unbounded requests, which can lead to `OutOfMemoryError` in both synchronous and asynchronous scenarios.

The problem was raised in [this StackOverflow question](https://stackoverflow.com/questions/50106725/rx-java2-floawable-rebatchrequests-doesnt-rebatch-requests) and apart from the obvious solutions, having such unbounded operator over an asynchronous boundary can be a problem regardless.

Possible resolutions:

- **Behavior change**: bound it to `bufferSize()` in 2.x
- (and/or) Introduce overload(s) with `prefetch` but leave the existing ones unbounded

Alternatively, this could be deferred until 3.x without worrying of the behavior change.

RomanWuattier:
`Flowable.blockingSubscribe(Consumer,…)` is using `Functions.MaxRequestSubscription()` which could lead to `OutOfMemoryError`. Is it the issue?
In that case, I suggest to:
* Keep existing `Flowable.blockingSubscribe(Consumer,…)` unbounded
* Deprecate existing `Flowable.blockingSubscribe(Consumer,…)`
* Introduce new overloads with `bufferSize()` and `prefetch`

WDYT? Can I work on this?

akarnokd:
Create a new overload:

`blockingSubscribe(Consumer<? super T> onNext, Consumer<? super Throwable> onError, Action onComplete, int bufferSize)`

have a dedicated implementation servicing it and leave the other overloads unbounded.

PR #6026, opened 2018-05-29
Fix Flowable.blockingSubscribe is unbounded and can lead to OOME

RomanWuattier:
Create and bound new `blockingSubscribe` overloads to `bufferSize`.
* Create new overloads with `bufferSize`
* Create a `boundedConsumer`
* Create a `BoundedSubsciber`

Close: #5988

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -5856,6 +5856,38 @@ public abstract class Flowable<T> implements Publisher<T> {
     /**
      * Subscribes to the source and calls the given callbacks <strong>on the current thread</strong>.
      * <p>
+     * If the Flowable emits an error, it is wrapped into an
+     * {@link io.reactivex.exceptions.OnErrorNotImplementedException OnErrorNotImplementedException}
+     * and routed to the RxJavaPlugins.onError handler.
+     * Using the overloads {@link #blockingSubscribe(Consumer, Consumer)}
+     * or {@link #blockingSubscribe(Consumer, Consumer, Action)} instead is recommended.
+     * <p>
+     * Note that calling this method will block the caller thread until the upstream terminates
+     * normally or with an error. Therefore, calling this method from special threads such as the
+     * Android Main Thread or the Swing Event Dispatch Thread is not recommended.
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator consumes the source {@code Flowable} in an bounded manner (up to bufferSize
+     *  outstanding request amount for items).</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code blockingSubscribe} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param onNext the callback action for each source value
+     * @param bufferSize the size of the buffer
+     * @since 2.1.15 - experimental
+     * @see #blockingSubscribe(Consumer, Consumer)
+     * @see #blockingSubscribe(Consumer, Consumer, Action)
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @Experimental
+    public final void blockingSubscribe(Consumer<? super T> onNext, int bufferSize) {
+        FlowableBlockingSubscribe.subscribe(this, onNext, Functions.ON_ERROR_MISSING, Functions.EMPTY_ACTION, bufferSize);
+    }
+
+    /**
+     * Subscribes to the source and calls the given callbacks <strong>on the current thread</strong>.
+     * <p>
      * Note that calling this method will block the caller thread until the upstream terminates
      * normally or with an error. Therefore, calling this method from special threads such as the
      * Android Main Thread or the Swing Event Dispatch Thread is not recommended.
@@ -5877,6 +5909,32 @@ public abstract class Flowable<T> implements Publisher<T> {
         FlowableBlockingSubscribe.subscribe(this, onNext, onError, Functions.EMPTY_ACTION);
     }
 
+    /**
+     * Subscribes to the source and calls the given callbacks <strong>on the current thread</strong>.
+     * <p>
+     * Note that calling this method will block the caller thread until the upstream terminates
+     * normally or with an error. Therefore, calling this method from special threads such as the
+     * Android Main Thread or the Swing Event Dispatch Thread is not recommended.
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator consumes the source {@code Flowable} in an bounded manner (up to bufferSize
+     *  outstanding request amount for items).</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code blockingSubscribe} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param onNext the callback action for each source value
+     * @param onError the callback action for an error event
+     * @param bufferSize the size of the buffer
+     * @since 2.1.15 - experimental
+     * @see #blockingSubscribe(Consumer, Consumer, Action)
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @Experimental
+    public final void blockingSubscribe(Consumer<? super T> onNext, Consumer<? super Throwable> onError,
+        int bufferSize) {
+        FlowableBlockingSubscribe.subscribe(this, onNext, onError, Functions.EMPTY_ACTION, bufferSize);
+    }
 
     /**
      * Subscribes to the source and calls the given callbacks <strong>on the current thread</strong>.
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -5853,6 +5853,37 @@ public final void blockingSubscribe(Consumer<? super T> onNext) {
...
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>The operator consumes the source {@code Flowable} in an bounded manner (up to bufferSize
+     *  outstanding request amount for items).</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code blockingSubscribe} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param onNext the callback action for each source value
+     * @param bufferSize the size of the buffer
+     * @since 2.0
```
akarnokd: Please designate as `2.1.15 - experimental` here and to the other methods.

src/main/java/io/reactivex/Flowable.java
```diff
@@ -5853,6 +5853,37 @@ public final void blockingSubscribe(Consumer<? super T> onNext) {
...
+     * @see #blockingSubscribe(Consumer, Consumer)
+     * @see #blockingSubscribe(Consumer, Consumer, Action)
+     */
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final void blockingSubscribe(Consumer<? super T> onNext, int bufferSize) {
```
akarnokd: Please add `@Experimental` annotation to these new methods.

src/main/java/io/reactivex/internal/subscribers/BoundedSubscriber.java
```diff
@@ -0,0 +1,139 @@
...
+public final class BoundedSubscriber<T> extends AtomicReference<Subscription>
+        implements FlowableSubscriber<T>, Subscription, Disposable, LambdaConsumerIntrospection {
+
+    private static final long serialVersionUID = -7251123623727029452L;
+    final Consumer<? super T> onNext;
+    final Consumer<? super Throwable> onError;
+    final Action onComplete;
+    final Consumer<? super Subscription> onSubscribe;
+
+    private int bufferSize;
```
akarnokd: Please no private instance fields.

src/main/java/io/reactivex/internal/subscribers/BoundedSubscriber.java
```diff
@@ -0,0 +1,139 @@
...
+    private final AtomicInteger internalBuffer;
```
akarnokd: Why is this atomic?

src/main/java/io/reactivex/internal/subscribers/BoundedSubscriber.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    public void onNext(T t) {
+        if (!isDisposed()) {
+            try {
+                if (internalBuffer.getAndIncrement() < bufferSize) {
+                    onNext.accept(t);
+                }
+                if (internalBuffer.get() == bufferSize) {
```
akarnokd: Why would you end the flow when the bufferSize number of elements have been received?

src/test/java/io/reactivex/internal/operators/flowable/FlowableBlockingTest.java
```diff
@@ -121,6 +185,56 @@ public void run() throws Exception {
...
+            }
+        };
+
+        Flowable.range(1, 5)
+                .subscribeOn(Schedulers.computation())
+                .blockingSubscribe(cons, cons, action, 3);
+
+        assertEquals(Arrays.asList(1, 2, 3, 100), list);
+    }
+
```
akarnokd: Please add an unit test that transfers 1 million items.

akarnokd (dismissed):
There are some small missing pieces and one logical error.

src/main/java/io/reactivex/internal/subscribers/BoundedSubscriber.java
```diff
@@ -0,0 +1,139 @@
...
+                RxJavaPlugins.onError(new CompositeException(t, e));
+            }
+        } else {
+            RxJavaPlugins.onError(t);
+        }
+    }
+
+    @Override
+    public void onComplete() {
+        if (get() != SubscriptionHelper.CANCELLED) {
```
davidmoten, line 105: if we are going to check the value of `get()` may as well do a compareAndSet loop for concurrent cancellation?

akarnokd: Don't. That level of precision it is not really worth it here.

src/main/java/io/reactivex/internal/subscribers/BoundedSubscriber.java
```diff
@@ -69,12 +67,14 @@ public void onSubscribe(Subscription s) {
...
-                if (internalBuffer.getAndIncrement() < bufferSize) {
-                    onNext.accept(t);
-                }
-                if (internalBuffer.get() == bufferSize) {
-                    onComplete.run();
-                    get().cancel();
+                onNext.accept(t);
+
+                int c = consumed + 1;
+                if (c == bufferSize) {
```
akarnokd: It is better to use a low-watermark than requesting only when the entire buffer has been consumed. Define a `final int limit` field, in the constructor set it as `this.limit = bufferSize - (bufferSize >> 2);`, then compare and request as `limit`.

akarnokd (commented):
One algorithmic adjustment and good to go.

akarnokd:
Also it would be great if you copied the unit tests that crash the callbacks so that the coverage on the new class gets very high.

akarnokd:
https://codecov.io/gh/ReactiveX/RxJava/pull/6026/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9pby9yZWFjdGl2ZXgvaW50ZXJuYWwvc3Vic2NyaWJlcnMvQm91bmRlZFN1YnNjcmliZXIuamF2YQ==

RomanWuattier:
I created a `BoundedSubscriberTest.java` to trigger `Consumer` and `Action` exceptions, but the code coverage remains unchanged. Did I miss a test?

src/main/java/io/reactivex/internal/operators/flowable/FlowableBlockingSubscribe.java
```diff
@@ -108,4 +108,23 @@ private FlowableBlockingSubscribe() {
...
+    /**
+     * Subscribes to the source and calls the given actions on the current thread.
+     * @param o the source publisher
+     * @param onNext the callback action for each source value
+     * @param onError the callback action for an error event
+     * @param onComplete the callback action for the completion event.
+     * @param bufferSize the number of elements to prefetch from the source Publisher
+     * @param <T> the value type
+     */
+    @SuppressWarnings("ResultOfMethodCallIgnored")
```
akarnokd: Why is this needed. It never appears anywhere else in RxJava. Please remove.

src/main/java/io/reactivex/internal/subscribers/BoundedSubscriber.java
```diff
@@ -0,0 +1,141 @@
...
+
+    public BoundedSubscriber(Consumer<? super T> onNext, Consumer<? super Throwable> onError,
+                            Action onComplete,
+                            Consumer<? super Subscription> onSubscribe) {
+        super();
+        this.onNext = onNext;
+        this.onError = onError;
+        this.onComplete = onComplete;
+        this.onSubscribe = onSubscribe;
+        this.bufferSize = ((Functions.BoundedConsumer) onSubscribe).getBufferSize();
```
akarnokd: Why is this cast needed? Can't you could simply pass the bufferSize into the constructor?

src/test/java/io/reactivex/internal/subscribers/BoundedSubscriberTest.java
```diff
@@ -0,0 +1,292 @@
...
+
+            BoundedSubscriber<Object> o = new BoundedSubscriber<Object>(new Consumer<Object>() {
+                @Override
+                public void accept(Object o) throws Exception {
+                    received.add(o);
+                }
+            }, new Consumer<Throwable>() {
+                @Override
+                public void accept(Throwable throwable) throws Exception {
+                    throw new TestException("Inner");
```
akarnokd, line 128: Place a breakpoint here and debug the test. It should reveal why the code coverage doesn't seem to execute that code in `BoundedSubscriber`.

akarnokd (changes requested):

src/test/java/io/reactivex/internal/subscribers/BoundedSubscriberTest.java
```diff
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) 2016-present, RxJava Contributors.
- *
+ * <p>
```
akarnokd: Please don't change the license headers.

src/test/java/io/reactivex/internal/subscribers/BoundedSubscriberTest.java
```diff
@@ -0,0 +1,388 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ * <p>
```
akarnokd: Please don't change the license headers.

src/test/java/io/reactivex/internal/subscribers/BoundedSubscriberTest.java
```diff
@@ -0,0 +1,388 @@
...
+import io.reactivex.functions.Action;
+import io.reactivex.functions.Consumer;
+import io.reactivex.internal.functions.Functions;
+import io.reactivex.internal.subscriptions.BooleanSubscription;
+import io.reactivex.plugins.RxJavaPlugins;
+import io.reactivex.processors.PublishProcessor;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import org.testng.annotations.Test;
```
akarnokd: Please use `org.junit.Test` as the TestNG tests are not part of the code coverage.

akarnokd (changes requested):
Fix headers and use org.junit.Test please.

akarnokd (approved):

PR #6028, opened 2018-05-29
2.x: Improve Observable.takeUntil

akarnokd:
This PR upgrades the `Observable.takeUntil` to a newer algorithm (the `Flowable` version is up-to-date).

Some unit test remnants from the v1 era were upgraded too as the new algorithm no longer disposes the source or other if they terminate on their own (the Reactive Streams specification doesn't allow that anyway).

PR #6029, opened 2018-05-30
2.x: Upgrade to Gradle 4.3.1, add TakeUntilPerf

akarnokd:
The JMH plugin 0.4.4 has some shortcomings that prevent the execution of unit tests on Windows 10 (and on some very restricted Linuxes). Version 0.4.5 has been fixed in this regard but it also requires Gradle 4.3.x. Unfortunately, there are no newer versions to the JMH plugin so this is likely as far as we can go with versions.

I've also added a new benchmark to measure the overhead in `takeUntil`. Here are the results:

i7 4790, Windows 10 x64, Java 8u172, JMH 1.20:

```
Benchmark                        Mode  Cnt         Score       Error  Units
TakeUntilPerf.flowable          thrpt    5    341933,112   13328,573  ops/s
TakeUntilPerf.flowable:items    thrpt    5  21128765,548  677544,832  ops/s
TakeUntilPerf.observable        thrpt    5    366931,108    8202,644  ops/s
TakeUntilPerf.observable:items  thrpt    5  10733841,643  112853,358  ops/s
```

The `observable` here uses an older algorithms & structure and is generally relaying half the items under the same time amonut than the `flowable` version. PR #6028  can then be evaluated with this benchmark.


vanniktech (approved):

src/jmh/java/io/reactivex/TakeUntilPerf.java
```diff
@@ -0,0 +1,95 @@
...
+        final CountDownLatch cdl = new CountDownLatch(1);
+
+        flowable.subscribe(this, Functions.emptyConsumer(), new Action() {
+            @Override
+            public void run() throws Exception {
+                cdl.countDown();
+            }
+        });
+
+        while (cdl.getCount() != 0) { }
```
artem-zinnatullin, line 79: @akarnokd I wonder if this loop affects benchmark in positive or negative way, it consumes single core to max

Should `cdl.await()` be used instead? it parks the thread

I see this pattern in few other benchmarks but idk if it's intentional

artem-zinnatullin:
```
TakeUntilPerf.flowable:items    thrpt    5  21128765,548  677544,832  ops/s
TakeUntilPerf.observable:items  thrpt    5  10733841,643  112853,358  ops/s
```

Holy moly

src/jmh/java/io/reactivex/TakeUntilPerf.java
artem-zinnatullin: // It shouldn't affect the comparison though since loop is present in both Flowable and Observable benchmarks

akarnokd: The problem `await` is that on my machines they can't wake up that fast and the throughput usually caps out around 200k ops/s, even if the underlying work finished much faster. 

For sorter sequences, typically in the range of 1000-10000 items, a spin-wait will result in a much higher throughput number as the end-of-work is detected more eagerly.

artem-zinnatullin: Ahhhh damn, I think you've mentioned that problem before

What if JMH would be event/callback based so we could benchmark async code like RxJava in a reactive manner?

ie:

```java
@Benchmark
public void flowable(BenchmarkObserver benchmarkObserver) {
    flowable.subscribe(this, Functions.emptyConsumer(), new Action() {
        @Override
        public void run() throws Exception {
            benchmarkObserver.onComplete();
        }
    });
}
```

PR #6047, opened 2018-06-17
2.x: Add the wiki pages as docs

akarnokd:
This PR adds the `RxJavaWiki.git` subgit markdown files under the `docs` directory, allowing external contribution through the same workflow as for the code.


vanniktech (commented):
Way too much text to check line by line. Were they exported automatically or you copy pasted them?

akarnokd:
I've checked out the RxJavaWiki.git and copied the files over to `RxJava/docs` from within the file system. No copy-pasting.

vanniktech (approved):

Issue #5980, opened 2018-04-29
2.x: subscribeActual() accidental public


akarnokd:
Looks like the following `protected void subscribeActual()` implementations have unintentionally become public:

- [`PublishSubject`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/subjects/PublishSubject.java#L134)
- [`PublishProcessor`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/processors/PublishProcessor.java#L95)

There is no reason to call `subscribeActual` from outside of RxJava, however, making these `protected` counts as a breaking change. However, it should be extremely unlikey somebody was actually using these methods directy.

The question is: do we take the risk and fix the visibility in 2.x or it has to wait for 3.x?

Note that many operators are also having this method as public but those are not really accessible from the outside.

vanniktech:
In this case I'd consider making the fix with version 2.2

sircelsius:
I don't know If this is still relevant, but I opened a PR for it, let me know if it helps

PR #6052, opened 2018-06-20
#5980 made subscribeActual protected

sircelsius:
### Proposal

Changed `subscribeActual` from `public` to `protected`, as described in #5980.

### How I tested

Ran tests locally, please let me know if I missed part of the contributing guidelines 👼 

Fixes: #5980.


akarnokd (approved):

akarnokd:
/cc @vanniktech, @artem-zinnatullin, @davidmoten 

Strictly speaking, this is technically a breaking change on something that shouldn't be in use at all by external developers. I'm fine with this change for 2.1.15.

vanniktech (approved):
I am as well.

davidmoten (approved):

Issue #5927, opened 2018-03-23
[2.1.11] NPE in SpscLinkedArrayQueue.lvElement


boris-petrov:
I've seen already a couple of those after upgrading. Unfortunately I cannot give you any reproduction as I have zero idea where it might come from but here are a couple of stacktraces:

```
java.lang.NullPointerException
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.lvElement(SpscLinkedArrayQueue.java:249)
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.newBufferPoll(SpscLinkedArrayQueue.java:154)
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.poll(SpscLinkedArrayQueue.java:144)
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.clear(SpscLinkedArrayQueue.java:184)
 at io.reactivex.internal.operators.mixed.ObservableConcatMapMaybe$ConcatMapMaybeMainObserver.dispose(ObservableConcatMapMaybe.java:152)
 at io.reactivex.internal.observers.BasicFuseableObserver.dispose(BasicFuseableObserver.java:152)
 at io.reactivex.internal.observers.BasicFuseableObserver.dispose(BasicFuseableObserver.java:152)
 at io.reactivex.internal.operators.observable.ObservableTake$TakeObserver.dispose(ObservableTake.java:90)
 at io.reactivex.internal.disposables.DisposableHelper.dispose(DisposableHelper.java:125)
 at io.reactivex.internal.observers.LambdaObserver.dispose(LambdaObserver.java:100)
 at java.util.HashMap$Values.forEach(HashMap.java:981)
```

```
java.lang.NullPointerException
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.lvElement(SpscLinkedArrayQueue.java:249)
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.newBufferPoll(SpscLinkedArrayQueue.java:154)
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.poll(SpscLinkedArrayQueue.java:144)
 at io.reactivex.internal.queue.SpscLinkedArrayQueue.clear(SpscLinkedArrayQueue.java:184)
 at io.reactivex.internal.operators.mixed.ObservableConcatMapMaybe$ConcatMapMaybeMainObserver.drain(ObservableConcatMapMaybe.java:200)
[...]

PR #6060, opened 2018-06-22
2.x: Fix concatMap{Single|Maybe} null emission on dispose race

### LINKED ISSUE #6059: [2.1.15] map method called with null
I have something like:

```java
observable.concatMapMaybe(...).map(x -> ...);
```

This `x` is `null` sometimes (our tests reproduce it consistently in one case). Didn't happen on `2.1.14`. I can't give more information for now, just to let you know there is a problem. If you can't figure it out soon, I'll try to create some reproduction.

akarnokd:
This PR fixes a bug in all 4 specialized `concatMap` implementation that allows `null` to be emitted when the success signal of the inner source races with the dispose signal of the sequence.

Likely fixes: #6059


vanniktech (approved):

akarnokd:
A very similar thing happened in #5927 just after a release but this bug was there between 2.1.11 and 2.1.14.

PR #6064, opened 2018-06-26
Add createScan operator to Observable

Goshik92:
Imagine a situation when your app has some state and wants it to be synchronized with the state of a server. The app sends its state to the server and gets back the difference between that state and the actual state of the server. Then, it applies the difference to its current state, sends the new state to the server and so on. This algorithm may be implemented using Observable.scan() operator. However, this operator seems to be redundant for the task because it acts on an upstream. In this example we will need to use a dummy Observable.range(...) stream as the upstream for Observable.scan(). The proposed operator provides a more elegant solution for the described problem and perhaps for some other problems.

Observable.scan() based solution:
```Java
Observable<State> streamOfAppStates =
        Observable.range(0, Long.MAX_VALUE - 1)
                .scan(initialState, (dummyLong, state) -> state.applyDiff(makeRequest(state)));
```

Observable.createScan() based solution:
```Java
Observable<State> streamOfAppStates =
        Observable.createScan(initialState, state -> state.applyDiff(makeRequest(state)));
```

createscan

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -1558,6 +1558,41 @@ public abstract class Observable<T> implements ObservableSource<T> {
         ObjectHelper.requireNonNull(source, "source is null");
         return RxJavaPlugins.onAssembly(new ObservableCreate<T>(source));
     }
+    
+    /**
+     * Returns an Observable that applies a specified {@code accumulator} function to the {@code seed} value
+     * then feeds the result of that function into the same function, and so on until the subscription is disposed,
+     * emitting the result of each of these iterations.
+     * <p>
+     * Note that the ObservableSource that results from this method will emit {@code seed} as its first
+     * emitted item.
+     * <p>
+     * Note that the {@code seed} is shared among all subscribers to the resulting ObservableSource
+     * and may cause problems if it is mutable. To make sure each subscriber gets its own value, defer
+     * the application of this operator via {@link #defer(Callable)}:
+     * <pre><code>
+     * Observable.defer(
+     *         () -&gt; Observable.createScan(new ArrayList&lt;&gt;(), list -&gt; list.add(0)));
+     * </code></pre>
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code scan} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param seed the initial value to be emitted and fed into the {@code accumulator}
+     * @param accumulator an accumulator function to be sequentially invoked with the {@code seed},
+     * with the result of that call, with that result, and so on
+     * @param <T> the type of the seed, the accumulated value, and the emitted items
+     * @return an Observable that emits {@code seed} followed by the results of each call to the
+     * accumulator function
+     * @see #scan(Object, BiFunction)
+     */
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public static <T> Observable<T> createScan(T seed, Function<T, T> accumulator) {
+        ObjectHelper.requireNonNull(seed, "seed is null");
+        ObjectHelper.requireNonNull(accumulator, "accumulator is null");
+        return RxJavaPlugins.onAssembly(new ObservableCreateScan<T>(seed, accumulator));
+    }
 
     /**
      * Returns an Observable that calls an ObservableSource factory to create an ObservableSource for each new Observer
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableCreateScan.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableCreateScan.java
@@ -1,0 +1,64 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


akarnokd:
This sounds like a very peculiar operator which can be implemented on top of existing operators as well as, being a static factory method, on any parent class. I don't see any reason this should be added to RxJava.

akarnokd:
For example:

```java
class MyClass {
    public static <T> Observable<T> createScan(T initial, Function<T, T> stateTransition) {
        return Observable.create(emitter -> {
            T state = initial;
            emitter.onNext(state);
            while (!emitter.isDisposed()) {
                state = stateTransition.apply(state);
                emitter.onNext(state);
            }
        });
    }
}
```

akarnokd:
or with `generate` that can support backpressure on `Flowable`s:

```java
class MyClass {
    public static <T> Flowable<T> createScan(T initial, Function<T, T> stateTransition) {
        return Flowable.generate(() -> initial, (state, emitter) -> {
            emitter.onNext(state);
            return stateTransition.apply(state);
        });
    }
}
```

Goshik92:
> I don't see any reason this should be added to RxJava

Well, I did not find any set of rules that allows to decide whether or not an operator should be a part of RxJava, so I'll just share my thoughts with you.

1. `createScan` may be useful for generating sequences defined by recurrence relations. For example, [this page](https://github.com/ReactiveX/RxJava/wiki/Problem-Solving-Examples-in-RxJava#generate-the-fibonacci-sequence) suggests using `scan` operator for generating the Fibonacci Sequence as a more functional way rather than using `create`. However, the article still considers it to be a janky way. `createScan` might be a more clear operator so solve the problem because it does not require confusing `Observable.from(0).repeat()` thing. When I faced the problem that I described in my commit, I knew that `scan` is a good fit. But I spent a great deal of time trying to understand what to do with the upstream. That is why I think `createScan` may be more intuitive in some cases.

2. As for your implementation of `createScan` with `generate`, it looks elegant and concise. But we can also do the opposite: implement `generate` with `createScan`. If we assume that we have `createScan` that honors backpressure, the code might look like this:
```Java
class StateHolder<S, T>{
    final S state;
    final MyEmitter<T> emitter;

    StateHolder(S state, MyEmitter<T> emitter) {
        this.state = state;
        this.emitter = emitter;
    }
}

class MyEmitter<T> implements Emitter<T> {

    Notification<T> emitterState;
    
    @Override
    public void onNext(T value) {
        emitterState = Notification.createOnNext(value);
    }

    @Override
        emitterState = Notification.createOnError(error);
    }

    @Override
    public void onComplete() {
        emitterState = Notification.createOnComplete();
    }
}

class MyClass{
    public static <S, T> Observable<T> generate(S initialState, BiFunction<S, Emitter<T>, S> generator) {
        return Observable
                .createScan(
                        new StateHolder<S, T>(initialState, new MyEmitter<T>()),
                        holder -> {
                            MyEmitter<T> e = new MyEmitter<T>();
                            return new StateHolder<>(generator.apply(holder.state, e), e);
                        })
                .skip(1)
                .map(holder -> holder.emitter.emitterState)
                .takeWhile(emitterState -> !emitterState.isOnComplete())
                .map(emitterState -> emitterState.getValue());
    }
}
```

It is a lot of code, but notice that the verbosity is mostly created by defining the data structures such as `MyEmitter` and `StateHolder`. What is interesting here is that this implementation is basically a chain of standard operators. This leads me to the conclusion that `createScan` is more fundamental than `generate`. In other words, `generate` = `createScan` + data structures + other operators. What is also important is that `accumulator` (or `stateTransition` in your terms) can be a pure function, whereas `generator` has to make side effects by calling `emitter.onNext(...)`.

To conclude, I think that `createScan` should be included in RxJava because it is more intuitive than `scan` for some problems and more fundamental than `generate`. Since both functions are the part of RxJava, why should we not include `createScan`?

akarnokd:
The existing `create` and `generate` operators I suggested are well tested, safe and efficient way for doing their intended behavior as well as enabling your suggested operator with far less convolution. 

In contrast, your suggested `generate` implementation relies on a bunch of other operators, lots of lambdas, indirection, allocation and wrapping and thus it is far less efficient and convoluted than the existing `generate` operator. Just look at how much shorter the `createScan` is implemented with `generate` than the other way around.

The rule for operators is that if it is source-like, such as `create`, `generate`, `from`, etc. it can live on any class on any library as using it is no more inconvenient than using something static from `Observable`.

To conclude, the proposed `createScan` can be considered, at best, a syntax convenience over a `create` or `generate` method for a peculiar value generation scheme but does not bring enough merit to be included in RxJava proper. If you have use for such convenience, I recommend having it in your own projects and/or releasing your own library having it.

Goshik92:
@akarnokd, thank you for your consideration. I understand that you do not want to clutter up RxJava API with some strange operators. But I still disagree that my operator is peculiar. I just found out that Java Stream API provides exactly the same operator in some of their classes. For example, in Stream class: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#iterate-T-java.util.function.UnaryOperator-

Search through GitHub shows that people use it a lot (please see the picture). I hope this can help to persuade you. Thank you again.
image

akarnokd:
That is what `generate` is for.

Goshik92:
Alright, I see. Thank you for the explanation.

Issue #5788, opened 2018-01-05
3.x: Single incorrect or missing images (tracking issue)


akarnokd:
- [x] `amb`: missing diagram
- [x] `ambArray`: missing diagram
- [x] `concat(arg)`: missing diagram
- [x] `concat(s1..s4)`: add 3rd and 4th source, indicate backpressure
- [x] `concatArray`: missing diagram
- [x] `concatArrayEager`: missing diagram
- [x] `concatEager`: missing diagram
- [x] `create`: missing diagram
- [x] `defer`: missing diagram
- [x] `error(Callable)`: missing diagram
- [x] `error(Throwable)`: indicate error in the box
- [x] `fromCallable`: missing diagram
- [x] `fromPublisher`: missing diagram
- [x] `fromObservable`: missing diagram
- [x] `merge`: missing diagram
- [x] `merge(SingleSource<SingleSource<T>>)`: indicate main element is a `SingleSource`
- [x] `merge(s1..s4)`: add 3rd and 4th source, indicate backpressure
- [x] `mergeDelayError`: Iterable/Publisher missing diagrams, S2,S3,S4 wrong diagrams
- [x] `never`: missing diagram
- [x] `timer`: missing diagram
- [x] `equals`: missing diagram
- [x] `unsafeCreate`: missing diagram
- [x] `using`: missing diagram
- [x] `wrap`: missing diagram
- [x] `ambWith`: missing diagram
- [x] `as`: missing diagram
- [x] `hide`: missing diagram
- [x] `compose`: missing diagram
- [x] `cache`: missing diagram
- [x] `cast`: missing diagram
- [x] `delay`: missing diagram
- [x] `delaySubscription`: missing diagram
- [x] `dematerialize`: missing diagram
- [x] `doAfterSuccess`: missing diagram
- [x] `doAfterTerminate`: indicate error case
- [x] `doFinally`: missing diagram
- [x] `doOnSubscibe`: missing diagram
- [x] `doOnSuccess`: missing diagram
- [x] `doOnEvent`: missing diagram
- [x] `doOnError`: missing diagram
- [x] `doOnDispose`: missing diagram
- [x] `filter`: one source item, indicate both true and false outcomes
- [x] `flatMapPublisher`: operator name
- [x] `blockingGet`: missing diagram
- [x] `lift`: missing diagram
- [x] `contains`: missing diagram
- [x] `mergeWith`: indicate backpressure
[...]

PR #6075, opened 2018-07-08
Add marble diagram to the Single.filter method

UMFsimke:
Hopefully this should tick off the `filter` operator in #5788

Please let me know if I need to change anything on the diagram itself. If not, please send me URL when you upload image within the project so I can change URL for the marble in the PR.

Although, I had a doubt and wanted to double check. Operator `filter` returns `Maybe` which essentially will call `onSuccess` or `onComplete`. By running jUnit tests, I could observe that this is a behavior but if I use `TestObserver` then both `assertValue()` and `assertComplete()` are invoked when there are elements that satisfy `filter` condition.

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -2371,7 +2371,7 @@ public abstract class Single<T> implements SingleSource<T> {
      * Filters the success item of the Single via a predicate function and emitting it if the predicate
      * returns true, completing otherwise.
      * <p>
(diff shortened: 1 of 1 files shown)


akarnokd:
`https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.filter.png`

`TestObserver` implements multiple consumer types and an `onSuccess` is translated into `onNext`+`onComplete` for an easier verification.

UMFsimke:
PR is updated. Thanks for explanation 👍

akarnokd (approved):

vanniktech (approved):

PR #6077, opened 2018-07-08
Add marble diagram for Single.hide operator

UMFsimke:
This should complete `hide` operator in #5788

Please let me know if I need to change anything on the diagram itself. If not, please send me URL when you upload image within the project so I can change URL for the marble in the PR.

Although I wasn't sure if I should represent somehow that `Disposable` is hidden as well. If so, can you please tell me how to represent that on marble?

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -1876,7 +1876,7 @@ public abstract class Single<T> implements SingleSource<T> {
      * Hides the identity of the current Single, including the Disposable that is sent
      * to the downstream via {@code onSubscribe()}.
      * <p>
-     * <img width="640" height="470" src="https://raw.githubusercontent.com/UMFsimke/RxJava/Images/Images/Hide.png" alt="">
+     * <img width="640" height="458" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.hide.png" alt="">
      * <dl>
      * <dt><b>Scheduler:</b></dt>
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -1875,6 +1875,8 @@
     /**
      * Hides the identity of the current Single, including the Disposable that is sent
      * to the downstream via {@code onSubscribe()}.
+     * <p>
+     * <img width="640" height="470" src="https://raw.githubusercontent.com/UMFsimke/RxJava/Images/Images/Hide.png" alt="">
```
akarnokd: There should be a different superhero(ine) per reactive type who is relatively recognizable from the stylized head and doesn't require complicated drawings.

UMFsimke: Is this one suitable?

https://github.com/UMFsimke/RxJava/blob/Images/Images/Hide2.png

akarnokd: Yes.

https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.hide.png

UMFsimke: Thanks, updated

akarnokd (approved):

UMFsimke:
I dont get it, why is CI failing. I've run tests locally and I dont get that error which CI is reporting.

akarnokd:
Sometimes CI is a bit noisy and some of our millisecond-level tests can fail. I've restarted the build.

PR #6079, opened 2018-07-10
2.x: Add Completable.takeUntil(Completable) operator

akarnokd:
This PR adds the missing dedicated `takeUntil` operator to `Completable`.

image

Previously, the same effect could be achieved via the `ambWith` but there are two benefits of a dedicated operator:
- easier to discover based on the `takeUntil` operator in other types
- more direct implementation unlike `ambWith` which is built upon an N-ary `amb` operator with additional overhead.

There was a feature request in #3708 some time ago but apparently the issue got closed off after `Single.takeUntil` was implemented.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -2032,6 +2032,32 @@ public abstract class Completable implements CompletableSource {
     }
 
     /**
+     * Terminates the downstream if this or the other {@code Completable}
+     * terminates (wins the termination race) while disposing the connection to the losing source.
+     * <p>
+     * <img width="640" height="468" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Completable.takeuntil.c.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code takeUntil} does not operate by default on a particular {@link Scheduler}.</dd>
+     *  <dt><b>Error handling:</b></dt>
+     *  <dd>If both this and the other sources signal an error, only one of the errors
+     *  is signaled to the downstream and the other error is signaled to the global
+     *  error handler via {@link RxJavaPlugins#onError(Throwable)}.</dd>
+     * </dl>
+     * @param other the other completable source to observe for the terminal signals
+     * @return the new Completable instance
+     * @since 2.1.17 - experimental
+     */
+    @CheckReturnValue
+    @Experimental
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Completable takeUntil(CompletableSource other) {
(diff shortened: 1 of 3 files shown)


vanniktech (approved):

src/main/java/io/reactivex/internal/operators/completable/CompletableTakeUntilCompletable.java
```diff
@@ -0,0 +1,145 @@
...
+            if (once.compareAndSet(false, true)) {
+                DisposableHelper.dispose(other);
+                downstream.onError(e);
+            } else {
+                RxJavaPlugins.onError(e);
+            }
+        }
+
+        void innerComplete() {
+            if (once.compareAndSet(false, true)) {
```
artem-zinnatullin, line 104: Emit error if called more than once?

artem-zinnatullin (approved):

src/main/java/io/reactivex/Completable.java
```diff
@@ -2031,6 +2031,32 @@ public final Completable subscribeOn(final Scheduler scheduler) {
...
+     *  error handler via {@link RxJavaPlugins#onError(Throwable)}.</dd>
+     * </dl>
+     * @param other the other completable source to observe for the terminal signals
+     * @return the new Completable instance
+     * @since 2.1.17 - experimental
+     */
+    @CheckReturnValue
+    @Experimental
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public final Completable takeUntil(CompletableSource other) {
```
davidmoten, line 2054: I guess the bird has already flown with this one but the operator `take` refers to `onNext` events and we are more or less using it for all event types (take till terminates or the other thing terminates). I'm not suggesting a change perhaps a naming review for 3.x.

davidmoten (approved):

src/main/java/io/reactivex/internal/operators/completable/CompletableTakeUntilCompletable.java
akarnokd: No. It could legitimately lose a race against `onComplete` and we don't do it for any other operator either.

src/main/java/io/reactivex/Completable.java
akarnokd: I find it more discoverable this way and besides, the main behavior of `takeUntil` is to stop the main source and complete the downstream, regardless of there could be items from the main source or not.

kirk-l:
Shouldn't there be a `Completable.takeUntil(Publisher)` operator, too? `Single` has overloads for both `Publisher` and `CompletableSource`.

akarnokd:
@sailorpsy PR welcome.

PR #6095, opened 2018-07-18
Add marbles for Single.timer, Single.defer and Single.toXXX operators

UMFsimke:
Here are marbles for `defer`, `timer(delay, unit)`, `timer(delay, unit, scheduler)`, `toFlowable`, `toMaybe`, `toFuture` operators in #5788 

Marble for `defer`:
defer

Marble for `timer(delay, unit)`:
timer

Marble for `timer(delay, unit, scheduler)`:
timerScheduler

Marble for `toFuture`, although for this one I believe we should add case when there is no event sent from Single as there will be timeout exception thrown:
toFuture

Marble for `toMaybe`:
toMaybe

Marble for `toFlowable`, although here I didnt figured out what will happen if we request more than 1 event from flowable:
toFlowable

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -436,6 +436,8 @@ public abstract class Single<T> implements SingleSource<T> {
     /**
      * Calls a {@link Callable} for each individual {@link SingleObserver} to return the actual {@link SingleSource} to
      * be subscribed to.
+     * <p>
+     * <img width="640" height="423" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Single.defer.png" alt="">
      * <dl>
      * <dt><b>Scheduler:</b></dt>
      * <dd>{@code defer} does not operate by default on a particular {@link Scheduler}.</dd>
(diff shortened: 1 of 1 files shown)


akarnokd:
defer: I would have drawn it with a curly brace listing the 3 options below each other. Please add the subscribe() indicators as well.

toFuture: I indicate Future<T> with a 6-sided shape and indicate get() call and a return. See Completable.fromFuture for an example.

toFlowable: simply indicate request(N). Any positive N has the same effect with this operator. However, an error is always pushed through even without a request; just remove the request(1) from that side of the diagram.

akarnokd:
@UMFsimke are you still around?

UMFsimke:
Yes. I have in plan to wrap it up today... sorry for delay I was on vacation

UMFsimke:
@akarnokd Are these ok now?

I'm a bit confused by what you exactly had in mind for `defer` and `toFuture`.

Please update me if they are fine as I do need to update image sizes in javadoc.
tofuture
toflowable
defer

akarnokd:
Yes, these are okay.

UMFsimke:
Thanks. Updated javadoc.

akarnokd (approved):

PR #6100, opened 2018-07-19
Makes blockingGet on Single have an optional non-blocking fast return path

jnlopar:

  - [ ] Please give a description about what and why you are contributing, even if it's trivial.

This allows blockingGet to quickly return known values instead of blocking and creating a Subscription. Implements first on SingleSubject and SingleJust. I imagine it will have a much greater effect on SingleSubject.

N/A

The existing unit tests seemed sufficient to verify current behavior. I did add some variations in the JMH benchmark tests to demonstrate usefulness of the change. The `Single.defer` behavior is essentially without the fast path, as it is wrapping the inner `Single` with one that does not have the fast path.

Local benchmark results show an order of magnitude improvement on `blockingGet` where this change applies.
```
Benchmark                        Mode  Cnt          Score         Error  Units
BlockingGetPerf.single          thrpt    5  325290806.912 ± 6329789.491  ops/s
BlockingGetPerf.singleDeferred  thrpt    5   29213041.202 ± 1734102.589  ops/s
BlockingGetPerf.singleSubject   thrpt    5  296488077.286 ± 9972222.042  ops/s
```

### DIFF
--- a/src/jmh/java/io/reactivex/BlockingGetPerf.java
+++ b/src/jmh/java/io/reactivex/BlockingGetPerf.java
@@ -13,6 +13,8 @@
 
 package io.reactivex;
 
+import io.reactivex.subjects.SingleSubject;
+import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
 import org.openjdk.jmh.annotations.*;
@@ -29,6 +31,8 @@ public class BlockingGetPerf {
     Observable<Integer> observable;
 
     Single<Integer> single;
+    Single<Integer> singleDeferred;
+    SingleSubject<Integer> singleSubject;
 
     Maybe<Integer> maybe;
 
@@ -41,6 +45,15 @@ public class BlockingGetPerf {
         observable = Observable.just(1);
 
         single = Single.just(1);
+
+        singleDeferred = Single.defer(new Callable<Single<Integer>>() {
+          public Single<Integer> call() {
+            return Single.just(1);
+          }
+        });
+
+        singleSubject = SingleSubject.create();
+        singleDeferred.subscribe(singleSubject);
 
         maybe = Maybe.just(1);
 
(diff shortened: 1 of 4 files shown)


akarnokd:
First of all, you don't have to retain the welcome text on a PR.

Second, this is a peculiar optimization that is rarely necessary, it is very rare one wants the single value of a `Single.just` back. Besides, `SingleSubject` already has the [`getValue`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/subjects/SingleSubject.java#L237) method for quickly accessing the terminal value if any.

Third, we do such optimizations via indicator interfaces like `ScalarCallable` and `Callable`, but they would have minimal value within `Single` (they are very valuable within `Flowable` and `Observable` though).

In conculsion, I don't think this change is worth it in general.

jnlopar:
Good to know on the PR welcome text, thanks.

I think you're right about `Single.just`, it's not likely to be used in this way. The `SingleSubject` is where this is more compelling for me. I was inspired to do it while attempting to clean up some code where the author was attempting to store the value as a local instance, providing a getter that would block if not present, but in general exposing a couple potential awkward race conditions in the implementation. Making `blockingGet` fast for the Subject means that can be used without incurring new subscriptions and also without potentially errant concurrency code at the transition points between reactive and imperative flows, where `Subject`s seem to be primarily used. I also just naively assumed that `blockingGet` would do a quick return on a subject where the final value was cached. I can understand it not having extremely wide appeal but I have also seen a few cases where it seems valid without adding any significant overhead to the other paths.

akarnokd:
`hasValue` and `getValue` can get you that without changing the library:

```java
public static <T> T getOrBlock(SingleSubject<T> source) {
    if (source.hasValue()) {
       return source.getValue();
    }
    return source.blockingGet();
}
```

This is also thread-safe unlike reading the value field of the `SingleSubject` without checking for termination.

PR #6113, opened 2018-07-29
Encoding Bug

FlXME:
Use InputStreamReader to specify encoding otherwise the default encoding will be used leading to strange bugs for some users.


akarnokd:
Doesn't compile. Have you actually tried your suggested solution first?

akarnokd:
Also did you run into problems with a particular default encoding? We have English text and practically ASCII/ISO-8859-1 text so the default reader should not run into problems.

PR #6114, opened 2018-07-29
Encoding Bug

FlXME:
Use InputStreamReader to specify encoding otherwise the default encoding will be used leading to strange bugs for some users.


akarnokd:
Doesn't compile. Have you actually tried your suggested solution first?

PR #6116, opened 2018-07-30
6108 changed README to use Gradle's implementation instead of compile

### LINKED ISSUE #6108: Gradle build error just as the error info from the part of README
Pls change the read me about the setup library into the project
The reason as below:
Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018，pls see http://d.android.com/r/tools/update-dependency-configurations.html,
So pls use implementation instead of compile in the correct position...

sircelsius:
fixes #6108 

Changed `README.md` to reflect Gradle's more recent version of the [library plugin](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation).

I believe that RxJava should be using `implementation`, but I'm not a Gradle expert, feel free to let me know if clients should be using `api`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The [1.x version](https://github.com/ReactiveX/RxJava/tree/1.x) is end-of-life a
 The first step is to include RxJava 2 into your project, for example, as a Gradle compile dependency:
 
 ```groovy
-compile "io.reactivex.rxjava2:rxjava:2.x.y"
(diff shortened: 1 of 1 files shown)


akarnokd (approved):

artem-zinnatullin (approved):

vanniktech (commented):
:+1:

PR #6121, opened 2018-08-01
2.x: Flowable.onErrorResumeNext improvements

akarnokd:
This PR improves the structure of the `onErrorResumeNext` operator by inlining the subscription-arbiter and not updating the outstanding request amount one by one. The change reduces the per-item overhead, some indirection cost and the memory cost (+8 bytes for the field, -16/-28 bytes due to the inlined header+field)

In addition, the null-check has been inlined into the resume function call so that the crash will now include the original exception as well as the NPE in a `CompositeException`, just like almost all other cases throughout RxJava.


vanniktech (commented):
:+1:

davidmoten (commented):
:+1

PR #6123, opened 2018-08-01
2.x: Improve Completable.onErrorResumeNext internals

akarnokd:
This PR improves the internals of the `Completable.onErrorResumeNext` operator:

- Inline the `Disposable` management.
- Reuse the same instance for observing the fallback `CompletableSource`.
- Report null return of the `errorMapper` the same way as a crash via `CompositeException`.
- The exceptions in the composite should be in the order **original** -> **mapper exception**.


artem-zinnatullin (approved):

vanniktech (commented):
:+1:

davidmoten (commented):
:+1:

PR #6129, opened 2018-08-02
2.x: Cleaunp - rename fields to upstream and downstream

akarnokd:
This PR cleans up the field namings and some local variable namings:

- Use `upstream` for the connection `Disposable` or `Subscription` instead of `d`, `s`, etc., including `AtomicReference<*>` declarations.
- Use `downstream` for the consumer field name in operators instead of `actual` for example.
- Fix one-two letter abbreviated naming to match the type, such as `Disposable s` -> `Disposable d`, etc.

The `CheckLocalVariablesInTests` has been extended with the relevant regexp checks.


src/main/java/io/reactivex/MaybeEmitter.java
```diff
@@ -68,14 +68,14 @@
 
     /**
      * Sets a Disposable on this emitter; any previous Disposable
-     * or Cancellation will be unsubscribed/cancelled.
-     * @param s the disposable, null is allowed
+     * or Cancellable will be disposed/cancelled.
```
artem-zinnatullin: `{@link`?

src/test/java/io/reactivex/validators/CheckLocalVariablesInTests.java
```diff
@@ -26,11 +26,21 @@
  * <li>{@code TestObserver} named as {@code ts*}</li>
  * <li>{@code PublishProcessor} named as {@code ps*}</li>
  * <li>{@code PublishSubject} named as {@code pp*}</li>
+ * <li>{@code Subscription} with single letter name such as "s" or "d"</li>
```
artem-zinnatullin, line 29: 👍

artem-zinnatullin (approved):
Very welcomed change, thanks

akarnokd:
Thanks @artem-zinnatullin for your diligent review of this PR.

vanniktech (commented):
Very nice :+1:

davidmoten:
:+1: very helpful thanks

src/main/java/io/reactivex/internal/observers/BasicFuseableObserver.java
```diff
@@ -28,13 +28,13 @@
 public abstract class BasicFuseableObserver<T, R> implements Observer<T>, QueueDisposable<R> {
 
     /** The downstream subscriber. */
-    protected final Observer<? super R> actual;
```
codefromthecrypt, line 31: fyi these are protected fields so will break folks. We'll try and work around this, just saying. cc @kojilin

akarnokd: Internal.

https://github.com/ReactiveX/RxJava#ioreactivexinternal

codefromthecrypt:
>
> In
> src/main/java/io/reactivex/internal/observers/BasicFuseableObserver.java
> <https://github.com/ReactiveX/RxJava/pull/6129#discussion_r225166066>:
>
> Internal.
>
> https://github.com/ReactiveX/RxJava#ioreactivexinternal
>
YIKES we have a lot of use of internal types, just didn't notice until you
mentioned. Will go through and rewrite a bunch of things. Thanks for the
pointer.

Issue #6132, opened 2018-08-03
Update wiki to reflect 4.x (tracking issue)


akarnokd:
This is the overview of the suggested/planned changes to the Wiki. *Text in bold are for extra considerations/options*.

- [ ] [`Additional-Reading.md`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Additional-Reading.md): review links, add newer links to blogs and topics
- [ ] [`Alphabetical-List-of-Observable-Operators.md`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Alphabetical-List-of-Observable-Operators.md): remove non-standard, deleted & otherwise project-external operators, add newer operators, link to the categorized other `.md` files and sections in the wiki.
- [ ] *`Alphabetical-List-of-3rd-party-operators.md`: create page and list operators that used to be in RxJava and/or interesting enough, but available in 3rd party libraries, such as rxjava2-extras.*
- [ ] [`Async-Operators`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Async-Operators.md): review the operator set, point to RxJava2Extensions project, *add fancy descriptions like in #6131*.
- [ ] [`Backpressure`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Backpressure.md): update wording, remove unavailable operators, fix image links, make sure examples are in **Java**.
- [ ] *[`Backpressure-(2.0).md`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Backpressure-(2.0).md): maybe review, should be okay as it has been recently added*
- [ ] [`Blocking-Observable-Operators.md`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Blocking-Observable-Operators.md): update to `blockingX` methods, *add fancy descriptions like in #6131*, existing operators only, fix image link
- [ ] [`Combining-Observables.md`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Combining-Observables.md): expand operator list (concat/eager/withLatestFrom missing?), *fancy like #6131*, link to RxJava2Extensions or explain join patterns here.
- [ ] [`Conditional-and-Boolean-Operators.md`](https://github.com/ReactiveX/RxJava/blob/3.x/docs/Conditional-and-Boolean-Operators.md): review operators, link/explain extensions ops, *fancy like #6131*
[...]

PR #6131, opened 2018-08-03
2.x: Expand Creating-Observables.md wiki

akarnokd:
This PR expands and updates the `Creating-Observables.md` wiki page to 2.x terminology and feature set. 

This also serves as a discussion basis about how to detail the various operators. The PR suggests the following structure:

- Top outline with links to the family of operators
- Each operator lists:
  - an availability list for the various reactive types: image available, image not available
  - ReactiveX documentation link if available
  - Short description
  - Example, named as `operator example` so that the heading generated can be linked to directly (i.e., `Creating-Observables.md#justexample`.

Related tracking issue: #6132

PR #6136, opened 2018-08-04
2.x: Update Getting started docs

ashishkrishnan:
This patch the following changes to the Getting started docs as per Issue No: #6132 
- Remove Troubleshooting guide
- Bump RxJava versions to 2.2.0 for Gradle, Maven etc.

Questions: 
- ~Looks like JFrog for RxJava 2 is broken or invalid now. Can we add Jitpack?~
- Need suggestions for `Getting started from readme.md`. Although, `How to use RxJava` or `Alphabetical List of Operators`, `Additional-Reading.md` can be great next steps.

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -1,20 +1,20 @@
 ## Getting Binaries
 
-You can find binaries and dependency information for Maven, Ivy, Gradle, SBT, and others at [http://search.maven.org](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22io.reactivex%22%20AND%20a%3A%22rxjava%22).
+You can find binaries and dependency information for Maven, Ivy, Gradle, SBT, and others at [http://search.maven.org](http://search.maven.org/#search%7Cga%7C1%7Cg%3A"io.reactivex.rxjava2"%20AND%20"rxjava2").
 
 Example for Maven:
 
 ```xml
 <dependency>
-    <groupId>io.reactivex</groupId>
+    <groupId>io.reactivex.rxjava2</groupId>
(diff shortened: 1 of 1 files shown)


akarnokd:
JFrog: The main readme contains the proper links: https://github.com/ReactiveX/RxJava#binaries

akarnokd:
> Getting started from readme.md

Just copy in the from [this](https://github.com/ReactiveX/RxJava#hello-world) inclusive to [this](https://github.com/ReactiveX/RxJava#further-reading) exclusive.

> How to use RxJava

Perhaps change the non-java examples into Java after all. However, since the text should also be reviewed for 2.x terminology, I don't recommend taking this one.

> Alphabetical List of Operators

There links there should point to the various category-wiki pages, such as https://github.com/ReactiveX/RxJava/blob/2.x/docs/Creating-Observables.md#just most of such pages hasn't been updated yet. I'd leave this last.

> Additional-Reading.md

Beyond verifying the existing links, one should vet the various blog posts. Some of the existing links should be also checked for how much they are version agnostic. I'd leave this to somebody with a longer history with RxJava.

--------

Perhaps the easiest next step is to work on one of the operator-category listings such as https://github.com/ReactiveX/RxJava/blob/2.x/docs/Filtering-Observables.md - review, remove nonexistent operators, fancy like #6131

akarnokd (approved):

ashishkrishnan:
@akarnokd Thanks for the next steps. Will pick them. Anything more to be done on this PR?

akarnokd:
We can always expand it later, the important changes are already okay with it.

ashishkrishnan:
Yes. That's true. Looking forward to it.

PR #6148, opened 2018-08-08
feat: add support for just operator

BirjuVachhani:
just operator was overloaded to support maximum 10 items

replaced all the overloaded methods with a method which supports varargs so that just operator can take any number of value supported by java varargs, not just 10.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) 2016-present, RxJava Contributors.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
  * compliance with the License. You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software distributed under the License is
  * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
  * the License for the specific language governing permissions and limitations under the License.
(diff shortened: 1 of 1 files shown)


JakeWharton:
This API already exists:

https://github.com/ReactiveX/RxJava/blob/fd63c492c576a2784a5665925457b75355ecffaa/src/main/java/io/reactivex/Observable.java#L1738

Also you reformatted the entire file making it unreviewable.

akarnokd:
Why, why, why? 

> replaced all the overloaded methods

You practically broke binary compatibility with this which we do not allow. Besides, there is a reason for the 1..10 argument variants' existence, namely to avoid the unchecked conversion warnings due to Java 6 limitations in typical usage cases.

BirjuVachhani:
As far as I know, Binary compatibility will break if the method is not found but I think binary compatibility will not break because the replaced method has the same name as the overloaded methods and it takes varargs as arguments so it can replace all the overloaded method without breaking binary compatibility. 

And I didn't get your second point about unchecked conversion warnings. it would be great if you can explain it.

And why overload it to maximum of 10?

akarnokd:
> binary compatibility will not break because the replaced method has the same name

Wrong. Compatibility requires the same number and types of arguments to remain the same. A varargs argument counts as one argument.

> And I didn't get your second point about unchecked conversion warnings. 

Whenever there is a `method(T... arg)` in Java 6, you'll most likely get a compiler warnings about unchecked conversion. Suggested reading: https://stackoverflow.com/questions/21132692/java-unchecked-unchecked-generic-array-creation-for-varargs-parameter

> And why overload it to maximum of 10?

We had to stop at some number to not bloat the API. You can define static methods on some class of yours which can define more than 10 arguments if you need.

BirjuVachhani:
So if new code is added to the file then the binary compatibility will break or not?

akarnokd:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

BirjuVachhani:
Thanks for the information. I really appreciate it and it helped to learn new things as well. 👍

Issue #6151, opened 2018-08-09
`BaseTestConsumer#assertValueSet` doesn't verify that all listed values were received


Stephan202:
[`BaseTestConsumer#assertValueSet`](https://github.com/ReactiveX/RxJava/blob/e8930c2830869f1089ac7627dda044e8d861fb6b/src/main/java/io/reactivex/observers/BaseTestConsumer.java#L567) is documented and implemented as:

```java
/**
 * Assert that the TestObserver/TestSubscriber received only the specified values in any order.
 * <p>This helps asserting when the order of the values is not guaranteed, i.e., when merging
 * asynchronous streams.
 *
 * @param expected the collection of values expected in any order
 * @return this
 */
@SuppressWarnings("unchecked")
public final U assertValueSet(Collection<? extends T> expected) {
    if (expected.isEmpty()) {
        assertNoValues();
        return (U)this;
    }
    for (T v : this.values) {
        if (!expected.contains(v)) {
            throw fail("Value not in the expected collection: " + valueAndClass(v));
        }
    }
    return (U)this;
}
```

Within our team we have several tests using `assertValueSet`, each attempting to assert that _all_ of the listed values are emitted. Today we found out that `assertValueSet` verifies a weaker condition, namely that no "unexpected" values are emitted. It does not verify that all listed values are emitted. This took us by surprise. If this is the intended behaviour, perhaps the documentation could be clarified on this point.

NB: the RxJava code base itself contains several tests using `assertValueSet`. Some of these also seem to assume the stronger interpretation (e.g. `FlowableFlatMapSingleTest#normalAsync`), while others would surely fail under the stronger semantics (e.g.`FlowableFlatMapSingleTest#takeAsync`).

NB2: I do observe that (nearly?) all tests in the RxJava code base invoking `assertValueSet` involve some form of nondeterminism, and a benefit of the current weaker semantics is that these tests are now less likely to fail on a system under heavy load. Still, as-is `assertValueSet` also passes if _no_ values are emitted; surely these tests wish to assert something more definitive?

akarnokd:
It is meant to assert received values are all in the collection. You can make a stronger assumption by adding assertValueCount.

PR #6152, opened 2018-08-10
2.x: Clarify TestObserver.assertValueSet in docs and via tests

akarnokd:
Clarify that `{TestObserver|TestSubscriber}.assertValueSet` checks for all received items to be in the provided set, but that set can contain additional elements that could have been received. Unit tests have been added to demonstrate the expected behavior.

Resolves: #6151.

### DIFF
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -566,7 +566,7 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
 
     /**
      * Assert that the TestObserver/TestSubscriber received only items that are in the specified
-     * collection as well irrespective of the order they were received.
+     * collection as well, irrespective of the order they were received.
      * <p>
      * This helps asserting when the order of the values is not guaranteed, i.e., when merging
      * asynchronous streams.
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -565,9 +565,14 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
     }
 
     /**
-     * Assert that the TestObserver/TestSubscriber received only the specified values in any order.
-     * <p>This helps asserting when the order of the values is not guaranteed, i.e., when merging
+     * Assert that the TestObserver/TestSubscriber received only items that are in the specified
+     * collection as well irrespective of the order they were received.
+     * <p>
+     * This helps asserting when the order of the values is not guaranteed, i.e., when merging
      * asynchronous streams.
+     * <p>
+     * To ensure that only the expected items have been received, no more and no less, in any order,
+     * apply {@link #assertValueCount(int)} with {@code expected.size()}.
      *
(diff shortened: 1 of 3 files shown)


src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -565,9 +565,14 @@ public final U assertValuesOnly(T... values) {
...
-     * Assert that the TestObserver/TestSubscriber received only the specified values in any order.
-     * <p>This helps asserting when the order of the values is not guaranteed, i.e., when merging
+     * Assert that the TestObserver/TestSubscriber received only items that are in the specified
+     * collection as well irrespective of the order they were received.
+     * <p>
+     * This helps asserting when the order of the values is not guaranteed, i.e., when merging
      * asynchronous streams.
+     * <p>
+     * To ensure that only the expected items have been received, no more and no less, in any order,
+     * apply {@link #assertValueCount(int)} with {@code expected.size()}.
```
artem-zinnatullin, line 575: I guess ideally method name would indicate that by using a word like `contains`, maybe add it to 3.x list of changes?

artem-zinnatullin (approved):

src/main/java/io/reactivex/observers/BaseTestConsumer.java
```diff
@@ -565,9 +565,14 @@ public final U assertValuesOnly(T... values) {
     }
 
     /**
-     * Assert that the TestObserver/TestSubscriber received only the specified values in any order.
-     * <p>This helps asserting when the order of the values is not guaranteed, i.e., when merging
+     * Assert that the TestObserver/TestSubscriber received only items that are in the specified
+     * collection as well irrespective of the order they were received.
```
Stephan202: > collection as well irrespective of the order they were received.

->

> collection as well, irrespective of the order in which they were received.

src/main/java/io/reactivex/observers/BaseTestConsumer.java
Stephan202, line 575: As-is this method "ignores" duplicates, while `assertValueCount` doesn't. If a user wishes to assert that all values were received while allowing for duplicates, they could to `.distinct().test()`. Is this the preferred approach, and if so, is it worth documenting that too? (Maybe have an example of that in the test below?)

Stephan202 (commented):
One grammar nit and one question :)

Thanks for the rapid response!

src/main/java/io/reactivex/observers/BaseTestConsumer.java
akarnokd: The aim is to provide some typical set of assertions with `TestObserver`. You can always modulate the input flow or simply enumerate contents of the `TestObserver` manually. I usually try to minimize the cases where additional operators are mentioned in the javadocs because it triggers "what about when using X" type of comments.

sircelsius: `assertValuesIn(Collection<? extends T> expected)`?

Stephan202 (approved):

src/main/java/io/reactivex/observers/BaseTestConsumer.java
akarnokd: Posted #6153.

Issue #5973, opened 2018-04-24
2.x: Maybe.flatMapXXX overload with onSuccessMapper, onErrorMapper, onCompleteSupplier


GeorgePetri:
Hello

When working with maybes i find this overload useful.
```java
<R> Maybe<R> flatMap(
            Function<? super T, ? extends MaybeSource<? extends R>> onSuccessMapper,
            Function<? super Throwable, ? extends MaybeSource<? extends R>> onErrorMapper,
            Callable<? extends MaybeSource<? extends R>> onCompleteSupplier)
```
Having the same overload for flatmapCompletable, flatmapSingle and maybe the others would make working with maybes easier. At the moment it is necessary to add toMaybe() and convert various stream.

Thanks!

akarnokd:
I'm not convinced. You could use Kotlin and extension methods or have a respective transformer that does the conversion inside the lambdas and use `compose`.

GeorgePetri:
Yup, kotlin extensions do help.

And  rxjava is wonderful for modeling stuff as streams, but it gets harder when there is business logic of the form "If this stream emits do X if it does not do Y". For those situations I think those operators would help.

akarnokd:
PR welcome, also consider it for `Single` and `Completable` for feature parity.

sorianog:
Would love to review the PR or help out with it 😄

akarnokd:
FYI, I added most of the operators to [RxJava2Extensions](https://github.com/akarnokd/RxJava2Extensions#flatmap-signal)

PR #6154, opened 2018-08-10
2.x: Add Maybe.flatMap that maps signal types to Singles

akarnokd:
This PR proposes a new `flatMap` overload for `Maybe` that maps the upstream event types, `onSuccess`, `onError` or `onComplete` into a `SingleSource` and relays its signal to the downstream.

image

There is already a variant for Maybe -> Maybe flatMapping of event types.

I'm not 100% convinced the need for such an operator and could lead to all sorts of cross-mapping operators `Maybe -> Completable`, `Maybe -> Observable`, `Maybe -> Flowable`, as well as operators on the other reactive classes.

The alternative is to convert either the upstream and/or convert the mapped inner source to the right type.

Requested in: #5973

### DIFF
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -3097,6 +3097,41 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     }
 
     /**
+     * Maps the {@code onSuccess}, {@code onError} or {@code onComplete} signals of this {@code Maybe} into a {@link SingleSource} and emits that
+     * {@code SingleSource}'s signals.
+     * <p>
+     * <img width="640" height="354" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Maybe.flatMap.sss.png" alt="">
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code flatMapSingle} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     *
+     * @param <R>
+     *            the result type
+     * @param onSuccessMapper
+     *            a function that returns a SingleSource to merge for the onSuccess item emitted by this Maybe
+     * @param onErrorMapper
+     *            a function that returns a SingleSource to merge for an onError notification from this Maybe
+     * @param onCompleteSupplier
+     *            a function that returns a SingleSource to merge for an onComplete notification this Maybe
+     * @return the new Single instance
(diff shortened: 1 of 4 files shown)


vanniktech (approved):

artem-zinnatullin:
Let's hold on a bit on this one, it doesn't seem to match current RxJava flatMap API across other types.

Couldn't this use case be solved with `materialize()` operator? 
(we'll need to add it to the Maybe tho):

```kotlin
Maybe
  .create { … }
  .materialize()
  .flatMapSingle { notification -> 
    when {
      notification.isOnSuccess() -> Single.just(1)
      notification.isOnError() -> Single.just(2)
      notification.isOnComplete() -> Single.just(3)
    }
  }
```

akarnokd:
We don't have `Maybe.materialize` and even if, `Notification` requires allocation for an item or error outcome.

artem-zinnatullin:
I know that we don't have `materialize()` for `Maybe` and stated that in original comment above. Neither did we merge this operator yet, right :)

`Maybe` doesn't emit more than one notification, this allocation unlikely will be noticeable in practice. 

I'm worried about API consistency, that's it.

JakeWharton:
I, too, find this a very weird inclusion. Doing `.map { 1 }.toSingle(3).onErrorReturnItem(2)` already covers this. The special-casing of error and completion handling here is contrary to how the rest of the API is structured.

akarnokd:
Generally, such signal mapping should help the case where one wants to map the original error if any and not any error from mapping a success item to something:

```java
source.flatMap(v -> someOtherSource).onErrorResumeNext(e -> recoverSource);
```

here, `onErrorResumeNext` would act on `source` or `someOtherSource` too, which is really difficult to avoid if one doesn't want that.

The reverse order can be also unwanted:

```java
source.onErrorResumeNext(e -> recoverSource).flatMap(v -> someOtherSource);
```

now `flatMap` may act on the success of `source` or the `recoverSource`.

JakeWharton:
Sure, so you put the operator on the inner source. This is an extremely common thing to prevent errors from the inner stream from stopping the outer stream.

akarnokd:
I think I'll implement the full cross-conversion (5 x 4) set of operators in my extensions project instead so the standard API doesn't get bloated.

artem-zinnatullin:
Thanks, David

akarnokd:
FYI, I added most of the operators to [RxJava2Extensions](https://github.com/akarnokd/RxJava2Extensions#flatmap-signal).

PR #6171, opened 2018-08-24
2.x: Add explanation text to Undeliverable & OnErrorNotImplemented exs

akarnokd:
This PR adds detailed error message to the `UndeliverableException` and `OnErrorNotImplementedException` as they are the most common exception-related questions around.

#### UndeliverableException

> The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | `<original exception message>`

#### OnErrorNotImplementedException

> The onError handler in the subscribe() method is not provided. Please specify it to avoid this type of exception. Further reading: https://github.com/ReactiveX/RxJava/wiki/Error-Handling | `<original exception message>`

### DIFF
--- a/src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
+++ b/src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
@@ -48,6 +48,6 @@ public final class OnErrorNotImplementedException extends RuntimeException {
      *          the {@code Throwable} to signal; if null, a NullPointerException is constructed
      */
     public OnErrorNotImplementedException(@NonNull Throwable e) {
-        super(e != null ? e.getMessage() : null, e != null ? e : new NullPointerException());
+        this("The onError handler in the subscribe() method is not provided. Please specify it to avoid this type of exception. Further reading: https://github.com/ReactiveX/RxJava/blob/2.x/docs/Error-Handling.md | " + (e != null ? e.getMessage() : ""), e);
     }
 }
--- a/src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
+++ b/src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
@@ -48,6 +48,6 @@ public final class OnErrorNotImplementedException extends RuntimeException {
      *          the {@code Throwable} to signal; if null, a NullPointerException is constructed
      */
     public OnErrorNotImplementedException(@NonNull Throwable e) {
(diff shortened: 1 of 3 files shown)


vanniktech (approved):

src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
```diff
@@ -48,6 +48,6 @@ public OnErrorNotImplementedException(String message, @NonNull Throwable e) {
      *          the {@code Throwable} to signal; if null, a NullPointerException is constructed
      */
     public OnErrorNotImplementedException(@NonNull Throwable e) {
-        super(e != null ? e.getMessage() : null, e != null ? e : new NullPointerException());
+        this("The onError handler in the subscribe() method is not provided. Please specify it to avoid this type of exception. Further reading: https://github.com/ReactiveX/RxJava/blob/2.x/docs/Error-Handling.md | " + (e != null ? e.getMessage() : ""), e);
```
artem-zinnatullin: >Please specify it to avoid this type of exception.

I'm afraid this advice is too similar to generic error handling like this

```java
try {
  method();
} catch (Throwable t) {
  t.printStackTrace();
}
```

What about something like:

>An error occurred in the flow but no error handler was provided. Further reading: https://github.com/ReactiveX/RxJava/blob/2.x/docs/Error-Handling.md

artem-zinnatullin (approved):

src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
akarnokd: > "no error handler was provided" 

I'm worried users will think having or adding `doOnError` will solve the error, hence the explicit mention of `subscribe()`.

artem-zinnatullin: Oh, I see

Well, I'm not against mentioning `subscribe()` as long as we don't just say to "set this error handler to avoid this exception"

We can also mention `onErrorXXX`, might be a bit overwhelming tho

akarnokd: How about this:

`The exception was not handled due to missing onError handler in the subscribe() method call.`

isfaaghyth (approved):

src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
artem-zinnatullin: 👍

artem-zinnatullin (approved):

Issue #6178, opened 2018-08-27
Error-Handling on Dispose / extend use of tryOnError in Emitter?


brainbytes42:
Hi there,

thanks for providing reactiveX for java!

I ran into an issue with RxJava 2.2.0 when I tried to run a Callable to get an Observable/Single and disposed that. 

The issue was "kinda" already discussed here: #4880 / #4991 - the new tryOnError-Method is great and in my eyes really makes the whole thing more usable than having some Exception on some global handler "out of nowhere". 

Before I come to the core issue, a short remark: I do understand that this is some kind of intended by design - but I think, maybe there could be tweaks to not get bitten by that design... ;-) It does make sense, that a disposed subscriber doesn't get an Exception, but on first glance, it's not intuitive, if that Exception happens right when I'm disposing something...

What I'm doing was the following:
I have a Task-Class implementing Callable, which represents a long(er) running computation. I was quite happy about Observable.fromCallable() (or Single.fromCallable respectively).
But it blew up, when I disposed the created Observable, as the Task uses an Exception if it is canceled. 
(In my opinion, this is totally fine, as it's similar to Java's InterruptedException and Callable allows for checkedExceptions.)

Here's the code showing the relevant part. The task checks on each iteration for Thread.interrupted(), and if it is, it throws an Exception. The manual disposing is for testing only, a real usecase would be for example a switchMap, to restart some processing if new data arrived before the previous calculation was finished.

    Disposable disposable = Single.fromCallable(task)
                                  .subscribeOn(Schedulers.computation())
                                  .subscribe(s -> System.out.println("Observed: " + s),
                                                throwable -> System.out.println("Observed Exception: " + throwable));

    TimeUnit.MILLISECONDS.sleep(50);
    disposable.dispose();

But this leads to an UndeliverableException thrown, which is more than unintended. :-(

Is there any way around that? (Or at least any way to catch that?) I didn't find any, and the referenced issues above didn't show any, too - with one exception:

Using the new tryOnError:

    Disposable disposable = Single.create(emitter -> {
        // emitter.setCancellable(() -> task.requestCancel = true); // edit: not necessary, forgot to remove
        try {
            emitter.onSuccess(task.call());
        } catch (Throwable t) {
            emitter.tryOnError(t);
        }
    }).subscribeOn(Schedulers.computation())
      .subscribe(s -> System.out.println("Observed: " + s),
                 throwable -> System.out.println("Observed Exception: " + throwable));

[...]

Issue #6179, opened 2018-08-27
Add an "Error handling" JavaDocs section to fromCallable & co


akarnokd:
The operators of `fromCallable`, `fromAction` and `fromRunnable` in the reactive classes could use an `<dt><b>Error handling:</b></dt>` section in the JavaDocs describing the regular and canceled/disposed behavior.

> If the `{@link Callable/Action/Runnable}` throws an exception/exception/unchecked exception, the respective `{@link Throwable}` is delivered to the downstream via `{@link Subscriber/Observer/.../#onError()}`, except when the downstream has canceled/disposed this `{@code Flowable/Observable/Single/Maybe/Completable}` source. In this latter case, the Throwable is delivered to the global error handler via `{@link RxJavaPlugins#onError(Throwable)}` as `{@link io.reactivex.exceptions.UndeliverableException UndeliverableException}`.

PR #6183, opened 2018-08-28
Documentation and custom ErrorHandler for Single.fromCallable(.)

brainbytes42:
Relating to #6178, a proposal for a solution containing a custom ErrorHandler.

- Added doc for Single.fromCallable(.) to clarify on handling of
UndeliverableExceptions.
- Proposing custom ErrorHandler for Single.fromCallable to have a more
granular Interface than just to have some global RxJavaPlugins class,
which handles UndeliverableExceptions. They shouldn't be swallowed, if
they aren't expected, but for expected ones, it should be possible to
handle them individually.

If approved, this should be extended to Observable, etc.

  - [ ✔ ] Please give a description about what and why you are contributing, even if it's trivial.

  - [ ✔ ] Please include the issue list number(s) or other PR numbers in the description if you are contributing in response to those.

  - [ ✔ ] Please include a reasonable set of unit tests if you contribute new code or change an existing one. If you contribute an operator, (if applicable) please make sure you have tests for working with an `empty`, `just`, `range` of values as well as an `error` source, with and/or without backpressure and see if unsubscription/cancellation propagates correctly.

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -583,6 +583,11 @@ public abstract class Single<T> implements SingleSource<T> {
      * <dl>
      *   <dt><b>Scheduler:</b></dt>
      *   <dd>{@code fromCallable} does not operate by default on a particular {@link Scheduler}.</dd>
+     *   <dt><b>Error handling:</b></dt>
+     *   <dd>If the {@code Single} was already disposed, Exceptions thrown by the {@link Callable} are forwarded as
+     *   {@link io.reactivex.exceptions.UndeliverableException} to {@link RxJavaPlugins}. This could occur for example, if the callable reacts on
+     *   {@link Thread#interrupted()} by throwing an {@link InterruptedException} to cancel it's computation, triggered by the disposing.
+     *   (Think of {@link Observable#switchMapSingle(Function)} for example, which disposes previous Singles if new items appear.)</dd>
      * </dl>
      *
      * @param callable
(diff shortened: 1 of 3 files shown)


akarnokd:
Updating the docs is fine, provided you go along the line proposed by #6179.

However, I don't support adding a new overload and changing the existing operator implementation.

akarnokd (changes requested):
Please just update the docs only.

brainbytes42:
ok, never mind - would just have been useful... will have to continue using "workarounds" then...

akarnokd:
You could implement it yourself the way you see fit, even taking RXJava's implementation by source and having it in your project/library.

brainbytes42:
yeah - and rebasing each and every new "original" version? no thanks... ;-)
(which means, it's good that this project is actively maintained - I just don't get, what's bad on _not_ having to fumble with some global ErrorHandling or otherwise not being able to use fromCallable with callables throwing Exceptions as their signature allows... The other way around using create works, but why do we need fromCallable then? In my opinion, thats not consistent then.)

akarnokd:
> yeah - and rebasing each and every new "original" version? no thanks... ;-)

Assuming this PR would have been accepted, doesn't it mean you'd have to go around and add all those handlers anyway?

>  The other way around using create works, but why do we need fromCallable then? In my opinion, thats not consistent then.)

Just because you don't see the value of some operator, it doesn't mean its superfluous. Unfortunately, many would see RxJava tailored to their exact local needs, which is generally unsustainable.

brainbytes42:
yup, but that would be once.

For my local needs, I built my own extensions - I was asking for tailoring to JAva-Specification, wich has Callable throwing an Exception. (Not allowing an Exception would mean forcing a return-value, even if cancelled...) But it's ok, I built another utility-class to fix that issue on my own.

PR #6187, opened 2018-08-30
2.x: Fix refCount termination-reconnect race

### LINKED ISSUE #6185: Possible concurrency issue with .refCount()? (2.2.0)
First, I want to thank you for this wonderful library which I gladly use in many professional applications!

While I was writing a component for one of our company applications, I wrote a multithreaded stress test, because the component itself introduces shared state. Sometimes the test failed. Of course, I suspected my component to be the culprit. After further investigation I could completely rule out my own component and still reproduce the problem. 
I think there could be a problem with the `refCount()` operator.

Less words, more code:

The problem is reproducible with the following unit test (well, at least as reproducible as a test in a multithreaded manner can be...)

```
public void replayRefCountShallBeThreadSafe() {
        for (int i = 0; i < 10000; i++) {
            Observable<Object> observable = Observable.just(new Object()).replay(1).refCount();

            TestObserver<Object> observer1 = observable
                    .subscribeOn(Schedulers.io())
                    .test();

            TestObserver<Object> observer2 = observable
                    .subscribeOn(Schedulers.io())
                    .test();

            assertThat(observer1.awaitTerminalEvent(5, TimeUnit.SECONDS), is(true));
            assertThat(observer2.awaitTerminalEvent(5, TimeUnit.SECONDS), is(true));
        }
}
```

As you can see, I am repeating the logic 10000 times. This (mostly) guarantees that the test fails on my environment. One of the observers does NOT receive the completed event and therefore the `awaitTerminalEvent` times out.

On the other hand, the following unit test where I manually `.connect()` the `ConnectableObservable`  returned vom `.replay(1)` passes:

```
public void replayShallBeThreadSafe() {
        for (int i = 0; i < 10000; i++) {
            ConnectableObservable<Object> observable = Observable.just(new Object()).replay(1);
            Disposable connection = observable.connect();

            TestObserver<Object> observer1 = observable
                    .subscribeOn(Schedulers.io())
                    .test();

            TestObserver<Object> observer2 = observable
                    .subscribeOn(Schedulers.io())
                    .test();

            assertThat(observer1.awaitTerminalEvent(5, TimeUnit.SECONDS), is(true));
            assertThat(observer2.awaitTerminalEvent(5, TimeUnit.SECONDS), is(true));
            connection.dispose();
        }
}
```

Please tell me if I am missing something...

Version: RxJava 2.2.0
Tests are JUnit tests, with hamcrest assertions.

Thanks!

akarnokd:
This PR modifies the `refCount` operator (in both `Flowable` and `Observable` types) to avoid certain termination-reconnection races.

The original race could happen when the refCounted source terminated at the same time as new observers arrived, leaving those new observers hanging as they practically joined a dying connection and got possibly undercut by the `dispose` call.

The change involve a new internal interface `ResettableConnection` that will allow resetting the connection object inside the connectable source if it is equal to the connection object known by the initiator of the original `connect` call.

Fixes #6185

### DIFF
--- a/src/main/java/io/reactivex/internal/disposables/ResettableConnectable.java
+++ b/src/main/java/io/reactivex/internal/disposables/ResettableConnectable.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
+ * distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.internal.disposables;
+
+import io.reactivex.annotations.Experimental;
+import io.reactivex.flowables.ConnectableFlowable;
+import io.reactivex.observables.ConnectableObservable;
+
+/**
+ * Interface allowing conditional resetting of connections in {@link ConnectableObservable}s
+ * and {@link ConnectableFlowable}s.
+ * @since 2.2.2 - experimental
+ */
+@Experimental
+public interface ResettableConnectable {
+
+    /**
+     * Reset the connectable if the current internal connection object is the
+     * same as the provided object.
+     * @param connectionObject the connection object identifying the last known
+     * active connection
+     */
+    void resetIf(Object connectionObject);
+}
--- a/src/main/java/io/reactivex/internal/disposables/ResettableConnectable.java
+++ b/src/main/java/io/reactivex/internal/disposables/ResettableConnectable.java
@@ -14,6 +14,7 @@
 package io.reactivex.internal.disposables;
 
 import io.reactivex.annotations.Experimental;
+import io.reactivex.disposables.Disposable;
 import io.reactivex.flowables.ConnectableFlowable;
 import io.reactivex.observables.ConnectableObservable;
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/internal/disposables/ResettableConnectable.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Interface allowing conditional resetting of connections in {@link ConnectableObservable}s
+ * and {@link ConnectableFlowable}s.
+ * @since 2.2.2 - experimental
+ */
+@Experimental
+public interface ResettableConnectable {
+
+    /**
+     * Reset the connectable if the current internal connection object is the
```
vanniktech: Find it hard to read and follow. Could we link the param in here so it's understandable what's the internal and what's the provided object?

vanniktech (approved):

src/main/java/io/reactivex/internal/disposables/ResettableConnectable.java
akarnokd: Changed the description.

vanniktech (approved):
Much better.

davidmoten (commented):
LGTM, thanks

davidmoten:
Thanks for the fix @akarnokd and the report @BernhardReu. A release in the next couple of weeks with this change would be very welcome (`.share` is a critical path for me). /cc @SuperEvenSteven @philipgroom.

akarnokd:
@davidmoten Sure. #6194

jhansche:
@akarnokd This seems to fix the issue for `.replay().refCount()`, but I'm seeing the same test fail intermittently with `.publish().refCount()` (aka `share()`):

```
    @Test
    fun `test refcount race conditions`() {
        for (i in 0 until 10_000) {
            val observable = Observable.just(1).publish().refCount()

            val observer1 = observable
                .subscribeOn(Schedulers.io())
                .test()

            val observer2 = observable
                .subscribeOn(Schedulers.io())
                .test()

            observer1
                .withTag("observer1 $i")
                .assertResult(1)

            observer2
                .withTag("observer2 $i")
                .assertResult(1)
        }
    }
```

The actual failure is obviously unpredictable due to the race conditions, but it usually fails pretty early.  In this case, it failed at `i=14`:

```
java.lang.AssertionError: Value count differs; expected: 1 [1] but was: 0 [] (latch = 1, values = 0, errors = 0, completions = 0, timeout!, disposed!, tag = observer2 14)
Expected :1 [1] 
Actual   :0 [] (latch = 1, values = 0, errors = 0, completions = 0, timeout!, disposed!, tag = observer2 14)
```

akarnokd:
@jhansche It is possible one of the observers connect right between onNext and onComplete, thus it may not receive the value. However, they should both complete, no timeout. I put in a fresh issue #6501 and will investigate further tomorrow.

PR #6199, opened 2018-09-03
Fix terminology of cancel/dispose in the JavaDocs

### LINKED ISSUE #6196: Fix terminology of cancel/dispose in the JavaDocs
`Flowable`s can be **canceled** and the other reactive types can be **disposed**. Change the JavaDocs of these latter types accordingly (via careful find/replace):

- [ ] Observable
- [ ] Single
- [ ] Maybe
- [ ] Completable

(Note: `Flowable` is okay).

luis-cortes:
Resolves #6196

Changing terminology from cancel -> dispose wherever appropriate in the following classes:

- `Observable.java`
- `Single.java`
- `Maybe.java`
- `Completable.java`

Areas where I did _not_ change from cancel -> disposal include: 

- where the cancellation was in reference to a `Flowable`, `Publisher`, or a `Future`
- `test()` methods where the `TestObserver` returned actually has a `cancel()` method

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -432,7 +432,7 @@ public abstract class Completable implements CompletableSource {
      * <p>
      * <img width="640" height="628" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Completable.fromFuture.png" alt="">
      * <p>
-     * Note that if any of the subscribers to this Completable call dispose, this Completable will cancel the future.
+     * Note that if any of the observers to this Completable call dispose, this Completable will cancel the future.
      * <dl>
      *  <dt><b>Scheduler:</b></dt>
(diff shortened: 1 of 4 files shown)


src/main/java/io/reactivex/Completable.java
```diff
@@ -432,7 +432,7 @@ public static Completable fromCallable(final Callable<?> callable) {
      * <p>
      * <img width="640" height="628" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Completable.fromFuture.png" alt="">
      * <p>
-     * Note that cancellation from any of the subscribers to this Completable will cancel the future.
+     * Note that if any of the subscribers to this Completable call dispose, this Completable will cancel the future.
```
akarnokd: "Note that if any of the **observers** to this Completable call dispose,"

akarnokd (commented):
One small adjustment and its good to go.

vanniktech (approved):

akarnokd (approved):

PR #6215, opened 2018-09-17
2.x: Add Observable.distinctOfGroup(count) operator

lenarf:
Add 2 operators for Observable:
- distinctOfGroup(int count):
- distinctOfGroup(int count, Function<? super T, K> keySelector):

Works as a distinctUntilChanged operator, but not for one last element, but for a group of last elements.

Based on fixed size queue-collection (Functions.java):

```
    static final class FixedQueueCallable<T>
            implements Callable<Queue<T>> {
        final int capacity;

        FixedQueueCallable(int capacity) {
            this.capacity = capacity;
        }

        @Override
        public Queue<T> call() throws Exception {
            return new LinkedList<T>() {
                @Override
                public boolean add(T t) {
                    boolean retVal = contains(t);

                    if (size() >= capacity) {
                        removeFirst();
                    }
                    addLast(t);

                    return !retVal;
                }
            };
        }
    }

    public static <T> Callable<Queue<T>> createFixedQueue(int capacity) {
        return new FixedQueueCallable<T>(capacity);
    }
```

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -7820,6 +7820,74 @@ public abstract class Observable<T> implements ObservableSource<T> {
     }
 
     /**
+     * Returns an Observable that emits all items emitted by the source ObservableSource that are distinct from their
+     * immediate predecessors groups based on {@link Object#equals(Object)} comparison.
+     * <p>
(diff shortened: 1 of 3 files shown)


akarnokd:
This looks like a very peculiar, convenience overload of an existing operator and should most likely live as an `ObservableTransformer` in your local projects.

In addition, please open an issue first so that neither we nor you spend time on feature requests that will be very likely rejected.

PR #6230, opened 2018-09-29
Removed an unnecessary if statement.

Xenios91:
I found this unnecessary if statement.


akarnokd:
If that if is unnecessary, why is the [build broken](https://travis-ci.org/ReactiveX/RxJava/builds/434909847#L840)? Please don't contribute just out of boredom and make sure you understand what you are changing.

PR #6234, opened 2018-10-01
Add timeout and unit to TimeoutException message

artem-zinnatullin:
This is a small enhancement to help with crash/log debugging.

Right now often times you get a stacktrace that points only to RxJava:

```java
java.util.concurrent.TimeoutException
        at
io.reactivex.internal.operators.flowable.FlowableTimeoutTimed$TimeoutSubscriber.onTimeout(FlowableTimeoutTimed.java:137)
        at
io.reactivex.internal.operators.flowable.FlowableTimeoutTimed$TimeoutTask.run(FlowableTimeoutTimed.java:169)
        at
io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
        at
io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
        at
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748) 
```

Adding `timeout` and `unit` values can help find related user code faster.

I'm not advocating for particular message, it can be something even shorter, like `$timeout $unit`. 

It can also be extracted in a method if you see value in that.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableTimeoutTests.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableTimeoutTests.java
@@ -13,6 +13,7 @@
 
 package io.reactivex.internal.operators.flowable;
 
+import static io.reactivex.internal.util.ExceptionHelper.timeoutMessage;
 import static org.junit.Assert.*;
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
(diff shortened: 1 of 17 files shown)


vanniktech (approved):

akarnokd:
I'm for better error messages but I'm not convinced a timeout value is generally unique enough in a project that will allow you to locate the flow. Especially when the timeout value comes from configuration and thus it is not prevalent in the source code.

An alternative would be an overload on the operator with a String argument that would be passed onto the exception constructor, tagging the particular source location.

artem-zinnatullin:
I like your suggestion, I still however think that adding timeout and value is a good option for now and for the case when user didn't specify a message.

This will be a "free" improvement for existing users who will just receive it with an update of the library and 0 changes in their code.

>Especially when the timeout value comes from configuration and thus it is not prevalent in the source code.

Often times these configurations are traceable, thus knowing the values can be helpful enough. 

But still, optional message sounds good to me, I can work on it once we decide on the destiny of this PR

akarnokd:
Okay, let's have a default message for now, but more like along these lines:

`The source did not signal an event for $timeout ${unit.toLowerCase()}(s) and has been terminated.`

artem-zinnatullin:
sgtm, implementing

src/test/java/io/reactivex/internal/operators/flowable/FlowableTimeoutTests.java
```diff
@@ -82,26 +83,24 @@ public void shouldNotTimeoutIfSecondOnNextWithinTimeout() {
 
     @Test
     public void shouldTimeoutIfOnNextNotWithinTimeout() {
-        Subscriber<String> subscriber = TestHelper.mockSubscriber();
```
artem-zinnatullin, line 85: Here and below I'm not sure what's the purpose of mocking, TestSubscriber/TestObserver are capable of checking what's needed in these test cases

artem-zinnatullin:
I've updated the message and added tests, turns out in some cases we didn't check for `TimeoutException`, now these cases are covered

@vanniktech please take a look again

src/test/java/io/reactivex/internal/operators/flowable/FlowableTimeoutTests.java
akarnokd: Probably some v1 hybrid remnant.

akarnokd (approved):

vanniktech (approved):
Nice even getting rid of some mockito nonsense.

PR #6236, opened 2018-10-03
2.x: Adjust Undeliverable & OnErrorNotImpl message to use full inner exception

akarnokd:
With `UndeliverableException` and `OnErrorNotImplementedException`, the wrapped exception may not have a message and the top line simply contains an unhelpful `null`:

    io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the 
            consumer because it has already canceled/disposed the flow or the exception 
            has nowhere to go to begin with. Further reading: 
            https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | null
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
        at io.reactivex.internal.operators.observable.ObservableCreate$CreateEmitter
            .onError(ObservableCreate.java:73)
        at io.reactivex.internal.operators.observable.ObservableCreate
            .subscribeActual(ObservableCreate.java:43)
        at io.reactivex.Observable.subscribe(Observable.java:12090)
     Caused by: java.io.InterruptedIOException
 
This PR adds the full `toString()` value of the wrapped exception to the main line. This is also more useful when the user posting the stacktrace doesn't provide the `Caused by:` part for some reason.

### DIFF
--- a/src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
+++ b/src/main/java/io/reactivex/exceptions/OnErrorNotImplementedException.java
@@ -48,6 +48,6 @@ public final class OnErrorNotImplementedException extends RuntimeException {
      *          the {@code Throwable} to signal; if null, a NullPointerException is constructed
      */
     public OnErrorNotImplementedException(@NonNull Throwable e) {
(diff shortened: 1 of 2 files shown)


vanniktech (approved):

artem-zinnatullin (approved):
lgtm

PR #6238, opened 2018-10-05
Implement ReplayBufferStrategy strategy for ReplaySubject

hadilq:
With this pull request, I propose strategy design pattern for handling different policies to buffer onNext values of ReplaySubject. These strategies would be useful specifically in Android, and can replace LiveData behaviour on emitting buffered values to UI components after configuration changes. For instance, when a View is subscribed to its ViewModel, by ReplaySubject instead of LiveData, and user decide to rotate the View, the observers of View would be disposed and miss new events while rotating. Such event could be a response from server, then by ReplayBufferStrategy.NO_OBSERVER strategy after rotating, the View can be notified with the response of the server.  In this scenario, after subscribing all observers of View to ReplaySubject of ViewModel, the onNext must be called to clear the buffer.

Also there is ReplayBufferStrategy.NO_OBSERVER_EMIT_ONCE strategy, which is a consequence of adding ReplayBufferStrategy.NO_OBSERVER strategy. It would be useful if in the above example the View has exactly one observer, so after all subscription we don't need to call onNext to clear the buffer.

In case of any inappropriate or wrong code, please notify me to fix it. Thank you so much.

### DIFF
--- a/src/main/java/io/reactivex/ReplayBufferStrategy.java
+++ b/src/main/java/io/reactivex/ReplayBufferStrategy.java
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) 2016-present, RxJava Contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is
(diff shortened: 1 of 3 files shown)


akarnokd:
This is a very peculiar feature which is better implemented outside RxJava (i.e., just like RxRelay or RxReplayingShare and others). Also the changes look thread-unsafe (like clearing the buffer while a new observet arrives). Finally, you should have opened a discussion issue first to avoid spending time on code that is likely to be rejected.

Issue #5081, opened 2017-02-07
RxJava 2.x does not have delaySubscription API for Completable.


tonycosentini:
Seems like an oversight since this exists in 1.x.

akarnokd:
Could you be more specific?

tonycosentini:
Apoligies for the vague issue.

I thought Rx 1.x had `delaySubscription` available for Completable and it was missing in 2.x. It looks like that's not the case though, delaySubscription is not available in either 1.x or 2.x.

Would it be ok to add `delaySubscription` for Completable in 2.x? I'm happy to open a PR, just want to double check if there is any reason it doesn't exist.

akarnokd:
Delaying subscription to a `Completable` "b" by a `Completable` "a" is `a.andThen(b)`. The other types can be transformed into Completable to participate in this pattern.

If you want, go ahead with a PR.

tonycosentini:
Makes sense - the same thing can be achieved with that and it's one less API to maintain.

I'm going to close this - using `andThen` solves my needs and is much simpler IMO.

Thank you!

soshial:
@akarnokd , but why do we need to write this
`Completable.timer(1, TimeUnit.SECONDS).andThen(deleteDb()).subscribe();`
instead of just doing this:
`deleteDb().delaySubscription(1, TimeUnit.SECONDS).subscribe();`?

Also, we do not always have access to the beginning of Completable, we might want to alter its subscription time afterwards, isn't it?

akarnokd:
@soshial I offered the option to write a PR for a specific timed `delaySubscription` but the OP chose the readily available workaround.

> Also, we do not always have access to the beginning of Completable, we might want to alter its subscription time afterwards, isn't it?

This makes no sense. You need the `Completable` to be able to delay a subscription to it. Also you can do this:

```java
deleteDb.compose(c -> timer(1, SECONDS).andThen(c));
```

Either way, the `Completable` instance has to exist.

PR #6242, opened 2018-10-10
Add delaySubscription() methods to Completable #5081

soshial:
Since Observable, Single already have `delaySubscription()`, but Completable doesn't, I added these methods to the code.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
+++ b/src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
@@ -14,17 +14,24 @@
 package io.reactivex.internal.operators.completable;
 
 import static org.junit.Assert.assertNotEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
 
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.junit.Test;
 
-import io.reactivex.*;
+import io.reactivex.CompletableSource;
+import io.reactivex.Observer;
+import io.reactivex.TestHelper;
+import io.reactivex.Completable;
 import io.reactivex.exceptions.TestException;
 import io.reactivex.functions.*;
 import io.reactivex.observers.TestObserver;
-import io.reactivex.schedulers.*;
+import io.reactivex.schedulers.Schedulers;
+import io.reactivex.schedulers.TestScheduler;
 
 public class CompletableDelayTest {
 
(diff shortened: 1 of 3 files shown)


akarnokd:
Please [fix the mistakes](https://travis-ci.org/ReactiveX/RxJava/builds/439579164#L583) and add unit tests verifying these new methods.

soshial:
The code failed this test `ParamValidationCheckerTest` > checkCompletable FAILED.

Which should I better do:
1. put `ObjectHelper.verifyPositive(delay, "delay");` into all new methods of mine or
2. `return RxJavaPlugins.onAssembly(new CompletableTimer(Math.max(delay, 0L), unit, scheduler));` inside timer(...)

The first way looks more logical to me.

akarnokd:
Neither. Update the test to permit any integer like the other `delaySubscription` implementations, e.g., https://github.com/ReactiveX/RxJava/blob/2.x/src/test/java/io/reactivex/validators/ParamValidationCheckerTest.java#L149

soshial:
Got it. I will fix the tests. What about `timer()`, which way is correct:

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Completable.java#L909
or
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Observable.java#L3906
?

akarnokd:
You don't have to `max`, the one in `Observable` version is old and unnecessary.

soshial:
Should I remove `max` then from `Observable`?

Can I add `@since 2.3` to the new methods?

akarnokd:
> Should I remove max then from Observable?

Sure.

> Can I add @since 2.3 to the new methods?

Add `@since 2.2.3 - experimental` and `@Experimental`

akarnokd (changes requested):
Please add unit tests that verify the expected behavior.

src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
```diff
@@ -120,4 +127,84 @@ public void errorDelayed() {
...
+        result.subscribe(to);
+        to.assertEmpty();
+
+        scheduler.advanceTimeBy(90, TimeUnit.MILLISECONDS);
+        to.dispose();
+
+        scheduler.advanceTimeBy(15, TimeUnit.MILLISECONDS);
+
+        to.assertEmpty();
+        verify(o, never()).onComplete();
```
akarnokd: No need to talk to  a mocked observer, TestObserver.assertEmpty() already verifies these.

src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
```diff
@@ -120,4 +127,84 @@ public void errorDelayed() {
...
+        TestObserver<Void> to = new TestObserver<Void>(o);
+
+        result.subscribe(to);
+
+        to.assertEmpty();
+        scheduler.advanceTimeBy(90, TimeUnit.MILLISECONDS);
+        to.dispose();
+        scheduler.advanceTimeBy(15, TimeUnit.MILLISECONDS);
+
+        verify(o, never()).onComplete();
```
akarnokd: No need to talk to  a mocked observer, TestObserver.assertEmpty() already verifies these.

src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
```diff
@@ -120,4 +127,84 @@ public void errorDelayed() {
...
+        TestObserver<Void> to = new TestObserver<Void>(o);
+
+        result.subscribe(to);
+
+        scheduler.advanceTimeBy(90, TimeUnit.MILLISECONDS);
+        to.assertEmpty();
+        scheduler.advanceTimeBy(15, TimeUnit.MILLISECONDS);
+        to.assertComplete();
+
+        verify(o, never()).onError(any(Throwable.class));
```
akarnokd: No need to talk to  a mocked observer, TestObserver.assertEmpty() already verifies these.

src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
```diff
@@ -120,4 +127,84 @@ public void errorDelayed() {
...
+
+        Observer<Void> o = TestHelper.mockObserver();
+        TestObserver<Void> to = new TestObserver<Void>(o);
+
+        result.subscribe(to);
+
+        scheduler.advanceTimeBy(90, TimeUnit.MILLISECONDS);
+        to.assertEmpty();
+        scheduler.advanceTimeBy(15, TimeUnit.MILLISECONDS);
+        to.assertComplete();
```
akarnokd: `to.assertResult();` will check for the no-error condition.

akarnokd (changes requested):

soshial:
Thanks a lot for explanations! May I simply edit the last commit, because I gave a wrong title to the commit with tests (8233972)?

akarnokd:
Just commit the changes as fresh.

akarnokd (approved):

src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
```diff
@@ -120,4 +123,69 @@ public void errorDelayed() {
...
+    public void testDelaySubscription() {
+        TestScheduler scheduler = new TestScheduler();
+        Completable result = Completable.complete()
+                .delaySubscription(100, TimeUnit.MILLISECONDS, scheduler);
+        TestObserver<Void> to = result.test();
+
+        scheduler.advanceTimeBy(90, TimeUnit.MILLISECONDS);
+        to.assertEmpty();
+        scheduler.advanceTimeBy(15, TimeUnit.MILLISECONDS);
+        to.assertResult();
```
artem-zinnatullin, line 189: nit: here we rely just on completion signal, which as an implementation bug may be emitted by `delaySubscription` operator. I mean, right now there is obviously no bug, but in future if operator will be re-implemented, it might appear

Thus I'd recommend to add verifiable logic to the upstream, like `Completable.fromAction { if (!atomicBoolean.compareAndSet(false, true)) throw IllegalStateException() }`

and verify against that too:

```
to.assertResult()
assertTrue(atomicBoolean.get())
```

artem-zinnatullin (approved):

vanniktech (approved):

akarnokd:
@soshial do you want to address @artem-zinnatullin 's review?

src/test/java/io/reactivex/internal/operators/completable/CompletableDelayTest.java
akarnokd: I'll add the extra tests in a separate PR.

soshial: Thank you, @akarnokd

PR #6257, opened 2018-10-19
2.x: Add Completable.delaySubscription marble, fix some javadoc

akarnokd:
- Add marbles to the new `Completable.delaySubscription` operator.
- Fix the method name referenced in the `ConditionalSubscriber` javadoc.

image
image

### DIFF
--- a/src/main/java/io/reactivex/internal/fuseable/ConditionalSubscriber.java
+++ b/src/main/java/io/reactivex/internal/fuseable/ConditionalSubscriber.java
@@ -16,7 +16,7 @@ package io.reactivex.internal.fuseable;
 import io.reactivex.FlowableSubscriber;
 
 /**
- * A Subscriber with an additional onNextIf(T) method that
+ * A Subscriber with an additional {@link #tryOnNext(Object)} method that
  * tells the caller the specified value has been accepted or
  * not.
  *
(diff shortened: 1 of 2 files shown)


vanniktech (approved):

src/main/java/io/reactivex/internal/fuseable/ConditionalSubscriber.java
```diff
@@ -16,7 +16,7 @@
 import io.reactivex.FlowableSubscriber;
 
 /**
- * A Subscriber with an additional onNextIf(T) method that
+ * A Subscriber with an additional {@link #tryOnNext(Object)} method that
```
codefromthecrypt, line 19: thanks was going to raise this :P

src/main/java/io/reactivex/internal/fuseable/ConditionalSubscriber.java
```diff
@@ -16,7 +16,7 @@
...
  * tells the caller the specified value has been accepted or
  * not.
  *
```
codefromthecrypt, line 22: separate topic, but when reading this I was wondering if we should call out pattern/antipattern for calling `tryOnNext` *and also* `onNext`. Is there a case where this should happen? If so, or not, would that description end up in javadoc or somewhere else?

akarnokd: No mixing. See Flowable's fromArray and filter for usage examples.

PR #6258, opened 2018-10-24
2.x: Fix cancel/dispose upon upstream switch for some operators

akarnokd:
This PR extends the `SubscriptionArbiter` to optionally allow or disallow cancelling the current `Subscription` if it is replaced by a new one. Some operators do not need to cancel the current `Subscription`: `concat`, `concatMap`, `repeat`, `repeatWhen`, `retry` and `retryWhen`. 

In addition `repeatWhen` and `retryWhen` were cancelling when the handler sequence itself terminated. The code has been updated to disconnect the upstream upon the completion/failure but before signaling the handler.

The Reactive Streams specification also disallows synchronous cancellation after the terminal event anyway.

Others may actually need to cancel, such as `Timeout`. 

`Observable`s don't have a specific arbiter, they use the `DisposableHelper` methods and the relevant ones were changed to `replace()` instead of the disposing `set` call.

Some tests actually checking if the dispose/cancel happens and had to be updated.

The `Flowable.delaySubscription(Publisher)` also used `SubscriptionArbiter` but it was unnecessary. The code has been replaced with a more apt deferred requesting scheme as the downstream requests need to be delayed until the main subscription happens, the other publisher is always consumed unbounded.

Resolves: #6259

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableDelaySubscriptionOther.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableDelaySubscriptionOther.java
@@ -12,10 +12,16 @@
  */
 package io.reactivex.internal.operators.flowable;
 
-import org.reactivestreams.*;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
-import io.reactivex.*;
-import io.reactivex.internal.subscriptions.SubscriptionArbiter;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import io.reactivex.Flowable;
+import io.reactivex.FlowableSubscriber;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
 import io.reactivex.plugins.RxJavaPlugins;
 
 /**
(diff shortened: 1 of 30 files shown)


artem-zinnatullin:
David, can you please allow 2-3 days window for review on this one? 
I'm smashed by work, but would really like to check this PR carefully

akarnokd:
Sure Artem.

vanniktech (approved):

isfaaghyth (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableDelaySubscriptionOther.java
```diff
@@ -35,94 +37,105 @@ public FlowableDelaySubscriptionOther(Publisher<? extends T> main, Publisher<U>
...
+                    s.request(Long.MAX_VALUE);
+                }
             }
 
             @Override
-            public void onNext(T t) {
-                child.onNext(t);
+            public void onNext(Object t) {
+                Subscription s = get();
+                if (s != SubscriptionHelper.CANCELLED) {
```
artem-zinnatullin, line 115: offtopic: I'm wondering if we should drop `SubscriptionHelper.isCancelled()`

artem-zinnatullin (approved):

artem-zinnatullin (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowableDelaySubscriptionOther.java
akarnokd: If you want to inline all of them.

artem-zinnatullin: https://github.com/ReactiveX/RxJava/pull/6263

PR #6262, opened 2018-10-26
Use JUnit's assert format for assert messages to enable better suppor…

bjoernQ:
This changes the message format of "assert" to match that of JUnit. This way IDEs like IntelliJ IDEA are able to parse it and offer functionality like "compare values".

### DIFF
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -351,11 +351,11 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
     public final U assertValue(T value) {
         int s = values.size();
         if (s != 1) {
-            throw fail("expected:<" + valueAndClass(value) + "> but was:<" + values + ">");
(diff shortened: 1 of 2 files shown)


akarnokd:
I thought IntelliJ already recognized RxJava's Expected/Actual after this: #5258.

vanniktech:
I thought so too.

bjoernQ:
Unfortunately #5258 didn't made it into 2.x as it seems

vanniktech (approved):
That explains it

akarnokd:
Looks like it was never extended: https://github.com/JetBrains/intellij-community/blob/99614a63425774df58eea3ac92e2b20af1633663/plugins/junit_rt/src/com/intellij/junit4/ExpectedPatterns.java#L34

You don't need <> as this pattern matches `expected: ` and `but was: ` so you only need to replace `Actual: `.

bjoernQ:
Thanks - will remove the (somewhat awkward <>) ... just added it to match the JUnit style but it's really not needed

akarnokd (approved):

vanniktech (approved):

Issue #6272, opened 2018-10-30
Have materialize() and dematerialize() for Maybe, Single and Completable?


akarnokd:
`Observable` and `Flowable` already have these methods, it might be worth adding them to `Maybe`, `Single` and `Completable`.

The main benefit I see with `Maybe` and `Single` is for the `zip` operator which now could wait for all source outcomes and the `zipper` will be always invoked.

Note that `Completable.materialize()` would become `Single<Notification<T>>` as well. `T` could be `Void` or an arbitrary target type:

```java
Single<Notification<Void>> c = Completable.complete().<Void>materialize();
Single<Notification<Integer>> s = Single.just(1).materialize();
Single<Notification<Integer>> m = Maybe.<Integer>empty().materialize();

Single.zip(c, s, m, (a, b, c) -> 1);
```

PR #6278, opened 2018-11-01
 2.x: Add materialize() and dematerialize()

akarnokd:
This PR adds the `materialize` operator to `Maybe`, `Single` and `Completable` to turn their signals into the corresponding `Notification` object. This operator has been available for `Observable`s (and `Flowable`s) from the beginning of the Rx API. The methods return `Single<Notification<T>>`.

To complement, the `dematerialize` operator is only defined for `Single` and results in a `Maybe`.

If accepted, I'll draw the correct marble diagrams for them in a separate PR.

Resolves: #6272

### DIFF
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -2317,7 +2317,7 @@ public abstract class Single<T> implements SingleSource<T> {
      * a type argument on this method (see example below).
      * <dl>
      * <dt><b>Scheduler:</b></dt>
-     * <dd>{@code delaySubscription} does by default subscribe to the current Single
+     * <dd>{@code dematerialize} does by default subscribe to the current Single
      * on the {@link Scheduler} you provided, after the delay.</dd>
      * </dl>
      * <p>
--- a/src/main/java/io/reactivex/Single.java
+++ b/src/main/java/io/reactivex/Single.java
@@ -2303,6 +2303,33 @@ public abstract class Single<T> implements SingleSource<T> {
     }
 
     /**
+     * Maps the {@link Notification} success value of this Single back into normal
+     * {@code onSuccess}, {@code onError} or {@code onComplete} signals as a
+     * {@link Maybe} source.
+     * <p>
+     * Note that {@code this} should be of type {@code Single<Notification<T>>} or
+     * the transformation will result in an {@code onError} signal of
+     * {@link ClassCastException}. Currently, the Java language doesn't allow specifying
+     * methods for certain type argument shapes only (unlike extension methods would),
+     * hence the forced casting in this operator.
+     * <dl>
+     * <dt><b>Scheduler:</b></dt>
+     * <dd>{@code delaySubscription} does by default subscribe to the current Single
+     * on the {@link Scheduler} you provided, after the delay.</dd>
+     * </dl>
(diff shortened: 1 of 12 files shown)


src/main/java/io/reactivex/Single.java
```diff
@@ -2302,6 +2302,44 @@
...
+     * </code></pre>
+     * @param <T2> the type inside the Notification
+     * @return the new Maybe instance
+     * @since 2.2.4 - experimental
+     */
+    @SuppressWarnings("unchecked")
+    @CheckReturnValue
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @Experimental
+    public final <T2> Maybe<T2> dematerialize() {
```
vanniktech: I find this API very awkward since it's available on every type but should only be used on `Single<Notification<T>>`. I don't have a better suggestion though. Except defining it yourself and using a Kotlin extension function.

vanniktech (approved):
:+1: to materialize

Mixed feelings about dematerialize

src/main/java/io/reactivex/Single.java
akarnokd: This has been the workaround for [`Observable.dematerialize`](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#dematerialize--) too.

The alternative is to have the user select the notification:

`Maybe<R> dematerialize(Function<T, Notification<R>> selector);`

thus if you have a `Single<Notification<T>>`, you only need an identity selector:

```java
Single<Notification<T>> source = ...
Maybe<T> source.dematerialize(v -> v);
```

However, this selector pattern is not an established one in RxJava right now. We could go for this but then the other existing `dematerialize` operators should also get a selector overload.

vanniktech: Didn't know about `dematerialize`, then I guess we stick to that convention. Providing a mapper sounds good. Then maybe we could deprecate the current `dematerialize` in Observable.

akarnokd: Okay then. I'll change this to use the selector and post a separate PR for adding a selector variant to `Observable`/`Flowable` and deprecate the type-fragile versions.

vanniktech: Sounds good!

src/test/java/io/reactivex/internal/operators/single/SingleDematerializeTest.java
```diff
@@ -18,62 +18,90 @@
...
+import io.reactivex.internal.functions.Functions;
 import io.reactivex.subjects.SingleSubject;
 
 public class SingleDematerializeTest {
 
     @Test
     public void success() {
         Single.just(Notification.createOnNext(1))
-        .<Integer>dematerialize()
+        .dematerialize(Functions.<Notification<Integer>>identity())
```
vanniktech, line 29: Too bad the identity function is inside an internal package :(

vanniktech (approved):

src/test/java/io/reactivex/internal/operators/single/SingleDematerializeTest.java
akarnokd: I suppose most users now can use lambdas in their project so this is just `v -> v` for them. Only we are still so unlucky to use such helper methods.

PR #6295, opened 2018-11-08
Fix the Multiple production threads cannot asynchronously invoke  synchronous delivery element methods at the same time

muyinchen:
When calling the generate method, when there are multiple threads that produce the delivery elements in the implementation logic of the Emitter, they cannot asynchronously call the synchronous delivery element method at the same time. Here, the error should be added to facilitate the developer to write the code bug. The problem is, a test is given later, temporarily written, you can refer to the operation.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java
@@ -130,7 +130,7 @@ public final class FlowableGenerate<T, S> extends Flowable<T> {
                     if (!hasNext) {
                         state = null;
                         onError(new IllegalStateException("The generator didn't call any of the " +
-                                "SynchronousSink method"));
+                                "SynchronousEmitter method"));
                         return;
                     }
                     e++;
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java
@@ -13,17 +13,20 @@
 
 package io.reactivex.internal.operators.flowable;
 
-import java.util.concurrent.Callable;
-import java.util.concurrent.atomic.AtomicLong;
-
-import org.reactivestreams.*;
-
-import io.reactivex.*;
+import io.reactivex.Emitter;
+import io.reactivex.Flowable;
 import io.reactivex.exceptions.Exceptions;
-import io.reactivex.functions.*;
-import io.reactivex.internal.subscriptions.*;
+import io.reactivex.functions.BiFunction;
+import io.reactivex.functions.Consumer;
+import io.reactivex.internal.subscriptions.EmptySubscription;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
 import io.reactivex.internal.util.BackpressureHelper;
 import io.reactivex.plugins.RxJavaPlugins;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicLong;
 
 public final class FlowableGenerate<T, S> extends Flowable<T> {
     final Callable<S> stateSupplier;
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java
```diff
@@ -122,7 +125,14 @@ public void request(long n) {
                         dispose(s);
                         return;
                     }
-
+                    //Multiple production threads cannot asynchronously
+                    //invoke synchronous delivery element methods at the same time
+                    if (!hasNext) {
```
akarnokd, line 130: Not sure what you want here as the multi-call to `onNext` is already covered: https://github.com/ReactiveX/RxJava/blob/a9fe6333099f47babbab1c622dca2685f9aaf011/src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java#L177

src/main/java/io/reactivex/internal/operators/flowable/FlowableGenerate.java
```diff
@@ -13,17 +13,20 @@
 
 package io.reactivex.internal.operators.flowable;
 
-import java.util.concurrent.Callable;
-import java.util.concurrent.atomic.AtomicLong;
-
-import org.reactivestreams.*;
-
-import io.reactivex.*;
+import io.reactivex.Emitter;
```
akarnokd, line 16: Please avoid unrolling `*` imports.

src/test/java/io/reactivex/internal/operators/flowable/FlowableGenerateTest.java
```diff
@@ -1,31 +1,39 @@
 /**
  * Copyright (c) 2016-present, RxJava Contributors.
- *
+ * <p>
```
akarnokd, line 3: Please make sure your IDE doesn't reformat the whole file.

src/test/java/io/reactivex/internal/operators/flowable/FlowableGenerateTest.java
```diff
@@ -279,7 +287,71 @@ public void accept(Emitter<Object> e) throws Exception {
...
             }
         })
-        .test(1)
-        .assertResult();
+                .test(1)
+                .assertResult();
+    }
+
+    @Test
+    public void flowable_bug_generate() {
```
akarnokd, line 295: Generate is not meant to be operated from multiple threads so this case is by definition invalid. Besides, multi-threaded races can't be properly detected even with your changes.

akarnokd:
Not sure what you tried here but it doesn't work in general. Please open an issue first the next time you encounter a problem so there is a use case we can investigate and help you with at least.

davidmoten:
Jiacheng asked me to have a look at this PR as well and I think seeks a bit more feedback.

I've had a quick scan of the PR. My impression of this is that this is not a bug you are fixing but rather you are trying to help the user detect that they called the emitter in a non-compliant way. I looked at the javadoc of Flowable#generate and it says:

>Returns a cold, synchronous, stateless and backpressure-aware generator of values.

I would say the javadoc needs more information in general about the use of `generate` and could do with the addition of a statement that there should be a happens-before relationship between calls to the `Emitter`. The word "synchronous" implies this perhaps but is vague (and presumably the calls can be asynchronous as long as the happens-before relationship exists).

Key comments from @akarnokd are:

> Besides, multi-threaded races can't be properly detected even with your changes.

If you want clarification on this then ask for it (or you could study the code looking for non-detected races which is always an educational experience!).

and 

>Please open an issue first the next time you encounter a problem so there is a use case we can investigate and help you with at least.

This is great advice. Discuss the use case first and suggest your ideas for handling it. It may save you a lot of time (analyzing code, preparing tests) as you would learn from this PR.

Your enthusiasm to contribute is very welcome, there's just a bit of a learning curve to get on top of in terms of both contribution conventions and the codebase itself. Keep at it!

akarnokd:
Thanks @davidmoten for giving a more detailed explanation.

A result worth investigating is that the code added has no coverage even though a test case was added to try and verify it: https://codecov.io/gh/ReactiveX/RxJava/compare/acd5466c4440961da905d0c1c0e78752bda155ef...a9fe6333099f47babbab1c622dca2685f9aaf011/changes#D1-130

I'll open an issue regarding the explicit mention of the non-concurrent use of the `Emitter` in the JavaDocs.

PR #6297, opened 2018-11-09
2.x: Fix refCount eager disconnect not resetting the connection

### LINKED ISSUE #6296: 2.x: refCount doesn't reset its source if one of its subscribers cancels immediately
The following tests fail because the cancel/dispose from the downstream comes before the actual connection is established which leaves the connectable source in a disposed state, refusing to take in new subscribers/observers:

```java
@Test
public void test() {
    BehaviorSubject<Integer> subject = BehaviorSubject.create();

    Observable<Integer> observable = subject
            .replay(1)
            .refCount();

    observable.takeUntil(Observable.just(1)).test();

    subject.onNext(2);

    observable.take(1).test().assertResult(2);
}


@Test
public void test2() {
    BehaviorProcessor<Integer> processor = BehaviorProcessor.create();

    Flowable<Integer> flowable = processor
            .replay(1)
            .refCount();

    // This line causes the test to fail.
    flowable.takeUntil(Flowable.just(1)).test();

    processor.onNext(2);

    flowable.take(1).test().assertResult(2);
}
```

akarnokd:
This PR fixes the case when an observer/subscriber disposes/cancels immediately upon subscribing to a `refCount` operator before it establishes the connection and ends up with a disposed but non-reset connection, preventing further interactions with the connectable source.

The fix is to detect this case and reset the connection when the `connect()` method signals the dispose handler.

Both `ConnectableFlowable.refCount` and `ConnectableFlowable.refCount` is affected.

Fixes: #6296

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
@@ -1394,4 +1394,20 @@ public class FlowableRefCountTest {
 
         assertTrue(((Disposable)o.source).isDisposed());
     }
+
+    @Test
+    public void disconnectBeforeConnect() {
+        BehaviorProcessor<Integer> processor = BehaviorProcessor.create();
+
+        Flowable<Integer> flowable = processor
+                .replay(1)
+                .refCount();
+
+        // This line causes the test to fail.
+        flowable.takeUntil(Flowable.just(1)).test();
+
+        processor.onNext(2);
+
+        flowable.take(1).test().assertResult(2);
+    }
 }
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
@@ -1403,7 +1403,6 @@ public class FlowableRefCountTest {
                 .replay(1)
                 .refCount();
 
-        // This line causes the test to fail.
         flowable.takeUntil(Flowable.just(1)).test();
 
         processor.onNext(2);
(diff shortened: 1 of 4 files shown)


vanniktech (approved):

src/test/java/io/reactivex/internal/operators/flowable/FlowableRefCountTest.java
```diff
@@ -1394,4 +1394,20 @@ public void timeoutDisposesSource() {
...
+
+    @Test
+    public void disconnectBeforeConnect() {
+        BehaviorProcessor<Integer> processor = BehaviorProcessor.create();
+
+        Flowable<Integer> flowable = processor
+                .replay(1)
+                .refCount();
+
+        // This line causes the test to fail.
```
artem-zinnatullin: nit: "caused"

src/test/java/io/reactivex/internal/operators/observable/ObservableRefCountTest.java
```diff
@@ -1345,4 +1345,19 @@ public void timeoutDisposesSource() {
...
+
+    @Test
+    public void disconnectBeforeConnect() {
+        BehaviorSubject<Integer> subject = BehaviorSubject.create();
+
+        Observable<Integer> observable = subject
+                .replay(1)
+                .refCount();
+
+        observable.takeUntil(Observable.just(1)).test();
```
artem-zinnatullin, line 1357: For consistency, this line either needs same comment as flowable test, or it should be removed from flowable test

artem-zinnatullin (approved):

src/test/java/io/reactivex/internal/operators/observable/ObservableRefCountTest.java
akarnokd: Removed from the other place.

artem-zinnatullin (approved):

PR #6304, opened 2018-11-12
[Proposal] Remove CheckReturnValue on subscribe() overloads

ZacSweers:
Proposal PR for my followup comment in https://github.com/ReactiveX/RxJava/issues/4878#issuecomment-437763406

Will close if the discussion raised there still settles on keeping them, just wanted to show a PoC of what the change scope would look like for reference.

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -2318,7 +2318,7 @@ public abstract class Completable implements CompletableSource {
      * @return the Disposable that can be used for disposing the subscription asynchronously
      * @throws NullPointerException if either callback is null
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(final Action onComplete, final Consumer<? super Throwable> onError) {
         ObjectHelper.requireNonNull(onError, "onError is null");
@@ -2345,7 +2345,7 @@ public abstract class Completable implements CompletableSource {
      * @param onComplete the runnable called when this Completable completes normally
      * @return the Disposable that allows disposing the subscription
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(final Action onComplete) {
         ObjectHelper.requireNonNull(onComplete, "onComplete is null");
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -14455,7 +14455,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      *             if {@code onNext} is null
      * @see <a href="http://reactivex.io/documentation/operators/subscribe.html">ReactiveX operators documentation: Subscribe</a>
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(Consumer<? super T> onNext) {
@@ -14486,7 +14486,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      *             if {@code onNext} is null, or
      *             if {@code onError} is null
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(Consumer<? super T> onNext, Consumer<? super Throwable> onError) {
@@ -14520,7 +14520,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      *             if {@code onComplete} is null
      * @see <a href="http://reactivex.io/documentation/operators/subscribe.html">ReactiveX operators documentation: Subscribe</a>
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(Consumer<? super T> onNext, Consumer<? super Throwable> onError,
@@ -14558,7 +14558,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      *             if {@code onSubscribe} is null
      * @see <a href="http://reactivex.io/documentation/operators/subscribe.html">ReactiveX operators documentation: Subscribe</a>
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @BackpressureSupport(BackpressureKind.SPECIAL)
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(Consumer<? super T> onNext, Consumer<? super Throwable> onError,
--- a/src/main/java/io/reactivex/Maybe.java
+++ b/src/main/java/io/reactivex/Maybe.java
@@ -4095,7 +4095,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
      *             if {@code onSuccess} is null
      * @see <a href="http://reactivex.io/documentation/operators/subscribe.html">ReactiveX operators documentation: Subscribe</a>
      */
-    @CheckReturnValue
+    @OptionalCheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     public final Disposable subscribe(Consumer<? super T> onSuccess) {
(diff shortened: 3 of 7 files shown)


vanniktech (changes requested):
I'd keep the annotation and in the rare cases suppress the warning. This annotation saved me more times than it harmed me.

ZacSweers:
Maybe you, but see the thread for plenty more cases where it's not useful :). Let's keep the discussion consolidated there

akarnokd:
This was discussed before. The no arg subscribe doesn't have it as being fire and forget, the rest do.

ZacSweers:
That's not true actually

`Observable#subscribe(Consumer<? super T> onNext, Consumer<? super Throwable> onError, Action onComplete, Consumer<? super Disposable> onSubscribe)`

also requires it despite accepting an onSubscribe Consumer, and presumably that shouldn't require it by that logic either since a handling consumer is passed in. I understand it was discussed before, but I feel I've offered a good set of examples in the linked issue that highlight where the original reasoning has not accounted for.

akarnokd:
> also requires it despite accepting an onSubscribe Consumer

It is to be removed in 3.x: #5622

Also the issue here is a general conflict between one's individual coding style and the hard coded coding style of a library. So you either ignore warnings individually as Jake suggested or create a separate subscribe method/converter in your project where you can ignore it once.

ZacSweers:
> or create a separate subscribe method/converter in your project where you can ignore it once.

This seems like a really harsh stance to take :/. Any specific thoughts on the examples I raised?

vanniktech:
Talked with Zac offline and I understand his points now. We both have different valid ways of using RxJava and the annotation is a constraint that should not be enforced per se. 

I'd be willing to give up the annotation if we have proper open source toolings like Zac suggested that simulate the old behavior.

ZacSweers:
And I'd be happy to contribute implementations (in existing or new projects) for error-prone and lint as a part of this :)

akarnokd:
This should have been part of some tooling all along, but then the lack of warnings out-of-box received by newcommers and beginners would have resulted in a lot of support requests. By the way, if tooling is created for this, it could have both the feature to add as well as remove via configuration for the entire project.

artem-zinnatullin:
my 2 cents: RxJava uses its own annotation for this, which means that it should be possible to disable it project (module) -wide without clashing with `@CheckReturnValue`s enforced by other libraries.

Thus, I'd rather keep it as is in RxJava (because there is value in it) and focus on tooling support for disabling such specific annotations.

vanniktech:
I think for Android Lint this might actually already be doable. For each issue there can be a regex defined. Might be enough to pass in the fully qualified name of the annotation.

ZacSweers:
A custom annotation for this would work, but we'd probably have to deprecate the existing one vs. remove it since it's part of the public API right?

@vanniktech have a link/sample? I didn't know that was an option!

@akarnokd when you say

> it could have both the feature to add as well as remove via configuration for the entire project.

Could you elaborate a bit? I'm not sure I followed

akarnokd:
I meant like a config file:

```
@CheckReturnValue
+ io.reactivex.Flowable.subscribe()
- io.reactivex.Flowable.subscribe(io.reactivex.function.Consumer<? super T>)
```

Some tooling would check this file and apply the annotation virtually to the given method signature or ignore it.

vanniktech:
Something like that might work (haven't tried it out myself):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<lint>
  <issue id="CheckResult">
    <ignore regexp="io.reactivex.annotations.CheckReturnValue"/>
  </issue>
</lint>
```

ZacSweers:
Let me see if ErrorProne's check allows for this. Unfortunately neither case covers the actual IDE tooling marking these (they don't read lint xml configurations and the intelliJ inspection doesn't support customization as far as I know)

vanniktech:
> they don't read lint xml configurations and the intelliJ inspection doesn't support customization as far as I know

I'm sure this can be fixed. Just like custom lint checks got fixed with 3.3

ZacSweers:
Maybe. That was a bug and this would be a FR :). lint xml files are entirely a CLI implementation, and not part of the main lint API that the CLI and IDE plugins both implement

ZacSweers:
Followup on error-prone - confirmed there's no way to configure this right now. I don't think focusing on disabling in other tooling is the right way to go about this, as it's much easier to enable custom tooling than it is to disable non-configurable existing tooling.

artem-zinnatullin:
One (me) could argue that changing a default behavior and requiring everyone who relies on it to reconfigure their tooling is also not great and hardly considerable as much easier :)

ZacSweers:
That’s why I was proposing having new tooling available out the gate, so
anyone using lint or error-prone could just add these. If they’re already
using either of these, it should be a straightforward addition. If they
aren’t, then they likely weren’t using tooling for the old behavior anyway.

In either case - like I mentioned above, adding tooling is a matter of
configuration. Disabling tooling is nigh-impossible without forking
error-prone or lint to configure them in these cases. I’d rather err on the
side of flexibility than imposing something that basically can’t be
disabled.

vanniktech:
> lint xml files are entirely a CLI implementation, and not part of the main lint API that the CLI and IDE plugins both implement

IntelliJ could read the lintConfig file from `android { lintOptions { .. } }`. Not seeing a problem with that

ZacSweers:
The problem is that's a significant feature request for functionality that doesn't exist today right?

vanniktech:
Maybe. Maybe not. It's worth asking for nonetheless ;) - https://issuetracker.google.com/issues/119660763

ZacSweers:
Even then, that still marries it to gradle android projects right?

vanniktech:
Probably yeah. Unless they find some universal way which would also work with other build systems. Feel free to suggest that in the issue since I'm only using Gradle.

ZacSweers:
Yeah I think that's a good thing to file in general, but as a separate async thing. Mind if we continue with using a renamed annotation for rx?

vanniktech:
Yeah I'm in for having both (proper tooling support) and new tooling to flag this. 

You want to rename it so standard tools don't know the name and hence don't flag it, right? Custom checks like you proposed would read it then.

ZacSweers:
Yeah rename, though we'd have to add a separate new one (and deprecate the existing one?) since it's part of the public API up to this point. Also - what should we call it? `PossiblyCheckReturnValue`? `RxCheckReturnValue`?

vanniktech:
I guess yeah but then in the end tools still need a way to exclude the current annotation since we can't remove it, right?

ZacSweers:
Oh no I meant you can't delete the class. You can still remove it from the methods. I was as a actually thinking it's fine in it's other uses, and maybe just make a custom 2nd one for like `CheckDisposableReturnValue`

vanniktech:
Sounds like a good tradeoff.

ZacSweers:
Rebased and switched over to a `OptionalCheckReturnValue`. Let me know if that works?

vanniktech (commented):
It would. I'd like to see the tooling in place before we merge this though.

ZacSweers:
I'll create a project for that now 👍

vanniktech:
You might have the possibility to reuse the existing Detector by creating your own and registering the new annotation and from there forward everything to the CheckResultDetector. That way you wouldn't need to duplicate all of the work.

I've also filed https://issuetracker.google.com/issues/119763072 - dunno if that's something they are willing to do though.

ZacSweers:
I've created a project with working checkers for both here: https://github.com/hzsweers/rxjava2-optionalcheckreturnvalue-checkers

Going to try to get review from @vanniktech and a couple others separately, but they're mostly straightforward adaptation of existing checkers just for this new annotation

vanniktech:
Lint implementation looks good!

JakeWharton:
Still opposed to this. Makes the common case worse for the benefit of the uncommon case.

ZacSweers:
Counterexample: RxLifecycle is nigh-ubiquitous in Android apps and negatively affected by this (since lifecycle handling is handled via upstream takeUntil), would that still be considered an uncommon case?

JakeWharton:
I think you vastly overestimate its use and the use of a view model (not
the library which has nothing to do with view models, of course, the actual
pattern) means you only have a single subscription inside a lifecycle
removing the need to use it anyway.

ZacSweers:
Maybe! Unfortunately neither of us can prove that one way or the other. What about any of the other cases I mentioned in https://github.com/ReactiveX/RxJava/issues/4878#issuecomment-437767153? Especially considering RxJava is synchronous by default, it seems to me like there's quite a few "uncommon" cases. Maybe not enough to be "common", but it gets to a larger point of when static analysis tools should be imposed.

To me, hard checks like this should only be imposed in black-and-white situations, like a factory or cases where it's otherwise pointless to call that method without capturing the result. This seems far from black-and-white. On top of that, it's a check that's not configurable in any built-in error-prone/lint tooling (it's even hardcoded into lint today) that would otherwise enforce it.

I think I have an alternative idea, but I'd like to get thoughts on the above first

Issue #6299, opened 2018-11-09
Javadoc: Explain explicitly that the Emitter methods should be called non-concurrently with generate


akarnokd:
The Javadocs of the `generate` methods as well the `Emitter` interface should explicitly mention that the methods should be called non-concurrently. Generate can be found in both `Flowable` and `Observable`.

#### generate:

```html
<p>
Note that the {@link Emitter#onNext}, {@link Emitter#onError} and 
{@link Emitter#onComplete} methods provided to the function should be called synchronously,
never concurrently and only while the method is executing. Calling them from multiple threads 
or outside the function call is not supported and leads to an undefined behavior.
```

#### Emitter:

```html
<p>
Note that the {@link Emitter#onNext}, {@link Emitter#onError} and 
{@link Emitter#onComplete} methods provided to the function should be called synchronously,
never concurrently. Calling them from multiple threads is not supported and leads to an 
undefined behavior.
```

davidmoten:
I believe calling from multiple threads is supported if there is a happens-before relationship between calls. Perhaps this:
```
<p>
Note that there must be a <i>happens-before</i> relationship between calls to the {@link Emitter#onNext}, {@link Emitter#onError} and 
{@link Emitter#onComplete} methods. Calling those methods from multiple threads without establishing that relationship is not supported and leads to undefined behavior.
```

akarnokd:
Fundamentally, the calls should not escape the callback method and one would have to block the callback if the `onNext` happens asynchronously to the method. Still you can only call it once and the blocking implicitly establishes this relationship.

davidmoten:
Sounds good, thanks @akarnokd

PR #6305, opened 2018-11-13
Javadoc : Explain explicitly about using Emitter interface methods synchronously

punitda:
Add note in Javadoc explicitly explaining about using Emitter interface methods synchronously.

- Add notes for `generate` methods in Observable and Flowable.
- Plus, add the similar note to Emitter interface.

Resolves: #6299

### DIFF
--- a/src/main/java/io/reactivex/Emitter.java
+++ b/src/main/java/io/reactivex/Emitter.java
@@ -17,6 +17,11 @@ import io.reactivex.annotations.NonNull;
 /**
  * Base interface for emitting signals in a push-fashion in various generator-like source
  * operators (create, generate).
+ * <p>
+ * Note that the {@link Emitter#onNext}, {@link Emitter#onError} and
+ * {@link Emitter#onComplete} methods provided to the function via the {@link Emitter} instance should be called synchronously,
+ * never concurrently. Calling them from multiple threads is not supported and leads to an
+ * undefined behavior.
  *
  * @param <T> the value type emitted
  */
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -2210,6 +2210,11 @@ public abstract class Flowable<T> implements Publisher<T> {
 
     /**
      * Returns a cold, synchronous, stateless and backpressure-aware generator of values.
+     * <p>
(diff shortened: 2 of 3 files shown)


src/main/java/io/reactivex/Emitter.java
```diff
@@ -17,6 +17,11 @@
 /**
  * Base interface for emitting signals in a push-fashion in various generator-like source
  * operators (create, generate).
+ * <p>
+ * Note that the {@link Emitter#onNext}, {@link Emitter#onError} and
+ * {@link Emitter#onComplete} methods provided to the function should be called synchronously,
```
akarnokd: "methods provided to the function **via the `{@link Emitter}` instance** should be"

src/main/java/io/reactivex/Flowable.java
```diff
@@ -2210,6 +2210,11 @@ public static int bufferSize() {
 
     /**
      * Returns a cold, synchronous, stateless and backpressure-aware generator of values.
+     * <p>
+     * Note that the {@link Emitter#onNext}, {@link Emitter#onError} and
+     * {@link Emitter#onComplete} methods provided to the function should be called synchronously,
```
akarnokd: "methods provided to the function **via the `{@link Emitter}` instance** should be"

src/main/java/io/reactivex/Flowable.java
```diff
@@ -2210,6 +2210,11 @@ public static int bufferSize() {
...
+     * never concurrently and only while the method is executing. Calling them from multiple threads
```
akarnokd: "while the **function body** is executing."

akarnokd (changes requested):
Thanks, could you rephrase all the sentences based on my suggestions in the comments?

src/main/java/io/reactivex/Flowable.java
punitda: Done.

src/main/java/io/reactivex/Flowable.java
punitda: Done.

src/main/java/io/reactivex/Emitter.java
punitda: Done.

akarnokd (approved):

vanniktech (approved):

Issue #6289, opened 2018-11-05
Javadoc: explain Flowable.groupBy should be consumed on all of its outputs


akarnokd:
In order for `Flowable.groupBy` to remain bounded in memory, it uses a prefetch and consumption-based replenishment strategy: when a group has consumed an item, a new item is requested from the source of `groupBy`.

Since a new item may create a fresh group, the only way to move forward is to have that group consumed as well. For this, the `GroupedFlowable` itself has to reach some operator and get subscribed to.

The typical use is with `flatMap` that will subscribe to the groups and keeps consuming them, however, `flatMap` by default works in a limited concurrency mode. It means as long as the groups themselves don't complete, `flatMap` won't accept a newer group, and thus that group will not get consumed, resulting in a potential livelock. The solution is to have a larger max-concurrency as there are number of groups expected, potentially going unbounded with `flatMap` completely.

This is not an issue with `Observable.groupBy` because there is no backpressure and non-consumed groups can't hinder active groups.

I suggest extending the JavaDocs of all `Flowable.groupBy` about this property, something along the lines of:

```html
<p>
Note that the {@link GroupedFlowable}s should be subscribed to as soon as possible, otherwise, 
the unconsumed groups may starve other groups due to the internal backpressure
coordination of the {@code groupBy} operator. Such hangs can be usually avoided by using 
{@link #flatMap} or {@link #concatMapEager} and overriding the default maximum concurrency 
value to be greater or equal to the expected number of groups, possibly using 
{@link Integer.MAX_VALUE} if the number of expected groups is unknown.
```

(Please when writing the `{@link #}`s out, pick the shortest overload of the respective method that also includes the `maxConcurrency` parameter. Your IDE's content assist should help you in this regard.)

akarnokd:
Ah, this has been done in #6308. Closing.

PR #6308, opened 2018-11-16
Javadoc updates for RXJava Issue 6289

pkamath2:
Updates to io.reactivex.Flowable javadocs as per #6289. 
Updated docs for all 6 Flowable.groupBy(...) methods.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -10346,15 +10346,12 @@ public abstract class Flowable<T> implements Publisher<T> {
      * {@code GroupedPublisher}s that do not concern you. Instead, you can signal to them that they may
      * discard their buffers by applying an operator like {@link #ignoreElements} to them.
      * <p>
-     * By default, a {@link GroupedFlowable} buffers only {@code Flowable.BUFFER_SIZE} items to emit.
-     * So, it is important that a subscription to consume these items begin as soon as the {@link GroupedFlowable}s are created.
-     * If not, it may result in groups upstream not being consumed.
-     * Typically {@code flatMap} or {@code concatMapEager} can be used to ensure all {@link GroupedFlowable} items are consumed.
-     * <p>
-     * Note that, while using {@link #flatMap(Function, int)} or {@link #concatMapEager(Function, int, int)} operators,
-     * we need to ensure that the maximum concurrency parameter is a value greater than
-     * or equal to the number of expected groups, or {@code Integer.MAX_VALUE} if the number of expected groups is unknown.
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/Flowable.java
```diff
@@ -10345,6 +10345,17 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
      * is subscribed to. For this reason, in order to avoid memory leaks, you should not simply ignore those
      * {@code GroupedPublisher}s that do not concern you. Instead, you can signal to them that they may
      * discard their buffers by applying an operator like {@link #ignoreElements} to them.
+     * <p>
+     * By default, a {@link GroupedFlowable} buffers only {@code Flowable.BUFFER_SIZE} items to emit.
```
akarnokd: The wording in these two paragraphs are odd and the terminology is slightly off. Could you please use this (revised) wording from the issue as the only paragraph instead:

```html
<p>
Note that the {@link GroupedFlowable}s should be subscribed to as soon as possible, otherwise, 
the unconsumed groups may starve other groups due to the internal backpressure
coordination of the {@code groupBy} operator. Such hangs can be usually avoided by using 
{@link #flatMap} or {@link #concatMapEager} and overriding the default maximum concurrency 
value to be greater or equal to the expected number of groups, possibly using 
{@link Integer.MAX_VALUE} if the number of expected groups is unknown.
```

src/main/java/io/reactivex/Flowable.java
```diff
@@ -10345,6 +10345,17 @@ public final Disposable forEachWhile(final Predicate<? super T> onNext, final Co
...
+     * So, it is important that a subscription to consume these items begin as soon as the {@link GroupedFlowable}s are created.
+     * If not, it may result in groups upstream not being consumed.
+     * Typically {@code flatMap} or {@code concatMapEager} can be used to ensure all {@link GroupedFlowable} items are consumed.
+     * <p>
+     * Note that, while using {@link #flatMap(Function, int)} or {@link #concatMapEager(Function, int, int)} operators,
+     * we need to ensure that the maximum concurrency parameter is a value greater than
```
akarnokd: > "we need to ensure"

This wording doesn't really fit into a JavaDoc.

akarnokd (changes requested):

src/main/java/io/reactivex/Flowable.java
pkamath2: Noted. Thanks for the tip.

akarnokd (approved):

vanniktech (approved):

Issue #6290, opened 2018-11-05
Javadoc: explain that distinctUntilChanged requires non-mutating data to work as expected


akarnokd:
`distinctUntilChanged` keeps the last value/key (depending on the overload) so it can compare it agains the newer value/key via `Object.equals`. However, if the value/key is mutable and gets mutated in between elements, the operator may not work as expected and filter out the seemingly same data:

```java
PublishSubject<List<Integer>> subject = PublishSubject.create();
List<Integer> list = new ArrayList<Integer>();
list.add(1);

subject.distinctUntilChanged().subscribe(System.out::println);

subject.onNext(list);
// prints [1]

list.add(2);

subject.onNext(list);
// does not print anything, but [1, 2] was expected.
```

In the example, the same reference is passed to `distinctUntilChanged` thus two subsequent items evaluate as same. This mistake by the user is also very common on Android and with text components using a mutable `CharSequence`.

I suggest adding a section to every `distinctUntilChanged` variant explaining the situation in short, something along the lines:

```
<p>
Note that if the element type of the flow is mutable, the comparison of the previous and current 
item may yield unexpected results if the items are mutated externally. Common cases are mutable
{@code CharSequence}s or {@code List}s where subsequent objects are actually the same 
references modified before/after {@code distinctUntilChanged}. It is recommended mutable data is
converted to an immutable one, such as `map(CharSequence::toString)` or 
`map(Collections::unmodifiableList)` for example, to avoid the situation.
```

ashishkrishnan:
Nice. Never actually thought of this case. Very valid. :tada: 

Few questions before the PR. 
1. This should reflect on withLatestForm variants for `Observable` & `Flowable` types right?
2. Additionally, will it be okay to link this issue to the doc for clarity?

akarnokd:
Hi.

1. It's about the `distinctUntilChanged` where this mutation is most obviously affecting a flow. All overloads in both `Observable` and `Flowable` should be updated.
2. We don't link to issues in the main javadoc.

PR #6311, opened 2018-11-16
Javadoc: explain that distinctUntilChanged requires non-mutating data to work as expected

punitda:
- Add note in javadoc for all `distinctUntilChanged()` methods in `Flowable` and `Observable` class explaining about unexpected results to expect when using mutable data sources like **Mutable CharSequence** or **Lists**.

Resolves: #6290


src/main/java/io/reactivex/Flowable.java
```diff
@@ -8740,6 +8740,13 @@ public final Completable concatMapCompletableDelayError(Function<? super T, ? ex
      * <p>
      * Note that the operator always retains the latest item from upstream regardless of the comparison result
      * and uses it in the next comparison with the next upstream item.
+     * <p>
+     * Note that if element {@code T} type in the flow is mutable, the comparison of the previous and current
```
akarnokd: "element type {@code T}"

src/test/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChangedTest.java
```diff
@@ -141,6 +141,84 @@ public void testDistinctUntilChangedOfSourceWithExceptionsFromKeySelector() {
         inOrder.verify(w, never()).onComplete();
     }
 
+    @Test
```
akarnokd: There is no reason to test this unwanted behavior.

akarnokd (changes requested):

src/test/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChangedTest.java
punitda: That's what I thought initially. Will remove all tests.

src/main/java/io/reactivex/Flowable.java
punitda: Oops. Will fix it.

punitda: Done.

src/test/java/io/reactivex/internal/operators/flowable/FlowableDistinctUntilChangedTest.java
punitda: Done.

akarnokd (approved):

vanniktech (approved):

src/main/java/io/reactivex/Flowable.java
```diff
@@ -8740,6 +8740,13 @@ public final Completable concatMapCompletableDelayError(Function<? super T, ? ex
      * <p>
      * Note that the operator always retains the latest item from upstream regardless of the comparison result
      * and uses it in the next comparison with the next upstream item.
+     * <p>
+     * Note that if element type {@code T} in the flow is mutable, the comparison of the previous and current
+     * item may yield unexpected results if the items are mutated externally. Common cases are mutable
+     * {@code CharSequence}s or {@code List}s where the objects will actually have the same
+     * references when they are modified and {@code distinctUntilChanged} will evaluate subsequent items as same.
+     * To avoid such situation, it is recommended that mutable data is converted to an immutable one,
+     * for example using `map(CharSequence::toString)` or `map(Collections::unmodifiableList)`.
```
Kiskae, line 8749: `Collections::unmodifiableList` isn't the right example to use since its a view to the list it is referencing, so if a mutable list is passed then any changes would still reflect the List it returns. Should probably be replaced with `ArrayList::new`

This documentation is linked to by `Collections::unmodifiableList`:

> Note that changes to the backing collection might still be possible, and if they occur, they are visible through the unmodifiable view. Thus, an unmodifiable view collection is not necessarily immutable. However, if the backing collection of an unmodifiable view is effectively immutable, or if the only reference to the backing collection is through an unmodifiable view, the view can be considered effectively immutable.

akarnokd: The best would be `ImmutableList::of` but we can't depend on external collections libraries. 

Perhaps we can have both by `list -> Collections.unmodifiableList(new ArrayList<>(list))`. This will make a copy of the list but also prevent it from being modified later in the pipeline. Unfortunately, we can't do much about the element mutability, which also affects `equals`.

punitda: As discussed, I think we should change `map(Collections::unmodifiableList)` to `map(list -> Collections.unmodifiableList(new ArrayList<>(list))`.  @akarnokd Let me know if we should go forward with that and I'll create a PR.

akarnokd: Yes, please modify the text as my suggestion.

Issue #6309, opened 2018-11-16
Lack of NonNull anotations in factory methods


MikolajKakol:
I was wondering if there shouldn't be an `@javax.annotation.Nonnull` it methods like:
`Single.just`
`Observable.just`
`Observable.fromX`
etc?

It would improve code quality on client side, since IDE might help detect early and compiler would print warning.

akarnokd:
JSR 305 where that annotation lives is incompatible with Java 9 modules so we use our own annotations. Most ide will let you define RxJava's annotations as custom nullability annotations or are supported out of box.

PR welcome.

MikolajKakol:
True, there are RxJava custom annotations and I see that most value for clients rely in `io.reactivex.functions` package. In order to make great improvements we should change 
`    R apply(@NonNull T t) throws Exception;`
to 
`    @NonNull R apply(@NonNull T t) throws Exception;`

Do you think that this is allowed?

akarnokd:
The `functions` are a special case because some usages could work with nulls and they just create warnings. This has been discussed before:

https://github.com/ReactiveX/RxJava/issues?utf8=%E2%9C%93&q=is%3Aissue+nullable+is%3Aclosed+annotation

freakomonk:
Hi. Can I take this up?

akarnokd:
@freakomonk Fine by me. Nobody else seems to be working on this at the moment.

freakomonk:
Sure @akarnokd . Any pointers to know before I start this up?

akarnokd:
Beyond this issue, nothing specifically. We'll review the PR anyway and give you feedback if something needs changing.

PR #6313, opened 2018-11-17
Adding NonNull annotations to Observable & Single

freakomonk:
Adding NonNull annotations in factory methods for improved code compilation

Resolves: #6309

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -118,6 +118,7 @@ public abstract class Completable implements CompletableSource {
      * @throws NullPointerException if sources is null
      */
     @CheckReturnValue
+    @NonNull
     @SchedulerSupport(SchedulerSupport.NONE)
     public static Completable ambArray(final CompletableSource... sources) {
         ObjectHelper.requireNonNull(sources, "sources is null");
@@ -146,6 +147,7 @@ public abstract class Completable implements CompletableSource {
      * @throws NullPointerException if sources is null
      */
     @CheckReturnValue
+    @NonNull
     @SchedulerSupport(SchedulerSupport.NONE)
     public static Completable amb(final Iterable<? extends CompletableSource> sources) {
         ObjectHelper.requireNonNull(sources, "sources is null");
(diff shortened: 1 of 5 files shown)


freakomonk:
#6309 @akarnokd Can you pls take a look at this and review ?

akarnokd:
Thanks for Contributing. A few points:
- You don't have to keep the welcome text of the PR, that's for telling contributors what to do. See the updated text for future reference.
- Didn't you mean to add `@NonNull` to the arguments rather than to the return value?

vanniktech (commented):
I like it. Kotlin currently warns about those since it's not capable of inferring whether they can be null or not. 

Annotation order should be the same for all usages. This will mix this up.

Can't we use a nonnull by default kind of annotation since 99 percent of the API only uses non null?

akarnokd:
Non-null by default came up in #5341 but not sure what happened. I'd think it was a JSR 305 annotation instead of just a convention thus the IDEs simply wouldn't have supported them.

@freakomonk Did you also want to add these to `Maybe`, `Flowable` and `Completable` static methods?

freakomonk:
@vanniktech Can you elaborate `nonNull by default` ?

freakomonk:
@akarnokd Thanks for the suggestions. Will keep in mind henceforth. 

I added the `@NonNull` annotation to methods with ` ObjectHelper.requireNonNull` check in them. As the input params are being checked for `null`, I am checking for the return value.

I have updated the `Flowable` static methods but do not see any `Maybe`, `Completable` that do not have already have this annotation.

vanniktech:
> Non-null by default came up in #5341 but not sure what happened. I'd think it was a JSR 305 annotation instead of just a convention thus the IDEs simply wouldn't have supported them.

This @freakomonk. I'm not sure though which annotations works where and what needs to be enabled/disabled or not. Either we look into this or we just use RxJava's NonNull annotation which does the job.

freakomonk:
@akarnokd  Hi David. Is there something else to do to get this PR merged ?

akarnokd:
@freakomonk I was waiting for you to add them to the other reactive base classes.

freakomonk:
@akarnokd Added annotation to further methods now. Can you pls check ?

freakomonk:
@akarnokd Is this now okay to be merged ? Your review pls

akarnokd (approved):

PR #6348, opened 2018-12-19
Add peek operator like java 8 streams

philberty:
I like the peek operator in Java 8 streams to provide a logging interface for streams or to apply to some static or global buffer etc.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -9712,6 +9712,21 @@ public abstract class Observable<T> implements ObservableSource<T> {
     }
 
     /**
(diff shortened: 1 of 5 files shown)


JakeWharton:
How is this different than the doOn* operators?

philberty:
Yeah good point i just find this simpler and it enforces final on each item.

akarnokd:
This operator already exist as `doOnNext`. Besides, looks like [final](https://stackoverflow.com/a/5380309/61158) is only cosmetic in this case.

Also please open a discussion before posting a PR with new a operator the next time so that unnecessary work can be avoided by you and us maintainers.

Issue #6331, opened 2018-12-06
Outdated java example in wiki 


hamedhr:
Outdated java example in [wiki](https://github.com/ReactiveX/RxJava/wiki/How-To-Use-RxJava) page :

```
public static void hello(String... names) {
    Observable.from(names).subscribe(new Action1<String>() {

        @Override
        public void call(String s) {
            System.out.println("Hello " + s + "!");
        }

    });
}
```

Change it to 

```
public static void hello(String... names) {
        Observable.fromArray(names).subscribe(s -> System.out.println("Hello " + s));
    }
```

Cuiernao:
java8......

ozcanyus:
Hello, I've created a pr about this issue.

PR #6351, opened 2018-12-27
Update outdated java example in wiki #6331

ozcanyus:
Updated java example in docs/How-To-Use-RxJava.md file with java 8 version.

Resolves: #6331

### DIFF
--- a/docs/How-To-Use-RxJava.md
+++ b/docs/How-To-Use-RxJava.md
@@ -11,15 +11,21 @@ You can find additional code examples in the `/src/examples` folders of each [la
 ### Java
 
 ```java
-public static void hello(String... names) {
-    Observable.fromArray(names).subscribe(s -> System.out.println("Hello " + s));
+public static void main(String[] args) {
+    Flowable.just("Hello world").subscribe(System.out::println);
 }
 ```
+If your platform doesn't support Java 8 lambdas (yet), you have to create an inner class of ```Consumer``` manually:
 
 ```java
-hello("Ben", "George");
-Hello Ben!
-Hello George!
+public static void main(String[] args) {
+    Flowable.just("Hello world")
+        .subscribe(new Consumer<String>() {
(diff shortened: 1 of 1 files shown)


docs/How-To-Use-RxJava.md
```diff
@@ -12,14 +12,7 @@ You can find additional code examples in the `/src/examples` folders of each [la
...
 public static void hello(String... names) {
-    Observable.from(names).subscribe(new Action1<String>() {
-
-        @Override
-        public void call(String s) {
-            System.out.println("Hello " + s + "!");
-        }
-
-    });
+    Observable.fromArray(names).subscribe(s -> System.out.println("Hello " + s));
```
akarnokd: I'd instead show a `Consumer<String>` and the lambda version, similar to the [hello world](https://github.com/ReactiveX/RxJava#hello-world) example.

ozcanyus: Updated similar to hello world example.

docs/How-To-Use-RxJava.md
```diff
@@ -11,22 +11,21 @@ You can find additional code examples in the `/src/examples` folders of each [la
...
-public static void hello(String... names) {
-    Observable.from(names).subscribe(new Action1<String>() {
-
-        @Override
-        public void call(String s) {
-            System.out.println("Hello " + s + "!");
-        }
-
-    });
+public static void main(String[] args) {
```
akarnokd: No, I meant show codes as anonym inner class and lambda, but keep the essence of the original example, i.e. fromArray

akarnokd (changes requested):

docs/How-To-Use-RxJava.md
ozcanyus: Oh ok, i'll fix.

akarnokd (approved):

vanniktech (approved):

PR #6356, opened 2019-01-04
2.x: Improve TestScheduler Javadoc

### LINKED ISSUE #6355: 2.x: parallel blocks forever with TestScheduler
io.reactivex.rxjava2:rxjava:2.2.3

 ```java
public class DemoApplication {
        List<Map.Entry<Integer, String>> entries = Arrays.asList(
                new AbstractMap.SimpleImmutableEntry<>(3, "hello"),
                new AbstractMap.SimpleImmutableEntry<>(3, "world")
        );

        Flowable.fromIterable(entries)
                .flatMap(entry -> Flowable.range(1, entry.getKey())
                        .parallel()
                        .runOn(Schedulers.io())
                        .flatMap(i -> logIt(entry.getValue()).toFlowable())
                        .sequential()
                )
                .all(success -> success)
                .doOnSuccess(success -> System.out.println("good"))
                .doOnError(Throwable::printStackTrace)
                .blockingGet();
    }

    private static Single<Boolean> logIt(String s) {
        System.out.println(s);

        return Single.just(true);
    }
}
```

This code works as expected. However, replacing `Schedulers.io()` with `TestScheduler` blocks forever. Gradle project attached.

[demo.zip](https://github.com/ReactiveX/RxJava/files/2722335/demo.zip)

### PR DESCRIPTION
**Pull request by asarkar**
Resolves #6355 by improving TestScheduler Javadoc

### DIFF
--- a/src/main/java/io/reactivex/schedulers/TestScheduler.java
+++ b/src/main/java/io/reactivex/schedulers/TestScheduler.java
@@ -23,9 +23,60 @@ import io.reactivex.internal.disposables.EmptyDisposable;
 import io.reactivex.internal.functions.ObjectHelper;
 
 /**
- * A special, non thread-safe scheduler for testing operators that require
- * a scheduler without introducing real concurrency and allows manually advancing
- * a virtual time.
+ * This special {@link Scheduler} allows testing asynchronous and/or timed operators
+ * and flows by programatically executing the scheduled tasks up to a certain virtual
+ * time, allowing checking for intermediate results, state changes or preparing the relevant
+ * context (global/captured object state) before executing the next set of tasks.
+ * <p>
+ * Tasks scheduled via {@link #scheduleDirect(Runnable)} or
+ * {@link Scheduler.Worker#schedule(Runnable)},
+ * with or without delay, won't be executed until the virtual time is moved forward
+ * via the {@link #advanceTimeBy} or {@link #advanceTimeTo} methods of this
+ * {@code TestScheduler}. Tasks scheduled without delay can be executed without
+ * moving the virtual time forward via the {@link #triggerActions} method.
+ * Note that these methods are not thread safe and only one thread should be
+ * allowed to call any of them at a time. Note also that these methods will only
+ * return control to the caller once all tasks scheduled up to the target virtual
+ * time have been executed.
+ * <p>
+ * All the standard {@code Scheduler} and {@link Scheduler.Worker} methods are thread safe.
+ * <p>
+ * Example:
+ * <pre><code>
+ * TestScheduler sch = new TestScheduler();
+ * AtomicLong offset = new AtomicLong();
+ *
+ * TestObserver<Long> to =
+ * Observable.interval(1, TimeUnit.HOURS, sch)
+ * .map(time -> offset.get() + time)
+ * .test();
+ *
+ * // the virtual time hasn't moved, no data should be present
+ * to.assertEmpty();
+ *
+ * // move time forward two hours
+ * // this won't actually wait two hours
+ * sch.advanceTimeBy(2, TimeUnit.HOURS);
+ *
+ * // we should have two values
+ * to.assertValues(0L, 1L);
+ *
+ * // change the offset value so future intervals are shifted
+ * offset.set(10L);
+ *
+ * // the scheduler has not moved its time, no new
+ * // values should emerge
+ * to.assertValues(0L, 1L);
+ *
+ * // move the time ahead again
+ * sch.advanceTimeBy(3, TimeUnit.HOURS);
+ *
+ * // more values should be received
+ * to.assertValues(0L, 1L, 12L, 13L, 14L);
+ *
+ * // we don't need further values, cleanup
+ * to.dispose();
+ * </code></pre>
  */
 public final class TestScheduler extends Scheduler {
     /** The ordered queue for the runnable tasks. */

### CODE REVIEW

**Review by akarnokd — CHANGES_REQUESTED** [review]
There are a couple of stylistic errors in this:

- worded like a blog post or casual talk instead of forward explanatory text. I.e. write it as "this is how it works" instead of "this doesn't work btw"
- uses a parallel flow as an example which is not beginner friendly
- code example should be in pre+code
- mentions of methods should be links

**Comment by asarkar** [comment]
@akarnokd The methods mentioned are present in `Observable`, `Flowable`, `Single` etc. Which ones would you like to link to?

**Comment by akarnokd** [comment]
None. Link to to `TestScheduler`'s own methods, such as `{@link #advanceTimeBy}`.

**Comment by asarkar** [comment]
Updated.

**Comment by akarnokd** [comment]
More like this:

```
<p>
Tasks scheduled via {@link #scheduleDirect(Runnable)} or 
{@link Scheduler.Worker#schedule(Runnable)}, 
with or without delay, won't be executed until the virtual time is moved forward 
via the {@link #advanceTimeBy} or {@link #advanceTimeTo} methods of this 
{@code TestScheduler}. Tasks scheduled without delay can be executed without
moving the virtual time forward via the {@link #triggerActions} method.
Note that these methods are not thread safe and only one thread is
allowed to call any of them at a time. Note also that these methods will only
return control to the caller once all tasks scheduled up to the target virtual
time have been executed.
<p>
All the standard {@code Scheduler} and {@link Scheduler.Worker} methods are thread safe.
<p>
This special {@link Scheduler} allows testing asynchronous and/or timed operators
and flows by programatically executing the scheduled tasks up to a certain virtual
time, allowing checking for intermediate results, state changes or preparing the relevant
context (global/captured object state) before executing the next set of tasks.
<p>
Example:
<pre><code>
TestScheduler sch = new TestScheduler();
AtomicLong offset = new AtomicLong();

TestObserver<Long> to = 
    Observable.interval(1, TimeUnit.HOURS, sch)
    .map(time -> offset.get() + time)
    .test();

// the virtual time hasn't moved, no data should be present
to.assertEmpty();

// move time forward two hours
// this won't actually wait two hours
sch.advanceTimeBy(2, TimeUnit.HOURS);

// we should have two values
to.assertValues(0L, 1L);

// change the offset value so future intervals are shifted
offset.set(10L);

// the scheduler has not moved its time, no new
// values should emerge
to.assertValues(0L, 1L);

// no move the time ahead again
sch.advanceTimeBy(3, TimeUnit.HOURS);

// more and offset values should be received
to.assertValues(0L, 1L, 12L, 13L, 14L);

// we don't need further values, cleanup
to.dispose();
</code></pre>
```

**Comment by asarkar** [comment]
Updated as per review comments.

**Review by akarnokd — APPROVED** [review]

File: src/main/java/io/reactivex/schedulers/TestScheduler.java [hunk 1]
```diff
@@ -23,9 +23,60 @@
...
 
 /**
- * A special, non thread-safe scheduler for testing operators that require
- * a scheduler without introducing real concurrency and allows manually advancing
- * a virtual time.
+ * This special {@link Scheduler} allows testing asynchronous and/or timed operators
+ * and flows by programatically executing the scheduled tasks up to a certain virtual
+ * time, allowing checking for intermediate results, state changes or preparing the relevant
+ * context (global/captured object state) before executing the next set of tasks.
+ * <p><p>
```
artem-zinnatullin: why `<p>` twice?

src/main/java/io/reactivex/schedulers/TestScheduler.java
```diff
@@ -23,9 +23,60 @@
...
+ * Tasks scheduled via {@link #scheduleDirect(Runnable)} or
+ * {@link Scheduler.Worker#schedule(Runnable)},
+ * with or without delay, won't be executed until the virtual time is moved forward
+ * via the {@link #advanceTimeBy} or {@link #advanceTimeTo} methods of this
+ * {@code TestScheduler}. Tasks scheduled without delay can be executed without
+ * moving the virtual time forward via the {@link #triggerActions} method.
+ * Note that these methods are not thread safe and only one thread should is
```
artem-zinnatullin: "should is allowed" doesn't seem to be grammatically correct

artem-zinnatullin (approved):
few small comments, otherwise LGTM

src/main/java/io/reactivex/schedulers/TestScheduler.java
asarkar: Seems like we don't get a blank line otherwise.

src/main/java/io/reactivex/schedulers/TestScheduler.java
asarkar: Fixed.

src/main/java/io/reactivex/schedulers/TestScheduler.java
```diff
@@ -23,9 +23,60 @@
...
+ * time, allowing checking for intermediate results, state changes or preparing the relevant
+ * context (global/captured object state) before executing the next set of tasks.
+ * <p>
+ * Tasks scheduled via {@link #scheduleDirect(Runnable)} or
+ * {@link Scheduler.Worker#schedule(Runnable)},
+ * with or without delay, won't be executed until the virtual time is moved forward
+ * via the {@link #advanceTimeBy} or {@link #advanceTimeTo} methods of this
+ * {@code TestScheduler}. Tasks scheduled without delay can be executed without
+ * moving the virtual time forward via the {@link #triggerActions} method.
+ * Note that these methods are not thread safe and only one thread should be
```
akarnokd, line 37: Please change it to `only one thread should call`, remove `be allowed to`.

PR #6357, opened 2019-01-04
2.x: Disposable already set! using delaySubscription with takeUntil/repeatWhen

eleventigers:
When using a combination of `takeUntil`/`repeatWhen` operators on an `Observable` with `delaySubscription` operator applied, it is possible to encounter a `ProtocolViolationException: Disposable already set!` thrown by `RepeatWhenObserver#onSubscribe`. It appears that if `delaySubscription` uses an `ObservableSource` which rapidly/concurrently emits multiple notifications a rare race condition leads to a double subscription issue. 

I am not sure if this issue is a user error (`take(1)` on `ObservableSource` supplied to `delaySubscription` alleviates it) or a bug in RxJava hence me opening this PR with a slow test case that demonstrates this issue.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/observable/ObservableDelayTest.java
+++ b/src/test/java/io/reactivex/internal/operators/observable/ObservableDelayTest.java
@@ -33,6 +33,7 @@ import io.reactivex.internal.functions.Functions;
 import io.reactivex.observers.*;
 import io.reactivex.schedulers.*;
 import io.reactivex.subjects.PublishSubject;
+import io.reactivex.subjects.Subject;
 
 public class ObservableDelayTest {
     private Observer<Long> observer;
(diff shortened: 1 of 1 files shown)


akarnokd:
Thanks for reporting a possible bug, but why haven't you just posted an issue with the unit test that fails? Posting a full PR without an actual resolution is not useful for us maintainers.

eleventigers:
@akarnokd no worries, I will re-open this as an issue.

eleventigers:
Tracking issue: https://github.com/ReactiveX/RxJava/issues/6358

vanniktech:
> Thanks for reporting a possible bug, but why haven't you just posted an issue with the unit test that fails? Posting a full PR without an actual resolution is not useful for us maintainers.

We might want to note that somewhere. Many other repositories/organizations like this.

PR #6359, opened 2019-01-04
2.x: Fix the error/race in Observable.repeatWhen due to flooding repeat signal

### LINKED ISSUE #6358: 2.x: Disposable already set! using delaySubscription with takeUntil/repeatWhen
Issue encountered in: io.reactivex.rxjava2:rxjava:2.2.4

When using a combination of takeUntil/repeatWhen operators on an Observable with delaySubscription operator applied, it is possible to encounter a ProtocolViolationException: Disposable already set! thrown by RepeatWhenObserver#onSubscribe. It appears that if delaySubscription uses an ObservableSource which rapidly/concurrently emits multiple notifications a rare race condition leads to a double subscription issue.

I am not sure if this issue is a user error (take(1) on ObservableSource supplied to delaySubscription alleviates it) or a bug in RxJava however the following, slow unit test demonstrates it:

```java

    @Test
    public void delaySubscriptionWithRepeatWhen() {
        List<Throwable> errors = TestHelper.trackPluginErrors();

        ExecutorService executor = Executors.newFixedThreadPool(2);
        try {
            Observable<Object> subscriptionSignal = Observable.merge(
                    Observable.just(new Object()).delay(0, TimeUnit.MILLISECONDS),
                    Observable.just(new Object()).delay(0, TimeUnit.MILLISECONDS)
            );

            final Subject<Boolean> boolStream = PublishSubject.<Boolean>create().toSerialized();

            Observable<Boolean> brokenStream = boolStream
                    .delaySubscription(subscriptionSignal)
                    .takeUntil(new Predicate<Boolean>()  {
                        @Override
                        public boolean test(Boolean state) throws Exception {
                            return state;
                        }
                    })
                    .repeatWhen(new Function<Observable<Object>, ObservableSource<?>>() {
                        @Override
                        public ObservableSource<?> apply(Observable<Object> objectObservable) throws Exception {
                            return boolStream.filter(new Predicate<Boolean>() {
                                @Override
                                public boolean test(Boolean state) throws Exception {
                                    return !state;
                                }
                            });
                        }
                    });

            final Random random = new Random();
            for (int i = 0; i < 1000000; i++) {
                final TestObserver<Boolean> testObserver = brokenStream.test();
                executor.execute(new Runnable() {
                    @Override
                    public void run() {
                        boolStream.onNext(random.nextBoolean());
                        testObserver
                                .assertNoErrors()
                                .dispose();
                    }
                });
                if (!errors.isEmpty()) {
                    fail("Uncaught fatal errors: " + errors);
                }
            }
        } finally {
            executor.shutdown();
            try {
                executor.awaitTermination(30, TimeUnit.SECONDS);
            }
        }
    }
```

akarnokd:
This PR fixes a race condition in the `ObservableRepeatWhen` operator for the case when the repeat signal is not 1-for-1 and a new subscription may race with the clearing of the previous disposable because `active` is set to `false` too early. The fix is to swap the two operations. 

`Observable.retryWhen` and the `Flowable` versions do not have this bug. Unit tests were added to all 4 operators to verify the correct behavior regardless.

Fixes: #6358

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableRepeatWhen.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableRepeatWhen.java
@@ -108,8 +108,8 @@ public final class ObservableRepeatWhen<T> extends AbstractObservableWithUpstrea
 
         @Override
         public void onComplete() {
-            active = false;
             DisposableHelper.replace(upstream, null);
+            active = false;
             signaller.onNext(0);
         }
 
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableRepeatTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableRepeatTest.java
@@ -29,6 +29,7 @@ import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.TestException;
 import io.reactivex.functions.*;
 import io.reactivex.internal.subscriptions.BooleanSubscription;
+import io.reactivex.plugins.RxJavaPlugins;
 import io.reactivex.processors.PublishProcessor;
 import io.reactivex.schedulers.Schedulers;
 import io.reactivex.subscribers.TestSubscriber;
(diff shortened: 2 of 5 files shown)


eleventigers:
Nice! Thanks @akarnokd for such a quick fix, I have verified that it fixes https://github.com/ReactiveX/RxJava/issues/6358 locally.

vanniktech (approved):

PR #6362, opened 2019-01-12
2.x: Fix Completable.andThen(Completable) not running on observeOn scheduler.

### LINKED ISSUE #6354: 2.2.5: observeOn might not work for Completable.andThen(Completable)
- RxJava 2.2.5

It seems that Completable passed to `Completable.andThen(...)` is not run on scheduler passed to observeOn. So
```
        Completable.fromAction(() -> System.out.println("subscribing in thread " + Thread.currentThread().getName()))
                .observeOn(Schedulers.io())
                .andThen(Completable.fromAction(() -> System.out.println("and then in thread " + Thread.currentThread().getName())))
                .subscribe(() -> System.out.println("observing in thread " + Thread.currentThread().getName()));
```
might produce
```
subscribing in thread main
and then in thread main
observing in thread main
```
I don't know if it's by design or not, but it seems to be caused by 'race condition' in `CompletableConcatArray` where we have do {...} while() loop, when next() method is run second time from io() thread and bumps up the atomic counter, and then second iteration of do/while loop starts, and executes completable from andThen operator from main thread. It can be easily reproduced if you add some work to that loop before checking condition in while clause so in `CompletableConcatArray`:
```
                a[idx].subscribe(this);
                // spin the cpu for a while to make ObserveOnCompletable run this next() method
                int i = 0, j = 0;
                while (i < 1000_000) {
                    i++;
                    while (j < 1000_000) {
                        j++;
                    }
                }
            } while (decrementAndGet() != 0);
```

pjastrz:
This PR fixes unpredictable scheduler/thread used by `Completable.observeOn(...)andThen(Completable)`, since sometimes scheduler defined in observeOn(...) was not used. It was inconsistent with similar `.andThen(Maybe)` and `.andThen(Single)`, since those are guaranteed to run second Completable using proper scheduler if defined with `.observeOn`.
Fixes: https://github.com/ReactiveX/RxJava/issues/6354.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenCompletableabTest.java
+++ b/src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenCompletableabTest.java
@@ -19,6 +19,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.Test;
 
 import io.reactivex.*;
+import io.reactivex.exceptions.TestException;
 import io.reactivex.functions.Action;
 import io.reactivex.observers.TestObserver;
 import io.reactivex.schedulers.Schedulers;
@@ -43,11 +44,11 @@ public class CompletableAndThenCompletableabTest {
     @Test
     public void andThenCompletableCompleteError() {
         Completable.complete()
-                .andThen(Completable.error(new RuntimeException("test")))
+                .andThen(Completable.error(new TestException("test")))
                 .test()
                 .assertNotComplete()
                 .assertNoValues()
-                .assertError(RuntimeException.class)
+                .assertError(TestException.class)
                 .assertErrorMessage("test");
     }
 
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenCompletableabTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Completable.complete()
+                .andThen(Completable.complete())
+                .test()
+                .assertComplete();
+    }
+
+    @Test
+    public void andThenCompletableCompleteError() {
+        Completable.complete()
+                .andThen(Completable.error(new RuntimeException("test")))
```
akarnokd: Please use `TestException` so we don't mistake them with any other `RuntimeException` failures.

src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
```diff
@@ -69,39 +63,4 @@ public void andThenMaybeError() {
             .assertError(RuntimeException.class)
             .assertErrorMessage("bla");
     }
-
-    @Test
-    public void andThenNoInterrupt() throws InterruptedException {
```
akarnokd, line 74: Why was this test removed without explanation?

akarnokd (changes requested):

src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenTest.java
pjastrz: I moved it to CompletableAndThenCompletableTest since it specifically tested `.andThen(Completable)`.

src/test/java/io/reactivex/internal/operators/completable/CompletableAndThenCompletableabTest.java
pjastrz: Ok, done.

akarnokd (approved):

vanniktech (approved):

Issue #6363, opened 2019-01-14
2.x: Deadlock when adding doOnSubscribe to Flowable


Whathecode:
For some time, I [have been investigating a deadlock that occurs in my codebase](https://stackoverflow.com/q/54116679/590790), seemingly after the default backbuffer size has been filled. While trying to create a repro, I ran into a similar issue which shows the same symptoms with a minimum amount of code. I see no obvious cause, and therefore believe this might be a potential RxJava 2.x bug (reproducible so far on 2.2.0 and 2.2.5).

```
fun main( args: Array<String> )
{
    // Override the default buffer size of 128 to 10 for easier testing.
    System.setProperty( "rx2.buffer-size", "10" )

    val numbers = Flowable
        .range( 0, Int.MAX_VALUE )
        // Commenting out '.doOnSubscribe' here prevents the deadlock.
        .doOnSubscribe { println( "Subscribed." ) }
        .share()
    val evenNumbers = numbers
        .filter { number -> number % 2 == 0 }
    val oddNumbers = numbers
        .filter { number -> number % 2 != 0 }
    val getNextOdd = oddNumbers.first( 0 )
    evenNumbers
        .concatMap { even ->
            Single.zip(
                Single.just( even ), getNextOdd,
                BiFunction { a: Int, b: Int -> Pair( a, b ) }
            ).toFlowable()
        }
        .takeWhile { it.first < 1000 }
        .doOnComplete { println( "Done." ) }
        .subscribe { println( it ) }

    readLine()
}
```
This halts depending on the backbuffer size. In the code example at:

> Subscribed.
> (0, 1)
> (2, 3)
> (4, 5)
> (6, 7)
> (8, 9)

As commented in the above code, commenting out the `doOnSubscribe` line will prevent the deadlock from occurring.

Is this a bug, or am I doing something inherently wrong? E.g., I imagine things get complicated the way I reuse the same stream to 'zip', but this is the use case which also is needed in my actual codebase. Regardless, I do not understand what `doOnSubscribe` would have to do with anything here.

akarnokd:
Hi and thanks for reporting. This is a bug in the `publish` operator (part of `share`) not requesting more if there was a change in the subscriber array (due to `first` in your example). The reason it didn't happen without `doOnSubscribe` is that `range` and `publish` operators fuse together when adjacent to each other and there is no request management involved there as an optimization. Preventing this fusion resulted in the request mismanagement bug. I'll post a fix shortly.

[...]

PR #6364, opened 2019-01-14
2.x: Fix publish not requesting upon client change

akarnokd:
Due to a bug in the `Flowable.publish` operator, it is not requesting more if there was a change in the subscriber array during an emission run in the non-fused operation mode.

Fixes: #6363

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
@@ -557,12 +557,20 @@ public final class FlowablePublish<T> extends ConnectableFlowable<T> implements 
                         InnerSubscriber<T>[] freshArray = subscribers.get();
                         if (subscribersChanged || freshArray != ps) {
                             ps = freshArray;
+
+                            // if we did emit at least one element, request more to replenish the queue
+                            if (d != 0) {
+                                if (sourceMode != QueueSubscription.SYNC) {
+                                    upstream.get().request(d);
+                                }
+                            }
+
                             continue outer;
                         }
                     }
 
                     // if we did emit at least one element, request more to replenish the queue
-                    if (d > 0) {
+                    if (d != 0) {
                         if (sourceMode != QueueSubscription.SYNC) {
                             upstream.get().request(d);
                         }
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
@@ -1368,4 +1368,73 @@ public class FlowablePublishTest {
     public void badRequest() {
         TestHelper.assertBadRequestReported(Flowable.range(1, 5).publish());
     }
+
+    @Test
+    @SuppressWarnings("unchecked")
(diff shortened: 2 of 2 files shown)


src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
```diff
@@ -557,12 +557,20 @@ void dispatch() {
...
+                                if (sourceMode != QueueSubscription.SYNC) {
+                                    upstream.get().request(d);
+                                }
+                            }
+
                             continue outer;
                         }
                     }
 
                     // if we did emit at least one element, request more to replenish the queue
```
vanniktech, line 572: Does not this comment need an update?

vanniktech (approved):

src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
akarnokd: No, this still applies if no change to the subscriber array happened.

src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
```diff
@@ -557,12 +557,20 @@ void dispatch() {
...
-                    if (d > 0) {
+                    if (d != 0) {
```
artem-zinnatullin, line 573: Sorry I don't understand this particular change, how can we end up with negative `d`?

Otherwise it's same as previously used condition 🤔

src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
```diff
@@ -1368,4 +1368,73 @@ public String apply(Integer t) throws Exception {
     public void badRequest() {
         TestHelper.assertBadRequestReported(Flowable.range(1, 5).publish());
     }
+
+    @Test
```
artem-zinnatullin, line 1372: nit: add link to original bug? https://github.com/ReactiveX/RxJava/issues/6363

artem-zinnatullin (commented):
Overall it looks good, but comment

src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
akarnokd: `d` should be never negative, thus `!=` is just fine.

src/test/java/io/reactivex/internal/operators/flowable/FlowablePublishTest.java
akarnokd: I don't usually include the issue number, that's what GitHub is for.

src/main/java/io/reactivex/internal/operators/flowable/FlowablePublish.java
artem-zinnatullin: but then `d > 0` is okay too, or is this a performance optimization?

akarnokd: It is slighly faster.

vanniktech (approved):

PR #6366, opened 2019-01-15
Fixing Javadoc duplicate with inheritDoc

oumaziz:
I'm a PhD Student from University of Bordeaux working on documentation copy-pastes (duplicates).
I've noticed that your project does have a few documentation duplicates.
For my research work, i've tried to fix this duplicate and i'd love to have your feedback about it:

Do you think that this duplicate documentation should co-evolve forever?
Are you aware of copy-pastes in your documentation?
If yes, how do you handle them?
Were you aware of JavaDoc reuse features such as {@inheritdoc}?
Would you like to have a tool that helps you maintain duplicate documentation?

### DIFF
--- a/src/main/java/io/reactivex/disposables/Disposable.java
+++ b/src/main/java/io/reactivex/disposables/Disposable.java
@@ -22,8 +22,8 @@ public interface Disposable {
     void dispose();
 
     /**
-     * Returns true if this resource has been disposed.
-     * @return true if this resource has been disposed
+     * Returns true if this resource has been disposed/cancelled.
+     * @return true if this resource has been disposed/cancelled
      */
     boolean isDisposed();
 }
--- a/src/main/java/io/reactivex/observers/ResourceCompletableObserver.java
+++ b/src/main/java/io/reactivex/observers/ResourceCompletableObserver.java
@@ -122,8 +122,7 @@ public abstract class ResourceCompletableObserver implements CompletableObserver
     }
 
     /**
-     * Returns true if this ResourceObserver has been disposed/cancelled.
-     * @return true if this ResourceObserver has been disposed/cancelled
+     * {@inheritDoc}
      */
     @Override
     public final boolean isDisposed() {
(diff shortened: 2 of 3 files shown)


src/main/java/io/reactivex/disposables/Disposable.java
```diff
@@ -22,8 +22,8 @@
     void dispose();
 
     /**
-     * Returns true if this resource has been disposed.
-     * @return true if this resource has been disposed
+     * Returns true if this resource has been disposed/cancelled.
```
akarnokd, line 25: Adding `cancelled` here is semantically incorrect in respect of `Disposable`.

src/main/java/io/reactivex/observers/ResourceCompletableObserver.java
```diff
@@ -122,8 +122,7 @@ public final void dispose() {
     }
 
     /**
-     * Returns true if this ResourceObserver has been disposed/cancelled.
-     * @return true if this ResourceObserver has been disposed/cancelled
+     * {@inheritDoc}
```
akarnokd, line 125: If we leave the comment out, the JavaDoc/IDE will happily use the parent docs, such as this: http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/observers/ResourceObserver.html#onSubscribe-io.reactivex.disposables.Disposable-

akarnokd:
> Do you think that this duplicate documentation should co-evolve forever?

No. See the comments in the changes view.

> Are you aware of copy-pastes in your documentation?

Yes. There are often slight modifications and there is no mechanism to share parts of the docs.

> If yes, how do you handle them?

If we find a mistake or get a report, we change the docs. It is not a high priority for us. Also we don't have many that can write the docs.

> Were you aware of JavaDoc reuse features such as {@inheritdoc}?

Yes.

> Would you like to have a tool that helps you maintain duplicate documentation?

IntelliJ can do this right via analysis tools?

oumaziz:
Thank you for your feedback!

PR #6371, opened 2019-01-16
2.x: Fix bounded replay() memory leak due to bad node retention

akarnokd:
This PR fixes a memory leak caused by node retention in the bounded `replay()` implementations. When the consumer disposed/canceled the flow, the `index` field was not cleared. In certain situations, such out-of-comission consumers were still reachable via thread local stack and through it, the node they were last refencing. Further items through it would then keep referencing an ever increasing number of linked nodes, causing OOME eventually.

The fix is to clear the `index` field upon dispose/cancellation. The subject/processor variants were working properly. Tests were added to verify them though as well.

Issue discovered on StackOverflow: https://stackoverflow.com/q/54191190/61158


vanniktech (approved):

artur-jablonski:
Thank you for fixing this! You guys rock!

PR #6380, opened 2019-01-21
Fix bug in CompositeException.getRootCause

guillermocalvo:
I found what I believe is a bug in `CompositeException.getRootCause`.

The [original code](https://github.com/ReactiveX/RxJava/commit/487a0ba52137e13d996ad12fc73bfeabb03c4fb7#diff-63dcdf1ecfc9f44d938e86154f465e41R306) use to be `if (root == null || root == e)`, but apparently after some refactoring it ended up as `if (root == null || cause == root)`.

As a side note, I think this method should be made `static` (that would have prevented the bug).

### DIFF
--- a/src/main/java/io/reactivex/exceptions/CompositeException.java
+++ b/src/main/java/io/reactivex/exceptions/CompositeException.java
@@ -280,7 +280,7 @@ public final class CompositeException extends RuntimeException {
      */
     /*private */Throwable getRootCause(Throwable e) {
         Throwable root = e.getCause();
-        if (root == null || cause == root) {
+        if (root == null || e == root) {
             return e;
         }
         while (true) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/exceptions/CompositeException.java
```diff
@@ -280,7 +280,7 @@ public int size() {
      */
     /*private */Throwable getRootCause(Throwable e) {
```
artem-zinnatullin, line 281: I think you should go ahead and make it static

artem-zinnatullin (approved):
Good catch 👍

akarnokd (approved):

vanniktech (approved):

src/main/java/io/reactivex/exceptions/CompositeException.java
guillermocalvo: Oops, already merged.

akarnokd: Just post a new PR.

PR #6383, opened 2019-01-24
Add globalTimeout operation on Flowable.

newkek:
Hi, 

While working on a project where we needed to set a finite timeout bound to a flowable and error out if the timeout was reached, I wasn't able to find an operation that would be able to do that in the Flowable class. Only `.timeout()` is available but it is to specify a timeout between each item, not the global Flowable.

After some help from the mailing list I ended up using `originalFlowable.takeUntil(Flowable.never().timeout())` which works very well for the intended purpose (shout-out to David from the mailing list who came up with it).
I think it would be valuable to other users who aren't aware of the `takeUntil(Publisher)` method, to have this operation available in the library. Please let me know if you think it would be useful as well.

I've included a simple unit test as well as Javadocs as best as I could, however I didn't know how to generate a marble schema...

Thanks.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -2409,6 +2409,56 @@ public abstract class Flowable<T> implements Publisher<T> {
     }
 
     /**
+     * Returns a Flowable that emits items until a global timeout time has passed after
+     * subscription.
+     * <p>
+     * If the Flowable is not completed within the defined time frame, the resulting Flowable
+     * terminates and the observer(s) will be notified with a {@link TimeoutException}.
+     * <dl>
+     *  <dt><b>Scheduler:</b></dt>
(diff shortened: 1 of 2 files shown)


akarnokd:
> shout-out to David from the mailing list who came up with it

And the same David will close this PR for two reasons: 1) the functionality can be achieved with a small set of existing operators, and 2) please open an issue to discuss feature requests first so that time is not wasted on either end.

newkek:
Ha - didn't realize @akarnokd was the David from the ML... 🤦‍♂️ 

Fair enough, although I thought this would be ok considering I've seen other methods in the Flowable class that are simple shortcuts to other operations, example: `Flowable.merge()`, `Flowable.mergeArray()`, `startWith()`, etc.. does that make sense?

Thanks for the quick reply. On both the ML and here.

PR #6392, opened 2019-01-31
Fix ignored tests due to internal structure changes of `ExecutorSched…

Se7soz:
Fix ignored tests due to internal structure changes of `ExecutorScheduler`

Expose a method to indicate if there are tasks scheduled for execution
to be used in tests making sure no more tasks after schedule retention.

  - [x] Please include the issue list number(s) or other PR numbers in the description if you are contributing in response to those.

  - [x] Please include a reasonable set of unit tests if you contribute new code or change an existing one. If you contribute an operator, (if applicable) please make sure you have tests for working with an `empty`, `just`, `range` of values as well as an `error` source, with and/or without backpressure and see if unsubscription/cancellation propagates correctly.

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/ExecutorScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/ExecutorScheduler.java
@@ -270,6 +270,13 @@ public final class ExecutorScheduler extends Scheduler {
             }
         }
 
+        public boolean hasTasks() {
+            if (tasks == null || tasks.size() == 0) {
+                return false;
+            }
+            return true;
+        }
+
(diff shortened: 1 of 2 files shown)


akarnokd (commented):
Any particular reason for doing this change far the sake of old tests? It doesn't seem to worth it.

Se7soz:
@akarnokd The reason I added those UTs is that they were commented with a comment to fix later so I thought of adding them and push this PR as a warm up for my contribution to the project .. hope its fine

akarnokd:
Thank for the contribution but I don't think reviving those particular tests are necessary as other tests for `ExecutorScheduler` already cover the class pretty well.

Also please open a regular issue next time if you wish to contribute something non-trivial to avoid unnnecessary work on either side.

PR #6398, opened 2019-02-05
2.x: Fix window() with start/end selector not disposing/cancelling properly

### LINKED ISSUE #6397: [2.2.6] window() operator leaks subscriptions
Hi,

I believe I encountered `window` operator leaking observable subscriptions. Here is a small test to reproduce the issue:
```
fun main(args: Array<String>) {
    observable(1).window(observable(2).filter { it == 1 }, Function { _: Int -> observable(3).filter { it == 1 } }, 1)
            .subscribe()
            .dispose()
}

private fun observable(n: Int): Observable<Int> {
    return Observable.create<Int> { emitter ->
        println("Subscribed! $n")
        emitter.onNext(1)
        emitter.onNext(2)
        emitter.onNext(3)
        emitter.setCancellable { println("Cancelled! $n") }
    }
}
```

The output is:
```
Subscribed! 2
Subscribed! 3
Cancelled! 3
Subscribed! 1
```

From the output you can see that cancellables for `observable(1)` and `observable(2)` are never called.

This may be severe if cancellables are used to dispose some (possibly) critical resources upon subscription disposal.

akarnokd:
This PR fixes the start-end selector variant of `Observable.window` and `Flowable.window` to properly 

- dispose the window-opening sequence upon disposing the main output flow and 
- disposing the main upstream upon disposing the main output flow provided there are no windows open.

Fixes: #6397

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableWindowWithStartEndFlowableTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableWindowWithStartEndFlowableTest.java
@@ -23,7 +23,7 @@ import org.junit.*;
 import org.reactivestreams.*;
 
 import io.reactivex.*;
-import io.reactivex.exceptions.TestException;
+import io.reactivex.exceptions.*;
 import io.reactivex.functions.*;
 import io.reactivex.internal.functions.Functions;
 import io.reactivex.internal.subscriptions.BooleanSubscription;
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableWindowWithStartEndFlowableTest.java
```diff
@@ -430,4 +431,58 @@ protected void subscribeActual(
...
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void mainWindowMissingBackpressure() {
+        PublishProcessor<Integer> source = PublishProcessor.create();
+        PublishProcessor<Integer> boundary = PublishProcessor.create();
+
+        TestSubscriber<Flowable<Integer>> ts = source.window(boundary, Functions.justFunction(Flowable.never()))
+        .test(0L)
+        ;
```
vanniktech, line 477: Why on it's own line?

src/test/java/io/reactivex/internal/operators/flowable/FlowableWindowWithStartEndFlowableTest.java
```diff
@@ -254,8 +255,8 @@ public void testUnsubscribeAll() {
 
         ts.dispose();
 
-        // FIXME subject has subscribers because of the open window
-        assertTrue(open.hasSubscribers());
+        // Disposing the outer sequence stops the opening of new windows
+        assertFalse(open.hasSubscribers());
```
vanniktech, line 259: Nice that we can fix this too!

vanniktech (approved):

src/test/java/io/reactivex/internal/operators/flowable/FlowableWindowWithStartEndFlowableTest.java
akarnokd: A habit if more methods are ever chained onto them.

PR #6416, opened 2019-02-20
optimize the ComputationScheduler createWorker() method implementation

JasonChen86899:
The previous method of **ComputationScheduler** createWorker() internally calling the innner **FixedSchedulerPool** class when creating the worker is to select the worker through a simple round-robin  that is modulo operation. Now make some optimizations that adds simple round-robin and non-round-robin two selection methods. Simple round-robin is different from the previous one. For the length of the workers array,when the array length is power of 2, the modulo operation is replaced by bit operation. non-round-robin method needs to used by setting the system property "rx2.no-round-robin"
with value "true" and only apply to subclasses of **NewThreadWorker** which essence is to select the minimum computational thread pool by comparing thread pool inner queue size.
 Unit test is in **ComputationSchedulerTests** testComputationSchedulerChooseWorker()

### DIFF
--- a/src/main/java/io/reactivex/internal/schedulers/ComputationScheduler.java
+++ b/src/main/java/io/reactivex/internal/schedulers/ComputationScheduler.java
@@ -23,6 +23,9 @@ import io.reactivex.annotations.NonNull;
 import io.reactivex.disposables.*;
 import io.reactivex.internal.disposables.*;
 import io.reactivex.internal.functions.ObjectHelper;
+import io.reactivex.internal.util.DefaultSchedulerWorkerChooserFactory;
+import io.reactivex.internal.util.SchedulerWorkerChooserFactory;
+import io.reactivex.internal.util.SchedulerWorkerChooserFactory.SchedulerWorkerChooser;
 
 /**
  * Holds a fixed pool of worker threads and assigns them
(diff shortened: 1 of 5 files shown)


akarnokd:
Hi. Why do you want this optimization? Asking for a computation scheduler is so rare that the 90 cycle modulo is only noticeable in highly targeted benchmarks. Also you added a lot of convoluted, indirect code that may not inline and exposed some internal details (such as queue size?). If you want different utilization, use `Schedulers.from` with an custom-configured executor.

Also please open an issue before posting non-trivial changes, especially performance related, to discuss if said contribution makes sense or is worth doing in the first place.

JasonChen86899:
Sorry for not open an issue in the first place. I want to discuss with you firstly to make sure worth doing this. Thanks.
The reason is that when the lib is used on the server side and used multiple times in the high concurrent scene, a computation scheduler worker will be selected by a scheduler selector is used as a subscriber executor for some times. At this time, if a lot of complex calculations are being performed (Android does not exist), simple round-robin will increase the burden of one cpu. This optimization mainly considers two aspects: if it is simple round-robin, it adopts bit operation because in many cases the core number of cpu is a power of 2. If the server is used multiple times and complex calculations are performed, it is necessary to select idle workers. On the server side, thread resources are very valuable, and computing can add adaptations to server-side scenarios, avoiding custom schedulers.

akarnokd:
Since you can specify custom schedulers for operators and you can implement your strategy in a custom scheduler, with which you can replace computation globally via hooks, I.see no reason to change the standard scheduler. Also schedulers can live in any library, unlike operators.

PR #6429, opened 2019-03-13
add doc for from action in new style

VovaStelmashchuk:
I start to implement new doc file for [Async-Operators.md](https://github.com/ReactiveX/RxJava/blob/2.x/docs/Async-Operators.md).
I check #6131; I can see some operations from the Async-Operators are equals with `Creating-Observables.md` 
Do we need to copy/paste the description of these operators from an existing source?
I talk about the following operator: `fromAction()`, `fromRunnable`, `fromCallable`


akarnokd:
> Do we need to copy/paste the description of these operators from an existing source?

No, those are not async sources by themselves.

akarnokd:
Closing due to inactivity.

PR #6431, opened 2019-03-15
2.x composite disposable docs

### LINKED ISSUE #6430: 2.x "d is null" message for CompositeDisposable add and addall
Affects version: 2.2.7 (and before)

Message "d is null" for errors in CompositeDisposable's add and addall methods is not very helpfull.

chronvas:
On CompositeDisposable add and addAll methods, if the param is null, currently the NPE error message (produced by ObjectHelper) is "d is null" which is not very helpful.

This is a small refactor for making the message a bit more helpful.

Resolves #6430

### DIFF
--- a/src/main/java/io/reactivex/disposables/CompositeDisposable.java
+++ b/src/main/java/io/reactivex/disposables/CompositeDisposable.java
@@ -93,7 +93,7 @@ public final class CompositeDisposable implements Disposable, DisposableContaine
      */
     @Override
     public boolean add(@NonNull Disposable d) {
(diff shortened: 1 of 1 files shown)


akarnokd:
If you want to improve, change the argument name to `disposable` instead.

akarnokd (changes requested):
Please rename the arguments/variables instead.

chronvas:
Sorry, went ahead and created another branch before you replied. Will close this one, please have a look at PR #6432

Issue #6402, opened 2019-02-10
Error in documentation of Observable .retry(n) 


RobLewis:
There appears to be an error in the 2.x Javadoc for the `retry( n )` operator of `Observable`. `n` is referred to as "number of retry attempts before failing" but the accompanying marble diagram shows only a single retry when `retry(2)` is called. 

Does `n` indicate the total number of subscriptions including the first one?

akarnokd:
Hello. You could write a small program to find it out and if there is a documentation mistake, you are welcome to post a PR.

RobLewis:
My test program (below) shows that the `n` argument does NOT include the initial attempt to subscribe, thus `retry( 10 )` gives a total of 11 subscriptions. 

The easiest doc fix would be to change the large `retry(2)` in the marble diagram to `retry(1)` but I have no facility for doing that. (The Javadoc for `Single` doesn't include a marble diagram, but the text description is correct.) 

```
Disposable retryDisp = Observable.error( new SocketTimeoutException() )

                .doOnSubscribe( disp -> System.out.println( "Subscribe count is now " + ++subCount ) )

                .doOnError( err -> System.out.println( "Error count is now " + ++errCount ) )

                .retry( 10 )  // gives a total of 11 subscriptions & 11 errors

                .subscribe(
                        item -> System.out.println( "We never see this" ),
                        err -> System.out.println( "Without this we get a 'missing error handler' crash" )
                );
```

RomanWuattier:
Does anyone work on this issue?
Also, the function parameter `times` is described as "the number of times to repeat" and could be mixed with the `repeat` operator. I think that's an opportunity to improve the Javadoc of `public final Observable<T> retry(long times, Predicate<? super Throwable> predicate)` at the same time.

akarnokd:
@RomanWuattier Nobody seems to work on this so you are welcome if you want.

akarnokd:
Closing via #5458.

PR #6458, opened 2019-04-10
Update the Javadoc of the `retry` operator

RomanWuattier:
Specify that the `times` function parameter describes "the number of times
to resubscribe if the current Observable fails".

Solves: #6402

### DIFF
--- a/src/main/java/io/reactivex/Completable.java
+++ b/src/main/java/io/reactivex/Completable.java
@@ -2090,7 +2090,7 @@ public abstract class Completable implements CompletableSource {
      *  <dt><b>Scheduler:</b></dt>
      *  <dd>{@code retry} does not operate by default on a particular {@link Scheduler}.</dd>
      * </dl>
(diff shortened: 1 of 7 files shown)


RomanWuattier:
Bellow is the marble with the number of retries decreased from 2 to 1. If the marble is correct, can you upload it to the repository?
retry o n

akarnokd:
Sure. Could you also make sure all the other counted `retry()` operators in the other base classes have the correct wording?

RomanWuattier:
Sure.
I updated every base classes. Also, I grabbed this opportunity to update some unit tests to illustrate the Javadoc wording.

Please, note:
* `Single` already has the right wording.
* Both `Maybe` and `Flowable` are using the default `retry` marble (https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/retry.png)

akarnokd (approved):

akarnokd:
Nice! Thanks!

vanniktech (approved):

Issue #6460, opened 2019-04-13
fromArray to report Element at index $i is null


akarnokd:
Currently, `Observable.fromArray` says `The $i th element is null`, which is a bit misleading as `i` is the element index. Update the error message to `The element at index $i is null`. The `Flowable` version should be updated if necessary. Note that the null checks appear multiple places.

akarnokd:
Closing via #6461

PR #6461, opened 2019-04-13
Change error message in ObservableFromArray

iruizm:
Changed the error message when an element at position i is null to what was suggested in #6460 .

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableFromArray.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableFromArray.java
@@ -126,7 +126,7 @@ public final class FlowableFromArray<T> extends Flowable<T> {
                 }
                 T t = arr[i];
                 if (t == null) {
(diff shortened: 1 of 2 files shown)


akarnokd:
Could you please update these locations as well?
 
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableFromArray.java#L129

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableFromArray.java#L159

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableFromArray.java#L212

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableFromArray.java#L242

iruizm:
Sure, it's done.

akarnokd (approved):

vanniktech (approved):

PR #6485, opened 2019-05-23
remove unused else from the Observable

VovaStelmashchuk:
Remove unused else from the Observable class.

### DIFF
--- a/src/main/java/io/reactivex/Observable.java
+++ b/src/main/java/io/reactivex/Observable.java
@@ -13,26 +13,70 @@
 
 package io.reactivex;
 
-import java.util.*;
-import java.util.concurrent.*;
-
 import org.reactivestreams.Publisher;
 
-import io.reactivex.annotations.*;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/Observable.java
```diff
@@ -13,26 +13,70 @@
 
 package io.reactivex;
 
-import java.util.*;
-import java.util.concurrent.*;
-
 import org.reactivestreams.Publisher;
 
-import io.reactivex.annotations.*;
+import java.util.Collection;
```
akarnokd: Please do not unroll star imports.

akarnokd (changes requested):

src/main/java/io/reactivex/Observable.java
VovaStelmashchuk: Fixed.

akarnokd (approved):

vanniktech (approved):

src/main/java/io/reactivex/Observable.java
VovaStelmashchuk: @akarnokd Maybe you can include some rule for the import bloc to `CONTRIBUTING.md` ?

akarnokd: @jordan1997 

> When submitting code, please make every effort to follow existing conventions and style in order to keep the code as readable as possible.

But don't worry, we will warn you if there is a stylistic problem.

PR #6486, opened 2019-05-25
Add few implementation for the Flowable.mergeDelayError

VovaStelmashchuk:
1. Add Flowable.mergeDelayError with 2-9 arguments. Now you can use it as a method merge.
2. Update tests which have annotation `@Ignore` for new logic.

### DIFF
--- a/src/main/java/io/reactivex/Flowable.java
+++ b/src/main/java/io/reactivex/Flowable.java
@@ -3921,6 +3921,303 @@ public abstract class Flowable<T> implements Publisher<T> {
     }
 
     /**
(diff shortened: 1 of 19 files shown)


akarnokd:
We decided not to implement the 5-9 argument variants because they are rarely needed and they are static methods that can live on any class in any external libraries.

Please next time open an issue first to discuss new operators.

VovaStelmashchuk:
@akarnokd What do you think about the tests which I updated?

akarnokd:
I don't see much value in them.

PR #6499, opened 2019-06-12
Null check for BufferExactBoundedObserver

JanKn:
- [*] Please give a description about what and why you are contributing, even if it's trivial:

**Improve stability of the library.**
Other variants of the onComplete method include this Null check already, e.g. BufferExactUnboundedObserver, this check should fix the case when there is a race condition and buffer is already set to "null" by the time onComplete is called.

It is causing 0.1% crashes in our production app, this should improve stability of other apps too.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferTimed.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferTimed.java
@@ -13,13 +13,19 @@
 
 package io.reactivex.internal.operators.flowable;
 
-import java.util.*;
-import java.util.concurrent.*;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
-import org.reactivestreams.*;
-
-import io.reactivex.*;
+import io.reactivex.Flowable;
+import io.reactivex.Scheduler;
 import io.reactivex.Scheduler.Worker;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.Exceptions;
@@ -27,7 +33,8 @@ import io.reactivex.internal.disposables.DisposableHelper;
 import io.reactivex.internal.functions.ObjectHelper;
 import io.reactivex.internal.queue.MpscLinkedQueue;
 import io.reactivex.internal.subscribers.QueueDrainSubscriber;
-import io.reactivex.internal.subscriptions.*;
+import io.reactivex.internal.subscriptions.EmptySubscription;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
 import io.reactivex.internal.util.QueueDrainHelper;
 import io.reactivex.subscribers.SerializedSubscriber;
 
@@ -501,13 +508,15 @@ public final class FlowableBufferTimed<T, U extends Collection<? super T>> exten
                 buffer = null;
             }
 
-            queue.offer(b);
-            done = true;
-            if (enter()) {
-                QueueDrainHelper.drainMaxLoop(queue, downstream, false, this, this);
-            }
+            if (b != null) {
+                queue.offer(b);
+                done = true;
+                if (enter()) {
+                    QueueDrainHelper.drainMaxLoop(queue, downstream, false, this, this);
+                }
 
-            w.dispose();
+                w.dispose();
+            }
         }
 
         @Override
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferTimed.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferTimed.java
@@ -13,19 +13,13 @@
 
 package io.reactivex.internal.operators.flowable;
 
-import org.reactivestreams.Subscriber;
-import org.reactivestreams.Subscription;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
+import java.util.*;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 
-import io.reactivex.Flowable;
-import io.reactivex.Scheduler;
+import org.reactivestreams.*;
+
+import io.reactivex.*;
 import io.reactivex.Scheduler.Worker;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.exceptions.Exceptions;
@@ -33,8 +27,7 @@ import io.reactivex.internal.disposables.DisposableHelper;
 import io.reactivex.internal.functions.ObjectHelper;
 import io.reactivex.internal.queue.MpscLinkedQueue;
 import io.reactivex.internal.subscribers.QueueDrainSubscriber;
-import io.reactivex.internal.subscriptions.EmptySubscription;
-import io.reactivex.internal.subscriptions.SubscriptionHelper;
+import io.reactivex.internal.subscriptions.*;
 import io.reactivex.internal.util.QueueDrainHelper;
 import io.reactivex.subscribers.SerializedSubscriber;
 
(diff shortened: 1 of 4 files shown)


akarnokd:
I don't see how that scenario would be possible with just the exact buffer. Only the terminal events can set `buffer` to null and `run` is protected against a null buffer.
Aren't you using `PublishSubject` as well from multiple threads, competing for termination?

akarnokd:
Also an unit test demonstrating the presumed bug has been fixed is required.

JanKn:
To be honest, I am trying to simulate the scenario past day and a half.
Being it likely a race condition, it makes it difficult.
I will give Unit test a shot, if you have a sample Test for some other operator, which is testing similar case, I would appreaciate if you can point me there, thanks.

Our app uses some PublishSubjects, but we switched mostly to PublishRelay. The scenario with disposing from multiple threads is possible.
The production StackTrace we are seeing is this:

```
Fatal Exception: java.lang.NullPointerException: Null is not a valid element
       at io.reactivex.internal.queue.MpscLinkedQueue.offer(MpscLinkedQueue.java:60)
       at io.reactivex.internal.operators.observable.ObservableBufferTimed$BufferExactBoundedObserver.onComplete(ObservableBufferTimed.java:507)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$MergeObserver.drainLoop(ObservableFlatMap.java:371)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$MergeObserver.drain(ObservableFlatMap.java:326)
       at io.reactivex.internal.operators.observable.ObservableFlatMap$MergeObserver.onComplete(ObservableFlatMap.java:303)
       at retrofit2.adapter.rxjava2.BodyObservable$BodyObserver.onComplete(BodyObservable.java:66)
       at retrofit2.adapter.rxjava2.CallExecuteObservable.subscribeActual(CallExecuteObservable.java:51)
       at io.reactivex.Observable.subscribe(Observable.java:12268)
       at retrofit2.adapter.rxjava2.BodyObservable.subscribeActual(BodyObservable.java:34)
       at io.reactivex.Observable.subscribe(Observable.java:12268)
       at io.reactivex.internal.operators.observable.ObservableFlatMap.subscribeActual(ObservableFlatMap.java:55)
       at io.reactivex.Observable.subscribe(Observable.java:12268)
       at io.reactivex.internal.operators.observable.ObservableBufferTimed.subscribeActual(ObservableBufferTimed.java:66)
       at io.reactivex.Observable.subscribe(Observable.java:12268)
       at io.reactivex.internal.operators.observable.ObservableMap.subscribeActual(ObservableMap.java:32)
       at io.reactivex.Observable.subscribe(Observable.java:12268)
       at io.reactivex.internal.operators.observable.ObservableDoOnEach.subscribeActual(ObservableDoOnEach.java:42)
       at io.reactivex.Observable.subscribe(Observable.java:12268)
       at io.reactivex.internal.operators.observable.ObservableSubscribeOn$SubscribeTask.run(ObservableSubscribeOn.java:96)
       at io.reactivex.Scheduler$DisposeTask.run(Scheduler.java:578)
       at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
       at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
       at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:784)
```

akarnokd:
That code path should not lead to NPE. The only way for `buffer` to remain null is that the supplier fails, but then you should get an error signal. Otherwise, calling `onComplete` twice would result in NPE. The exact-boundary doesn't release the buffer upon disposing it, unlike the other implementation in that file.

JanKn:
I agree that it can be something wrong with the supplier. 
Another example: If supplier breaches the contract and calls onError (buffer is set to null) and then onComplete, which it should not, null gets into MprcLinkedQueue.offer:

```
...
    public boolean offer(final T e) {
        if (null == e) {
            throw new NullPointerException("Null is not a valid element");
        }
...  
```     
It does not happen super often given the size of the app, but it is still a concern.
Will it be something we can merge?

If so, I can push the test which covers this (fails before change, passes after change):
```
 @Test
    public void bufferExactIssueXXX() {
        Scheduler.Worker w = Schedulers.io().createWorker();
        TestObserver<List<Integer>> observer = new TestObserver<List<Integer>>();
        BufferExactBoundedObserver<Integer, List<Integer>> buf = new BufferExactBoundedObserver<Integer, List<Integer>>(observer, ArrayListSupplier.<Integer>asCallable(), 100, TimeUnit.MILLISECONDS, 10, false, w);

        buf.onError(new Throwable());
        buf.onComplete();
    }
```

akarnokd:
In that case, use a failing supplier with an empty source. No threading necessary. Also  the Flowable variant may be prone to the same issue. Could you check and fix that too?

src/main/java/io/reactivex/internal/operators/flowable/FlowableBufferTimed.java
```diff
@@ -13,21 +13,28 @@
 
 package io.reactivex.internal.operators.flowable;
 
-import java.util.*;
-import java.util.concurrent.*;
+import org.reactivestreams.Subscriber;
```
akarnokd: Please don't unroll star imports.

src/test/java/io/reactivex/internal/operators/flowable/FlowableBufferTest.java
```diff
@@ -13,32 +13,65 @@
 
 package io.reactivex.internal.operators.flowable;
 
-import static org.junit.Assert.*;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.*;
+import org.junit.Before;
```
akarnokd: Please don't unroll star imports.

src/test/java/io/reactivex/internal/operators/flowable/FlowableBufferTest.java
```diff
@@ -2769,4 +2802,15 @@ public void timedSizeBufferAlreadyCleared() {
 
         sub.run();
     }
+
+    @Test
+    public void bufferExactFailingSupplier() {
+        Scheduler.Worker w = new TestScheduler().createWorker();
+        TestSubscriber<List<Integer>> observer = new TestSubscriber<List<Integer>>();
+
+        FlowableBufferTimed.BufferExactBoundedSubscriber<Integer, List<Integer>> buf = new FlowableBufferTimed.BufferExactBoundedSubscriber<Integer, List<Integer>>(observer, ArrayListSupplier.<Integer>asCallable(), 100, TimeUnit.MILLISECONDS, 10, false, w);
```
akarnokd: The list supplier should fail, calling onError and onCompleted this way is just a plain protocol violation.

Do something this instead:

```java
Flowable.empty()
.buffer(10, 10, TimeUnit.SECONDS, Schedulers.computation(), new Callable<List<Object>>() {
    @Override public void call() {
        throw new TestException();
    }
})
.test()
.assertFailure(TestException.class)
;
```

src/test/java/io/reactivex/internal/operators/observable/ObservableBufferTest.java
```diff
@@ -2136,4 +2165,15 @@ public void bufferSkipDoubleOnSubscribe() {
             }
         });
     }
+
+    @Test
+    public void bufferExactFailingSupplier() {
+        Scheduler.Worker w = new TestScheduler().createWorker();
```
akarnokd: The list supplier should fail, calling onError and onCompleted this way is just a plain protocol violation.

Do something this instead:

```java
Observable.empty()
.buffer(10, 10, TimeUnit.SECONDS, Schedulers.computation(), new Callable<List<Object>>() {
    @Override public void call() {
        throw new TestException();
    }
})
.test()
.assertFailure(TestException.class)
;
```

akarnokd (changes requested):

akarnokd (approved):

JanKn:
Hi, 
thank you for the test sample, that helped a lot, I just had to hit the right condition.
I think we can merge this change now.
When do you expect a release containing this?

akarnokd:
I usually wait for @vanniktech's review, then if you think this is urgent, we can release soon after.

vanniktech (commented):
Whops. Totally lost track of this! LGTM!

PR #6507, opened 2019-06-17
3.x: Sync up with 2.2.10 snapshot

akarnokd:
The Changes.md always comes up conflicting. I'll leave it as it is for now.


vanniktech (commented):
Why are we pushing for 3.x?

akarnokd:
I decided it is time to move forward towards the next major release.

PR #6511, opened 2019-06-17
3.x: Widen functional interface throws, replace Callable with Supplier

akarnokd:
This PR widens the `throws Exception` into `throws Throwable` in the functional interfaces and adjusts `catch (Exception` to `catch(Throwable` where needed.

The major change is the replacement of `java.util.concurrent.Callable` in almost all API with `io.reactivex.functions.Supplier` which is defined with `throws Throwable`. Since subinterfaces can't widen the throws clause, only narrow it, `Supplier` can't extend `Callable`.

`fromCallable` remained in all base classes and a separate PR will introduce `fromSupplier`.

The single-valued fusion now works with `Supplier` and `ScalarSupplier` types instead of `Callable` and `ScalarCallable` (removed).

### DIFF
--- a/src/main/java/io/reactivex/functions/Supplier.java
+++ b/src/main/java/io/reactivex/functions/Supplier.java
@@ -17,8 +17,8 @@ package io.reactivex.functions;
  * A functional interface (callback) that provides a single value or
  * throws an exception.
  * <p>
- * Thins interface was added to allow throwing any subclass of {@link Throwable}s,
- * which is not directly possible with the Java standard {@link java.util.Callable} interface. 
+ * This interface was added to allow throwing any subclass of {@link Throwable}s,
+ * which is not directly possible with the Java standard {@link java.util.concurrent.Callable} interface.
  * @param <T> the value type returned
  * @since 3.0.0
  */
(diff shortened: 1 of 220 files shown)


src/main/java/io/reactivex/functions/Supplier.java
```diff
@@ -0,0 +1,33 @@
...
+ * the License for the specific language governing permissions and limitations under the License.
+ */
+
+package io.reactivex.functions;
+
+/**
+ * A functional interface (callback) that provides a single value or
+ * throws an exception.
+ * <p>
+ * Thins interface was added to allow throwing any subclass of {@link Throwable}s,
```
lorenzpahl: Typo: This

src/main/java/io/reactivex/internal/fuseable/ScalarSupplier.java
```diff
@@ -22,16 +22,16 @@
  * <p>
  * Implementors of {@link #call()} should not throw any exception.
```
lorenzpahl: The link should now point to `get()`.

akarnokd: Thanks. Posted an update with additional cleanups.

PR #6512, opened 2019-06-17
2.x: Use @Code instead of using <code> </code> markup

JonathanSawyer:
Improves Javadoc @Code usage

- Uses `@Code` instead of `<code> </code>` markup
- Replaces escape characters inline with using `@Code` attribute
           1. `&gt;` changes to `>`
           2. `&lt;` changes to `<`


nirvanarsc (approved):

akarnokd:
Thanks, but I'm fine with the original `<code>` approach at these places.

PR #6513, opened 2019-06-18
2.x: Remove redundant type casts and type arguments

JonathanSawyer:
This PR mostly tidies up some test code.

- Removes `redundant type arguments` for various `fromArray` calls in `tests`.
   - ParallelFlowable.fromArray
   - Flowable.fromArray
   - Observable.fromArray
- Remove `redundant type casts` mostly in tests
- Remove `redundant type casts` from 
   - src/main/java/io/reactivex/Flowable.java 
   - src/main/java/io/reactivex/internal/operators/flowable/FlowableCombineLatest.java


nirvanarsc (approved):

akarnokd:
Thanks, but these simplifications break Eclipse compilation. Not sure what IDE are you using, but many of the removal of explicit casts are not correct in a Java 6 compilation settings. Many of these were corrected in Java 8 and above, which we can't use.

JonathanSawyer:
IntelliJ

PR #6525, opened 2019-06-20
3.x: remove `test` from `testXXX` method names, create a validator

### LINKED ISSUE #6518: 3.x: remove `test` from `testXXX` method names, create a validator.
A lot of test methods were named like `testXXX` in the 0.x-1.x era and we didn't want to change those for 2.x. Now is the time to get rid of the redundant naming from all test methods.

In addition, a validator should be added that checks all test Java files and reports if it finds the pattern `void\s+test[a-zA-Z0-9]` in any file. See [this validator](https://github.com/ReactiveX/RxJava/blob/3.x/src/test/java/io/reactivex/validators/CheckLocalVariablesInTests.java) for an example of finding and checking source code files.

maksim-m:
Fix #6518.

- [X] Create validator
- [X] Automatically rename methods "testXXX" to "xxx" (analog to [FixLicenseHeaders](https://github.com/ReactiveX/RxJava/blob/3.x/src/test/java/io/reactivex/validators/FixLicenseHeaders.java) check)
- [x] Rename methods

### DIFF
--- a/src/test/java/io/reactivex/validators/TestPrefixInMethodName.java
+++ b/src/test/java/io/reactivex/validators/TestPrefixInMethodName.java
@@ -104,7 +104,7 @@ public class TestPrefixInMethodName {
                                 in.close();
                             }
 
-                            if (found && System.getenv("CI") == null) {
+                            /*if (found && System.getenv("CI") == null) {
                                 PrintWriter w = new PrintWriter(new FileWriter(u));
 
                                 try {
@@ -114,7 +114,7 @@ public class TestPrefixInMethodName {
                                 } finally {
                                     w.close();
                                 }
-                            }
+                            }*/
                         }
                     }
                 }
--- a/src/test/java/io/reactivex/validators/TestPrefixInMethodName.java
+++ b/src/test/java/io/reactivex/validators/TestPrefixInMethodName.java
@@ -13,6 +13,7 @@
 
 package io.reactivex.validators;
 
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.*;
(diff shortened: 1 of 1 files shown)


src/test/java/io/reactivex/validators/TestPrefixInMethodName.java
```diff
@@ -0,0 +1,131 @@
...
+                                    } else {
+                                        lines.add(line);
+                                    }
+
+                                }
+                            } finally {
+                                in.close();
+                            }
+
+                            if (found && System.getenv("CI") == null) {
```
akarnokd: I'd rather not have this run unconditionally. There could be cases this introduces conflicts or other compilation errors without a good way of undoing the changes. Please comment this part out for now.

akarnokd (changes requested):

src/test/java/io/reactivex/validators/TestPrefixInMethodName.java
maksim-m: Updated

akarnokd:
I was doing a massive set of changes to the test files myself so it would be great you rolled back the renames, kept the validator code but not fail when violations are found. I'll merge in #6526, then just the validator, then we can enable the offline/CI failure mode and fix the names in a fresh PR.

maksim-m:
I added `@Ignore` annotation for now and removed a commit with the actual renaming.
I will open a new PR when #6526 and this one are merged.

akarnokd (approved):

vanniktech (approved):

Issue #6122, opened 2018-08-01
3.x: Disambiguate startWith(T|Iterable<T>)?


akarnokd:
We were using postfixes to avoid inference problems around operator methods (such as `fromIterable` vs `fromArray`) but there is at least one operator currently, `startWith`, that [may exhibit type ambiguity](https://twitter.com/HansWurst315/status/1024386078785105921). A dedicated postfixed name should resolve the issue, such as:

1. `startWith(T)` + `startWithIterable(Iterable<T>)`
2. `startWithItem(T)` + `startWith(Iterable<T>)` + remove `startWith(T)`
3. `startWithItem(T)` + `startWithIterable(Iterable<T>)` + remove `startWith(T)`.

As a preparation, we can deprecate `startWith(T)` within 2.2.x in case option 2 or 3 is chosen.

vanniktech:
The same also happens with some `onError*` methods.

I'm more inclined towards option 3 but no strong opinion.

davidmoten:
For 3.x I'm happy with 3 also.

davidmoten:
Perhaps instead of `startWithItem` we use `startWithJust` to match our `just` usage?

RomanWuattier:
That's just my two cents' worth I would opt for the option 3.

I think `startWithJust` could be confusing since other methods call `just` for example, `range` when the count value is one or `fromArray` when the length is one.

akarnokd:
3.x is now accepting PRs if somebody wants to tackle this.

akarnokd:
Closing via #6530.

PR #6530, opened 2019-06-21
3.x: disambiguate startWith+1 to startWithItem & startWithIterable

akarnokd:
`startWith(T)` and `startWith(Iterable)` was causing some trouble with `startWith(Publisher)`. This PR will rename the former two.

The diagrams will be updated in a separate PR.

Resolves: #6122


vanniktech (approved):

vanniktech:
This is one of the changes where library authors can run into troubles. The only way to fix this is to use 3.x publish a new version and drop 2.x support

akarnokd:
These are conveniences for `concatArray(wraper(first), this)` so those libraries affected can still support both versions with the same method call.

PR #6546, opened 2019-06-27
2.x: Simplify JUnit tests with more appropriate assert methods

slievrly:
Simplify unit tests by using the more appropriate JUnit assert methods.


akarnokd:
How did you detect these?

Could you also post these PRs against 3.x too?

slievrly:
@akarnokd I detected it through IDEA's Inspect Code plugin, and it works fine if it passes the unit test.
I can also submit this unit test simplified pr to 3.x

akarnokd (approved):

vanniktech (approved):

PR #6548, opened 2019-06-28
fix typo

okou19900722:


akarnokd:
RxJava 1 is no longer supported or maintained. You are welcome to post PRs against 3.x though if you want.

PR #6549, opened 2019-06-29
3.x: Simplify JUnit tests with more appropriate assert methods

slievrly:
Simplify JUnit tests with more appropriate assert methods.


akarnokd (approved):

akarnokd:
Thanks!

Please in the future, fill in the issue text and don't leave the default welcome text there. The better the issue text describes the change the easier search engines will find them.

vanniktech (approved):

PR #6570, opened 2019-07-12
3.x: Code verification fixes for javac's generated switchmap classes

akarnokd:
Java 13+ compilers generate a synthetic class for switch statements over enumerables now, so `Observable.class` (and a few others) now receive an `Observable$1.class` companion, which trips the no anonymous inner classes heuristic verification. This PR changes the respective unit test to filter out such classes in a heuristic manner too.

In addition, the `findSource` method has been moved into `TestHelper` for clarity and changed so that any capitalization of the "rxjava" host directory is accepted.

### DIFF
--- a/src/test/java/io/reactivex/testsupport/TestHelper.java
+++ b/src/test/java/io/reactivex/testsupport/TestHelper.java
@@ -17,7 +17,9 @@ import static org.junit.Assert.*;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 
+import java.io.File;
 import java.lang.reflect.*;
+import java.net.URL;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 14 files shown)


JakeWharton:
For what it's worth, javac has _always_ generated these, they just used to be named `YourClass$SwitchMap`.

JakeWharton:
Actually as far as I can tell this behavior hasn't changed at all. I get the same behavior with Java 14 as I get with Java 6.

akarnokd:
Looks like I've been tricked by Gradle/IntelliJ all along. Eclipsec doesn't generate this file and by default and all class files end up in the same bin directory for me -> no failures locally. Javac does generate this but Gradle/IntelliJ put test classes separately into build/classes/java/test or out/test so the directory returned by getResource naturally didn't contain such files. However, when I run the entire tests suite with IntelliJ 2019.2, the test somehow saw the $1 file and failed. 

In other terms, a bug in the test met a glitch/setting/bug in IntelliJ and suddenly worked as intended.

I've updated the test again to strip on "/test" instead as it seems to be the common theme.

Issue #6571, opened 2019-07-15
2.x: inconsistent error handling in onErrorResumeNext after completion


davideberlein:
Tested using RxJava `v2.2.10`

### Use Case: 
We want to handle a known exception that can occur anywhere in a long observable stream using `onErrorResumeNext`.
The Exception is thrown due to async behavior: A resource is closed which triggers the a `takeUntil` however due to async operation (`observeOn`) in the stream, an operator may still access the closed resource causing the exception. 

### Problem
 
1. The observable emits a value
2. During the execution of an operator A (in this case `map`) the `takeUntil` is triggered
3. The subscription completes
4. Operator A throws an Exception
5. XXX

Depending on which operators are between the operator that throws the exception and the `onErrorResumeNext` RX behaves differently in step **5**, not allowing us to properly handle the Error:

- `doOnNext`, `map`, `concatMap` and probably most others: `onErrorResumeNext` is called and handles the error
- `switchMap`: the Exception is completely swallowed, neither `onErrorResumeNext` is called nor an `UndeliverableException` is thrown or printed anywhere. 
- `flatMap`: A `UndeliverableException` is thrown.

For us all cases except for the `flatMap` behavior is okay.

### Example:
```Java
final Subject<String> takeUntil$ = BehaviorSubject.create();
Observable.just("Some value")
		.map(e -> {
			takeUntil$.onNext("complete");
			throw new RuntimeException("Exception after unsubscribe");
		})
		// Add any of these to test the different cases:
		//.doOnNext(e -> System.out.println("On Next"))
		//.map(v -> v)
		//.concatMap(o -> Observable.just(o))
		//.switchMap(o -> Observable.just(o))
		.flatMap(o -> Observable.just(o))
		.onErrorResumeNext(e -> {
			System.out.println(System.currentTimeMillis() + " onErrorResumeNext GOT ERROR: " + e);
			return Observable.empty();
		})
		.takeUntil(takeUntil$)
		.subscribe(v -> System.out.println("GOT NEXT" + v), v -> System.out.println("Subscribe GOT ERROR" + v), () -> System.out.println(System.currentTimeMillis() + " COMPLETED"));
```

### Exception Printed with flatMap:

```
[...]

PR #6572, opened 2019-07-15
3.x: fix switchMaps inconsistency swallowing errors when cancelled

akarnokd:
The various `switchMap` operators did not report any accumulated exceptions to the global error handler when the sequence was cancelled (and thus the errors would never be delivered through the regular channels).

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java
+++ b/src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java
@@ -14,6 +14,7 @@
 package io.reactivex.internal.operators.flowable;
 
 import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.*;
 
 import java.util.*;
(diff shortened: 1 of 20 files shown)


src/test/java/io/reactivex/internal/operators/flowable/FlowableSwitchTest.java
```diff
@@ -1202,4 +1203,34 @@ public Object apply(Integer w) throws Exception {
         .assertNoErrors()
         .assertComplete();
     }
+
+    @Test
+    public void undeliverableUponCancel() {
```
pwittchen, line 1208: I see that all `undeliverableUponCancel()` test methods in different test classes are quite similar. Do you think, does it make sense to extract this code, parametrize it and move it to the newly created `abstract class SwitchTest`, which these classes can extend to avoid code duplication? I haven't checked if it's doable.

akarnokd: Try it, you'll see.

vanniktech (approved):
Does 2.x also need a fix?

akarnokd:
It is certainly an inconsistency and goes against the no-error lost principle introduced via 2.x. It was reported (#6571)  in conjunction with 2.x but the changes introduce more places of undeliverable exceptions. I don't know what such change could mean for users of the 2.2.x line.

PR #6574, opened 2019-07-16
3.x: travis use openjdk8 as oraclejdk8 stopped working?

akarnokd:
The last [merge failed](https://travis-ci.org/ReactiveX/RxJava/builds/559419743#L166) and looks like oraclejdk8 target is broken?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,6 @@
 language: java
 jdk:
-- oraclejdk8
-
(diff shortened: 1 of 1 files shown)


akarnokd:
Looks like the new default is "xenial" which doesn't support oraclejdk8. The old one were "trusty" and the related [build matrix](https://github.com/akarnokd/RxJava3_BuildMatrix/blob/master/.travis.yml#L4) works on that.

PR #6588, opened 2019-07-23
Fix #6587

### LINKED ISSUE #6587: Inner errors within concatMapSingle vanish if disposed
Not sure if this intentional or a bug. The following code:
```java
package Test;

import io.reactivex.Single;

public class Test {
//		Observable.just(0,1)		// UndeliverableException - First concatMapSingle-block throws, interrupting second one
		Observable.just(2, 0)		// Just "MyError" output, exception in first block appears to vanish
				.concatMapSingle(workItem2 -> {
					return Single.just(workItem2).subscribeOn(Schedulers.computation()).map(workItem -> {
						try {
							Thread.sleep(1000);
							if (workItem == 1)
								throw new Exception("Something in first block failed");
							Thread.sleep(1000);
							System.out.println("InterruptedException in block 1");
							throw e;
						}
						return workItem;
					});
				}).concatMapSingle(workItem2 -> {
					return Single.just(workItem2).subscribeOn(Schedulers.computation()).map(workItem -> {
						try {
							Thread.sleep(1000);
							if (workItem == 2) {
								throw new Exception("Something in second block failed");
							}
							Thread.sleep(1000);
							System.out.println("InterruptedException in block 2");
							throw e;
						}
						return workItem;
					});
				}).blockingSubscribe(item -> System.out.println("Item finished " + item), err -> {
					System.out.println("MyError: " + err.toString());
					err.printStackTrace();
				});
	}
}
```
has 2 processing stages. If the first one throws an exception, the second one gets interrupted and throws an `UndeliverableException`. In the opposite case, if the 2nd block throws an exception, the 1st one gets interrupted, but its `InterruptedException` is completely ignored - not delivered to any error handler, nor thrown as `UndeliverableException`.

This is caused by some kind of race condition within `ConcatMapSingleMainObserver`. When the 2nd block throws, on the `ConcatMapSingleMainObserver` instance belonging to the 1st block, `dispose` is called before `innerError`. Because at this point `errors` is empty, `addThrowable` succeeds, but the `drain` loop is never called, because `dispose` has set the `AtomicInteger` to `1`. Therefore, the exception in `error` is never retrieved and forwarded.

This can be fixed (if it needs fixing, anyways) by having `innerError`, `dispose` and the `drain` loop check for `cancelled` and forward exceptions to `RxJavaPlugins.onError` if appropriate. I will make a PR...

Erlkoenig90:
In `ObservableConcatMapSingle`, if `dispose` is called before `innerError`, forward errors to `RxJavaPlugins.onError` instead of swallowing them. Since `dispose` and `innerError` can run on different threads, there can be various concurrency situations. Depending on which one executes faster, `dispose` or `innerError` will call `RxJavaPlugins.onError` via `errors.terminate`. If the `drain` loop is running while `dispose` is called, it will call `RxJavaPlugins.onError` instead.

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
+++ b/src/main/java/io/reactivex/internal/operators/flowable/FlowableConcatMapEager.java
@@ -172,6 +172,8 @@ public final class FlowableConcatMapEager<T, R> extends AbstractFlowableWithUpst
                 do {
                     cancelAll();
                 } while (decrementAndGet() != 0);
+
+                errors.tryTerminateAndReport();
             }
         }
 
@@ -279,6 +281,7 @@ public final class FlowableConcatMapEager<T, R> extends AbstractFlowableWithUpst
                         while (e != r) {
                             if (cancelled) {
                                 cancelAll();
+                                errors.tryTerminateAndReport();
                                 return;
                             }
 
@@ -305,6 +308,8 @@ public final class FlowableConcatMapEager<T, R> extends AbstractFlowableWithUpst
                                 current = null;
                                 inner.cancel();
                                 cancelAll();
+                                errors.addThrowable(ex);
+                                ex = errors.terminate();
                                 a.onError(ex);
                                 return;
                             }
@@ -333,6 +338,7 @@ public final class FlowableConcatMapEager<T, R> extends AbstractFlowableWithUpst
                         if (e == r) {
                             if (cancelled) {
                                 cancelAll();
+                                errors.tryTerminateAndReport();
                                 return;
                             }
 
--- a/src/main/java/io/reactivex/internal/operators/mixed/FlowableConcatMapCompletable.java
+++ b/src/main/java/io/reactivex/internal/operators/mixed/FlowableConcatMapCompletable.java
@@ -156,6 +156,7 @@ public final class FlowableConcatMapCompletable<T> extends Completable {
             inner.dispose();
             if (getAndIncrement() == 0) {
                 queue.clear();
+                errors.tryTerminateAndReport();
             }
         }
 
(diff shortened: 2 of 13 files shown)


akarnokd:
All of the `concatMap` have this issue and the fix looks like with `switchMap` in #6572. I was planning on fixing all within a few days, but if you want, you could extend this PR.

Erlkoenig90:
Okay, I've started with `FlowableConcatMapEager`, `FlowableConcatMapCompletable`, `FlowableConcatMapMaybe`, `FlowableConcatMapSingle`, `ObservableConcatMap`. I think `errors.terminate` or `errors.tryTerminateAndReport` should only be called if the respective instance is "locked" (via the `AtomicInteger`), or else a concurrently running `drain` loop might leak the `TERMINATED` instance. So, I do that in `dispose` inside the `if (getAndIncrement() == 0)` check. Also, in the `drain` loops, before every `return` there should be a call to any of the `terminate` functions, because such a premature `return` leaves the instance "locked" and prevents `dispose` from collecting the error.

akarnokd:
No, cancellation should not be ordered with onXXX signals.

The error should be terminated in [the cancel/dispose sections](https://github.com/ReactiveX/RxJava/pull/6572/files#diff-9ec65c8417057e78257c5f9c3a385b9cR180) and the tests should [look like this](https://github.com/ReactiveX/RxJava/pull/6572/files#diff-f4eb5ef6eb14238b88049838ae26a1e8R1208).

> might leak the TERMINATED instance

Yes, this is to be avoided [like this](https://github.com/ReactiveX/RxJava/pull/6572/files#diff-d0a55be4952bb5d37363f121ced81967R183) or possibly factored out similar to the `tryTerminateAndReport`.

akarnokd:
Thanks for your efforts, but more work needs to be done in this regard. I'll post a PR fixing all affected operators shortly.

Issue #6611, opened 2019-08-05
3.x: Revise AtomicThrowable.addThrowable usage


akarnokd:
Currently, many (if not all?) usage of `addThrowable` is as follows:

```java
if (errors.addThrowable(ex)) {
   // do something with the error case
} else {
    RxJavaPlugins.onError(ex);
}
```

The plugin error call could be inlined into the method call and renamed to `tryAddThrowable` to ensure all usage places are updated correctly.

RomanWuattier:
Hey, do you need help doing this?

akarnokd:
Each usage location has to be reevaluated and there are thousands of those. It would be easier if I do them.

PR #6612, opened 2019-08-05
3.x: Fix many operators swallowing undeliverable exceptions

### LINKED ISSUE #6587: Inner errors within concatMapSingle vanish if disposed

akarnokd:
Fix many operators to emit the collected exception in case of a cancel/dispose call to the plugin error handler (i.e., becoming undeliverable excptions).

In addition, the terminal event/error delivery has been unified in many other operators so that the terminal-indicator exception is never leaked.

There will be a separate PR about cleaning up the `addThrowable` usage cases (#6611).

Fixes: #6587

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/completable/CompletableMerge.java
+++ b/src/main/java/io/reactivex/internal/operators/completable/CompletableMerge.java
@@ -70,6 +70,7 @@ public final class CompletableMerge extends Completable {
         public void dispose() {
             upstream.cancel();
             set.dispose();
+            error.tryTerminateAndReport();
         }
 
         @Override
@@ -106,7 +107,7 @@ public final class CompletableMerge extends Completable {
 
                 if (error.addThrowable(t)) {
                     if (getAndSet(0) > 0) {
-                        downstream.onError(error.terminate());
+                        error.tryTerminateConsumer(downstream);
                     }
                 } else {
                     RxJavaPlugins.onError(t);
@@ -114,7 +115,7 @@ public final class CompletableMerge extends Completable {
             } else {
                 if (error.addThrowable(t)) {
                     if (decrementAndGet() == 0) {
-                        downstream.onError(error.terminate());
+                        error.tryTerminateConsumer(downstream);
                     }
                 } else {
                     RxJavaPlugins.onError(t);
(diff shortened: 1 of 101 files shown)


vanniktech (approved):

artem-zinnatullin:
Nice

Issue #6615, opened 2019-08-14
Flowable.fromIterable does not work as expected when used in switchMap


adrian-linca:
This issue was reproduces on RX-Java v2.2.6 and v3.0.0-RC1

I have this piece of code:
```java
Flowable
  .range(1, 2)
  .doOnNext(value -> {
    System.out.println("value1: " + value);
  })
  .switchMap(value -> Flowable.just(value * 10))
  .subscribe(
    value -> System.out.println("value2: " + value),
    throwable -> System.out.println("error: " + throwable),
    () -> System.out.println("complete")
   );
```
When run it prints:
value1: 1
value2: 10
value1: 2
value2: 20
complete

Then I have the exact same thing but with Flowable.fromIterable:
```java
Flowable
    .range(1, 2)
    .doOnNext(value -> {
        System.out.println("value1: " + value);
    })
    .switchMap(value -> Flowable.fromIterable(Arrays.asList(value * 10)))
    .subscribe(
        value -> System.out.println("value2: " + value),
        throwable -> System.out.println("error: " + throwable),
        () -> System.out.println("complete")
    );
```
This one prints:
value1: 1
value2: 10

And then nothing, no complete, no error, nothing.

akarnokd:
Hi and thanks for reporting. This is a bug in `switchMap` not playing nice with `fromIterable`. You can work around by applying `.hide()` to `fromIterable`. I'll post a fix shortly.

adrian-linca:
Thanks.

PR #6618, opened 2019-08-14
2.x: Fix switchMap incorrect sync-fusion & error management

akarnokd:
This PR fixes the incorrect request call inside `switchMap` when working with sync-fused sources.

Fixes: #6615


vanniktech (approved):
Déjà vu

Issue #6625, opened 2019-08-21
2.x Calling hasNext() on a disposed blocking iterator hangs forever


BrynCooke:
Calling `hasNext()` on a a blocking iterator never returns if the iterator has been disposed. 

Tested on 2.2.11

To reproduce:
```
Iterator<Integer> iterator = Flowable.fromArray(1, 2, 3).delay(1, TimeUnit.SECONDS).blockingIterable().iterator();

System.out.println("start");
((Disposable)iterator).dispose();
iterator.hasNext(); //Hangs forever
```

akarnokd:
Indeed, if you "hack" into that Iterator it won't work properly as it implementing Disposable is not part of the public intended API. I'll see what I can do about it in 2.x. For 3.x, it probably needs some redesign to work with try-with-resources for example.

akarnokd:
Closing via #6626 & #6627.

PR #6626, opened 2019-08-21
3.x: Fix blockingIterable not unblocking when force-disposed

akarnokd:
When the iterator was cast to `Disposable` and disposed, the subsequent `hasNext` would block indefinitely. That interface is not intended to be part of the public API and `Iterator` in general does not support any form of official cancellation (unlike Stream). This PR makes sure that if that dispose is called, it unblocks the iterator.

A separate PR will be posted for 2.x.

Related #6625

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableIterable.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableIterable.java
@@ -182,7 +182,7 @@ public final class BlockingFlowableIterable<T> implements Iterable<T> {
         @Override
         public void dispose() {
             SubscriptionHelper.cancel(this);
-            signalConsumer(); // just in case it is currently blocking in hasNext
+            signalConsumer(); // Just in case it is currently blocking in hasNext.
         }
 
         @Override
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableIterable.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableIterable.java
@@ -21,7 +21,7 @@ import org.reactivestreams.Subscription;
 
 import io.reactivex.rxjava3.core.*;
 import io.reactivex.rxjava3.disposables.Disposable;
-import io.reactivex.rxjava3.exceptions.MissingBackpressureException;
+import io.reactivex.rxjava3.exceptions.*;
 import io.reactivex.rxjava3.internal.queue.SpscArrayQueue;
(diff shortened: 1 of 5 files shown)


src/main/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableIterable.java
```diff
@@ -175,6 +182,7 @@ public void remove() {
         @Override
         public void dispose() {
             SubscriptionHelper.cancel(this);
+            signalConsumer(); // just in case it is currently blocking in hasNext
```
vanniktech:
```suggestion
            signalConsumer(); // Just in case it is currently blocking in hasNext.
```

src/main/java/io/reactivex/rxjava3/internal/operators/observable/BlockingObservableIterable.java
```diff
@@ -146,6 +153,7 @@ public void remove() {
         @Override
         public void dispose() {
             DisposableHelper.dispose(this);
+            signalConsumer(); // just in case it is currently blocking in hasNext
```
vanniktech:
```suggestion
            signalConsumer(); // Just in case it is currently blocking in hasNext.
```

src/test/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableToIteratorTest.java
```diff
@@ -163,4 +167,28 @@ protected void subscribeActual(Subscriber<? super Integer> s) {
 
         it.next();
     }
+
+    @Test(expected = NoSuchElementException.class)
+    public void disposedIteratorHasNextReturns() {
+        Iterator<Integer> it = PublishProcessor.<Integer>create()
+                .blockingIterable().iterator();
+        ((Disposable)it).dispose();
```
vanniktech, line 175: Is this really public API? Feels weird and awkward to use

vanniktech (approved):

src/test/java/io/reactivex/rxjava3/internal/operators/flowable/BlockingFlowableToIteratorTest.java
akarnokd: It is not the intended usage but force-casting "hacks" the Iterator would result in hangs.

There is no good way to have cancellation through Iterable/Iterator as it was never intended to work with resources. For-each doesn't work with iterators and add try-with-resources over Iterable requires tracking all Iterator spawns so that all of them can be cancelled if the for-each/iteration is broken out.

vanniktech (approved):

PR #6629, opened 2019-08-21
2.x: Fix refCount() not resetting when cross-canceled

### LINKED ISSUE #6608: ReplaySubject and buffer with debounce not emitting values after first subscription
The issue described below started with version 2.2.10 and continues with 2.2.11. It works with 2.2.7 through 2.2.9.

I have a test where a set a `ReplaySubject` like so:

```
locationPublisher = ReplaySubject.create()
locationPublisher.onNext(mockLocation)
locationPublisher.onComplete()

locationObservable = locationPublisher.share()
```

Then, I subscribe twice to the observable (I'm simplifying the test case just to demonstrate the bug):
```
        locationObservable
            .buffer(locationObservable.debounce(5, TimeUnit.SECONDS))
            .test()
            .assertValueCount(2)


        locationObservable
            .buffer(locationObservable.debounce(5, TimeUnit.SECONDS))
            .test()
            .assertValueCount(2)
```
When I run the above test, it fails on the second test subscription on RxJava 2.2.10 or 2.2.11:
> java.lang.AssertionError: Value counts differ; expected: 2 but was: 1 (latch = 0, values = 1, errors = 0, completions = 1)
Expected :2 
Actual   :1 (latch = 0, values = 1, errors = 0, completions = 1)

akarnokd:
This PR fixes the issue with `refCount` not resetting the connection when the termination triggers cross-cancellation over it.

Fixes #6608

The fix is more involved than #6609 because how 2.x uses two `publish()` implementation internally due to bugfix #6505. The old/classic implementation does not fail #6608 but the newer implementation fails #6608. If the fix is applied unconditionally, the old/classic implementation fails an older unit test verifying an error allows reconnection. Therefore, the PR checks and applies the new code path only if `refCount` isn't talking to the classic publish implementation.

As a reminder #6609 for 3.x has a redesigned `Connectable` with a much more clearer reset semantics and thus the restructuring of the termination handling had no trouble passing the aforementioned error-allows-reconnect unit test.

### DIFF
--- a/src/test/java/io/reactivex/internal/operators/observable/ObservableRefCountTest.java
+++ b/src/test/java/io/reactivex/internal/operators/observable/ObservableRefCountTest.java
@@ -532,7 +532,7 @@ public class ObservableRefCountTest {
         to2.assertValue(30);
     }
 
-    @Test(timeout = 10000)
+    @Test//(timeout = 10000)
     public void testUpstreamErrorAllowsRetry() throws InterruptedException {
         final AtomicInteger intervalSubscribed = new AtomicInteger();
         Observable<String> interval =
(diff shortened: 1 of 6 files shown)


src/test/java/io/reactivex/internal/operators/observable/ObservableRefCountTest.java
```diff
@@ -532,7 +532,7 @@ public Integer apply(Integer t1, Integer t2) {
         to2.assertValue(30);
     }
 
-    @Test(timeout = 10000)
+    @Test//(timeout = 10000)
```
vanniktech: Just remove it?

vanniktech (approved):

src/test/java/io/reactivex/internal/operators/observable/ObservableRefCountTest.java
akarnokd: Forgot to undo.

PR #6632, opened 2019-08-22
3.x: Upgrade Gradle & dependency versions

akarnokd:
This PR upgrades to:

- Gradle 5.6
- TestNG 7.0.0
- Mockito 3.0.0
- JMH 1.21
- JMH plugin 0.5.0-rc-2 (required by Gradle 5.5+)
- License 0.15.0
- Jacoco 0.8.4
- Guava 28

In addition, the `osgi` plugin has been deprecated in Gradle 5 and has been removed in Gradle 6, therefore a separate plugin called `bnd` is required to produce the necessary jar properties.
(Also the MockitoJunitRunner has been moved to a new location, hence the package change.)

This Gradle version should support working on a Java 11 runtime when executing `gradlew build`.

Note that this may require further "live" testing as the auto-release of snapshots as well as auto-release can't be properly tested offline.


akarnokd:
Snapshotting worked:
- https://travis-ci.org/ReactiveX/RxJava/builds/575569079#L4644
- https://oss.jfrog.org/libs-snapshot/io/reactivex/rxjava3/rxjava/3.0.0-SNAPSHOT/

Issue #6634, opened 2019-08-27
3.x: improve combineLatest method names


vanniktech:
Calling `combineLatest` from Kotlin is a bit awkward. The function has a lot of different parameters, including Iterable, Array as well as 2-x call sites for passing multiple Observables.

Is it possible to prefix these so we have some kind of `combineLatestFromIterable` which takes the `Iterable<T>`:

```diff
-public static <T, R> Observable<R> combineLatest(Iterable<? extends ObservableSource<? extends T>> sources, Function<? super Object[], ? extends R> combiner) {
+public static <T, R> Observable<R> combineLatestFromIterable(Iterable<? extends ObservableSource<? extends T>> sources, Function<? super Object[], ? extends R> combiner) {
```

and then `combineLatest` which takes the 2-x call sites.

```java
public static <T1, T2, R> Observable<R> combineLatest(
            ObservableSource<? extends T1> source1, ObservableSource<? extends T2> source2,
            BiFunction<? super T1, ? super T2, ? extends R> combiner) {
```

The array call we could call `combineLatestFromArray`.

```diff
-public static <T, R> Observable<R> combineLatest(ObservableSource<? extends T>[] sources, Function<? super Object[], ? extends R> combiner) {
+public static <T, R> Observable<R> combineLatestFromArray(ObservableSource<? extends T>[] sources, Function<? super Object[], ? extends R> combiner) {
```
What should we do with?

```java
public static <T, R> Observable<R> combineLatest(Function<? super Object[], ? extends R> combiner, int bufferSize, ObservableSource<? extends T>... sources) {
```

Is it really required? Callers could just use the array overload.

`combineLatestDelayError` is equivalent and could get the method name improvements as well.

Other methods, such as concat/merge could benefit from this change too.

akarnokd:
`combineLatestArray` would be reasonable and match the other `xArray` method names. What's the conflict between `combineLatest(Iterable, F)` and `combineLatest(O1, O2, F2)`?

The varargs variant could be removed.

vanniktech:
There should not be any issues with `combineLatest(Iterable, F)` and `combineLatest(O1, O2, F2)`. I'd just go for consistency. Looking at `startWithIterable`.

akarnokd:
`startWithIterable` was disambiguated due to `startWith(Publisher)` and/or `startWith(T)` so I'd say is an exception. So if the main trouble is with `combineLatest(O[], F)` then let's rename that to `combineLatestArray`.

[...]

PR #6635, opened 2019-08-27
3.x: Remove vararg overloads for combineLatest in Observable + Flowable

vanniktech:
Started removing the vararg function from #6634


vanniktech:
Hmm after looking at the tests the varag variant for `combineLatestDelayError` might be useful since there are no 2-x overloads available. What do you think?

akarnokd:
Let me guess, IntelliJ refactor?

akarnokd:
The varargs is a convenience over the array variant so `combineLatestDelayError`s can go.

akarnokd (approved):

PR #6637, opened 2019-08-28
3.x: Avoid using System.getProperties() due to security restrictions

### LINKED ISSUE #6636: System.properties access restriction.
I'm getting `access denied ("java.util.PropertyPermission" "*" "read,write")` error when using the IO Scheduler. 

Its coming from: 
https://github.com/ReactiveX/RxJava/blob/13772a173cb6a59643dfd5eaf023a1412f804096/src/main/java/io/reactivex/internal/schedulers/SchedulerPoolFactory.java#L93

```
static {
        Properties properties = System.getProperties();

        PurgeProperties pp = new PurgeProperties();
        pp.load(properties);

        PURGE_ENABLED = pp.purgeEnable;
        PURGE_PERIOD_SECONDS = pp.purgePeriod;

        start();
    }
```

All this seems to be from what I could see is that it's trying to get purge options from the system properties. 

```
static final class PurgeProperties {

        boolean purgeEnable;

        int purgePeriod;

        void load(Properties properties) {
            if (properties.containsKey(PURGE_ENABLED_KEY)) {
                purgeEnable = Boolean.parseBoolean(properties.getProperty(PURGE_ENABLED_KEY));
            } else {
                purgeEnable = true;
            }

            if (purgeEnable && properties.containsKey(PURGE_PERIOD_SECONDS_KEY)) {
                try {
                    purgePeriod = Integer.parseInt(properties.getProperty(PURGE_PERIOD_SECONDS_KEY));
                } catch (NumberFormatException ex) {
                    purgePeriod = 1;
                }
            } else {
                purgePeriod = 1;
            }
        }
    }
```

In my system I can't allow for read/write access to all System properties. Can this be changed to use `System.getProperty(PURGE_ENABLED_KEY)` or the like as I can allow for access to specific keys.

akarnokd:
Change the code in the `SchedulerPoolFactory` to use `System.getProperty` directly. Also fix property names to `rx3.` as now both v2 and v3 could run together.

Fixes #6636

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -44,7 +44,7 @@ import io.reactivex.rxjava3.subscribers.*;
  * Reactive Streams implementations.
  * <p>
  * The Flowable hosts the default buffer size of 128 elements for operators, accessible via {@link #bufferSize()},
- * that can be overridden globally via the system parameter {@code rx2.buffer-size}. Most operators, however, have
+ * that can be overridden globally via the system parameter {@code rx3.buffer-size}. Most operators, however, have
  * overloads that allow setting their internal buffer size explicitly.
  * <p>
  * The documentation for this class makes use of marble diagrams. The following legend explains these diagrams:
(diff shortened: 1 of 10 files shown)


cxbiao:
ok

vanniktech (approved):

PR #6620, opened 2019-08-16
3.x: elementAt, first - constrain upstream requests 

davidmoten:
As discussed in #6569 this PR constrains upstream requests for the `elementAt` and `first` overloads. Other operators will be covered in other PRs.

Issue #6569, opened 2019-07-12
3.x: unbounded requests from first, take and others


davidmoten:
In 2.x we talked about some surprising request patterns from operators like `first`, `take` and others where despite only one or a limited number of items being needed, Long.MAX_VALUE was requested of upstream and then cancelled after the desired number arrived (#5077). I believe this was a micro-optimization performance boost that improved the Scrabble benchmarks. Any change to the pattern was rejected based on it being a breaking API change.

In essence I'd like us not to be opinionated about the effect on the upstream of over-requesting (particularly over a network boundary). By over-requesting we are fundamentally *losing information that can be useful to optimizing upstream processing*. An example that springs to mind is that requesting a large number may be translated to an api call to a remote upstream that does a full sort (O(nlogn)) whereas requesting only one can be implemented upstream with a max scan (O(n)). This of course assumes one and only request to create the stream so is not a run-of-the-mill streaming case. I'd also suggest we are not opinionated about the ability of upstream to respond to cancellation (upstream may be performing cpu intensive actions in third-party libraries that aren't cancellable).

I think the effect on the benchmarks of reverting to naturally bounded requests where obvious (`first`, `take`, etc) will be very small.

 Can we revisit this one for 3.x?

akarnokd:
Okay, let's make limit the default and change first/single. 

What about the lambda subscribe?

davidmoten:
Thanks @akarnokd, so I'm suggesting we constrain upstream requests for:

* `elementAt`
* `elementAtOrError`
* `first` (calls `elementAt`)
* `firstOrError` (calls `elementAt`)
* `take`
* `single(default)`
* `singleOrError`
* `blockingSingle(default)` calls `single(default)`
* `blockingSingle` (calls `singleOrError`)

`limit` can then be an alias for `take`.

>What about the lambda subscribe?

You've lost me, what are you referring to?

akarnokd:
There is no reason to have `limit` separately then.

[lambda subscribe](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/Flowable.java#L14925) requests max value.

davidmoten:
>lambda subscribe requests max value.

Yes, that seems reasonable to me. Why would it be otherwise?

davidmoten:
> There is no reason to have limit separately then.

ok by me

akarnokd:
[...]

PR #6650, opened 2019-09-12
3.x: constrain upstream requests from take, remove limit operator

davidmoten:
As per discussion in #6569, this PR constrains upstream requests from the `take` operator and removes the `limit` operator.

I've also added a couple of unit tests for coverage of `elementAt` that I didn't include in #6620.


src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableTake.java
```diff
@@ -103,13 +103,29 @@ public void request(long n) {
             if (!SubscriptionHelper.validate(n)) {
                 return;
             }
-            if (!get() && compareAndSet(false, true)) {
-                if (n >= limit) {
-                    upstream.request(Long.MAX_VALUE);
+            while (true) {
```
akarnokd: Why not use the limit() internals?

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -11212,52 +11212,6 @@ public final Completable ignoreElements() {
         return RxJavaPlugins.onAssembly(new FlowableLift<R, T>(this, lifter));
     }
 
-    /**
-     * Limits both the number of upstream items (after which the sequence completes)
-     * and the total downstream request amount requested from the upstream to
-     * possibly prevent the creation of excess items by the upstream.
-     * <p>
```
akarnokd, line 11219: The Javadoc of take should contain these extra information.

akarnokd:
Why not simply delete `take` and rename `limit`? That way, the tests and documentation of limit would remain intact.

davidmoten:
Indeed! Will do.

davidmoten:
Copied code from `limit` to `take` and fixed javadoc.

akarnokd (approved):

vanniktech (approved):

PR #6674, opened 2019-10-17
3.x: Fix window(time) possible interrupts while terminating

### LINKED ISSUE #6672: Question: Why a running computation thread got interrupted while window is finished
Why an active computation thread got interrupted while window is finished?
```
        Observable
            .intervalRange(0, 2, 0, 1, TimeUnit.SECONDS)
            .window(1, TimeUnit.SECONDS)
            .doOnNext(i -> {
                long start = System.currentTimeMillis();
                while (System.currentTimeMillis() < start+1500) {
                }
                if (Thread.interrupted()) {
                    System.out.println("WTF? Thread is interrupted!");
                }
            })
            .ignoreElements()
            .blockingAwait();
```
Is it a bug?

akarnokd:
Fix the case in `window(time)` variants where the timer thread is busy with window emission and the upstream terminates on some other thread, the window emission is interrupted.

~~So instead of disposing the timer/worker right after the upstream termination, a `DISPOSE` message is queued up. Thus any ongoing drain loop from the timer thread can cleanup gracefully.~~

After some additional considerations, there is no need for `DISPOSE`. The drain loop will take care of disposing the timer and the main downstream can simply be terminated.

The 2.x fix will be in a separate PR shortly.

Fixes #6672

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableWindowTimed.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableWindowTimed.java
@@ -90,6 +90,8 @@ public final class FlowableWindowTimed<T> extends AbstractFlowableWithUpstream<T
 
         static final Object NEXT = new Object();
 
+        static final Object DISPOSE = new Object();
+
         volatile boolean terminated;
 
         WindowExactUnboundedSubscriber(Subscriber<? super Flowable<T>> actual, long timespan, TimeUnit unit,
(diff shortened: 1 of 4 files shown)


akarnokd:
The coverage complains about the new additions because it is very difficult to establish an interleaving that would skip over the done part the right time and find `done == false` and `o == DISPOSE`. I.e., executing the entire `onComplete` while the drain loop is between reading `done` and polling the queue.

vanniktech:
Let me know once you're done so I can review.

akarnokd:
That's it. Can't make the coverage better.

vanniktech (approved):

PR #6692, opened 2019-10-29
Add BlockHound integration

bsideup:
[BlockHound](http://github.com/reactor/BlockHound) is a tool for JVM to detect blocking calls in non-blocking threads.

It provides an SPI to integrate with, although it also comes with built-in RxJava2 integration. Since the introduction of SPI, we encourage JVM frameworks to provide an implementation of the SPI.
Netty project [just added](https://github.com/netty/netty/pull/9687) such integration, and even Project Reactor itself [uses the SPI](https://github.com/reactor/reactor-core/pull/1682) to integrate with BlockHound.
The main motivation is to be able to adjust the integration if the internals are changed, e.g. if RxJava decides to use some other marker interface for its threads.

BlockHound is Java 8+, but the SPI class is only loaded when BlockHound installs the integrations (using Java's `ServiceLoader`), hence there should be no problem with RxJava3's baseline (which is Java 6 at the moment). There are no Java 8 bytecodes in that class too, only JDK 8 class references. Netty is Java 6 too and it was fine for them.

A separate test source is used because BlockHound is global (since it is implemented as a Java agent) and cannot be disabled after the installation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,8 @@ buildscript {
   ext.jfrogExtractorVersion = "4.5.2"
   ext.bndVersion = "4.2.0"
   ext.checkstyleVersion = "6.19"
+  ext.blockhoundVersion = "1.0.1.RELEASE"
+  ext.testsetsVersion = "2.2.1"
 
   // --------------------------------------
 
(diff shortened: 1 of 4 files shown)


akarnokd:
No. I don't want to add Blockhound integration.

lanwen:
@akarnokd just tested this helper with micronaut and found it quite useful, can I ask the reason of rejection?

bsideup:
@akarnokd I understand that you're busy, but will appreciate a bit longer answer, since I believe that the community will benefit from RxJava3 getting a support for BlockHound, which is (despite the organization) is neutral, does not require you to have a dependency on it and does not add *any* performance overhead to the framework (unless the agent is explicitly enabled by the user, and it is a very low one, there are some public benchmarks too)

akarnokd:
The PR code demonstrates there is no need to add BlockHound to the library directly; any external companion library can setup that Thread identification. Also the PR doesn't work because the BlockHound API requires Java 8 which we don't support and may not support for months, but as an external support, users can rely on early-access desugaring on Android to workaround.

bsideup:
@akarnokd ok, thank you for clarifying the rejection, I respect your decision as a maintainer.

PR #6683, opened 2019-10-23
Update README.md

nithasha-samad:
Thank you for contributing to RxJava. Before pressing the "Create Pull Request" button, please consider the following points:

  - [ ] Please give a description about what and why you are contributing, even if it's trivial.

  - [ ] Please include the issue list number(s) or other PR numbers in the description if you are contributing in response to those.

  - [ ] Please include a reasonable set of unit tests if you contribute new code or change an existing one. If you contribute an operator, (if applicable) please make sure you have tests for working with an `empty`, `just`, `range` of values as well as an `error` source, with and/or without backpressure and see if unsubscription/cancellation propagates correctly.

Issue #6698, opened 2019-11-01
Version incompatibility with Android Studio when R8 is disabled


RPCarter53:
We use RxJava in our app and, following an update to version 2.2.13, a release build with R8 disabled fails. We have a need to disable R8 as the Amazon In-App billing libraries are currently incompatible and Amazon have not yet release a fix. I have gone back to RxJava 2.2.10 and the build works again.

akarnokd:
How exactly doesn't work for you? I'm not that familiar with the platform and infrastructure, could you post a Android Studio project that demonstrates the problem?

RPCarter53:
David,
	The project in question is quite large and not something that I would be allowed to share I’m  afraid. Essentially R8 is an alternative to ProGuard for code shrinking and obfuscation, it is the default tool with Android Studio and is still controlled by Proguard configuration files. If I disable R8, forcing the build to fall back to Proguard I get a exception in the build process, I do not get this with version 2.2.10. Can you see if the ProGuard definitions have changed for RxJava between these two releases?

Regards

Roy

akarnokd:
RxJava 2 doesn't need any ProGuard settings.

> The project in question is quite large and not something that I would be allowed to share I’m  afraid

If you created a minimal project with your setup, does that also fail to build with 2.2.13 and not with 2.2.10?

> If I disable R8, forcing the build to fall back to Proguard I get a exception in the build process

Are you allowed to share the exception at least?

RPCarter53:
David,
	For clarification 2.2.12 builds OK 2.2.13 fails.

	This actually causes an exception in the build process as detailed below and it is in the area of Proguard transformation. I also have warnings about “org.reactivestreams.FlowAdapters$FlowPublisherFromReactive: can't find referenced class java.util.concurrent.Flow$Publisher”. I believe concurrent.Flow is a Java 9 addition which Android does not as yet support. 

Cheers

Roy


org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:transformClassesAndResourcesWithProguardForFreeAmazonRelease'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:151)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:148)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:191)
[...]

PR #6699, opened 2019-11-01
Readme: Cherry-pick fixes from #6683, add ProGuard workaround entry

akarnokd:
- Pick the accepted changes from #6683.
- Add entry about ProGuard and Reactive Streams JAR warning issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ It extends the [observer pattern](http://en.wikipedia.org/wiki/Observer_pattern)
 - fixed API mistakes and many limits of RxJava 2
 - intended to be a replacement for RxJava 2 with relatively few binary incompatible changes
 - Java 8 lambda-friendly API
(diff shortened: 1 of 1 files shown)


vanniktech (commented):
Would be nice to embed them https://twitter.com/jakewharton/status/1004401938467876865?lang=en

akarnokd:
If I understand #6698, the issue was disabling R8, which had no problems to begin with, and the default ProGuard complaining. Looks like this embed would work only with R8.

vanniktech:
AFAIK, this only does work for R8 yes. Would still be super beneficial as it's the default shrinker for Android.

akarnokd:
Does R8 even complain without those `-dontwarn` entries?

vanniktech:
I haven't built a release version with the latest RxJava version. If I do, and I see those warnings, I'll come back to this issue.

PR #6700, opened 2019-11-02
3.x: Update javadoc for observeOn to mention its eagerness

### LINKED ISSUE #6697: 3.x: update javadoc for observeOn to mention its eagerness
`Flowable.observeOn` and `Observable.observeOn` try to do as much event delivery as they can.
Add a section explaining this property to all these multi-valued `observeOn` overloads:

"This operator keeps emitting as many signals as it can on the given Scheduler's Worker thread, which may result in a longer than expected occupation of this thread. In other terms, it does not allow per-signal fairness in case the worker runs on a shared underlying thread. If such fairness and signal/work interleaving is preferred, use the `delay` operator with zero time instead."

Also link to the `@see #delay(long,TimeUnit,Scheduler)`.

JosemyDuarte:
Update javadoc to reflect the eagerness property of `observeOn` in `Flowable` and `Observable` classes.

Resolves #6697

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -11418,6 +11418,11 @@ public abstract class Flowable<T> implements Publisher<T> {
      * Modifies a Publisher to perform its emissions and notifications on a specified {@link Scheduler},
      * asynchronously with a bounded buffer and optionally delays onError notifications.
      * <p>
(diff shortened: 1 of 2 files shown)


akarnokd (changes requested):
Please put the text after the image and the see as last.

akarnokd (approved):

akarnokd:
Thanks! Much appreciated!

Would you like to backport the changes to the 2.x branch?

PR #6704, opened 2019-11-07
2.x: Add ProGuard rule to avoid j.u.c.Flow warnings due to RS 1.0.3

akarnokd:
Add rule to ignore `java.util.concurrent.Flow` when using Android's R8 to prevent warnings about the `FlowAdapter.class` contained within the Reactive Streams JAR file.

### DIFF
--- a/src/main/resources/META-INF/proguard/rxjava2.pro
+++ b/src/main/resources/META-INF/proguard/rxjava2.pro
@@ -1,0 +1,1 @@
+-dontwarn java.util.concurrent.Flow


vanniktech (approved):

akarnokd:
I'll merge this and pull in the snapshot in a Android project.

akarnokd:
It didn't work.

image

akarnokd:
This was AS 3.5. If I put the `-dontwarn` into the `proguard-rules.pro` of the project, it didn't work either.

Upgraded to AS 3.5.2. If I put `-dontwarn` into the `proguard-rules.pro` it works. Having only the `rxjava2.pro` still doesn't.

Edit. Running clean and make, the warnings are there. Running make again results in no warnings.

PR #6705, opened 2019-11-07
2.x: ProGuard R8 dont warn Flow* classes

akarnokd:
#6704 didn't work, probably because the filter has to match a filename `java.util.concurrent.Flow$Publisher`, not some parent class name. 

Having `-dontwarn java.util.concurrent.Flow*` did work in a test project's proguard file locally.


vanniktech (approved):
Ups. Yes. You need that `*`

akarnokd:
This works.

PR #6740, opened 2019-12-03
3.x: Change Flowable.groupBy to signal MBE instead of possibly hanging

### LINKED ISSUE #6641: 3.x: Change the behavior of Flowable.groupBy to signal MBE if no main requests
Currently, the `Flowable.groupBy` queues up groups until the downstream requests more groups. Unfortunately, this can lead to hangs because if there are more groups than requested, active groups may starve out as they don't trigger replenishments or those replenishments result in more groups to be created. Example:

```java
Flowable.range(1, 1000)
.groupBy(v -> v)
.flatMap(v -> v, 16)
.test()
.assertValueCount(1000)
.assertComplete();
```
This case works if `flatMap`'s concurrency level is greater or equal to the expected group count.

I propose changing the `Flowable.groupBy` operator to signal `MissingBackpressureException` if the direct downstream is not ready to receive more groups.

akarnokd:
This PR changes the backpressure behavior of `Flowable.groupBy` to signal `MissingBackpressureException` instead of silently hanging if the produced groups are not ready to be accepted by the downstream. 

This can happen if one `flatMap`s a `groupBy` but there are more groups produced than the concurrency level of `flatMap`. Since replenishment is tied to item consumption from the groups, not consuming them can result in none of the groups receiving any further items and the whole operator hangs.

The following changes have been applied:

- Removed the queue from the main operator since it will now try to emit directly and not buffer groups.
- The main `Flowable`, lacking a queue, no longer supports operator fusion. Tests checking this property have been removed as well.
- When a group is drained, consumed items are replenished in batch if possible. Detecting a cancellation will also trigger a replenishment.
- When a group is pulled (fusion mode), now all `pull`, `isEmpty` and `clear` will trigger replenishment so that other groups can make progress too.
- Unit tests have been modified to have large enough bufferSize/prefetch amounts to allow them to pass.

Fixes #6641

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -19,7 +19,7 @@ import org.reactivestreams.*;
 
 import io.reactivex.rxjava3.annotations.*;
 import io.reactivex.rxjava3.disposables.Disposable;
-import io.reactivex.rxjava3.exceptions.Exceptions;
+import io.reactivex.rxjava3.exceptions.*;
 import io.reactivex.rxjava3.flowables.*;
 import io.reactivex.rxjava3.functions.*;
 import io.reactivex.rxjava3.internal.functions.*;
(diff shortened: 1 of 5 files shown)


vanniktech (approved):
Never used that operator and I remember @davidmoten had lots of use cases so let's wait for his approval?

akarnokd:
Sure.

akarnokd:
Let's merge this into the release and get feedback from the field.

PR #6754, opened 2019-12-10
2.x: Zip, CombineLatest, and Amb operators throw when supplied with ObservableSource implementation that doesn't subclass Observable

### LINKED ISSUE #6753: 2.x: Zip, CombineLatest, and Amb operators throw when supplied with ObservableSource implementation that doesn't subclass Observable
Issue found in version 2.2.15.

The Zip, CombineLatest, and Amb operators of the `Observable` class each have an overload that accepts an `Iterable<ObservableSource>` and a `Function<Array<Any>, R>` where `R` is a type parameter. The problem is that the elements of the `Iterable` are copied into an `Observable[]` internally by the `Observer` classes implementing the operators. This causes an `ArrayStoreException` if one supplies any custom `ObservableSource` implementation that doesn't subclass `Observable`.

The equivalent functions for the other stream types don't have this restriction. I.e, the Zip implementation in `Maybe` copies the source values into a `MaybeSource[]` and avoids this problem.

Example kotlin code which reproduces the issue:
```kotlin
import io.reactivex.Observable
import io.reactivex.ObservableSource

val sourceA = ObservableSource<String> { Observable.just("A").subscribe(it) }
val sourceB = ObservableSource<String> { Observable.just("b", "B").subscribe(it) }

val combined =
    Observable.combineLatest(
       listOf(sourceA, sourceB),
       { (it[0] as String) + (it[1] as String) }
    )

fun main() {
    combined.subscribe(::println)
}
```
A simple workaround is to wrap any `ObservableSource`. To fix the above example pass `listOf(sourceA, sourceB).map { Observable.wrap(it) }` to `combineLatest`.

mgsholte:
Fixes #6753

### DIFF
--- a/src/main/java/io/reactivex/internal/operators/observable/ObservableAmb.java
+++ b/src/main/java/io/reactivex/internal/operators/observable/ObservableAmb.java
@@ -36,7 +36,7 @@ public final class ObservableAmb<T> extends Observable<T> {
         ObservableSource<? extends T>[] sources = this.sources;
         int count = 0;
         if (sources == null) {
-            sources = new Observable[8];
+            sources = new ObservableSource[8];
             try {
                 for (ObservableSource<? extends T> p : sourcesIterable) {
                     if (p == null) {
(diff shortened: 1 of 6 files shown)


akarnokd:
Please add unit tests so this doesn't happen again in case of any future change.

mgsholte:
Added unit tests.

akarnokd (approved):

PR #6758, opened 2019-12-12
3.x: Fix Flowable.window (size, time) cancellation and abandonment

akarnokd:
This PR fixes the `Flowable.window` operator (with size and time boundaries) so that

- cancelling the inner windows allows cancelling the upstream once neither the main output nor other windows are being consumed further
- ignoring a window still allows cancelling the upstream.

These fixes are necessary similar to how `groupBy` had to be updated in #6740.

The publisher-boundary and start-stop publisher boundary as well as the `Observable` variants will be updated in subsequent PRs due to the sheer number of changes.


vanniktech (commented):
👍

Issue #6695, opened 2019-10-30
3.x release vs. Java 8 support on Android 


akarnokd:
Recently, the news came that [Android Studio 4 will support](https://developer.android.com/studio/preview/features#j8-desugar) a so-called "desugaring" of Java 8 APIs (i.e., some kind of auto-backport for target devices not supporting newer API levels as I understand). This would allow RxJava 3 to increase the baseline support to Java 8 on its surface API:

- `Stream` support
- Stream `Collector`s support
- `Optional` support
- `CompletableFuture` support
- Use site non-null annotations

Some features won't be supported:

- `java.time.Duration`: allocates, we'll stick to time+unit
- `java.util.function`: these can't throw `Throwable`s, overloads would create bloat and/or ambiguity

The main issue with Android's Java 8 support is timing related to the pledged release schedule of RxJava 3.x, **end of December 2019**. AGP 4 may not release for months, inclining us to:

- Wait for AGP's release, effectively parking RxJava 3 for months as release candidate. Drawback: AGP may slip a lot.
- Release on schedule with Java 6 baseline. Drawback: AGP may release shortly after and thus
  - we'd be stuck on 3.x for years;
  - we'd have to start working on 4.x undermining 3.x in a sense, also adding more maintenance burden;
  - we release 3.1 with baseline 8, abandoning 3.0 and possibly violating some pledge;
- Release on schedule with Java 8 baseline. Drawback: 3.x improvements won't be available to many at release.
- Release on schedule with Java 8 bytecode output but no 8 API use, add those in 3.1. Drawback: despite best efforts, this could cause problems with existing desugaring/tooling.

The same issue is likely to arise with the next step, Java 9 support on Android. From our perspective, support for the Flow.* interfaces would be just enough as none of the other 9 features would work (VarHandles - field reflection anomalies?, Cleaner - portable?, Modules?). Beyond 9, the next major change to RxJava would have to come due to value types (no version ETA yet). Loom/Fibers is likely simply an API expansion, not a fundamental change.

Suggestions, ideas welcome.

slisaasquatch:
Just some of my thoughts.

I'm not a fan of the 1st option, because literally nobody wins. People can still use RxJava 2 before AGP 4 comes out. I think the 3rd option makes a lot more sense than the 1st.

For Java 9 support, only supporting Flow.* interfaces to me doesn't justify requiring Java 9, as interoperating between the interfaces is pretty easy.

[...]

PR #6765, opened 2019-12-17
3.x: [Java 8] Upgrade to Java 8, add Flowable.fromX operators

akarnokd:
This PR upgrades RxJava to use and compile with Java 8.

- Build target set to 8
- AnimalSniffer set to 8
- JavaDocs link to the official JDK set to 8
- Implemented operators:
  - `Flowable.fromOptional`
  - `Flowable.fromCompletionStage`
  - `Flowable.fromStream`
- Applied `@SafeVarargs` to many methods
- Created marbles

Related #6695

image
image
image

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -18651,6 +18651,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.FULL)
     @SchedulerSupport(SchedulerSupport.NONE)
+    @NonNull
     public static <T> Flowable<T> fromOptional(@NonNull Optional<T> optional) {
         ObjectHelper.requireNonNull(optional, "optional is null");
         return optional.map(Flowable::just).orElseGet(Flowable::empty);
@@ -18687,6 +18688,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.FULL)
     @SchedulerSupport(SchedulerSupport.NONE)
+    @NonNull
     public static <T> Flowable<T> fromCompletionStage(@NonNull CompletionStage<T> stage) {
         ObjectHelper.requireNonNull(stage, "stage is null");
         return RxJavaPlugins.onAssembly(new FlowableFromCompletionStage<>(stage));
@@ -18732,6 +18734,7 @@ public abstract class Flowable<T> implements Publisher<T> {
     @CheckReturnValue
     @BackpressureSupport(BackpressureKind.FULL)
     @SchedulerSupport(SchedulerSupport.NONE)
+    @NonNull
     public static <T> Flowable<T> fromStream(@NonNull Stream<T> stream) {
         ObjectHelper.requireNonNull(stream, "stream is null");
         return RxJavaPlugins.onAssembly(new FlowableFromStream<>(stream));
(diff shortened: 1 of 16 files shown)


src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -18620,4 +18619,121 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
...
+     * @param optional the optional value to convert into a {@code Flowable}
+     * @return the new Flowable instance
+     * @see #just(Object)
+     * @see #empty()
+     * @since 3.0.0
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public static <T> Flowable<T> fromOptional(@NonNull Optional<T> optional) {
```
JakeWharton, line 18654: +NonNull for return type

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -18620,4 +18619,121 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
...
+     * </dl>
+     * @param <T> the element type of the CompletionStage
+     * @param stage the CompletionStage to convert to Flowable and signal its terminal value or error
+     * @return the new Flowable instance
+     * @since 3.0.0
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public static <T> Flowable<T> fromCompletionStage(@NonNull CompletionStage<T> stage) {
```
JakeWharton, line 18690: +NonNull for return type

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -18620,4 +18619,121 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
...
+     * @param <T> the element type of the source {@code Stream}
+     * @param stream the {@code Stream} of values to emit
+     * @return the new Flowable instance
+     * @since 3.0.0
+     * @see #fromIterable(Iterable)
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    public static <T> Flowable<T> fromStream(@NonNull Stream<T> stream) {
```
JakeWharton, line 18735: +NonNull for return type

JakeWharton (approved):

src/main/java/io/reactivex/rxjava3/core/Flowable.java
akarnokd: Done. I suppose it should be added to all other methods too.

vanniktech: That plus I think Jake meant `<T> Flowable<@NonNull T>`

akarnokd: Okay, the other methods will get these in some other PRs.

src/main/java/io/reactivex/rxjava3/annotations/NonNull.java
```diff
@@ -19,10 +19,10 @@
 import java.lang.annotation.*;
 
 /**
- * Indicates that a field/parameter/variable/return type is never null.
```
vanniktech, line 22: Do we want to mention type use too?

vanniktech (approved):

src/main/java/io/reactivex/rxjava3/annotations/NonNull.java
akarnokd: It's not that important; type parameter should imply its use too.

He-Pin:
With this move, there are many we can help.

He-Pin:
Should remove the usage of `Optional.of` with `Optional.ofNullable`.

PR #6769, opened 2019-12-17
[3.x] Remove deprecated method from 2.x

JakeWharton:


akarnokd:
I kept this so that when writing a standard method, one doesn't accidentally null-check an auto-boxed primitive.

JakeWharton:
I see. I was about to remove `ObjectsHelper.requireNonNull` in favor of `Objects.requireNonNull`.

akarnokd:
Well, if all uses get replaced with `Objects.requireNonNull`, this overload wouldn't come into play. I suppose `Objects.requireNonNull` is desugared and inlined across all API levels already, right?

JakeWharton:
Yep! It's actually the first API that was desugared and has been supported since AGP 3.2.

akarnokd:
Okay then.

PR #6770, opened 2019-12-17
[3.x] Replace some ObjecsHelpers methods with Java 8 APIs

JakeWharton:


akarnokd (approved):

JakeWharton:
These are all desugared as of AGP 3.5.

PR #6774, opened 2019-12-17
Extends function

He-Pin:
on top of https://github.com/ReactiveX/RxJava/pull/6773


akarnokd:
No. These were kept separately deliberately so that converters and transformers for different types can be implemented in the same class.

PR #6775, opened 2019-12-18
3.x: [Java 8] Implement mapOptional, collector, first/last/single stage

akarnokd:
This PR implements the following `Flowable` operators:

- `mapOptional` - essentially a map-filter via `Optional`s.
- `collect(Collector)` - use the Java 8 `Collector` API, see [`Collectors`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html).
- `firstStage` - return the first item or a default value as a `CompletionStage`.
- `firstOrErrorStage` - return the first item or error as a `CompletionStage`.
- `singleStage` - return the only item or a default value as a `CompletionStage`.
- `singleOrErrorStage` - return the only item or error as a `CompletionStage`.
- `lastStage` - return the last item or a default value as a `CompletionStage`.
- `lastOrErrorStage` - return the last item or error as a `CompletionStage`.

Related: #6776

Diagrams:

image
image
image
image
image
image
image
image

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -6930,6 +6930,7 @@ public abstract class Flowable<T> implements Publisher<T> {
      * @return a Single that emits the result of collecting the values emitted by the source Publisher
      *         into a single mutable data structure
      * @see <a href="http://reactivex.io/documentation/operators/reduce.html">ReactiveX operators documentation: Reduce</a>
+     * @see #collect(Collector)
      */
     @CheckReturnValue
     @NonNull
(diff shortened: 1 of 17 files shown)


src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -18739,4 +18741,260 @@ public final Completable switchMapCompletableDelayError(@NonNull Function<? supe
...
+     * @param defaultItem the item to signal if the upstream is empty
+     * @return the new CompletionStage instance
+     * @since 3.0.0
+     * @see #firstOrErrorStage()
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.FULL)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @NonNull
+    public final CompletionStage<T> firstStage(@Nullable T defaultItem) {
```
JakeWharton, line 18836: I wonder if these are actually useful compared to just using the regular `Single`-returning operators and then calling a `toStage()`.

JakeWharton: Or, put another way, would you add `blockingFirstOptional()` and `blockingLastOptional()` and the like or would you expect people to using the `Single`-returning operators and then call `blockingOptional()`?

JakeWharton (approved):

src/main/java/io/reactivex/rxjava3/core/Flowable.java
akarnokd: I'm not sure. We have `first` and `blockingFirst` so in case of a `firstStage` -> `first().toCompletionStage()` change, wouldn't the API consistency require us dropping `blockingFirst` for `first().blockingGet()`?

JakeWharton: I don't mind that. It keeps you inside the reactive APIs for transforming the data into the shape you want and then you can use a conversion or blocking operator to escape.

Up to you though. Just thinking of how to minimize the number of operators needed to support these new types.

akarnokd: I'd keep them for now and see if I can beef up `blockingGet` with macro-fusion (as their backing specific implementation is to be reused there). If it can be done reasonably well, I'm willing to drop `blockingX`, and `xStage` operators.

PR #6781, opened 2019-12-19
3.x: Migrate Disposables static factories to Disposable interface

JakeWharton:


src/main/java/io/reactivex/rxjava3/disposables/Disposable.java
```diff
@@ -26,4 +35,111 @@
      * @return true if this resource has been disposed
      */
     boolean isDisposed();
+
+    /**
+     * Construct a {@code Disposable} by wrapping a {@link Runnable} that is
+     * executed exactly once when the {@code Disposable} is disposed.
+     * @param run the Runnable to wrap
+     * @return the new Disposable instance
+     */
```
akarnokd, line 45: I think these additions would count as `@since 3.0.0`.

JakeWharton: Yep. Done.

akarnokd (approved):

PR #6806, opened 2019-12-29
3.x: Prettify Flowable javadoc + validator

akarnokd:
This PR unifies the Javadoc style of the `Flowable` type:
- Mention of `Flowable` has to be `{@code Flowable}`
- First mention of a set of types has to be `{@link TheType}`
- Subsequent mention of a set of types has to be `{@code TheType}`
- Some keywords and method names have to be `{@code methodName}`

In addition, `JavadocWording` has been modified to detect an additional set of wrong typename use.


akarnokd:
Diff view due to a lot of changes: https://patch-diff.githubusercontent.com/raw/ReactiveX/RxJava/pull/6806.diff

PR #6845, opened 2020-01-14
3.x: Add operator matrix wiki + auto-extractor

akarnokd:
Adds the `Operator-Matrix.md` wiki page and some code to generate the matrix of operators available in various base classes.

It should also help see missing or inconsistent operator presence. Note that many operators do not make sense everywhere.


vanniktech (commented):
Nice

PR #6846, opened 2020-01-14
Deleted unnecessary NPE tests

### LINKED ISSUE #6830: 3.x: Remove duplicate unit tests that check null arguments on standard operators
The `ParamValidationCheckerTest` verifies the `null` checks on arguments so there is no longer need for individual unit test methods doing it too. These kinds tests should be identifiable by `expected = NullPointerException.class` or `catch (NullPointerException`

- [x] `Flowable`
- [x] `Observable`
- [x] `Maybe`
- [x] `Single`
- [x] `Completable`
- [x] `ParallelFlowable`

slisaasquatch:
Resolves #6830

I have kept tests that arbitrarily throw NPEs like [this one](https://github.com/ReactiveX/RxJava/blob/0bd2a0152b64723ac3daf3e2fbad3d59c7f3bbb8/src/test/java/io/reactivex/rxjava3/completable/CompletableTest.java#L375), and tests with `Iterable`s that return null `Iterator`s like [this one](https://github.com/ReactiveX/RxJava/blob/0bd2a0152b64723ac3daf3e2fbad3d59c7f3bbb8/src/test/java/io/reactivex/rxjava3/flowable/FlowableNullTests.java#L61).


src/test/java/io/reactivex/rxjava3/completable/CompletableTest.java
```diff
@@ -424,23 +394,6 @@ public Completable get() {
         normal.assertSubscriptions(1);
     }
 
-    @Test(expected = NullPointerException.class)
-    public void deferNull() {
-        Completable.defer(null);
-    }
-
-    @Test(expected = NullPointerException.class)
```
akarnokd, line 432: There are many of these testing the inner null checks, these should be kept.

akarnokd (changes requested):
Many null tests were deleted that should have been kept.

akarnokd (approved):

PR #6848, opened 2020-01-15
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
As one of the most popular JVM projects on GitHub, this security measure aims to help protect your users from supply chain attacks.

See: https://github.com/gradle/wrapper-validation-action

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push]
+
+jobs:
+  validation:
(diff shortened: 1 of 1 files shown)


akarnokd:
How is this supposed to work? I don't see any output regarding the validation in the CI log.

JLLeitschuh:
Hi @akarnokd,

You won't see it because your master branch doesn't have any existing GitHub Actions. As soon as you do, this will show up just like the Travis CI status checks.

Here's an example of what a failure would look like.
image (5) (5)

akarnokd:
Thanks!

PR #6858, opened 2020-01-22
3.x: Verify the use of base interfaces in operator inputs & lambdas

akarnokd:
This PR adds a test that verifies standard operator methods use base interfaces as their input and lambdas capable of returning base types do as well. Consequently, it fixes two declarations of `Flowable.buffer` having the wrong input type.

Basically, detect the following declarations:

```java
// direct use of the class, should be Publisher
    public void method1(Flowable<?> f)

// lambda returns the class, should return a Publisher
    public void method2(Callable<Flowable<?>> c)

// The inner Publisher emits the class, should emit Publisher
    public void method3(Supplier<Publisher<Flowable<?>>> c)

// Direct use of the class as array type, should be Publisher[]
    public void method4(Flowable<?>[] array)

// Lambda returns an array of the class, should return Publisher[]
    public void method5(Callable<Flowable<?>[]> c)

// The inner Publisher emits an array of the class, should emit Publisher[]
    public void method6(Callable<Publisher<Flowable<?>[]>> c)
```

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -6790,7 +6790,7 @@ public abstract class Flowable<@NonNull T> implements Publisher<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     @NonNull
     public final <TOpening, TClosing> Flowable<List<T>> buffer(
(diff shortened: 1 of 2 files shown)


vanniktech (approved):
Really nice!

PR #6891, opened 2020-01-28
3.x: Swap Maybe.flatMapSingle and Maybe.flatMapSingleElement

### LINKED ISSUE #6878: 3.x: Swap Maybe.flatMapSingle and Maybe.flatMapSingleElement ?
Currently, `flatMapSingle` returns `Single` and signals `NoSuchElementException` if the source `Maybe` is empty. `flatMapSingleElement` remains `Maybe`.

I'd consider renaming `flatMapSingle` to `flatMapSingleOrError` and `flatMapSingleElement` to `flatMapSingle` as I think a default `Maybe` -> `Maybe` transformer should better match the other such transformers.

Alternatively, drop `flatMapSingle` and rename `flatMapSingleElement` to `flatMapSingle` so that the original behavior could be achieved via `flatMapSingle.toSingle())`.

slisaasquatch:
Resolves #6878

Dropped the old `Maybe.flatMapSingle` and replaced it with `Maybe.flatMapSingleElement`


slisaasquatch:
Hmm for XFlatMapTest.maybeSingle, if I replace `flatMapSingle().toSingle()` with `toSingle().flatMap()`, then the test will pass.

akarnokd:
The test checks if disposing while in the function, the returned `Single` never gets subscribed to and thus no error is signaled to the downstream or the global error handler. Put an `@Ignore` on it for now and I'll look into the operator tomorrow.

akarnokd (approved):

PR #6905, opened 2020-02-13
Fix Document Error

myungjunChae:
# Description
In `docs/Filtering-Observables.md`, Example code of `.timeout()` has unexpected result.

It creates Observable with Thread.sleep(), but subscribed by just `.subscribe()` not `.blockingSubscribe()`.

So It needs to be changed `.subscribe()`to `.blockingSubscribe()` for expected result.

### DIFF
--- a/docs/Filtering-Observables.md
+++ b/docs/Filtering-Observables.md
@@ -747,7 +747,7 @@ Observable<String> source = Observable.create(emitter -> {
 });
 
 source.timeout(1, TimeUnit.SECONDS)
-        .subscribe(
+        .blockingSubscribe(
                 item -> System.out.println("onNext: " + item),
                 error -> System.out.println("onError: " + error),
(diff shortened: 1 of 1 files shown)


akarnokd:
The example works as intended and prints the expected result. Have you tried the example?

> It creates Observable with Thread.sleep(), but subscribed by just .subscribe() not .blockingSubscribe().

This blocks on the main thread after "C" thus the timeout message will still print. The `main()` can't quit until the `Thread.sleep(1200)` and  calls after that end.

myungjunChae:
Sorry, akarnkd. It's my fault. 
Actually, whole my code is like under
```
 stream.subscribeOn(Schedulers.io())
        .timeout(1, TimeUnit.SECONDS)
        .subscribe(
            {println(it)},
            {err->println(err)}
        )
```

It has `.subscribeOn(Schedulers.io())`additionally.
When I eliminate that, Code works right.

PR #6944, opened 2020-04-02
Update image urls to non-transparent version.

### LINKED ISSUE #6883: 3.x: Marbles should have a proper background
The transparent ones look terrible with dark-themed IDEs or non-light background:

image

Only looks fine in Eclipse:

image

Write a program that goes through the images in `https://github.com/ReactiveX/RxJava.wiki.git` `images/rx-operators`, checks if the image has transparency, draw it onto a white background then save it as `originalName.v3.png` (because some of those are linked all around). Find the filename in the source code and replace them.

tomislavhofman:
Updated image urls to use non-transparent ones as per #6883

Resolves #6883


akarnokd (approved):

akarnokd:
Thanks. The new v3 images load in properly.

tomislavhofman:
Glad to help.

PR #6946, opened 2020-04-04
3.x: Fix Observable.flatMap with maxConcurrency hangs

### LINKED ISSUE #6945: Deadlock
RXJava 3.0.1.
I randomly got into a weird situation where my code entered a deadlock and I don't know why.
I simplified the code and got to the below part which will enter a deadlock and never return.
This seems like a bug, but maybe I'm missing something.
The code:
```java
Observable
    .range(1, 20)
    .flatMap(
            integer -> {
                if (integer % 5 != 0) {
                    return Observable
                            .just(integer);
                }

                return Observable
                        .just(-integer)
                        .observeOn(Schedulers.io());
            },
            false,
            1
    )
    .ignoreElements()
    .blockingAwait();
```

akarnokd:
Queued up scalar values were not counted as completed, thus the subsequent sources where not subscribed to to fill up the allowed concurrency level.

Fixes #6945

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableFlatMap.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableFlatMap.java
@@ -318,6 +318,7 @@ public final class ObservableFlatMap<T, U> extends AbstractObservableWithUpstrea
                 if (checkTerminate()) {
                     return;
                 }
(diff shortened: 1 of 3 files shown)


vanniktech (approved):
Flowable works?

akarnokd:
Yes. Added the test for it too anyway.

vanniktech (approved):

Issue #748, opened 2014-01-14
StackOverflowError is swallowed 


zsxwing:
The following code should have thrown StackOverflowError:

``` java
        final PublishSubject<Integer> a = PublishSubject.create();
        final PublishSubject<Integer> b = PublishSubject.create();
        a.subscribe(new Observer<Integer>() {

            @Override
            public void onCompleted() {

            }

            @Override
            public void onError(Throwable e) {
                e.printStackTrace();
            }

            @Override
            public void onNext(Integer args) {
                System.out.println(args);
            }
        });
        b.subscribe(new Observer<Integer>() {

            @Override
            public void onCompleted() {

            }

            @Override
            public void onError(Throwable e) {
                e.printStackTrace();
            }

            @Override
            public void onNext(Integer args) {
                System.out.println(args);
            }
        });
        a.subscribe(new Observer<Integer>() {

            @Override
            public void onCompleted() {

            }

            @Override
            public void onError(Throwable e) {
                e.printStackTrace();
            }

            @Override
            public void onNext(Integer args) {
                b.onNext(args + 1);
            }
        });
        b.subscribe(new Observer<Integer>() {

            @Override
            public void onCompleted() {

            }

            @Override
            public void onError(Throwable e) {
                e.printStackTrace();
            }

            @Override
            public void onNext(Integer args) {
                a.onNext(args + 1);
            }
        });
        a.onNext(1);
```

However, StackOverflowError is swallowed.
The problem is the following line in SafeObserver: https://github.com/Netflix/RxJava/blob/0b1b6e7a91d89ad63263b80747f0bd4683fe4ba2/rxjava-core/src/main/java/rx/operators/SafeObserver.java#L117

``` java
            RxJavaPlugins.getInstance().getErrorHandler().handleError(e);
```

When StackOverflowError is thrown, there is few stack space. However, this line will generate too big stack frame and cause the thread crash. If I comment out this line, I can observe StackOverflowError.

Reported by Samuel at https://groups.google.com/forum/#!topic/rxjava/eraZ-32w1gQ

PR #6956, opened 2020-04-11
3.x: Fix scheduled tasks' fatal exception behavior

### LINKED ISSUE #6954: 3.x: Exceptions will be swallowed when schedulers are involved

akarnokd:
Fatal exceptions may be lost with scheduled direct & scheduled periodic direct tasks because `FutureTask` simply treats them as exceptional outcomes. For regular tasks, [ScheduledRunnable](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledRunnable.java#L67) already avoids rethrowing fatal errors as it would go nowhere.

This PR adds this behavior to the direct runnable tasks.

Resolves #6954

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
@@ -18,11 +18,9 @@ import java.util.concurrent.TimeUnit;
 
 import io.reactivex.rxjava3.annotations.*;
 import io.reactivex.rxjava3.disposables.Disposable;
-import io.reactivex.rxjava3.exceptions.Exceptions;
 import io.reactivex.rxjava3.functions.Function;
 import io.reactivex.rxjava3.internal.disposables.*;
 import io.reactivex.rxjava3.internal.schedulers.*;
-import io.reactivex.rxjava3.internal.util.ExceptionHelper;
 import io.reactivex.rxjava3.plugins.RxJavaPlugins;
 import io.reactivex.rxjava3.schedulers.SchedulerRunnableIntrospection;
 
@@ -542,9 +540,9 @@ public abstract class Scheduler {
                 try {
                     run.run();
                 } catch (Throwable ex) {
-                    Exceptions.throwIfFatal(ex);
-                    worker.dispose();
-                    throw ExceptionHelper.wrapOrThrow(ex);
+                    // Exceptions.throwIfFatal(ex); nowhere to go
+                    dispose();
+                    RxJavaPlugins.onError(ex);
                 }
             }
         }
@@ -588,8 +586,8 @@ public abstract class Scheduler {
             try {
                 decoratedRun.run();
             } finally {
-                dispose();
                 runner = null;
+                dispose();
             }
         }
 
--- a/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
@@ -543,6 +543,7 @@ public abstract class Scheduler {
                     // Exceptions.throwIfFatal(ex); nowhere to go
                     dispose();
                     RxJavaPlugins.onError(ex);
+                    throw ex;
                 }
             }
         }
@@ -584,7 +585,13 @@ public abstract class Scheduler {
         public void run() {
             runner = Thread.currentThread();
             try {
-                decoratedRun.run();
+                try {
+                    decoratedRun.run();
+                } catch (Throwable ex) {
+                    // Exceptions.throwIfFatal(e); nowhere to go
+                    RxJavaPlugins.onError(ex);
+                    throw ex;
+                }
             } finally {
                 dispose();
                 runner = null;
--- a/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
@@ -586,8 +586,8 @@ public abstract class Scheduler {
             try {
                 decoratedRun.run();
             } finally {
-                runner = null;
                 dispose();
+                runner = null;
             }
         }
 
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
@@ -37,7 +37,6 @@ public final class ScheduledDirectPeriodicTask extends AbstractDirectTask implem
         try {
             runnable.run();
             runner = null;
-            lazySet(FINISHED);
         } catch (Throwable ex) {
             // Exceptions.throwIfFatal(ex); nowhere to go
             runner = null;
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
@@ -42,6 +42,7 @@ public final class ScheduledDirectPeriodicTask extends AbstractDirectTask implem
             runner = null;
             dispose();
             RxJavaPlugins.onError(ex);
+            throw ex;
         }
     }
 }
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
@@ -16,7 +16,6 @@
 
 package io.reactivex.rxjava3.internal.schedulers;
 
-import io.reactivex.rxjava3.exceptions.Exceptions;
 import io.reactivex.rxjava3.plugins.RxJavaPlugins;
 
 /**
@@ -39,7 +38,7 @@ public final class ScheduledDirectPeriodicTask extends AbstractDirectTask implem
             runnable.run();
             runner = null;
         } catch (Throwable ex) {
-            Exceptions.throwIfFatal(ex);
+            // Exceptions.throwIfFatal(ex); nowhere to go
             runner = null;
             lazySet(FINISHED);
             RxJavaPlugins.onError(ex);
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
@@ -37,10 +37,11 @@ public final class ScheduledDirectPeriodicTask extends AbstractDirectTask implem
         try {
             runnable.run();
             runner = null;
+            lazySet(FINISHED);
         } catch (Throwable ex) {
             // Exceptions.throwIfFatal(ex); nowhere to go
             runner = null;
-            lazySet(FINISHED);
+            dispose();
             RxJavaPlugins.onError(ex);
         }
     }
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/ExecutorScheduler.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/ExecutorScheduler.java
@@ -320,6 +320,10 @@ public final class ExecutorScheduler extends Scheduler {
                 }
                 try {
                     actual.run();
+                } catch (Throwable ex) {
+                    // Exceptions.throwIfFatal(ex); nowhere to go
+                    RxJavaPlugins.onError(ex);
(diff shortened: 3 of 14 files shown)


src/main/java/io/reactivex/rxjava3/internal/schedulers/ScheduledDirectPeriodicTask.java
```diff
@@ -39,7 +38,7 @@ public void run() {
             runnable.run();
             runner = null;
         } catch (Throwable ex) {
-            Exceptions.throwIfFatal(ex);
+            // Exceptions.throwIfFatal(ex); nowhere to go
```
jxdabc, line 41: But it seems make #748 a problem again. It seems tricky to balance this issue and #748. After all, if we passed StackOverflowError to RxJavaPlugins.onError(), RxJavaPlugins.onError() should throw it and it will be swallowed again.

I think customizing the future task will completely solve #748 and this issue. JDK supports that and at least as I know Android are using that approach. And I think it is not complex as the core code is only implementing the JDK's interface. It even simplifies the schedulers implementation, as I see ; )

jxdabc (commented):
I'm sorry, I think I made a mistake.
I'll check it.

akarnokd:
> this will halt the application when stack overflow:

I don't understand. As far as I know, calling the uncaught exception handler on Android kills your app. On desktop, such calls do nothing but the repetition of the `schedulePeriodicallyDirect` should end.

jxdabc:
> > this will halt the application when stack overflow:
> 
> I don't understand. As far as I know, calling the uncaught exception handler on Android kills your app. On desktop, such calls do nothing but the repetition of the `schedulePeriodicallyDirect` should end.

I am sorry. I think I made a mistake when simplifies #748 case. The key point is when StackOverflowError ocurrs calling error hander without throw may cause stack full again. That is also why I selected a complex approach. I'll check it.

jxdabc:
> > this will halt the application when stack overflow:
> 
> I don't understand. As far as I know, calling the uncaught exception handler on Android kills your app. On desktop, such calls do nothing but the repetition of the `schedulePeriodicallyDirect` should end.

Hi, akarnokd, sorry to the above mistake.

Personally,  I think this approach has some flaws.


1 - This approach means we'll catch fatal exceptions instead of throwing it through the call stack. But the functions on the call stack may need this exception to clean/stop itself. For example, this change will make periodic tasks unable to stop (at least on desktop or server):


    @Test // Fail
    public void periodicTaskShouldStopOnError() throws Exception {
        AtomicInteger repeatCount = new AtomicInteger();
        new ComputationScheduler().schedulePeriodicallyDirect(new Runnable() {
            @Override
            public void run() {
                repeatCount.incrementAndGet();
                if (true) {
                    throw new OutOfMemoryError();
                }
            }
        }, 0, 500, TimeUnit.MILLISECONDS);

        Thread.sleep(1000 * 3);
        assertEquals(1, repeatCount.get());
    }


2 - To solve the above problem, we need to rethrow the exception after RxJavaPlugins.onError() or make RxJavaPlugins.onError() to throw after calling the uncaughtExceptionHandler of the thread. This actullay makes some exceptions to be handled more than once. And I think it will make things more and more complex as we solve those cases one by one.

3 - If we are able to check exceptions in the Future as #6954 mentioned (and JDK has given us all we need). Exception handing in RxJava will be simpler and easier as we do elsewhere: make RxJavaPlugins.onError() handle/propagate the exception and continue the work, or make RxJavaPlugins.onError() throws and exceptions go through the stack to do whatever clean/stop needed and finally to the uncaughtExceptionHandler of the thread.

jxdabc:
Disposing the periodic task on error may be a little tricky workaround. 

1 - The periodic FutureTask is canceled asynchronously, until PeriodicDirectTask.setFuture is done. This may cause the periodic task to run some more rounds before actually stop. The following test almost can't success on my laptop:

    @Test
    public void periodicTaskShouldStopOnError() throws InterruptedException {
        try {
            // suppress log
            RxJavaPlugins.setErrorHandler((e) -> {});

            CountDownLatch latch = new CountDownLatch(1000);

            AtomicInteger assertCount = new AtomicInteger();
            for (int i = 0; i < 1000; i++) {
                new Thread(new Runnable() {
                    @Override
                    public void run() {
                        try {
                            periodicTaskShouldStopOnErrorOnce();
                        } catch (AssertionError e) {
                            e.printStackTrace();
                            assertCount.incrementAndGet();
                        } catch (Exception e) {
                        } finally {
                            latch.countDown();
                        }
                    }
                }).start();
            }

            latch.await();
            assertEquals(0, assertCount.get());
        } finally {
            RxJavaPlugins.reset();
        }
    }

    private void periodicTaskShouldStopOnErrorOnce() throws Exception {
        AtomicInteger repeatCount = new AtomicInteger();

        Schedulers.computation().schedulePeriodicallyDirect(new Runnable() {
            @Override
            public void run() {
                repeatCount.incrementAndGet();
                if (true) {
                    throw new OutOfMemoryError();
                }
            }
        }, 0, 1, TimeUnit.NANOSECONDS);

        Thread.sleep(200);

        assertEquals(1, repeatCount.get());
    }

2 - I wonder whether we should route InterruptibleRunnable's & BooleanRunnable's errors to RxJavaPlugins.onError() as well. If not, I have to handle errors both in RxJavaPlugins.onError() and thread's uncaught exception handlers. When schedulers are involved, errors may be pass to either of the two. 

3 - My original PR was actually intending to make things easy for my team: we only need to set an error handle via RxJavaPlugins.setErrorHandler which prints a log for not-a-matter error and rethrow for fatal erros and every thing goes fine. However, maybe I could replace all schedulers with RxJavaPlugins.setXXXSchedulerHandler without change the main stream though I think my original PR simplifies implemetation of the main stream without side effect as I could see.

akarnokd:
1. Looks like the periodic task should be crashed nonetheless.
2. Yes, task wrapper behavior should be consistent across.
3. "make things easy for my team" - that's why I encourage people to write their own hooks and components to workaround very specific corner cases.

jxdabc:
> 2. Yes, task wrapper behavior should be consistent across.

Would you like to do this in this PR or later?

jxdabc (approved):

jxdabc (approved):

jxdabc (approved):

jxdabc (approved):

src/main/java/io/reactivex/rxjava3/core/Scheduler.java
```diff
@@ -542,9 +540,10 @@ public void run() {
                 try {
                     run.run();
                 } catch (Throwable ex) {
-                    Exceptions.throwIfFatal(ex);
-                    worker.dispose();
-                    throw ExceptionHelper.wrapOrThrow(ex);
+                    // Exceptions.throwIfFatal(ex); nowhere to go
+                    dispose();
+                    RxJavaPlugins.onError(ex);
+                    throw ex;
```
akarnokd, line 546: @JakeWharton Does this syntax work with Android desugar (rethrowing final `Throwable`s)?

JakeWharton: Hmm this looks like a normal throw. I don't understand what might be a compatibility concern here.

akarnokd: The Java 7 compiler feature that let's you rethrow constant cheched or `Throwable` exceptions from within a `catch` block, even if the method didn't specify a `throws` clause.

JakeWharton: Ah, missed it was `Throwable`. Checked exceptions are only a feature of the Java compiler not the Java bytecode so yeah the Android toolchain won't care 👍 .

jxdabc:
Great, that fixed all corner cases I could see.
But I wonder whether calling RxJavaPlugins.onError(ex) (calling uncaught exception handler) in the middle of exception throw chain is good and intuitional. A little tricky as I see.
If you'd have a look on #6955, the original PR. It requires no more change than this one, handling Future related erros in ONE place and not calling uncaught exception handler in the middle of exception throw chain. It should pass all tests in this PR as well without modification. :)

PR #6975, opened 2020-05-09
3.x: Fix Flowable.groupBy eviction logic double decrement and hang

### LINKED ISSUE #6974: 3.0.3 - Hangs when using combination of flatMap and publish
Hi.

After few days of desperate attempts to find the cause of my servers' spontaneous hangs, I managed to reproduce the issue, but still don't have a clear idea about the reason, but I think I have identified a possible direction.

All problems has started when I started using very promising operator groupBy with evicting map provider possibility:
```java
<K, V> Flowable<GroupedFlowable<K, V>> groupBy(@NonNull Function<? super T, ? extends K> keySelector,
            @NonNull Function<? super T, ? extends V> valueSelector,
            boolean delayError, int bufferSize,
            @NonNull Function<? super Consumer<Object>, ? extends Map<K, Object>> evictingMapFactory)
```

First I realised that it doesn't work well with Caffeine cache, as it performs evictions on a separate thread pool, and with async eviction notification groupBy goes crazy.

After moving to Guava things got better but only to some extent.

The tests I want to demonstrate I believe reproduce the issue I have. Definitely chain of operators is way more complicated in the production code, but here with certain variant of .publish() operator it's reproducible as well.

Basically, I'm not sure if there is issue in the groupBy(...evictingMapFactory) or in the publish(Function...), or in both.

Also tests are demonstrating that eviction policy doesn't work very well when crossing async boundaries downstream.

Please take a look at the attached file, it contains few simple tests, I kindly ask to check carefully all parameters, as it's important.

[FlowableOperators.java.zip](https://github.com/ReactiveX/RxJava/files/4604420/FlowableOperators.java.zip)

akarnokd:
When a group is evicted, the group is synchronously `onComplete`d during an `onNext`. If a cancellation happened during this time (for example, when a `publish` cancels its upstream because its output completes), the cancellation would unconditionally decrement the group counter even though the `groups` map had no longer the group (because it was evicted before). Then once the eviction logic finishes, it decremented the group counter again, leading to invalid internal state and hangs.

The fix is to only decrement if there was a group actually removed from the map.

Fixes #6974

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableGroupBy.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableGroupBy.java
@@ -270,9 +270,10 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
 
         public void cancel(K key) {
             Object mapKey = key != null ? key : NULL_KEY;
-            groups.remove(mapKey);
-            if (groupCount.decrementAndGet() == 0) {
-                upstream.cancel();
+            if (groups.remove(mapKey) != null) {
+                if (groupCount.decrementAndGet() == 0) {
+                    upstream.cancel();
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


davidmoten (approved):
LGTM, thanks @akarnokd

vanniktech (approved):

PR #6988, opened 2020-05-19
3.x: Fix Flowable.groupBy eviction-completion-replenishment problems

### LINKED ISSUE #6982: Flowable.groupBy hangs with group consumed via observeOn and take(timed)
Continuing evaluating the problems from #6974

The following test hangs after a while:

```java
    @Test
    public void issue6974Part2Case1ObserveOnNoCap() {
        final int groups = 20;
        int flatMapMaxConcurrency = 1_000_000;

        Flowable
        .range(1, 500_000)
        .map(i -> i % groups)
        .doOnRequest(v -> {
            System.out.println("Source: " + v);
        })
        .groupBy(i -> i)
        .flatMap(gf -> gf
                     .take(10, TimeUnit.MILLISECONDS)
            , flatMapMaxConcurrency)
        .test()
        .assertNoErrors()
        .assertComplete();
    }
```

Adding `hide` makes it pass which indicates the fusion between the group and `observeOn` seems to stop requesting more items.

akarnokd:
This PR fixes the two cases from https://github.com/ReactiveX/RxJava/issues/6982#issuecomment-630290658

### Case 1: completions missed

When the upstream completed and some groups were waiting in the eviction queue, the eviction queue was cleared without completing those groups. Since they were no longer in the main map, those groups never completed and downstream operators relying on all groups completing (i.e., flatMap) would never complete.

The fix is to call `completeEvictions` upon upstream termination.

### Case 2: no replenisment on evicted groups

When an eviction happens, the groups get completed. However, unlike with the non-evicting case, a group completion no longer implies the upstream has completed. Consequently, the evicted groups emitted their items but never issued a replenishment for them, hanging the still-alive main operator.

The fix is to replenish emitted (and polled) counts whenever the group detects a completion. (Errors still imply the main operator has terminated so no need to replenish then.) 

Resolves #6982

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableGroupBy.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableGroupBy.java
@@ -214,9 +214,7 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                 g.onError(t);
             }
             groups.clear();
-            if (evictedGroups != null) {
-                evictedGroups.clear();
-            }
+            completeEvictions();
             downstream.onError(t);
         }
 
@@ -226,10 +224,10 @@ public final class FlowableGroupBy<T, K, V> extends AbstractFlowableWithUpstream
                 for (GroupedUnicast<K, V> g : groups.values()) {
                     g.onComplete();
                 }
+
                 groups.clear();
-                if (evictedGroups != null) {
-                    evictedGroups.clear();
-                }
+                completeEvictions();
+
                 done = true;
                 downstream.onComplete();
             }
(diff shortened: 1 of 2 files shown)


davidmoten (approved):
thanks @akarnokd

sobersanta:
Thank you! Starting tests, will let you know if I find any new cases ;)

akarnokd:
I'll merge this once the previous PR clears, then I will run it on the Build_Matrix to see if any flakyness arises.

sobersanta:
I'm afraid I got another case of hang during normal sequence shutdown via onComplete.
Still trying to have a reproducing test case... From my logs I can see that eviction of group happened during shutdown sequence, that's all I got for now.

akarnokd:
Thanks. No rush here with the release. If you find something, it will go in the next PR.

sobersanta:
Oh, I'm so sorry. Seems to be a false alarm, I was a bit lost with versions!

davidmoten:
Are the groupBy fixes something that should be backported to 2.x?

akarnokd:
The abandonment can't be backported because that's a breaking behavior change. The group termination might be, but the unit tests in these 3.x fixes may hang anyway due to not having a properly overarching group state management. We'll only know if a backport is attempted. PR welcome.

davidmoten:
Thanks @akarnokd. I haven't seem symptoms in our applications of problems so not urgent for me yet but I'll put it on our backlog.

PR #6992, opened 2020-05-20
Removed unnecessary upstream.cancel() call for casually finished upstream sequences.

### LINKED ISSUE #6989: 3.x: unnecessary call to upstream .cancel() in FlowablePublishMulticast operator
I think I should add it as a separate issue.

In a casual completion chain from the upstream, when it is calling `onComplete`/`onError`, `FlowablePublishMulticast` calls `SubscriptionHelper.cancel(upstream);` in the `dispose()` method (line 196).
I believe that upstream cancellation should be done only from `.cancel()` method not from `.onComplete` or `.onError` as it causes issues in some scenarios including extra groups creation in the groupBy operator.

sobersanta:
- no upstream.cancel() in FlowablePublishMulticast when the sequence is finished normally via onComplete/onError from upstream;
- minor code cleanup - unnecessary Disposable implementation to avoid method name clash
- Fixes https://github.com/ReactiveX/RxJava/issues/6989
- FlowablePublishFunctionTest.java  refactored to use lambdas

### DIFF
--- a/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowablePublishFunctionTest.java
+++ b/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowablePublishFunctionTest.java
@@ -19,6 +19,7 @@ package io.reactivex.rxjava3.internal.operators.flowable;
 import static org.junit.Assert.*;
 
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.junit.*;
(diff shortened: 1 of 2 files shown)


src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowablePublishFunctionTest.java
```diff
@@ -538,4 +539,68 @@ public boolean test(Integer w) throws Exception {
         .test()
         .assertResult(1000000);
     }
+
+    @Test
+    public void noUpstreamCancelOnCasualChainClose() {
+        AtomicBoolean parentUpstreamCancelled = new AtomicBoolean(false);
+        Flowable.range(1, 10)
+                .doOnCancel(new Action() {
```
akarnokd: You can use lambdas in tests now.

sobersanta: Cool!
Should I change it?

akarnokd: Yes.

sobersanta: If that's fine, I've refactored whole file to lambdas.

akarnokd (approved):

PR #6994, opened 2020-05-20
3.x: Fix recent groupBy tests sometimes failing with MBE

akarnokd:
These tests can fail with `MissingBackpressureException` because groups may not complete fast enough so `flatMap` can request more groups in time. This doesn't happen consistently but could fail the test on CI. The workaround is to allow any termination, not just normal completion. The reasoning is that the aim of the tests were to verify the operator doesn't hang.

The underlying complication is that whenever there is an item replenishment, any subsequent item can result in a fresh group being created. If the concurrency level of `flatMap` is not high enough, this will result in a MBE and the sequence terminates.


akarnokd:
The matrix passed so https://travis-ci.org/github/akarnokd/RxJava3_BuildMatrix/builds/689534225

PR #6996, opened 2020-05-23
Suppress UndeliverableException handling in tests (#6987)

### LINKED ISSUE #6987: 3.x: Suppress undeliverables in some tests
Some tests trigger a lot of undeliverable exceptions flooding the console output. Example:

```
at io.reactivex.rxjava3.internal.subscribers.FutureSubscriberTest
    .onErrorCancelRace(FutureSubscriberTest.java:178)

at io.reactivex.rxjava3.internal.operators.observable.ObservableWindowWithObservableTest
    .disposeMainBoundaryErrorRace(ObservableWindowWithObservableTest.java:590)
```

These are generally irrelevant from those particular tests and should be safely ignored. The difficulty is, locating these tests if the console doesn't retain too many lines.

To resolve the issue, perhaps the easiest way is to write a `TestHelper.withUndeliverableSuppressed(Action)` method and keep wrapping the problematic test bodies.

akluball:
I used a JUnit TestRule to suppress handling of UndeliverableException for test methods annotated with @SuppressUndeliverable. This approach is coarser than the suggested approach (it suppresses the handling of all UndeliverableExceptions for the entire annotated test method).
I thought this approach was clean but I can switch to the suggested approach if preferred.

I added @SuppressUndeliverable to most tests that were logging UndeliverableExceptions with the test method in the stack trace. This may be more suppression than was required so just let me know.

Fixes #6987

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
@@ -32,6 +32,15 @@ import io.reactivex.rxjava3.schedulers.Schedulers;
  * Utility class to inject handlers to certain standard RxJava operations.
  */
 public final class RxJavaPlugins {
+    @NonNull
+    private static final java.util.function.Consumer<? super Throwable> defaultErrorHandler = new java.util.function.Consumer<Throwable>() {
+        @Override
+        public void accept(Throwable throwable) {
+            throwable.printStackTrace(); // NOPMD
+            uncaught(throwable);
+        }
+    };
+
     @Nullable
     static volatile Consumer<? super Throwable> errorHandler;
 
@@ -175,6 +184,15 @@ public final class RxJavaPlugins {
     @Nullable
     public static Function<? super Scheduler, ? extends Scheduler> getComputationSchedulerHandler() {
         return onComputationHandler;
+    }
+
+    /**
+     * Returns the default error handler.
+     * @return the default error handler
+     */
+    @NonNull
+    public static java.util.function.Consumer<? super Throwable> getDefaultErrorHandler() {
+        return defaultErrorHandler;
     }
 
     /**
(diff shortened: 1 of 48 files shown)


src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
```diff
@@ -32,6 +32,15 @@
  * Utility class to inject handlers to certain standard RxJava operations.
  */
 public final class RxJavaPlugins {
+    @NonNull
```
akarnokd: Please do not change RxJavaPlugins.

src/test/java/io/reactivex/rxjava3/internal/jdk8/ParallelCollectorTest.java
```diff
@@ -33,6 +33,9 @@
 
 public class ParallelCollectorTest extends RxJavaTest {
 
+    @Rule
+    public final SuppressUndeliverableRule suppressUndeliverableRule = new SuppressUndeliverableRule();
```
akarnokd: Can these rules moved into `RxJavaTest` so you don't have to add them to all test classes?

src/test/java/io/reactivex/rxjava3/testsupport/SuppressUndeliverableRule.java
```diff
@@ -0,0 +1,59 @@
...
+public class SuppressUndeliverableRule implements TestRule {
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        if (description.getAnnotation(SuppressUndeliverable.class) != null) {
+            return new Statement() {
+                @Override
+                public void evaluate() throws Throwable {
+                    try {
+                        RxJavaPlugins.setErrorHandler(new Consumer<Throwable>() {
```
akarnokd: You can now use Java 8 lambdas in tests.

akarnokd (changes requested):

akarnokd (approved):

akarnokd:
Thanks!

PR #7046, opened 2020-08-02
Allow users to specify a consumer with which to consume the dropped v…

MarcusGrass:
…alue on a buffer overflow when using onBackpressureBuffer.

The reason for this PR is that previously you could not specify an action to take on the dropped item using Flowable.onBackpressureBuffer(...). This change overloads the method to allow the user of the API to provide a consumer which will accept the dropped item. An example use-case would be to send overflowing messages to another storage than the heap. The specific situation that prompted this request was producing to a message queue, in case of network error between the producer and the queue an OOM would quickly occur. With this change those messages could be sent to a colder storage than the heap and later be retrieved when the network error subsides.

There are a few tests added to FlowableOnBackpressureBufferStrategyTest.java that firstly ensures that the consumer is activated correctly, and secondly that the operator works the same, in all relevant aspects, as the already existing variant where an Action is provided to be executed.

In addition to this there is one more notable change. The documentation of `onBackpressureBuffer(long capacity, @Nullable Action onOverflow, @NonNull BackpressureOverflowStrategy overflowStrategy)` states: 

> <li>{@link BackpressureOverflowStrategy#ERROR} (default) will call {@code onError} dropping all undelivered items,
> canceling the source, and notifying the producer with {@code onOverflow}. </li>
However, in the case of BackpressureOverflowStrategy#ERROR the onOverflow Action was never invoked. This was corrected and the testoverflowError was modified to reflect this change.

This change means that the possible exception that could come from the invocation of the Action(or consumer) needs to be handled. My solution was that if that method throws an exception that exception is the one that will be propagated instead of the MissingBackpressureException which would previously have been thrown. This is an opinionated implementation and my reasoning is that regardless of which backpressure strategy you chose, an exception in the Action should propagate down, and kill the stream. Getting the exception thrown by your user-defined Action(or consumer) would be more helpful to debug your application. You can of course disagree with this, in the end it is fairly arbitrary.

Lastly the javadoc for `onBackpressureBuffer(long capacity, @Nullable Action onOverflow, @NonNull BackpressureOverflowStrategy overflowStrategy)` was updated to remove the note that the method throws a NPTR if invoked with a null Action. This was not true, not present on the other overloads involving Action, and disproved by a preexisting test.

Ps. I'm not entirely happy with the method signature `onBackpressureBuffer(long capacity, @NonNull BackpressureOverflowStrategy overflowStrategy, @NonNull Consumer<? super T> onDrop)`
but as having the same order as `onBackpressureBuffer(long capacity, @Nullable Action onOverflow, @NonNull BackpressureOverflowStrategy overflowStrategy)` results in an ambiguous method call if the action is provided as null.

Pps. in `onBackpressureBuffer(long capacity, @NonNull BackpressureOverflowStrategy overflowStrategy, @NonNull Consumer<? super T> onDrop)` onDrop was declared as `@NonNull` to be more compliant with the convention on NonNull on onBackpressureDrop where null is not allowed on a `Consumer<? super T>`

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -12562,7 +12562,7 @@ public abstract class Flowable<@NonNull T> implements Publisher<T> {
      * @param onOverflow action to execute if an item needs to be buffered, but there are no available slots, {@code null} is allowed.
      * @param overflowStrategy how should the resulting {@code Flowable} react to buffer overflows, {@code null} is not allowed.
      * @return the new {@code Flowable} instance
-     * @throws NullPointerException if {@code onOverflow} or {@code overflowStrategy} is {@code null}
+     * @throws NullPointerException if {@code overflowStrategy} is {@code null}
      * @throws IllegalArgumentException if {@code capacity} is non-positive
      * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
      * @since 2.0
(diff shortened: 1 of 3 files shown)


akarnokd:
Hi. What is the reason and/or use case for this operator to be added to RxJava proper? (I.e, just because it was not possible to specify a consumer for the dropped item, doesn't automatically mean there would be a use for the operator.)

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -12574,7 +12574,52 @@ public final Completable ignoreElements() {
...
+     * </dl>
+     *
+     * @param capacity number of slots available in the buffer.
+     * @param overflowStrategy how should the resulting {@code Flowable} react to buffer overflows, {@code null} is not allowed.
+     * @param onDrop action to execute upon the dropped item if an item needs to be buffered, but there are no available slots, {@code null} is not allowed.
+     * @return the new {@code Flowable} instance
+     * @throws NullPointerException if {@code onDrop} or{@code overflowStrategy} is {@code null}
+     * @throws IllegalArgumentException if {@code capacity} is non-positive
+     * @see <a href="http://reactivex.io/documentation/operators/backpressure.html">ReactiveX operators documentation: backpressure operators</a>
+     * @since 2.0
```
akarnokd, line 12612: If this operator is to be accepted, the version tag should be the next upcoming version and also marked as experimental.

```
@since 3.0.6 - experimental

//...

@Experimental
```

MarcusGrass: Fixed

MarcusGrass:
Added an example case to the edit.
Specifically to avoid OOM in case of spurious slowdowns or stops downstream
by enabling sending of messages out of the heap until the blockage clears.

akarnokd:
The given example makes no sense to me. Backpressure is there to limit items generated thus one wouldn't use onBackpressureBuffer. If you need lossless buffering of a much larger size than the memory, the rxjava-extras project has a disk-based buffer.

MarcusGrass:
In the application in question, disk storage is not available. Storing to
disk is the same problem as storing on the heap except for the fact that it
is even less predictable. You most likely know the heap size of an
application that you write, but in my target Infrastructure I do not know
the disk size, I only know that the space is limited and I don't have
monopoly on it.
What I specifically need is a heap buffer because production is not uniform
but the downstream consumer consumes uniformly. Then I run into the very
real possibility that my consumer could for some reason stop consuming
completely (downtime, network, etc.) for some non-trivial duration. In that
case I do not want an OOM, I'd rather not drop messages, and sending them
to disk is not an option.

akarnokd:
> production is not uniform

What is your source of data that you need such convoluted backpressure mechanics for? By the way, you should have opened an issue first to explore options without trying to change RxJava for what seems like a peculiar local problem stemming from your use case and/or setup.

In general, adding the means to see dropped items leads down the rabbit hole of resource aware flows and item lifecycle, which aren't well explored at the moment and would require the redesign of the library from the ground up for sure.

I don't support this change to be included in RxJava proper but can offer alternative options to solve your original problem:

- post an issue detailing your flow and requirements to see if existing tools could be applied instead,
- take the operator by source and implement the changes in the target project,
- rethink your flow to avoid the need for such a solution or increase the memory allocated to your app to not need an off-heap storage.

MarcusGrass:
I suggest adding that information to the CONTRIBUTION.md. 
I did implement this in an operator but figured, if I'm using open source I might aswell contribute, that's the reason for this PR.
You should still fix the broken documentation.

MarcusGrass:
> In general, adding the means to see dropped items leads down the rabbit hole of resource aware flows and item lifecycle, which aren't well explored at the moment and would require the redesign of the library from the ground up for sure.

Could you elaborate on this a bit?
`onBackpressureDrop` takes a consumer that in my understanding does what, to me, seems to be the same as supplying a Consumer instead of an action in `onBackpressureBuffer`. If you have the time or some information that you could link on why one becomes problematic while the other doesn't that would be much appreciated!

akarnokd:
The underlying problem is that with resource-aware flows and item lifecycle support, items can't be just lost in face of all the signals exchanged between operators. For that, item handover and cleanup has to be carefully implemented which affects operators as well as schedulers. [It can become quite complicated pretty fast](https://github.com/akarnokd/akarnokd-misc/tree/master/src/main/java/hu/akarnokd/reactive).

`onBackpressureDrop` does not buffer thus either it [hands the item along or hands it to the callback](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureDrop.java#L82-L94) and as such is not affected by other signals at that point.

There is no such clean handover with the standard `onBackpressureBuffer`. Adding a callback API without changing the entire underlying behavior just invites the wrong usage and a myriad of caveats. Therefore, we don't go there with RxJava proper and delegate it to 3rd party implementations where the consequences can be contained much better.

PR #7083, opened 2020-09-23
3.x: Fix toFlowable(ERROR) not cancelling on MBE

### LINKED ISSUE #7081: 2.x - Observable.toFlowable(ERROR) not releasing upstream on MissingBackPressure
Hi

Code sample :
```
Observable.interval(1, TimeUnit.MILLISECONDS)
				.doFinally(() -> System.err.println("finally obs"))
				.toFlowable(BackpressureStrategy.ERROR)
				.doFinally(() -> System.err.println("finally flow"))
				.map(any -> {
					System.out.println("any = " + any);
					Thread.sleep(1000);
					return any;
				})
				.subscribe(any -> {
				}, err -> {
					System.err.println(err.getMessage());
				});
```
With this code, a MissingBackPressureException is quickly generated and propagated through the downstream of `.toFlowable(BackpressureStrategy.ERROR)`, but it looks like the upstream is never disposed. Am I missing something ?

Tested with rxjava 2.2.19

akarnokd:
There was a missing `upstream.cancel()` for when the overflow happens.

Fixes #7081

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureError.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureError.java
@@ -65,6 +65,7 @@ public final class FlowableOnBackpressureError<T> extends AbstractFlowableWithUp
                 downstream.onNext(t);
                 BackpressureHelper.produced(this, 1);
             } else {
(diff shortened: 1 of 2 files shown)


vanniktech (approved):
Also here, does Observable behave correctly?

PR #7110, opened 2020-11-12
3.x: Flowable scan/scanWith backpressure documentation update

### LINKED ISSUE #7109: Flowable scan operators backpressure inconsistency
Hello,

Flowable operator scan (without initial value) pulls values according to downstream requests, but other scan variations will pull in buffersize() chunks.

RxJava version: 3.0.7

Code (if kotlin is not acceptable, let me know and I will amend to a Java version):

```
class ScanTest {
	@Test
	fun `scan with initial`() {
		var requested = 0L
		Flowable.never<Int>()
			.doOnRequest { requested += it }
			.scan(0, Int::plus)
			.test(2)

		assert(requested == 1L) { "Requested was $requested and not 1" }
	}

	@Test
	fun `scan without initial`() {
		var requested = 0L
		Flowable.never<Int>()
			.doOnRequest { requested += it }
			.scan(Int::plus)
			.test(1)

		assert(requested == 1L) { "Requested was $requested and not 1" }
	}

	@Test
	fun `scanWith`() {
		var requested = 0L
		Flowable.never<Int>()
			.doOnRequest { requested += it }
			.scanWith({ 0 }, Int::plus)
			.test(2)

		assert(requested == 1L) { "Requested was $requested and not 1" }
	}
}
```

only the second test passes.

Trace:

```
java.lang.AssertionError: Requested was 127 and not 1

	at ScanTest.scanWith(ScanTest.kt:35)


java.lang.AssertionError: Requested was 127 and not 1

	at ScanTest.scan with initial(ScanTest.kt:13)
```

My expectation would be that after supplying the initial value, the remaining values would be requested up as the downstream requires (akin to the no initial value scan).

fsbarata:
Documented Flowable initial value scan variants upstream consumption pattern as it is different from the variant with no initial value.

Resolves #7109

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -14735,9 +14735,9 @@ public abstract class Flowable<@NonNull T> implements Publisher<T> {
      * <dl>
      *  <dt><b>Backpressure:</b></dt>
      *  <dd>The operator honors downstream backpressure and expects the current {@code Flowable} to honor backpressure as well.
-     *  Violating this expectation, a {@link MissingBackpressureException} <em>may</em> get signaled somewhere downstream.</dd>
-	 *  <dd>The downstream request pattern is not preserved across this operator.
-	 *  The upstream is requested {@link #bufferSize()} - 1 upfront and 75% of {@link #bufferSize()} thereafter.</dd>
+     *  Violating this expectation, a {@link MissingBackpressureException} <em>may</em> get signaled somewhere downstream.
+     *  The downstream request pattern is not preserved across this operator.
+     *  The upstream is requested {@link #bufferSize()} - 1 upfront and 75% of {@link #bufferSize()} thereafter.</dd>
      *  <dt><b>Scheduler:</b></dt>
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -14736,6 +14736,8 @@ public final void safeSubscribe(@NonNull Subscriber<@NonNull ? super T> subscrib
      *  <dt><b>Backpressure:</b></dt>
      *  <dd>The operator honors downstream backpressure and expects the current {@code Flowable} to honor backpressure as well.
      *  Violating this expectation, a {@link MissingBackpressureException} <em>may</em> get signaled somewhere downstream.</dd>
+	 *  <dd>The downstream request pattern is not preserved across this operator.
```
akarnokd, line 14739: You don't have to start a new `<dd>` here, just add the text on a new line inside the previous `<dd>`.

fsbarata: - [x] merge `<dd>` sections

akarnokd (approved):

akarnokd:
Thanks!

fsbarata:
My pleasure!

PR #7117, opened 2020-11-25
Add NonNull annotations to parameters which are null checked

hqzxzwb:
This PR is adding NonNull annotations to method parameters that are null checked.
It will help Kotlin compiler to recognize code that might produce a NullPointerException.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/processors/ReplayProcessor.java
+++ b/src/main/java/io/reactivex/rxjava3/processors/ReplayProcessor.java
@@ -368,7 +368,7 @@ public final class ReplayProcessor<T> extends FlowableProcessor<T> {
     }
 
     @Override
-    public void onNext(T t) {
+    public void onNext(@NonNull T t) {
         ExceptionHelper.nullCheck(t, "onNext called with a null value.");
 
         if (done) {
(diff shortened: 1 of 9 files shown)


akarnokd:
[The type annotation](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/core/Observable.java#L102) on the parent type should be still in effect, hence the lack of `@NonNull` on these methods. Do you have a problem in your IDE without them?

hqzxzwb:
Yes I have some code that is calling onNext with nullable arguments, and passes the build. But just now I realized that calling Observer.onNext with explicit `@NonNull` on the parameter is still compiling, which is just weird. I will try to find more about it.

hqzxzwb:
Well. Maybe the real problem is Kotlin compiler does not know about RxJava3 annotation yet.
I checked [this file](https://github.com/JetBrains/kotlin/blob/master/core/compiler.common.jvm/src/org/jetbrains/kotlin/load/java/JvmAnnotationNames.kt) of Kotlin and found that it does not have RxJava3 annotation in the list. Would the IDE really work with these annotations at all?

akarnokd:
I'd say this is a matter for the Kotlin compiler side. I recommend opening an issue/PR with them first.

hqzxzwb:
Agreed. I searched and found an open issue in Kotlin project.
https://github.com/JetBrains/kotlin/pull/3591

PR #7124, opened 2020-12-05
3.x: Add onBackpressureReduce operator

ansip:
Added one more backpressure strategy. It works like backpressure latest but instead of just dropping values it merge/reduce excessive items


src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        @Override
+        public void onNext(T t) {
+            T v = current.get();
+            if (v == null) {
+                current.lazySet(t);
+            } else if ((v = current.getAndSet(null)) != null) {
+                try {
+                    current.lazySet(reducer.apply(v, t));
```
akarnokd: The reducer's return value should be checked for `null`.

src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceTest.java
```diff
@@ -0,0 +1,236 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.reactivestreams.Publisher;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+public class FlowableOnBackpressureReduceTest extends RxJavaTest {
+
+    static final BiFunction<Integer, Integer, Integer> TEST_INT_REDUCER = new BiFunction<Integer, Integer, Integer>() {
```
akarnokd: You can use lambdas in tests.

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
```diff
@@ -0,0 +1,68 @@
...
+import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.exceptions.Exceptions;
+import io.reactivex.rxjava3.functions.BiFunction;
+import org.reactivestreams.Subscriber;
+
+import java.util.Objects;
+
+public final class FlowableOnBackpressureReduce<T> extends AbstractFlowableWithUpstream<T, T> {
+
+    private final BiFunction<T, T, T> reducer;
```
akarnokd: Please remove `private`.

akarnokd (commented):
Technically a good PR, but I'm not yet convinced such an operation is in high enough demand to be included in RxJava.

In addition, the idea of reducing while being backpressured usually involves two types: The upstream type `T` and a collection type `C` that holds the aggregate and is the output type. The most common quoted use case involves `T` and `List<T>`; this has the drawback that individual `T`s have to be always wrapped to be able to reach downstream.

The operator in this PR is limited to `T` -> `T` transformations of which not many standard types can do meaningful reduction. Of course, people could try with first mapping `T` into an `ArrayList<T>`, then the reducer just adds to this list, however, this can result in unbounded memory usage as the list grows in a backpressured situation.

Consequently, the demand and usage considerations of this operator doesn't strike me as a prime candidate for RxJava proper, but may live as a `FlowableTransformer` in a 3rd party library. There is a similar operator [coalesce](https://github.com/akarnokd/RxJavaExtensions#flowabletransformerscoalesce) available for example.

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
```diff
@@ -0,0 +1,68 @@
...
+            if (v == null) {
+                current.lazySet(t);
+            } else if ((v = current.getAndSet(null)) != null) {
+                try {
+                    current.lazySet(reducer.apply(v, t));
+                } catch (Throwable throwable) {
+                    Exceptions.throwIfFatal(throwable);
+                    cancel();
+                    onError(throwable);
+                    return;
```
akarnokd: The failure path is not covered: https://codecov.io/gh/ReactiveX/RxJava/pull/7124/diff?src=pr&el=tree#D4-55

ansip:
> Technically a good PR, but I'm not yet convinced such an operation is in high enough demand to be included in RxJava.
> 
> In addition, the idea of reducing while being backpressured usually involves two types: The upstream type `T` and a collection type `C` that holds the aggregate and is the output type. The most common quoted use case involves `T` and `List<T>`; this has the drawback that individual `T`s have to be always wrapped to be able to reach downstream.
> 
> The operator in this PR is limited to `T` -> `T` transformations of which not many standard types can do meaningful reduction. Of course, people could try with first mapping `T` into an `ArrayList<T>`, then the reducer just adds to this list, however, this can result in unbounded memory usage as the list grows in a backpressured situation.
> 
> Consequently, the demand and usage considerations of this operator doesn't strike me as a prime candidate for RxJava proper, but may live as a `FlowableTransformer` in a 3rd party library. There is a similar operator [coalesce](https://github.com/akarnokd/RxJavaExtensions#flowabletransformerscoalesce) available for example.

Ok, let me explain why I decided to open this PR here

Real usage example in our company (and I suppose and other companies as well): typically we have many applications with grids, tables of data
The tables have live updates coming after snapshot received on the client side
So the first update is the snapshot and after it server sends only deltas (row updates)
That means that here we are not ready to loose any data and operators like latest and drop are useless
The only operator from standard library which we can use in this case is backpressure buffer
But the problem with this operator is that if we have slow consumer client/UI or we have slow network etc. 
We can just simple receive MissingBackpressureException (in best case) or OOM in worst case
Because this operator collects data - and that is we don't want to do

Why not 3rd party

3rd party is a good point, but experience shows nobody knows about them :)
As the result I saw many times how developers reinvent wheel trying to solve this problem

Coalesce is too generic and I would say it is not the similar operator, it looks more like backpressure buffer and collects items (so it has the same drawbacks as I described above)
Yes it can solve our problem but as you mentioned there are additional allocations for list and increased gc pressure so it does not look like a good choice
And if we think in this way coalesce can replace latest and drop as well... You can just keep single first element in list and this would be drop implementation 
or you can replace first element in list and this would be latest operator implementation
Does it mean that latest and drop should be moved from RxJava to extentions as well?

Why only T type

This was done intentionally to keep single responsibility and enforce user to return the same type
This operator is not a mapper it just allows user to choose what to do with updates, keep first or second or calculate new items
If user need to collect updates nobody restrcits to make T as a container itself

Update {
	List rows = ....
}

and in reduce user can merge to lists in a single Update with rows list
If user needs mapping he can do this right after this operator
.onBackpressureReduce()
.map()


So that was my thoughts when I raised this PR
Of course I think you have more experince in RxJava than me and more clear view on how it should develop
I don't mind to move this to 3rd party lib, extensions or decline.
We already have implementation inside our company so this problem solved for us
This PR was just an attempt to share my experience

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
ansip: done
just wondering is there any reason or best practice to make all field package-private?

src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceTest.java
ansip: changed to lambdas

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
ansip: added

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
akarnokd: To avoid any accidental private accessors generated.

akarnokd:
I see. The no-dataloss possibility and the low complexity of the operator is a plus. @vanniktech, @davidmoten, anybody else, any opinions?

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduce.java
ansip: added tests

davidmoten:
>I see. The no-dataloss possibility and the low complexity of the operator is a plus. @vanniktech, @davidmoten, anybody else, any opinions?

I'm happy for this one to be included especially as the complexity is low. I wouldn't mind seeing a bit more consideration given to the typing as it is very unusual for an aggregating operator to not offer overloads that support a different downstream type.

vanniktech (approved):
I'm fine with adding this new operator.

Should not this be @Experimental?

akarnokd:
> Should not this be @experimental?

Yes, also target `3.0.9`. 

> I wouldn't mind seeing a bit more consideration given to the typing as it is very unusual for an aggregating operator to not offer overloads that support a different downstream type.

Agree, [reduce(BiFunction)](http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Flowable.html#reduce-io.reactivex.rxjava3.functions.BiFunction-) and [reduceWith(Supplier, BiFunction)](http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Flowable.html#reduceWith-io.reactivex.rxjava3.functions.Supplier-io.reactivex.rxjava3.functions.BiFunction-) tend to go together. Here, the `seed` variant wouldn't make much sense so both can be called `onBackpressureReduce`.

Okay then, we accept this new operator. I'll add some comments about where to adjust things.

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -12676,6 +12676,39 @@ public final Completable ignoreElements() {
...
+     *  <dd>The operator honors backpressure from downstream and consumes the current {@code Flowable} in an unbounded
+     *  manner (i.e., not applying backpressure to it).</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>{@code onBackpressureReduce} does not operate by default on a particular {@link Scheduler}.</dd>
+     * </dl>
+     * @param reducer the bi-function to call when there is more than one non-emitted value to downstream,
+     * the first argument of the bi-function is previous item and the second one is currently emitting from upstream
+     * @return the new {@code Flowable} instance
+     * @throws NullPointerException if {@code reducer} is {@code null}
+     * @since 3.1.0
```
akarnokd, line 12702: Please use `@since 3.0.9 - experimental`.

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -12676,6 +12676,39 @@ public final Completable ignoreElements() {
...
+     */
+    @CheckReturnValue
+    @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
+    @SchedulerSupport(SchedulerSupport.NONE)
+    @NonNull
+    public final Flowable<T> onBackpressureReduce(@NonNull BiFunction<T, T, T> reducer) {
```
akarnokd, line 12708: Please annotate with `@Experimental`.

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -12676,6 +12676,39 @@ public final Completable ignoreElements() {
         return RxJavaPlugins.onAssembly(new FlowableOnBackpressureLatest<>(this));
     }
 
+    /**
+     * Reduces a sequence of two not emitted values via a function into a single value if the downstream is not ready to receive
+     * new items (indicated by a lack of {@link Subscription#request(long)} calls from it) and emits this latest
+     * item when the downstream becomes ready.
+     * <p>
+     * <img width="640" height="245" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/bp.obp.reduce.v3.png" alt="">
```
akarnokd, line 12684: I've created the marble for this operator: `https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Flowable.onBackpressureReduce.png`. Scaled-down size `640 x 354`.

image

akarnokd (changes requested):

src/main/java/io/reactivex/rxjava3/core/Flowable.java
ansip: done

src/main/java/io/reactivex/rxjava3/core/Flowable.java
ansip: added

src/main/java/io/reactivex/rxjava3/core/Flowable.java
ansip: Thanks!

ansip:
> > Should not this be @experimental?
> 
> Yes, also target `3.0.9`.
> 
> > I wouldn't mind seeing a bit more consideration given to the typing as it is very unusual for an aggregating operator to not offer overloads that support a different downstream type.
> 
> Agree, [reduce(BiFunction)](http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Flowable.html#reduce-io.reactivex.rxjava3.functions.BiFunction-) and [reduceWith(Supplier, BiFunction)](http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Flowable.html#reduceWith-io.reactivex.rxjava3.functions.Supplier-io.reactivex.rxjava3.functions.BiFunction-) tend to go together. Here, the `seed` variant wouldn't make much sense so both can be called `onBackpressureReduce`.
> 
> Okay then, we accept this new operator. I'll add some comments about where to adjust things.

Ok, so should I add `onBackpressureReduce(Supplier<R> seedSupplier, BiFunction<R, ? super T, R> reducer)`?

If yes could you please advise on the method semantic
Should `seedSupplier` be called for each new coming T from upstream in case if there is no backpressure (downstream requested MAX_VALUE)? Or it should be called once and we should store this value inside this operator and pass it to reducer each time?

akarnokd:
The output type is `R` thus you can't relay `T`s anymore. Consequently, even if the downstream is ready, you have to call the supplier and the reducer.

akarnokd:
> Ok, so should I add onBackpressureReduce(Supplier<R> seedSupplier, BiFunction<R, ? super T, R> reducer)?

Yes, you can do it in this PR or in a fresh PR after this gets merged.

`https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Flowable.onBackpressureReduce.ff.png` 640 x 315

image

ansip:
> he reducer.

Ok got it

ansip:
> > Ok, so should I add onBackpressureReduce(Supplier seedSupplier, BiFunction<R, ? super T, R> reducer)?
> 
> Yes, you can do it in this PR or in a fresh PR after this gets merged.
> 
> `https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/Flowable.onBackpressureReduce.ff.png` 640 x 315
> 
> image

Yes I would prefer to raise separate PR, I suppose it would be easier to review smaller changeset
Will do this today or tomorrow

akarnokd (approved):

PR #7128, opened 2020-12-07
Implement Observable.publish with completeOnDisposal

sebek64:
This is a sketch of a feature I'd often find useful. ConnectableObservable.connect returns a Disposable, but disposing it actually leaves the observers in a strange state. They have not been disposed, and no terminal event has been delivered to them either. However, any newly-subscribed observer will actually receive a terminal event.

This PR adds a switch to Observable.publish method, enabling a "more correct" behavior.

This PR is just a draft. I will happily adjust it according to the feedback. I don't want to spend hours of polishing it to just get a rejection in the end. Maybe I miss something important and this change cannot be done, or the solution with the switch is incorrect and we should rather change the behavior of the publish() method.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Observable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Observable.java
@@ -10842,6 +10842,29 @@ public abstract class Observable<@NonNull T> implements ObservableSource<T> {
     }
 
     /**
+     * Returns a {@link ConnectableObservable}, which is a variety of {@link ObservableSource} that waits until its
+     * {@link ConnectableObservable#connect connect} method is called before it begins emitting items to those
+     * {@link Observer}s that have subscribed to it.
+     * <p>
+     * <img width="640" height="510" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/publishConnect.v3.png" alt="">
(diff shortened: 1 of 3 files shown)


akarnokd:
This is inherently unsafe because `onNext` events may be in-flight and calling `onComplete` from an arbitrary thread concurrently is forbidden. 

Also this PR is a significant behavior change which we don't generally do within a minor/patch release.

The intended dispose strategy for connectables is to first have all observers dispose, then dispose the upstream. To emulate the desired behavior, use `takeUntil` with a subject that will complete observers properly while disposing as intended:

```java
PublishSubject<Object> terminate = PublishSubject.create();

ConnectableObservable<T> conn = ...

Observable<T> source = conn.takeUntil(terminate);

source.subscribe(/* ... */);

Disposable d = conn.connect();

// ....

terminate.onNext(0);
d.dispose();
```

sebek64:
Thanks for the quick reaction. I didn't realize this problem with Observable contract. It is probably unfixable.

According to the code snippet you provided, I think the last two lines should be swapped to get the correct behavior, otherwise the upstream will continue producing elements which will be discarded by takeUntil. But I get the point, it will be quite a sufficient workaround I guess.

Closing the PR.

PR #7129, opened 2020-12-09
3.x: Add onBackpressureReduce operator

ansip:
added one more overloaded version of onBackpressureReduce operator
```
@Experimental
    @CheckReturnValue
    @NonNull
    public final <R> Flowable<R> onBackpressureReduce(@NonNull Supplier<R> supplier, @NonNull BiFunction<R, ? super T, R> reducer) {
        return RxJavaPlugins.onAssembly(new FlowableOnBackpressureReduceWith<>(this, supplier, reducer));
    }
```

Changes:
- Added `FlowableOnBackpressureReduceWith`
- `AbstractBackpressureThrottlingSubscriber` has two types now T - upstream, R- downstream and inherited by 3 classes: `BackpressureLatestSubscriber`, `BackpressureReduceSubscriber` and `BackpressureReduceWithSubscriber`

suggest to review and merge after [this ](https://github.com/ReactiveX/RxJava/pull/7124)PR merged


akarnokd:
Could you please rebase this onto a fresh `3.x`?

ansip:
> Could you please rebase this onto a fresh `3.x`?

done

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -12710,6 +12710,45 @@ public final Completable ignoreElements() {
         return RxJavaPlugins.onAssembly(new FlowableOnBackpressureReduce<>(this, reducer));
     }
 
+    /**
+     * The operator applies {@code reducer} to each emitted item from upstream. This operator uses {@code supplier} to
+     * create new items of type R to pass it with value of type T from upstream to {@code reducer} or in case if the downstream is not ready to receive
+     * new items (indicated by a lack of {@link Subscription#request(long)} calls from it) it uses the latest returned value
+     * from {@code reducer} for previous not consumed item from upstream.
+     * When the downstream becomes ready it emits the latest value returned by {@code reducer} to the downstream.
```
akarnokd: This is not a good first sentence to describe what the operator is, and I'd probably not write down these details at all. Compare to the previous `onBackpressureReduce`:

> Reduces a sequence of two not emitted values via a function into a single value if the downstream is not ready to receive
> new items (indicated by a lack of {@link Subscription#request(long)} calls from it) and emits this latest
> item when the downstream becomes ready.

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWith.java
```diff
@@ -0,0 +1,80 @@
...
+                    ));
+                } else if ((v = current.getAndSet(null)) != null) {
+                    current.lazySet(Objects.requireNonNull(reducer.apply(v, t), "The reducer returned a null value"));
+                } else {
+                    current.lazySet(Objects.requireNonNull(
+                            reducer.apply(Objects.requireNonNull(supplier.get(), "The supplier returned a null value"), t),
+                            "The reducer returned a null value"
+                    ));
+                }
+                drain();
```
akarnokd: `drain()` should be outside the `try-catch`.

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWith.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Objects;
+
+public final class FlowableOnBackpressureReduceWith<T, R> extends AbstractFlowableWithUpstream<T, R> {
+
+    final BiFunction<R, ? super T, R> reducer;
+    final Supplier<R> supplier;
+
+    public FlowableOnBackpressureReduceWith(Flowable<T> source, Supplier<R> supplier, BiFunction<R, ? super T, R> reducer) {
+        super(source);
+        this.reducer = Objects.requireNonNull(reducer, "reducer is null");
```
akarnokd: Null-checks should be part of the `Flowable` method.

akarnokd: Same goes for `FlowableOnBackpressureReduce()` constructor.

akarnokd (changes requested):

src/main/java/io/reactivex/rxjava3/core/Flowable.java
akarnokd: > Reduces upstream values into an aggregate value, provided by a supplier and combined via a reducer function, while the downstream is not ready to receive items, then emits this aggregate value when the downstream becomes ready.
>
> ...image...
>
> Note that even if the downstream is ready to receive an item, the upstream item will always be aggregated into the output type, calling both the supplier and the reducer to produce the output value.

src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -12710,6 +12710,45 @@ public final Completable ignoreElements() {
...
+     * </dl>
+     * @param supplier the factory to call to create new item of type R to pass it as the first argument to {@code reducer}.
+     * It is called when previous returned value by {@code reducer} already sent to downstream or the very first update from upstream received.
+     * @param reducer the bi-function to call to reduce excessive updates which downstream is not ready to receive.
+     * The first argument of type R is the object returned by {@code supplier} or result of previous {@code reducer} invocation.
+     * The second argument of type T is the current update from upstream.
+     * @return the new {@code Flowable} instance
+     * @throws NullPointerException if {@code supplier} or {@code reducer} is {@code null}
+     * @since 3.0.9 - experimental
+     */
```
akarnokd, line 12742: `@see #onBackpressurReduce(BiFunction)`

ansip: added

src/main/java/io/reactivex/rxjava3/core/Flowable.java
ansip: Thanks, changed

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWith.java
ansip: moved outside

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWith.java
ansip: moved to Flowable methods

akarnokd (approved):

vanniktech (approved):

PR #7151, opened 2021-01-17
3.x: onReduceBackpressure internals cleanup

akarnokd:
- Simplify empty/non-empty cases.
- Fix cancellation order and wrong fall-through.
- Add deterministic test to cover the reduction cases.

### DIFF
--- a/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWithTest.java
+++ b/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWithTest.java
@@ -95,7 +95,7 @@ public class FlowableOnBackpressureReduceWithTest extends RxJavaTest {
         PublishProcessor<Integer> source = PublishProcessor.create();
         TestSubscriberEx<Integer> ts = new TestSubscriberEx<>(0L);
 
-        source.onBackpressureReduce(() -> 0, (a, b) -> a + b).subscribe(ts);
+        source.onBackpressureReduce(() -> 0, Integer::sum).subscribe(ts);
 
         source.onNext(1);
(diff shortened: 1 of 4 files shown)


src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWithTest.java
```diff
@@ -90,6 +90,30 @@ public void simpleBackpressure() {
         ts.assertNotComplete();
     }
 
+    @Test
+    public void reduceBackpressuredSync() {
+        PublishProcessor<Integer> source = PublishProcessor.create();
+        TestSubscriberEx<Integer> ts = new TestSubscriberEx<>(0L);
+
+        source.onBackpressureReduce(() -> 0, (a, b) -> a + b).subscribe(ts);
```
ansip: Could be replaced with `Integer::sum`

src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableOnBackpressureReduceWithTest.java
```diff
@@ -90,6 +90,30 @@ public void simpleBackpressure() {
         ts.assertNotComplete();
     }
 
+    @Test
+    public void reduceBackpressuredSync() {
```
ansip, line 94: what is the difference from `synchronousDrop` test, maybe just rename it to `reduceBacpressuredSync`?

akarnokd: Different request pattern.

vanniktech (approved):

PR #7160, opened 2021-01-25
Io Scheduler, Scheduled worker release.

### LINKED ISSUE #7153: 3.x and 2.x - Schedulers.io() fails to produce new thread when needed, leading to deadlock
I've tested this with the latest versions of 3.x and 2.x (v3.0.9 and v2.2.20 at the time)

If my understanding is correct, `Schedulers.io()` should create a new thread if all the other threads are busy. This is not happening here, and it leads to a deadlock.

```
Flowable.just("item")
        .observeOn(Schedulers.io())
        .firstOrError()
        .subscribe(__ -> {
            for (int i = 0; i < 200; i++) {
                Completable.complete()
                        .observeOn(Schedulers.io())
                        .blockingAwait();
            }
            System.out.println("Will never reach this");
        });
```
I managed to simplify our code to this minimal reproducible example. I'm aware that `.blockingAwait()` doesn't make sense here but it does in the original code

The issue is easier to reproduce if the IO thread pool contains just a few threads, but it can still happen with bigger pools.

I think it can be related to some operator releasing the worker too soon. It will work fine if I replace the `firstOrError` operator like this

```
Single.just("item")
        .observeOn(Schedulers.io())
        .subscribe(__ -> {
            for (int i = 0; i < 200; i++) {
                Completable.complete()
                        .observeOn(Schedulers.io())
                        .blockingAwait();
            }
            System.out.println("NO problem here");
        });
```

Any help with this is much appreciated.
Cheers

sgc-code:
`IoScheduler` releases the workers back to the pool to be reused optimistically to reduce the number of threads being created. This could lead to deadlocks in some situations like this one:

```
Flowable.just("item")
        .observeOn(Schedulers.io())
        .firstOrError()
        .subscribe(__ -> {
            for (int i = 0; i < 200; i++) {
                Completable.complete()
                        .observeOn(Schedulers.io())
                        .blockingAwait();
            }
            System.out.println("Will never reach this");
        });
```
As discussed on https://github.com/ReactiveX/RxJava/issues/7153, this PR allow to configure this behaviour.
Enabling `rx3.io-scheduled-release` will guarantee that workers are not reused until all the processing has finished. 

Once we merge this we, I can create a similar PR for the `2.x` branch


src/main/java/io/reactivex/rxjava3/internal/schedulers/IoScheduler.java
```diff
@@ -212,11 +218,20 @@ public void dispose() {
             if (once.compareAndSet(false, true)) {
                 tasks.dispose();
 
-                // releasing the pool should be the last action
-                pool.release(threadWorker);
+                if (USE_SCHEDULED_RELEASE) {
+                    threadWorker.scheduleActual(this, 0, TimeUnit.NANOSECONDS, null);
```
sgc-code, line 222: I'm not sure if this is preferred to `threadWorker.scheduleDirect(this, 0, TimeUnit.NANOSECONDS);`
Any comment about this?

src/test/java/io/reactivex/rxjava3/schedulers/IoScheduledReleaseTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Test;
+
+import java.util.concurrent.TimeUnit;
+
+public class IoScheduledReleaseTest extends RxJavaTest {
+
+    /* This test will be stuck in a deadlock if rx3.io-scheduled-release is not set */
+    @Test
+    public void scheduledRelease() {
+        System.setProperty("rx3.io-scheduled-release", Boolean.TRUE.toString());
```
akarnokd: This will not work because the classes have been already initialized. For a workaround, change

`private static final boolean USE_SCHEDULED_RELEASE;`

to

`static boolean USE_SCHEDULED_RELEASE;`

move this `IoScheduledReleaseTest` to `package io.reactivex.rxjava3.internal.schedulers;`,

then:

```java
boolean save = IoScheduler.USE_SCHEDULED_RELEASE;
IoScheduler.USE_SCHEDULED_RELEASE = true;
try {
  // ...
} finally {
    IoScheduler.USE_SCHEDULED_RELEASE = save;
}
```

src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
```diff
@@ -32,6 +32,9 @@
  * <ul>
  * <li>{@code rx3.io-keep-alive-time} (long): sets the keep-alive time of the {@link #io()} Scheduler workers, default is {@link IoScheduler#KEEP_ALIVE_TIME_DEFAULT}</li>
  * <li>{@code rx3.io-priority} (int): sets the thread priority of the {@link #io()} Scheduler, default is {@link Thread#NORM_PRIORITY}</li>
+ * <li>{@code rx3.io-scheduled-release} (boolean): Changes release behaviour mode for {@link #io()} Scheduler.
```
akarnokd: No need to be so verbose here.

```
<li>{@code rx3.io-scheduled-release} (boolean): {@code true} sets the worker release mode of the
{@link #io()} Scheduler to <em>scheduled</em>, default is {@code false} for <em>eager</em> mode.</li>
```

src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
```diff
@@ -159,6 +162,9 @@ public static Scheduler computation() {
      * <ul>
      * <li>{@code rx3.io-keep-alive-time} (long): sets the keep-alive time of the {@code io()} Scheduler workers, default is {@link IoScheduler#KEEP_ALIVE_TIME_DEFAULT}</li>
      * <li>{@code rx3.io-priority} (int): sets the thread priority of the {@code io()} Scheduler, default is {@link Thread#NORM_PRIORITY}</li>
+     * <li>{@code rx3.io-scheduled-release} (boolean): Changes release behaviour mode for {@code #io()} Scheduler.
```
akarnokd: No need to be so verbose here.

```
<li>{@code rx3.io-scheduled-release} (boolean): {@code true} sets the worker release mode of the
{@link #io()} Scheduler to <em>scheduled</em>, default is {@code false} for <em>eager</em> mode.</li>
```

src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
```diff
@@ -159,6 +162,9 @@ public static Scheduler computation() {
...
+     *    Enabling this guarantees that workers are not released before they finish all processing, therefore preventing deadlocks.
```
akarnokd: Add a new paragraph at the end of this doc instead:

```
<p>
When the {@link Scheduler.Worker} is disposed, the underlying worker can be released to the cached worker pool in two modes:
<ul>
<li>In <em>eager</em> mode (default), the underlying worker is returned immediately to the cached worker pool
 and can be reused much quicker by operators. The drawback is that if the currently running task doesn't
respond to interruption in time or at all, this may lead to delays or deadlock with the reuse use of the
underlying worker.
</li>
<li>In <em>scheduled</em> mode (enabled via the system parameter {@code rx3.io-scheduled-release}
set to {@code true}), the underlying worker is returned to the cached worker pool only after the currently running task
has finished. This can help prevent premature reuse of the underlying worker and likely won't lead to delays or
deadlock with such reuses. The drawback is that the delay in release may lead to an excess amount of underlying
workers being created.
</li>
</ul>
```

akarnokd (changes requested):

src/main/java/io/reactivex/rxjava3/internal/schedulers/IoScheduler.java
akarnokd: Doesn't really matter.

src/test/java/io/reactivex/rxjava3/schedulers/IoScheduledReleaseTest.java
sgc-code: I don't know how it was running on my machine, but this makes sense, 👍

src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
sgc-code: done

src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
sgc-code: Changed, Please double check it's in the right place now

akarnokd (approved):

vanniktech (approved):

PR #7169, opened 2021-01-27
3.x: Introduce property rx3.scheduler.use-nanotime

### LINKED ISSUE #7154: 3.x Feature: introduce abstract time source for scheduler impl
Version:
rxjava:3.0.9

Related:
https://github.com/ReactiveX/RxJava/issues/2943

Precondition:
* Task with delay scheduled in a `Scheduler`

Problem:
When the linux kernel wakes up from suspension `System.currentTimeMillis` will be adjusted on linux x86 OpenJDK. In my case this is bad, because watchdogs/ timer will fire, which they shouldn't. I would like to use a different time-source, but can't, because `System.currentTimeMillis` is hard-coded.

```java
    public long now(@NonNull TimeUnit unit) {
        return unit.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
    }
```

Solution:
I would like to introduce a `TimeSource` interface, which can be set via `RxJavaPlugins`.

```kotlin
interface TimeSource {
    fun now(unit : TimeUnit) : Long
}
```

Impl
```kotlin
class NanoSource : TimeSource {
    override fun now(unit: TimeUnit): Long {
        return unit.convert(System.nanoTime(), TimeUnit.NANOSECONDS)
    }
}
```

In order to not break the current behavior the default TimeSource would still be `System.currentTimeMillis()`. 

This idea was pitched by: @artem-zinnatullin in https://github.com/ReactiveX/RxJava/issues/2943#issuecomment-102455813

If this is fine request is fine with the community I would try to implement it quickly and create a PR for it.

SergejIsbrecht:
Fixes #7154

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Scheduler.java
@@ -60,8 +60,9 @@ import io.reactivex.rxjava3.schedulers.SchedulerRunnableIntrospection;
  * interface which can grant access to the original or hooked {@code Runnable}, thus, a repeated {@code RxJavaPlugins.onSchedule}
  * can detect the earlier hook and not apply a new one over again.
  * <p>
- * The default implementation of {@link #now(TimeUnit)} and {@link Worker#now(TimeUnit)} methods to return current
- * {@link System#currentTimeMillis()} value in the desired time unit. Custom {@code Scheduler} implementations can override this
+ * The default implementation of {@link #now(TimeUnit)} and {@link Worker#now(TimeUnit)} methods to return current {@link System#currentTimeMillis()}
+ * value in the desired time unit, unless {@code rx3.scheduler.use-nanotime} (boolean) is set. When the property is set to
+ * {@code true}, the method uses {@link System#nanoTime()} as its basis instead. Custom {@code Scheduler} implementations can override this
  * to provide specialized time accounting (such as virtual time to be advanced programmatically).
  * Note that operators requiring a {@code Scheduler} may rely on either of the {@code now()} calls provided by
  * {@code Scheduler} or {@code Worker} respectively, therefore, it is recommended they represent a logically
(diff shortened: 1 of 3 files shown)


SergejIsbrecht:
@akarnokd ,

I got a problem, when I try to test the code, because when the class is loaded, one can not change set variable:

`private static final String DRIFT_USE_NANOTIME = "rx3.scheduler.drift-use-nanotime"; 

Resolution:
* no tests
* define a own test set, which tests the nanoTime use-case and configure this test-src-set to run each class in a new JVM
* find a Junit-runner which loads the test-classes via a own ClassLoader
* remove "final" for `DRIFT_USE_NANOTIME` and set it for testing, which might not be desirable, because it can be changed later.

What are your thoughts on this issue regarding testing?

src/main/java/io/reactivex/rxjava3/core/Scheduler.java
```diff
@@ -88,6 +89,28 @@
  * All methods on the {@code Scheduler} and {@code Worker} classes should be thread safe.
  */
 public abstract class Scheduler {
+    static final boolean IS_DRIFT_USE_NANOTIME = Boolean.getBoolean("rx3.scheduler.drift-use-nanotime");
```
akarnokd: This can't be tested because the class gets initialized much earlier. Please drop the `final`. Also please document the field with Javadocs.

akarnokd (changes requested):

akarnokd (commented):
Since this affects now(), let's call the property just `rx3.scheduler.use-nanotime`.

akarnokd (approved):

SergejIsbrecht:
@akarnokd , I'll changed the commit message as-well, to match the new property name.

akarnokd:
No need for that.

vanniktech (approved):

PR #7170, opened 2021-01-28
2.x: Introduce property rx2.scheduler.use-nanotime (#7154)

SergejIsbrecht:
Fixes #7169


SergejIsbrecht:
@akarnokd , I had to create a new `SchedulerTest` class, because on the 2.x branch the package did not match the package of `Scheduler`, therefore I could not access the field (package-private).

akarnokd (approved):

vanniktech (approved):

PR #7173, opened 2021-01-30
3.x: Workaround for FutureTask.toString + JDK 11 build

### LINKED ISSUE #7172: 3.x StackOverflowError on prinln of Disposable returned from Scheduler
Version:
io.reactivex.rxjava3:rxjava:3.0.6

Example
```kotlin
    @Test
    fun stackoverflow() {
        val scheduleDirect = Schedulers.single().scheduleDirect {
            while (true) { }
        }

        println(scheduleDirect)
    }
```

This fails as-well:
```kotlin
    @Test
    fun stackoverflow() {
        val scheduleDirect = Schedulers.single().scheduleDirect({
            while (true) {
                // break
            }
        }, 1000, TimeUnit.MILLISECONDS)

        println(scheduleDirect)
    }
```

but not this:
```kotlin
    @Test
    fun stackoverflow() {
        val scheduleDirect = Schedulers.single().scheduleDirect({
            while (true) {
                break
            }
        }, 0, TimeUnit.MILLISECONDS)

        Thread.sleep(1)

        println(scheduleDirect)
    }
```

StackTrace:
```
java.lang.StackOverflowError
	at java.base/java.util.concurrent.FutureTask.toString(FutureTask.java:497)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.util.concurrent.atomic.AtomicReference.toString(AtomicReference.java:275)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:168)
	at java.base/java.util.concurrent.FutureTask.toString(FutureTask.java:513)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.util.concurrent.atomic.AtomicReference.toString(AtomicReference.java:275)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:168)
	at java.base/java.util.concurrent.FutureTask.toString(FutureTask.java:513)
	at java.base/java.lang.String.valueOf(String.java:2951)
	at java.base/java.util.concurrent.atomic.AtomicReference.toString(AtomicReference.java:275)
```

Analysis:
kind of looks like the toString() has a self reference, which triggers a `StackOverflowError`

Note: when I set some breakpoints and debug slowly the issue will not happen.

akarnokd:
Starting from JDK 10, the `FutureTask.toString` can now print the inner callable routine, which if somehow references the parent `FutureTask` again, it leads to `StackOverflowError`. This can happen in RxJava because `AbstractDirectTask` and `ScheduledRunnable` store the `Future` object returned by the executor in a reference field that gets accessed by their default `toString` implementation.

The fix is to define custom `toString`s that break this cycle. Note that we can't print the underlying `Runnable` either because that could also reference something in a circular manner. In addition, the release of the `runner` `Thread` marker has been moved closer to the exit of the wrappers which helps with the state determination and self-cancellation.

To verify the fix works, a new GitHub Action has been added, targeting JDK 11 as the issue does not manifest itself under the main target JDK 8.

Resolves #7172

### DIFF
--- a/.github/workflows/gradle_jdk11.yml
+++ b/.github/workflows/gradle_jdk11.yml
@@ -1,0 +1,33 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: JDK 11
+
+on:
+  push:
+    branches: [ 3.x ]
+  pull_request:
+    branches: [ 3.x ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Cache Gradle packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.gradle/caches
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+        restore-keys: ${{ runner.os }}-gradle
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - name: Build PR
+      run: ./gradlew -PreleaseMode=pr build --stacktrace
+    #- name: Upload to Codecov  
+    #  uses: codecov/codecov-action@v1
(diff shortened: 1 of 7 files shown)


akarnokd:
[Works across the board too.](https://github.com/akarnokd/RxJava3_BuildMatrix/actions/runs/524114162)

vanniktech (approved):

PR #7175, opened 2021-02-01
3.x: Separate GitHub Action for JavaDocs pushback for releases

akarnokd:
As of late, releases tend to get stuck between BinTray and Sonatype with Validation Failure. This freezes out the bintray plugin and the release process stops progressing to the codecov & JavaDocs pushbacks.

This PR adds a separate GitHub Action (GHA) to generate and push back the JavaDocs in parallel with the BinTray release process so when it hangs, there is no extra work needed to manually generate and upload the Javadocs to the gh-pages branch.

Unfortunately, since the release does reach BinTray but not Sonatype, simply restarting the Release GHA would then fail with duplicate release on BinTray itself. There are currently [no options](https://github.com/bintray/gradle-bintray-plugin/issues/317) to timeout or ignore errors with the bintray plugin.

Also unfortunate that there is no good way to test this action outside of a full or pre-release triggered from GH so the next release for RxJava has to do release-candidates (RC1,...) first.

### DIFF
--- a/.github/workflows/gradle_release_docs.yml
+++ b/.github/workflows/gradle_release_docs.yml
@@ -38,7 +38,7 @@ jobs:
       run: chmod +x push_javadoc.sh
     - name: Extract version tag
       run: echo "BUILD_TAG=${GITHUB_REF:10}" >> $GITHUB_ENV
-    - name: Build and Release
+    - name: Assemble and Generate JavaDocs
       run: ./gradlew assemble javadoc --stacktrace
     - name: Push Javadocs
       run: ./push_javadoc.sh
--- a/.github/workflows/gradle_release_docs.yml
+++ b/.github/workflows/gradle_release_docs.yml
@@ -1,7 +1,7 @@
 # This workflow will build a Java project with Gradle
 # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
 
-name: Release
+name: Release JavaDocs
 
 on:
   release:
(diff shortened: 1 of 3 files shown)


.github/workflows/gradle_release_docs.yml
```diff
@@ -0,0 +1,44 @@
...
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Cache Gradle packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.gradle/caches
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+        restore-keys: ${{ runner.os }}-gradle
+    - name: Grant execute permission for gradlew
```
vanniktech, line 35: Can't we do this step locally, make it executable & push the changes? Or is there any other reason this is needed

.github/workflows/gradle_release_docs.yml
```diff
@@ -0,0 +1,44 @@
...
+      run: chmod +x gradlew
+    - name: Grant execute permission for push
+      run: chmod +x push_javadoc.sh
```
vanniktech, line 38: Same

vanniktech (approved):

.github/workflows/gradle_release_docs.yml
akarnokd: Windows checkouts and pushes tend to break this so it is safer to have the script grant execute rights as it goes.

akarnokd:
Bintray goes away so this won't be necessary.

PR #7181, opened 2021-02-04
3.x: Release to Sonatype directly

### LINKED ISSUE #7179: Bintray is closing down in May 2021
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

I don't fully understand what this means for release processes using `gradle-bintray-plugin` or how to migrate.

akarnokd:
Update the release process to use Sonatype directly via the help of the `com.vanniktech:gradle-maven-publish-plugin`.

In order for it to work, two secrets has to be set in the [settings](https://github.com/ReactiveX/RxJava/settings/secrets/actions):

 - `SIGNING_PRIVATE_KEY`
- `SIGNING_PASSWORD`

The `SIGNING_PRIVATE_KEY` should be the ascii-armored private key to sign the JAR files and can be generated from a pgp keychain:

```
gpg --export-secret-key --armor ___key_name_here___ > private_key.asc
```

The `SIGNING_PASSWORD` is then the password for that `private_key.asc`.

I tested the process on my local machine as much as I could, but we'll only know if it really works when doing an RxJava release.

Resolves #7179

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,15 +13,14 @@ buildscript {
   ext.jacocoVersion = "0.8.4"
   ext.animalSnifferVersion = "1.5.2"
   ext.licenseVersion = "0.15.0"
-  ext.bintrayVersion = "1.8.5"
   ext.jfrogExtractorVersion = "4.20.0"
   ext.bndVersion = "5.2.0"
   ext.checkstyleVersion = "8.26"
+  ext.vanniktechPublishPlugin = "0.13.0"
 
   // --------------------------------------
 
   repositories { 
-    jcenter()
     mavenCentral()
     maven {
         url "https://plugins.gradle.org/m2/"
@@ -31,9 +30,9 @@ buildscript {
     classpath "ru.vyarus:gradle-animalsniffer-plugin:$animalSnifferVersion"
     classpath "gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:$licenseVersion"
     classpath "me.champeau.gradle:jmh-gradle-plugin:$jmhGradleVersion"
-    classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:$bintrayVersion"
     classpath "org.jfrog.buildinfo:build-info-extractor-gradle:$jfrogExtractorVersion"
     classpath "biz.aQute.bnd:biz.aQute.bnd.gradle:$bndVersion"
+    classpath "com.vanniktech:gradle-maven-publish-plugin:$vanniktechPublishPlugin"
   }
 }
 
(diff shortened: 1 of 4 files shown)


ohbus (approved):

build.gradle
```diff
@@ -48,11 +47,14 @@ if (releaseTag != null && !releaseTag.isEmpty()) {
...
     }
     version = releaseTag;
-    project.properties.put("release.version", releaseTag);
+    project.setProperty("release.version", releaseTag);
+    project.setProperty("VERSION_NAME" , releaseTag);
 
     println("Releasing with version " + version);
 }
 
+println(project.properties["VERSION_NAME"])
```
vanniktech: Do you still want this?

gradle.properties
```diff
@@ -1,2 +1,25 @@
 release.scope=patch
 release.version=3.0.0-SNAPSHOT
+VERSION_NAME=3.0.0-SNAPSHOT
```
vanniktech, line 2: Can we unify this with the variable above?

akarnokd: Done.

build.gradle
```diff
@@ -154,60 +138,16 @@ license {
...
-publishing.publications.all {
-  pom.withXml {
-    asNode().dependencies."*".findAll() {
-      it.scope.text() == "runtime" && project.configurations.compile.allDependencies.find { dep ->
-        dep.name == it.artifactId.text()
-      }
-    }.each { it.scope*.value = "compile"}
-  }
-}
+// publishing.publications.all {
```
akarnokd: @vanniktech  Do we need these with your plugin?

vanniktech (approved):

build.gradle
vanniktech: Yes. Runtime dependencies aren't changed to compile.

akarnokd: I commented this out because it wouldn't work in its current location. Where should this be moved?

vanniktech: I haven't used that syntax but you should be able to move it after applying the plugin? It might need an afterEvaluate though

mostroverkhov:
FYI default plugins shipped with gradle can do that without external dependencies - `maven-publish` and `signing` are enough.

Just apply corresponding plugins on [modules](https://github.com/jauntsdn/netty-websocket-http2/blob/c74589de351f43d9e20bf0db83792d4117a0542f/netty-websocket-http2/build.gradle#L18) that need to be released - their [publishing](https://github.com/jauntsdn/netty-websocket-http2/blob/c74589de351f43d9e20bf0db83792d4117a0542f/gradle/publishing.gradle#L60) and [signing](https://github.com/jauntsdn/netty-websocket-http2/blob/c74589de351f43d9e20bf0db83792d4117a0542f/gradle/publishing.gradle#L89) configurations are rather tiny

External config is 
```
signing.keyId
signing.password
signing.secretKeyRingFile

ossrhUsername
ossrhPassword
```

May be worth considering to reduce future maintenance burden.

build.gradle
akarnokd: Done.

PR #7196, opened 2021-02-24
3.x: Allow Single.zip and Maybe.zip result to be garbage collected

### LINKED ISSUE #7195: GC friendly `Single.zip` operator
Version 3.0.10

Hey there, 

My use case requires me to load huge datasets into memory. I need multiple of those datasets to cross compare them in their entirety. I decided to use the `Single.zip` operator to combine them.

Example:

```java
public static void main(String[] args)
{
  Single.zip(
      Single
          .defer(() -> Single.just(buildHumongousCollection()))
          .subscribeOn(Schedulers.computation()),

      Single
          .defer(() -> Single.just(buildHumongousCollection()))
          .subscribeOn(Schedulers.computation()),

      (z1, z2) -> new Collection[]{ z1, z2 }
  )
      .map(zip ->
      {
        // do anything with any operators, for example building new objects from the source set
        return new String[]{ "Foo", "Bar", "Baz" };
      })
      .blockingSubscribe(result ->
      {
        // just here to allow setting a breakpoint
        for ( final var msg : result ) { System.out.println(msg); }
      });
}

private static Collection<byte[]> buildHumongousCollection()
{
  final var rand = new Random();

  final var result = new LinkedList<byte[]>();
  for ( int i = 0; i < 50_000_000; i++ )
  {
    final var buff = new byte[32];
    rand.nextBytes(buff);
    result.add(buff);
  }

  return result;
}
```

I noticed that when I reach `onNext` of my Observer within `subscribe` in my execution chain, The memory is still allocated and will never be collected. So I started to investigate and found the cause of it.

The `ZipCoordinator` within `SingleZipArray` does not drop the local value collection after completion:

https://github.com/ReactiveX/RxJava/blob/09b2b1b4c9428616b2634a67bea0ee5a09201007/src/main/java/io/reactivex/rxjava3/internal/operators/single/SingleZipArray.java#L106-L121

Since the ZipCoordinator will always reside within the stack of one of the living threads, the garbage collector will never collect the it and therefore also the dataset even though it can never be used again. The only way to access it is through private fields.

I'd suggest to insert
```java
values = null;
```
at 118 - it should allow gc to free the memory.

zireael-0:
Resolves #7195

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/maybe/MaybeZipArray.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/maybe/MaybeZipArray.java
@@ -117,12 +117,12 @@ public final class MaybeZipArray<T, R> extends Maybe<R> {
                     v = Objects.requireNonNull(zipper.apply(values), "The zipper returned a null value");
                 } catch (Throwable ex) {
                     Exceptions.throwIfFatal(ex);
-                    values = null;
+                    this.values = null;
                     downstream.onError(ex);
                     return;
                 }
 
-                values = null;
+                this.values = null;
                 downstream.onSuccess(v);
             }
         }
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/maybe/MaybeZipArray.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/maybe/MaybeZipArray.java
@@ -74,7 +74,7 @@ public final class MaybeZipArray<T, R> extends Maybe<R> {
 
         final ZipMaybeObserver<T>[] observers;
 
-        final Object[] values;
+        Object[] values;
 
         @SuppressWarnings("unchecked")
         ZipCoordinator(MaybeObserver<? super R> observer, int n, Function<? super Object[], ? extends R> zipper) {
@@ -101,7 +101,7 @@ public final class MaybeZipArray<T, R> extends Maybe<R> {
                     d.dispose();
                 }
 
-                Arrays.fill(values, null);
+                values = null;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/reactivex/rxjava3/internal/operators/maybe/MaybeZipArray.java
```diff
@@ -100,6 +101,8 @@ public void dispose() {
                     d.dispose();
                 }
             }
+
+            Arrays.fill(values, null);
```
akarnokd: This should go inside the condition.

src/main/java/io/reactivex/rxjava3/internal/operators/maybe/MaybeZipArray.java
```diff
@@ -115,6 +118,7 @@ void innerSuccess(T value, int index) {
                     return;
                 }
 
+                Arrays.fill(values, null);
```
akarnokd: This should also be in the catch block above, between the `throwIfFatal` and `onError` calls. Similarly, `innerError` should be updated.

akarnokd (changes requested):
Please adjust the PR for both classes.

akarnokd (approved):

zireael-0:
@akarnokd Sorry, hit re-review accidentally :see_no_evil:

akarnokd (changes requested):
Looks like I forgot about the case when the consumer retains the whole array reference, thus clearing it out can cause late errors.

Let's make `values` non-final and null it out instead as you originally suggested.

akarnokd (approved):
I committed a few changes to fix race conditions with the changes

PR #7209, opened 2021-03-11
Optimize code

Goooler:
- Use `java5`, `java7`, `java8` syntax (include lambdas).
- Add `final` & `static`
- Refine control flow
- Inline redudant local variables
- Add `Nullable` , `NonNull` annotations
- Refine comment & regex
- Optimize imports

And more.


akarnokd:
Please don't do such PRs.

- For one, this is not reviewable unlike the header change due to the very large number of changes.
- We didn't want to change existing function use to lambdas en masse, only occasionally or when adding new tests.
- Some changes, such as the inlining of variables, is wrong. Those were introduced to avoid field reads.

PR #7213, opened 2021-03-13
Fix intermittently failing unit tests

benjamintboyle:
Fix several unit tests that are intermittently failing on slower clients due to timeouts. All fixes
involve increasing timeouts.

All other changes are adjustments to formatting within the updated tests.


akarnokd (changes requested):
If you need more time in tests, that's fine. However, don't refactor unrelated parts of the tests.

Also no need to use `Duration` so that you decompose it to milliseconds.

akarnokd (approved):

PR #7218, opened 2021-03-25
3.x: Project Loom support in standard schedulers via hooks

akarnokd:
This PR adds a hook point for when the underlying `ScheduledExecutorService`s are created for the standard `Scheduler` implementations. This could support the upcoming Project Loom features without RxJava 3 to go beyond Java 8.

In Project Loom, to be able to cheaply suspend in code, that code has to run in a Virtual Thread. Such virtual threads can have a backing carrier thread, i.e., via a regular threaded `ExecutorService`. In order for RxJava to allow any suspending function in any scheduler, one has to override the underlying default `ScheduledExecutorService` with one that has the virtual thread intermediate layer:

```java
// Requires JDK 17 Loom build
var factory = Thread.ofVirtual().scheduler(executor).factory();
var newExecutor = Executors.newSingleThreadScheduledExecutor(1, factory);
```

However, shutting down this `newExecutor` would not have any effect on the original executor thus one needs to intercept the `shutdown` methods and forward those calls to the original `executor` too:

```java
return new ScheduledExecutorService() {
   // ...
   @Override
   public List<Runnable> shutdownNow() {
        var list = newExecutor.shutdownNow();
        executor.shutdownNow();
        return list;
   }
   @Override
   public void shutdown() {
        newExecutor.shutdown();
        executor.shutdown();
   }
};
```

Some caveats:
- The hook has to be set early before any Scheduler get initialized.
- I'm not 100% sure the above use of a virtual factory actually works. In the [Loom API](https://download.java.net/java/early_access/loom/docs/api/java.base/java/util/concurrent/Executors.html), the new virtual and unowned factory methods return `ExecutorService` which doesn't support timed scheduling.
- This can't be tested in RxJava without significant hacks to the CI (i.e., hack in the Loom runtime, enable jdk-dependent inclusion of unit test classes). The closest place to test this would be in https://github.com/akarnokd/RxJavaFiberInterop .

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/SchedulerPoolFactory.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/SchedulerPoolFactory.java
@@ -19,6 +19,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import io.reactivex.rxjava3.exceptions.Exceptions;
 import io.reactivex.rxjava3.functions.Function;
+import io.reactivex.rxjava3.plugins.RxJavaPlugins;
 
 /**
  * Manages the creating of ScheduledExecutorServices and sets up purging.
@@ -142,7 +143,8 @@ public final class SchedulerPoolFactory {
      * @return the ScheduledExecutorService
      */
     public static ScheduledExecutorService create(ThreadFactory factory) {
-        final ScheduledExecutorService exec = Executors.newScheduledThreadPool(1, factory);
+        ScheduledExecutorService exec = Executors.newScheduledThreadPool(1, factory);
+        exec = RxJavaPlugins.onCreateExecutor(exec);
         tryPutIntoPool(PURGE_ENABLED, exec);
         return exec;
     }
--- a/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
@@ -15,7 +15,7 @@ package io.reactivex.rxjava3.plugins;
 
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.Objects;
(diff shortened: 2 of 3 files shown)


akarnokd:
Logically, this might not be as straightforward as I first thought.

#### Schedulers.single()

There is only one FIFO execution queue, so if a task suspends, we still can't allow other tasks (direct or worker) to run in the meantime. Those tasks might just block that single carrier thread anyway.

#### Schedulers.io()

First, worker reuse is dependent upon releasing them, not when they block. Second, a new worker would create a new carrier thread still. With resumption, the flow essentially becomes CPU oriented, thus having more carrier threads than cores is wasteful. Third, if a carrier thread gets blocked due to non-suspendable API use, the scheduler should create a new carrier to avoid clogging up the other carriers. 

The Java `ForkJoinPool` may be smart enough to detect blocked threads and is otherwise bound in numer of threads, so we could use it as carrier for otherwise freshly created virtual executors per worker. Also worker reuse would have no benefit. A drawback is that thread confinement would be lost as any worker might run/resume on any carrier thread. Practically we'd need some single-threaded carrier management on our own.

#### Schedulers.newThread()

Similar problems as with `io`. If we allow it to be unbounded, the work might just block, as it would do today. If we bound this type of Scheduler too, there would be no practical difference to `io`.

#### Schedulers.computation()

This is similar to `io` but with two options: we ignore blocking and always have up to N carrier threads or we do the same as in `io`: create a new carrier on demand. The drawback is that we might end up with more carriers than cores and no timely way of reducing the carrier count back to the core count.

#### Conclusion

This hook might not be good/enough to support virtualizing these schedulers and we'd need a dedicated `Scheduler` implementation instead (in a loom-enabled repo).

akarnokd:
After further consideration, I think a simple hook would not work. Custom Scheduler(s) have to be implemented which requires a Loom runtime, thus can't be done RxJava.

PR #7219, opened 2021-03-26
Modernize gradle plugin block, change maven to maven-publish

benjamintboyle:
Move gradle plugins to block at top of file per gradle recommendations.
Change deprecated maven plugin to maven-publish as in Gradle 7.0 it will
be removed. Several other minor updates to gradle file including, removing
semi-colons, changing println to use built-in logger.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ ext {
     testNgVersion = "7.3.0"
     mockitoVersion = "3.8.0"
     jmhLibVersion = "1.21"
-    guavaVersion = "30.1-jre"
+    guavaVersion = "30.1.1-jre"
     jacocoVersion = "0.8.4"
     checkstyleVersion = "8.41"
 }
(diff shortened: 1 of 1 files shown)


akarnokd (dismissed):

build.gradle
```diff
@@ -1,72 +1,47 @@
...
-  ext {
+ext {
     reactiveStreamsVersion = "1.0.3"
     junitVersion = "4.13.2"
     testNgVersion = "7.3.0"
     mockitoVersion = "3.8.0"
     jmhLibVersion = "1.21"
-    jmhGradleVersion = "0.5.3"
-    guavaVersion = "30.1.1-jre"
+    guavaVersion = "30.1-jre"
```
akarnokd: Why going back to 30.1-jre here?

benjamintboyle: I was working on an older branch. I will correct it

akarnokd (approved):

PR #7220, opened 2021-03-26
Remove gradle garbage collection tasks

benjamintboyle:
Removed 2 gradle garbage collection tasks. After move to github actions,
these tasks should no long be needed. Formatted other lines to match rest of file.


benjamintboyle:
Needs some more work

PR #7235, opened 2021-04-14
3.x: Migrate Gradle to use Java Toolchain

benjamintboyle:
Use Gradle java toolchain functionality, instead of relying on JAVA_HOME
version used to run the Gradle instance. Remove target and compatibility
Java versions properties. Make the Java version configurable in gradle.properties
with default version set to 8.

Fix bug where GitHub JDK11 Action was using Java 8 due to hardcoded
target/compatibility properties of 1.8.


akarnokd:
The JDK 11 job is meant to be built as target 8. There was a bug due to JDK internals change we need to test against.

PR #7241, opened 2021-04-19
3x: Add full module descriptor

### LINKED ISSUE #7240: 3.x Consider adding a full Java module descriptor

aalmiray:
As described at https://github.com/ReactiveX/RxJava/issues/7240, add a full Java module descriptor to version 3.x.

This setup lets the 3.x series remain Java 8 compatible while the `module-info.class` is compiled with Java 9+ and stored inside the jar at `/META-INF/versions/9`.

Resolves #7240.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ plugins {
     id("com.github.hierynomus.license") version "0.16.1"
     id("biz.aQute.bnd.builder") version "5.3.0"
     id("com.vanniktech.maven.publish") version "0.17.0"
+    id("org.beryx.jar") version "1.2.0"
 }
 
 ext {
@@ -29,6 +30,10 @@ if (releaseTag != null && !releaseTag.isEmpty()) {
 
     logger.lifecycle("Releasing with version: " + project.version)
 }
+
+group = "io.reactivex.rxjava3"
+version = project.properties["VERSION_NAME"]
+description = "RxJava: Reactive Extensions for the JVM – a library for composing asynchronous and event-based programs using observable sequences for the Java VM."
 
 repositories {
     mavenCentral()
@@ -52,6 +57,7 @@ java {
     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
 }
+ext.javaCompatibility = JavaVersion.VERSION_1_8
 
 tasks.withType(JavaCompile) {
     options.compilerArgs << "-parameters"
@@ -86,6 +92,8 @@ animalsniffer {
 }
 
 jar {
+    // Cover for bnd still not supporting MR Jars: https://github.com/bndtools/bnd/issues/2227
+    bnd('-fixupmessages': '^Classes found in the wrong directory: \\\\{META-INF/versions/9/module-info\\\\.class=module-info}$')
     bnd(
             "Bundle-Name": "rxjava",
             "Bundle-Vendor": "RxJava Contributors",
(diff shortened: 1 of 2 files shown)


aalmiray:
Converting to draft as there are 2 standing issues:
- `org.reactivestreams` is not a full module, yet. See https://github.com/reactive-streams/reactive-streams-jvm/issues/531
- bnd Gradle plugin 5.3.0 can't handle multi-release JARs correctly https://github.com/bndtools/bnd/issues/2227

See build error: `error  : Classes found in the wrong directory: {META-INF/versions/9/module-info.class=module-info}`

A248:
@aalmiray , regarding the two issues you mentioned:

* It isn't a requirement that reactive-streams become a full module in order for RxJava to do so. There's nothing stopping a full module from depending on an automatic one; rather, that's the purpose of automatic modules.

* The last time I ran into this warning from the Bnd gradle plugin, it was an annoying message in the build log but didn't fail the build. Regardless, you should be able to fix it by adapting this example for gradle kotlin: https://github.com/bndtools/bnd/issues/2227#issuecomment-491310275. 

Also, I'm not a veteran RxJava user, but I believe the `flowables` package should be exported, and the requirement on org.reactivestreams should be `transitive`.

aalmiray:
@A248 as mentioned in #7240 the motivation for adding a full module descriptor is to facilitate the use of RxJava with custom Java Runtimes created with jlink. Custom modular JRs only support full modules, not automatic modules.

ZacSweers:
This seems like an unnecessarily roundabout way to do this. Here's a simple example that you could just adapt instead: https://github.com/melix/jdoctor/blob/main/jdoctor-build-logic/src/main/kotlin/me.champeau.java-module.gradle.kts

aalmiray:
That would require the use of toolchains and ensuring a Java 9+ JDK to be available at all times. Toolchains are not stable enough, just have a look at the Gradle issue tracker.

ZacSweers:
Can you offer a stronger citation? We are using it in several projects with no issues.

aalmiray:
Browse Gradle's issue tracker, they have a label matching toolchains. Look at the number of issues and their assigned milestone. The engineer in charge of toolchains left the company in early June 2021.

ZacSweers:
Unless you can point to a specific issue that prohibits its use here, I don't think "go look at the issue tracker" is very constructive. No tool is perfect. We use this in about a dozen projects with varying degrees of complexity and it's been fine. Happy to toss up a PR if that would quell your concerns.

aalmiray:
As you wish. I'm more interested in the outcome (RxJava as a Java module) than how that goal it's achieved.

A maintainable option for the project must be found, after all it's "easy" for me to suggest ModiTect as I know it but the RxJava team would have to live with it if the PR is accepted as is.

If ModiTect it's not the way then so be it, again the "what" is more important to me than the means to achieve it.

One more thing, regardless of how the build is made modular you'd still have to find a solution for reactive-streams becoming modular which is currently blocked by an OSGi compatibility, due to bnd and its companion Gradle plugin.

ZacSweers:
Sure, opened a PR here #7332. 

> One more thing, regardless of how the build is made modular you'd still have to find a solution for reactive-streams becoming modular which is currently blocked by an OSGi compatibility, due to bnd and its companion Gradle plugin.

I think I've worked around this just suppressing the warning message from bnd (per the original issue you linked), but I'm not sure if/what else would be needed. Considering the issue has been open for 4 years with no movement, it seems unlikely to change.

akarnokd:
If this only needs a module-info.class, I took #7332 and injected the file as resource: #7333. I don't expect the exported packages to change much in the future. Is this enough for the module system compatibility?

aalmiray:
Following this last option would require build instructions to re-generate `module-info.class` when updates to the descriptor are warranted, allowing anyone to test their changes otherwise you risk a bottle neck when only a few can make the change and push it before every one else may continue.

akarnokd:
> when updates to the descriptor are warranted

It is rare for the main export packages to change. We just had a new package `operators` added after 1.5 years. I don't expect any new package in the forseable future.

Another middle ground I see is to generate the module file via ASM so there is no need for 9+ compiler and those mockup classes to satisfy it.

aalmiray:
FWIW generating `module-info.class` via ASM is exactly what the ModiTect plugin does. There's more than one way to configure generation of that file, this PR presented but one option.

akarnokd:
I see. In that case, the BND failure can be suppressed. Unless the first point in https://github.com/ReactiveX/RxJava/pull/7241/#issuecomment-869087722 is still a concern, this would be good to go.

akarnokd (approved):

akarnokd:
I have updated the GHA for the JDK 11 build to verify the module-info. Could you rebase this onto the current 3.x branch?

aalmiray:
@akarnokd PR has been rebased and updated with hints for bnd to skip the error due to `module-info.class`, following what @ZacSweers did for #7332

akarnokd (approved):

PR #7258, opened 2021-05-05
3.x: Remove Javadoc not failing on error

benjamintboyle:
The 'javadoc' task should fail when there is an error. The
'javadocCleanup' task failed for PR [7239](https://github.com/ReactiveX/RxJava/pull/7239/checks) due to publish plugin issue [242](https://github.com/vanniktech/gradle-maven-publish-plugin/issues/242),
while it should have been the 'javadoc' task that failed the build.
Also, publishing when javadocs fail is not desired.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,6 @@ tasks.withType(JavaCompile) {
 apply from: file("gradle/javadoc_cleanup.gradle")
 
 javadoc {
-    failOnError = false
     exclude "**/internal/**"
     exclude "**/test/**"
     exclude "**/perf/**"


akarnokd:
If I understand correctly, #7239 would override the JDK that compiles and tests RxJava to be 8 always. This is not desired because the purpose of the JDK 11 build is to verify the fix of #7173, which needs an actual JDK 11 runtime. We don't care about the Javadoc generated for the JDK 11 build, only the build of JDK 8.

benjamintboyle:
Thanks for giving that history. I'll have to think through it now that I understand it.

I think that this pull request is valuable and independent of that though. Correct?

benjamintboyle:
Were you implying that the Javadoc task for JDK 11 Workflow the was added in the previous pull request should be removed? That would make sense

akarnokd:
> I think that this pull request is valuable

I think it was disabled because pre-Java 8 javadoc generation sometimes failed even though it managed to do its job. Let's have this PR.

> Were you implying that the Javadoc task for JDK 11 Workflow the was added in the previous pull request should be removed?

Let it be for now.

PR #7260, opened 2021-05-06
3.x: Move Gradle properties into gradle.properties file

benjamintboyle:
Moved version, group, and description into gradle.properties file.
Removed POM_ARTIFACT property as the 'name' property in the settings file
is used when building the POM. Removed 'release.scope' property as it
does not appear to be used. Removed VERSION_NAME property and references
to it, in favor of 'version' property.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,14 +27,10 @@ if (releaseTag != null && !releaseTag.isEmpty()) {
     if (releaseTag.startsWith("v")) {
         releaseTag = releaseTag.substring(1)
     }
-    project.setProperty("VERSION_NAME" , releaseTag)
+    project.version = releaseTag
 
-    logger.info("Releasing with version: {}", version)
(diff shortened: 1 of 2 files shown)


benjamintboyle:
Also verified that the POM created does not change. Generated POM before changes and diff'd it with POM after changes. The POMs are identical.

akarnokd:
Removing some of the properties might break the release plugin: https://github.com/vanniktech/gradle-maven-publish-plugin#setting-properties

benjamintboyle:
Replaced property for release plugin.

benjamintboyle:
VERSION_NAME and GROUP default to 'version' and 'group' as seen on Line 9 and 10 from [Plugin source](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/22ad42bb35e9cadd5775659b1045b70db6bc806f/src/main/kotlin/com/vanniktech/maven/publish/legacy/Coordinates.kt)

PR #7261, opened 2021-05-07
3.x: Remove toolVersion for jacoco and checkstyle

benjamintboyle:
Remove versioning of Jacoco and Checkstyle. Instead they will be updated
by Gradle when it is updated. Dependabot does not update tool versions.
If issues come up, they can be corrected when Gradle is updated.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,8 +18,6 @@ ext {
     mockitoVersion = "3.9.0"
     jmhLibVersion = "1.21"
     guavaVersion = "30.1.1-jre"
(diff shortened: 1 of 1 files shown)


benjamintboyle:
Note for Gradle 7.0, the current version is:
Jacoco: 0.8.6
Checkstyle: [8.37](https://github.com/gradle/gradle/blob/c4d09af3db073776225e705307d93284df8753c0/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/CheckstylePlugin.java)

PR #7263, opened 2021-05-07
3.x: Consolidate test logging

benjamintboyle:
Move test logging to consolidated block. Change all 'rootProject' to
'project'.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -144,7 +144,9 @@ tasks.withType(Test) {
         warn.exceptionFormat = "full"
     }
 
-    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+    if (System.getenv("CI") == null) {
+        maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+    }
 }
 
 jacocoTestReport {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -120,51 +120,36 @@ jmh {
...
+
         warn.events = ["failed", "skipped"]
-        warn.exceptionFormat="full"
-     }
-} 
+        warn.exceptionFormat = "full"
+    }
 
-check.dependsOn testng 
+    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
```
akarnokd: This was supposed to run parallel only in non-CI environments.

benjamintboyle: I understand that was the intent, but it seems that it has not for quite some time. I wasn't able to find "CI" set anywhere. Should testing on Github be single threaded?

akarnokd: https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
<html>
<body>


Environment variable | Description
-- | --
CI | Always set to true.


</body>
</html>

benjamintboyle: Ah, thanks. I wasn't aware of that. I'll submit an update.

PR #7284, opened 2021-06-19
Add an overloaded version assertValues that takes predicates

wcat:
An overloaded version of assertValues, that takes predicates, has been added to BaseTestConsumer. This is consistent with the other asserts on values (like assertValue and assertValueAt) which are already allowing assertions with predicates.

The same behaviour can already be obtained by using multiple assertValueAt:
```
streamUnderTest.test()  
   .assertValueAt(0, somePredicate)
   .assertValueAt(1, someOtherPredicate);
```
but being able to use assertValues for the same checks makes (imho) the assertion clearer and less cluttered:
```
streamUnderTest.test()
   .assertValues(somePredicate, someOtherPredicate);
```

NOTE: I'm not very fond of the javadoc, I think it can be improved

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/rxjava3/observers/BaseTestConsumer.java
@@ -464,8 +464,8 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


akarnokd:
No. We intend to keep the test API minimal and this suggestion tries to add convenience that isn't widely needed. This type of value assertion is better suited as a static method in one's own test environment.

PR #7291, opened 2021-07-07
Fix Bug

ycxGHub:
ObservableCreate should return immediately when dispose the flow.


akarnokd:
This is not a bug. You can do this check inside the `subscriber` yourself:

```java
Observable.create(emitter -> {
   if (emitter.isDisposed()) return;
});
```

PR #7293, opened 2021-07-15
Removes the purge thread in favour of standard ScheduledThreadPoolExecutor APIs

chggr:
The RXSchedulerPurge thread is currently enabled by default and runs every second
to call purge() on each executor in the pool. This is causing significant issues
for battery powered devices (e.g. mobile phones), because it needs to periodically
wake up the CPU to perform purging. The RXSchedulerPurge thread could be completely
removed in favour of using the standard setRemoveOnCancelPolicy() API on the
ScheduledThreadPoolExecutor which became available in Java 7 and offers removal
of cancelled tasks at the moment they are cancelled in O(1).

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/schedulers/SchedulerPoolFactory.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/schedulers/SchedulerPoolFactory.java
@@ -31,86 +31,11 @@ public final class SchedulerPoolFactory {
 
     static final String PURGE_ENABLED_KEY = "rx3.purge-enabled";
 
-    /**
-     * Indicates the periodic purging of the ScheduledExecutorService is enabled.
-     */
     public static final boolean PURGE_ENABLED;
-
-    static final String PURGE_PERIOD_SECONDS_KEY = "rx3.purge-period-seconds";
-
-    /**
-     * Indicates the purge period of the ScheduledExecutorServices created by create().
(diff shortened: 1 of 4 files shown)


akarnokd:
The question is, how does this affect compatibility and desugaring?

/cc @JakeWharton

JakeWharton:
I do not recall the minimum Android version we decided on for 3.x. It doesn't seem explicitly documented anywhere.

The API in question (`setRemoveOnCancelPolicy`) is available from Android API 21 and newer (which corresponds to Android 5.0 as a user-facing version). API 21 is a relatively safe minimum, but I suspect prior to this you were able to use 3.x on even lower API levels.

It's not unheard of for libraries to bump their minimum-supported Android API level in minor releases. So I'm supportive for the next minor version provided that the minimum is also documented somewhere.

akarnokd:
Okay then, from 3.0.14, we require API 21. I'll add this to the readme.

JakeWharton:
Sorry by minor I meant for a 3.1.0 rather than patch release in the 3.0.x series. At least then someone might look a bit more closely at the changes than a patch release where they expect only fixes and not new functionality / changes.

Ultimately you can decide though since it's not explicitly documented anywhere and this is really about making it explicit.

akarnokd:
Indeed 3.1.0 would be more appropriate.

PR #7300, opened 2021-07-22
Clearly document side effects of calling Schedulers.from()

chggr:
Calls to Schedulers.from() APIs have the side effect of instantiating all the
default RxJava Schedulers if they have not already been instantiated. This is
currently not documented and should be documented to make sure that users of
these APIs do not call them in latency sensitive execution paths.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
+++ b/src/main/java/io/reactivex/rxjava3/schedulers/Schedulers.java
@@ -376,6 +376,8 @@ public final class Schedulers {
      * execute those tasks "unexpectedly".
      * <p>
      * Note that this method returns a new {@code Scheduler} instance, even for the same {@code Executor} instance.
+     * Calling this method will have the side effect of instantiating all the default RxJava Schedulers,
+     * if they have not already been instantiated.
      * @param executor
      *          the executor to wrap
      * @return the new {@code Scheduler} wrapping the {@code Executor}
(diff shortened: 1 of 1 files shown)


akarnokd:
I don't think that's something needing documentation. Anything triggering class initialization of `Schedulers` would have the same effect. That shouldn't be a problem in general and if you encountered it in your specific case, you can work around it by pre-initializing the class.

chggr:
Users of the Schedulers.from() API might expect that this is just translating an Executor to a Scheduler, rather than bringing up all RXJava thread pools. All other Schedulers APIs are for obtaining a Scheduler (io(), computation(), etc) so the extra cost of brining up the RXJava thread pools is justified for them. I don't think the extra cost is justified for Schedulers.from() though, so as a first step I think it makes sense to document the side effects of that API clearly in Javadoc. I can then raise a separate PR if that is agreeable, to move Schedulers.from() to SchedulerUtils.from() so that people can call the API without the side effects. What do you think?

akarnokd:
No.

I'd rather investigate why you have such a latency problem in your environment.

akarnokd:
To elaborate, documenting it has no practical use because by itself, knowing the init behavior doesn't give much option to avoid said init behavior.

There is a similar issue with customizing the standard schedulers by reusing the internal implementation before `Schedulers` is initialized. Therefore, we offer the `RxJavaPlugins.createComputationScheduler` method.

On a similar vein, I'm open to adding `RxJavaPlugins.createExecutorScheduler(Executor, interruptibleWorker, fair)` overload to get a fresh `ExecutorScheduler` and then reference it from the `Schedulers.from` methods:

> It is possible to wrap an `Executor` into a `Scheduler` without triggering the initialization of all the standard schedulers by using the RxJavaPlugins.createExecutorScheduler(Executor, boolean, boolean) method.

PR #7302, opened 2021-07-23
Fix missing NonNull on Observable.map

ZacSweers:
I noticed this while testing the upcoming Kotlin 1.5.30-M1 (which has improved support for type-use nullability annotations) https://youtrack.jetbrains.com/issue/KT-47833. This makes at least match `Single`, but `Maybe.map` is also missing this. I suspect there are others too.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Observable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Observable.java
@@ -10371,7 +10371,7 @@ public abstract class Observable<@NonNull T> implements ObservableSource<T> {
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     @NonNull
-    public final <R> Observable<R> map(@NonNull Function<? super T, ? extends R> mapper) {
+    public final <R> Observable<@NonNull R> map(@NonNull Function<? super T, ? extends R> mapper) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/rxjava3/core/Observable.java
```diff
@@ -10371,7 +10371,7 @@ public final Completable ignoreElements() {
     @CheckReturnValue
     @SchedulerSupport(SchedulerSupport.NONE)
     @NonNull
-    public final <R> Observable<R> map(@NonNull Function<? super T, ? extends R> mapper) {
+    public final <R> Observable<@NonNull R> map(@NonNull Function<? super T, ? extends R> mapper) {
```
akarnokd: Thanks. I think it should be on the type argument declaration as `<@NonNull R>`. I have to check other cases and get back to you shortly.

ZacSweers: Ah yes you're right

```suggestion
    public final <@NonNull R> Observable<R> map(@NonNull Function<? super T, ? extends R> mapper) {
```

akarnokd:
FYI, I've sweeped across the code base and clarified the nullability annotations on generic type arguments in #7303.

PR #7303, opened 2021-07-23
3.x: Move nullability annotations into the type argument declaration

akarnokd:
This PR makes sure generic type argument declarations on methods include the `@NonNull` (or `@Nullable`) annotation as well as source-checking test code to verify the expected style.

Before:
```java
public final <R> Maybe<R> map(@NonNull Function<? super T, ? extends R> mapper) {
// or
public final <R> Maybe<R> map(@NonNull Function<? super T, @NonNull ? extends R> mapper) {
```

After:
```java
public final <@NonNull R> Maybe<R> map(@NonNull Function<? super T, ? extends R> mapper) {
```

Related #7302


ZacSweers:
This is great!! Will be a big win for nullaway users and kotlin users starting in 1.5.30.

ZacSweers (approved):

vanniktech (approved):

PR #7306, opened 2021-08-02
3.x: Add RxJavaPlugins.createExecutorScheduler

### LINKED ISSUE #7305: 3.x: Add RxJavaPlugins.createExecutorScheduler
As per https://github.com/ReactiveX/RxJava/pull/7300#issuecomment-885037358

akarnokd:
This PR adds the `RxJavaPlugins.createExecutorScheduler` that instantiates the already existing `ExecutorScheduler` independent of the `Schedulers` class. The `from` methods in the `Schedulers` class are now pointing to this new method.

Resolves #7305
Related #7300

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
+++ b/src/main/java/io/reactivex/rxjava3/plugins/RxJavaPlugins.java
@@ -15,7 +15,7 @@ package io.reactivex.rxjava3.plugins;
 
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.Objects;
-import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.*;
 
 import org.reactivestreams.Subscriber;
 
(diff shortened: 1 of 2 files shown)


vanniktech (approved):

matthew-swift:
Hi there. I'm not sure this change is working as expected. The following code creates the standard default schedulers:

```java
ExecutorService executor = Executors.newSingleThreadExecutor();
Scheduler scheduler = RxJavaPlugins.createExecutorScheduler(executor, false, false);
Flowable.intervalRange(1, 10, 1, 1, TimeUnit.SECONDS, scheduler)
        .blockingForEach(System.out::println);
executor.shutdown();
```

But according to the new Javadoc it should not trigger the creation of the default schedulers:

> It is possible to wrap an Executor into a Scheduler without triggering the initialization of all the standard schedulers by using the RxJavaPlugins.createExecutorScheduler(Executor, boolean, boolean) method before the Schedulers class itself is accessed.

It clearly will though because the `ExecutorScheduler` class contains a static reference `HELPER` which references `Schedulers.single()` which then initializes all of the default schedulers as observed by putting a breakpoint in the static initializer block on line 80 of `Schedulers`.

Apologies if I'm misunderstanding the intent of this change. If I am then that may indicate that the new Javadoc is a bit misleading.

akarnokd:
Indeed the method should not trigger that initialization. I'll work out how to break that cycle and post a fix shortly.

matthew-swift:
Thanks for your quick response David. :+1:

PR #7309, opened 2021-08-04
Code cleanup

Goooler:
- Remove redundant cast and type.
- Replace `for i` with `for in`.
- Inline variables.


akarnokd:
Please don't apply IntelliJ suggestions. Also inlining locals causes overhead due to repeated field access.

PR #7332, opened 2021-09-06
3.x: Add module-info.java

### LINKED ISSUE #7240: 3.x Consider adding a full Java module descriptor

ZacSweers:
This uses basic Gradle APIs to configure an MR-Jar with a `module-info.java` file. Resolves #7240, continuation from #7241 but without the need for an extra plugin and works around the `bnd` issue (they lack support for this and seeing as the issue is 4 years old it's not clear that they ever will).

This uses Gradle toolchains for simpler control over compilations, allowing us to target Java 8 for the `main` java sources and use Java 11 (but targeting java 9) for the `module-info.java` file, then adding the compiled `module-info.class` to the final jar under the appropriate `META-INF/versions/9` path.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,12 +49,36 @@ dependencies {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(8)
+    }
+    modularity.inferModulePath.set(true)
 }
 
 tasks.withType(JavaCompile) {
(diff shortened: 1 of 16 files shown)


akarnokd:
It's no good. Eclipse gets confused by the module-info.java
image

IntelliJ also complains
image

akarnokd:
I suggest #7333.

akarnokd:
Thanks for the contribution. After exploring the options, I prefer #7241 because it doesn't confuse IDEs and requires only minimal changes to the build.

ZacSweers:
That's strange, I did this in IntelliJ and it worked fine

PR #7333, opened 2021-09-06
3.x: Inject prebuilt module-info.class for Module support

### LINKED ISSUE #7240: 3.x Consider adding a full Java module descriptor

akarnokd:
If I understand correctly, all we need is a compiled module-info.class in the META-INF/versions/9 directory.

So instead of doing compiler tricks, I took #7332, generated the jar, extracted the module-info.class, then put the class file into the resources directory as is.

Also the bnd plugin needed to suppress the misplaces class file error.

Replaces #7332
Replaces #7241
Resolves #7240


akarnokd:
Closing in favor of #7241

PR #7335, opened 2021-09-06
Module descriptor verify

akarnokd:
Manual rebase of #7241 to include #7334 otherwise the GHA would not build it.


akarnokd:
Looks like it passes: https://github.com/ReactiveX/RxJava/pull/7335/checks?check_run_id=3524928155#step:6:27

PR #7344, opened 2021-09-29
Update assert messages format to be compliant with JUnitTestRunner and GradleTestRunner

keytronic:
Currently assert messages match this regex `expected: (.*)\s*but was: (.*)` which is compliant with the `JUnitTestRunner` but not with the `GradleTestRunner`. (which is now default runner on Android Studio)

I propose to change regex to `\nexpected: (.*)\n\s*got: (.*)`.
This way IDEA will be able to enable diff feature no matter which runner you use.

[JUnitTestRunner available regexes](https://github.com/JetBrains/intellij-community/blob/99614a63425774df58eea3ac92e2b20af1633663/plugins/junit_rt/src/com/intellij/junit4/ExpectedPatterns.java#L27)
[GradleTestRunner available regexes](https://github.com/JetBrains/intellij-community/blob/213.3714/plugins/gradle/java/src/execution/GradleRunnerUtil.java#99)

### DIFF
--- a/src/main/java/io/reactivex/observers/BaseTestConsumer.java
+++ b/src/main/java/io/reactivex/observers/BaseTestConsumer.java
@@ -351,11 +351,11 @@ public abstract class BaseTestConsumer<T, U extends BaseTestConsumer<T, U>> impl
     public final U assertValue(T value) {
         int s = values.size();
         if (s != 1) {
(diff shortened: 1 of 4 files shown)


akarnokd:
Hi, thanks for the proposal but you targeted the wrong branch. We are on `3.x`. Please resubmit your proposal accordingly.

keytronic:
Sorry about that.

PR #7346, opened 2021-10-03
Minor code improvements

apokhyla-inc:
Changes summary:

Replaced `Collections::sort` usage with List::sort since it helps to simply avoid the extra method call (`Collections::sort` simply delegates `List::sort`). Also, replaced strange` for(;;)` construction followed by corresponding `break` keyword with more suitable in this case do-while loop.

Testing done:

Built the project.


akarnokd:
Hi.

Any Java 8 specific method calls are limited to operators interacting with Java 8 features, therefore, List::sort can't be used in Functions.

The `for (;;)` is an established style in RxJava and there is no reason to change it.

apokhyla-inc:
Got it - thanks!

PR #7351, opened 2021-10-13
Fix assert message format

keytronic:
This PR is related to a previous one I opened: #7345

I've made a mistake when I tested regex because `\nexpected: (.*)\n\s*got: (.*)` does not allow prefixes.
I changed assert messages to use suffixes instead of prefixes.

Sorry about that !

### DIFF
--- a/src/test/java/io/reactivex/rxjava3/observers/TestObserverTest.java
+++ b/src/test/java/io/reactivex/rxjava3/observers/TestObserverTest.java
@@ -999,7 +999,7 @@ public class TestObserverTest extends RxJavaTest {
 
     @Test
     public void assertValueAtIndexNoMatch() {
-        assertThrowsWithMessage("Values at position 2 differ;\nexpected: b (class: String)\ngot: c (class: String) (latch = 0, values = 3, errors = 0, completions = 1)", AssertionError.class, () -> {
+        assertThrowsWithMessage("\nexpected: b (class: String)\ngot: c (class: String); Value at position 2 differ (latch = 0, values = 3, errors = 0, completions = 1)", AssertionError.class, () -> {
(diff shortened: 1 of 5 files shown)


akarnokd (approved):

vanniktech (approved):
Should the error messages be tested against that regex to ensure that it works?

akarnokd (changes requested):
Indeed, please add tests that verify the error messages against the intended regex.

src/test/java/io/reactivex/rxjava3/observers/TestObserverTest.java
```diff
@@ -999,7 +1005,7 @@ public void assertValueAtIndexMatch() {
 
     @Test
     public void assertValueAtIndexNoMatch() {
-        assertThrowsWithMessage("Values at position 2 differ;\nexpected: b (class: String)\ngot: c (class: String) (latch = 0, values = 3, errors = 0, completions = 1)", AssertionError.class, () -> {
+        assertThrowsWithMessageMatchRegex(ASSERT_MESSAGE_REGEX, AssertionError.class, () -> {
```
akarnokd, line 1019: Please don't change this to the regex check but introduce separate tests.

akarnokd (changes requested):

src/test/java/io/reactivex/rxjava3/observers/TestObserverTest.java
keytronic: Good point. I fixed tests 👍

akarnokd (approved):

vanniktech (approved):

Akhenaton27 (approved):

PR #7384, opened 2022-01-11
3.x: Maybe.fromSupplier & fromCallable Nullable annotation fix

### LINKED ISSUE #7376: Maybe#fromSupplier expects not null value from callable
In documentation about Maybe::fromSupplier it's said it works different from other operators and don't throw `NullPointerException` in case we provide `null` like
```kotlin
Maybe.fromSupplier<Boolean> {
    null as? Boolean
}
``` 
The static code analyser marks it with warning as expected `Boolean but was Boolean?`
In case we change the code like
```kotlin
Maybe.fromSupplier<Boolean?> {
    null as? Boolean
}
``` 
We also got warning as T is marked as NonNull
```java
@CheckReturnValue
@NonNull
public static <@NonNull T> Maybe<T> fromSupplier(@NonNull Supplier<? extends T> supplier) {
     Objects.requireNonNull(supplier, "supplier is null");
     return RxJavaPlugins.onAssembly(new MaybeFromSupplier<>(supplier));
}
```
Also according documentation this method `@throws NullPointerException if {@code supplier} is {@code null}` that's wrong

akarnokd:
Adds a nullability override to `Maybe.fromSupplier` and `Maybe.fromCallable` to avoid warnings in Kotlin when returning nulls from the callbacks.

While coming up with the fix, it appears the Kotlin compiler misinterprets the `@NonNull` on the `supplier` parameter itself and still gives the warning. In my understanding, this might be a bug with the compiler. Reported https://youtrack.jetbrains.com/issue/KT-50734 .

Resolves #7376

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Maybe.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Maybe.java
@@ -1079,7 +1079,7 @@ public abstract class Maybe<T> implements MaybeSource<T> {
     @CheckReturnValue
     @NonNull
     @SchedulerSupport(SchedulerSupport.NONE)
-    public static <@NonNull T> Maybe<T> fromCallable(@NonNull Callable<? extends T> callable) {
+    public static <@NonNull T> Maybe<T> fromCallable(@NonNull Callable<? extends @Nullable T> callable) {
         Objects.requireNonNull(callable, "callable is null");
         return RxJavaPlugins.onAssembly(new MaybeFromCallable<>(callable));
     }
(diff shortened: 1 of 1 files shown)


vanniktech (approved):

akarnokd:
Looks like this fix is no good still because enabling nullability on the type argument that is already limited to nonnull has the wrong variance. 

I.e., a nullable supplier parameter can use a non-nullable supplier because whatever gets the value can deal with both null and non-null values.

`Supplier<T?> <- Supplier<T!>`

The reverse doesn't work as a non-nullable supplier can't use a nullable supplier as the get()-ter is not expecting a null.

Maybe the simplest fix is to drop the declaration site non-null annotation and move them to all the use sites.

PR #7391, opened 2022-02-03
refactor: Use diamond operator

zieka:
Uses diamond operator to infer type

---
Created with: [Moderne](https://app.moderne.io)


akarnokd:
Please don't, especially if you used some tool to auto-generate these. Some of those diamond usages result in compilation errors in Eclipse.

PR #7410, opened 2022-04-08
Set permissions for GitHub actions

neilnaveen:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,6 +1,9 @@
 name: "Validate Gradle Wrapper"
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   validation:
     name: "Validation"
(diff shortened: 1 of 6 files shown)


akarnokd:
We have scripts that post back javadoc. Would they still work?

neilnaveen:
> We have scripts that post back javadoc. Would they still work?

It does work with generating javadoc https://github.com/turrisxyz/RxJava/runs/5885975144?check_suite_focus=true

akarnokd:
Nope, it doesn't work: https://github.com/ReactiveX/RxJava/runs/6484285664?check_suite_focus=true#step:10:299

PR #7419, opened 2022-05-17
Declare `<? super @Nullable Foo>` instead of `<@Nullable ? super Foo>`.

cpovirk:
The latter results in runtime errors when using `-Xnullability-annotations=@io.reactivex.rxjava3.annotations:strict`: Kotlin interprets `? super Foo` to force the callback parameters to be non-nullable types, apparently effectively ignoring the `@Nullable` part?

Even if this is technically a Kotlin bug (and I don't know if it is), I recommend `<? super @Nullable Foo>`: That is the syntax that is preferred [by the Checker Framework](https://checkerframework.org/manual/#annotations-on-wildcards) and by [JSpecify](https://jspecify.dev/).

(In fact, I'd likewise recommend changing `class Foo<@NonNull T>` to `class Foo<T extends @NonNull Object>`. In that case, Kotlin handles both versions fine. But the latter version is better for Checker Framework and JSpecify. I've experimented with a version of that, and I'd be happy to send it as a pull request if you'd like to go that direction.)

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Completable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Completable.java
@@ -1868,7 +1868,7 @@ public abstract class Completable implements CompletableSource {
     @CheckReturnValue
     @NonNull
     @SchedulerSupport(SchedulerSupport.NONE)
-    public final Completable doOnEvent(@NonNull Consumer<@Nullable ? super Throwable> onEvent) {
+    public final Completable doOnEvent(@NonNull Consumer<? super @Nullable Throwable> onEvent) {
         Objects.requireNonNull(onEvent, "onEvent is null");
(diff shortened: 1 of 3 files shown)


akarnokd:
This could be a Kotlin problem. Please post on their issue tracker.

I see no compelling reason to change annotation ordering. Perhaps The Checker Framework and JSpecify should be more accepting with our style.

cpovirk:
I imagine this will get fixed by [KT-50734](https://youtrack.jetbrains.com/issue/KT-50734), so thanks for reporting that.

Do you happen to know of other projects or tools that go with annotations directly on the wildcard (`@Nullable ? super Foo`)? I may need to find some workarounds for us for other libraries that do so, or maybe I can establish enough demand for the tools to offer configuration to support that model.

akarnokd:
No idea.

PR #7441, opened 2022-06-08
update FlowableDistinct to support external storage keep distinct ticket

wenhaozhao:
Update FlowableDistinct, replace key collection with Function<? super T, Single<Boolean>> duplicateDetector, allow to use external storage,like redis, to keep distinct ticket


akarnokd:
Absolutely not.

You messed up the internals of an established operator, which is a big no-no. Also the way it is messed up doesn't work properly with the lifecycle of the rest of the operator.

PR #7459, opened 2022-08-08
3.x: Standardize MissingBackpressureException message, introduce QueueOverflowException

akarnokd:
Extend `MissingBackpressureException` with a standard message string and a static methods to create the exception with the standard message. This will make sure every case reports the same consistent message.

In addition, the queue overflows now have a separate exception class `QueueOverflowException`. This exception should not happen during normal operations and is mainly there to detect misuse or bugs in operators. Consequencly, even though many places used to signal `MissingBackpressureException` for such queue overflows, there was no reason to check for this specific exception class outside developing RxJava itself.

ThrottleLatest needs to be updated after #7457

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/exceptions/MissingBackpressureException.java
+++ b/src/main/java/io/reactivex/rxjava3/exceptions/MissingBackpressureException.java
@@ -21,6 +21,25 @@ public final class MissingBackpressureException extends RuntimeException {
     private static final long serialVersionUID = 8517344746016032542L;
 
     /**
+     * The default error message.
+     * <p>
+     * This can happen if the downstream doesn't call {@link org.reactivestreams.Subscription#request(long)}
+     * in time or at all.
+     * @since 3.1.6
+     */
+    public static final String DEFAULT_MESSAGE = "Could not emit value due to lack of requests";
+
+    /**
+     * The message for queue overflows.
+     * <p>
+     * This can happen if the upstream disregards backpressure completely or calls
+     * {@link org.reactivestreams.Subscriber#onNext(Object)} concurrently from multiple threads
+     * without synchronization. Rarely, it is an indication of bugs inside RxJava
+     * @since 3.1.6
+     */
+    public static final String QUEUE_OVERFLOW_MESSAGE = "Queue overflow due to illegal concurrent onNext calls or a bug in RxJava";
+
+    /**
      * Constructs a MissingBackpressureException without message or cause.
      */
     public MissingBackpressureException() {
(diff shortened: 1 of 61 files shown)


src/main/java/io/reactivex/rxjava3/exceptions/MissingBackpressureException.java
```diff
@@ -20,6 +20,25 @@ public final class MissingBackpressureException extends RuntimeException {
...
+
+    /**
+     * The message for queue overflows.
+     * <p>
+     * This can happen if the upstream disregards backpressure completely or calls
+     * {@link org.reactivestreams.Subscriber#onNext(Object)} concurrently from multiple threads
+     * without synchronization. Rarely, it is an indication of bugs inside RxJava
+     * @since 3.1.6
+     */
+    public static final String QUEUE_OVERFLOW_MESSAGE = "Queue overflow due to illegal concurrent onNext calls or a bug in RxJava";
```
vanniktech: These could be private though, right?

vanniktech (approved):

src/main/java/io/reactivex/rxjava3/exceptions/MissingBackpressureException.java
akarnokd: 3rd party implementations should be able to verify they got this exact message:

```java
assertEquals(ex.getMessage(), MissingBackpressureException.QUEUE_OVERFLOW_MESSAGE);
```

vanniktech: Hmm should not we then just have two different exceptions both having only one public constructor which contains the exact message and then you simply check for its type which is superior than the String message?

akarnokd: Alright, let's introduce a separate exception.

akarnokd: Done.

src/main/java/io/reactivex/rxjava3/exceptions/QueueOverflowException.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public QueueOverflowException() {
+        this(DEFAULT_MESSAGE);
+    }
+
+    /**
+     * Constructs a QueueOverflowException with the given message but no cause.
+     * @param message the error message
+     */
+    public QueueOverflowException(String message) {
```
vanniktech, line 46: Why would we need this?

vanniktech (approved):

src/main/java/io/reactivex/rxjava3/exceptions/QueueOverflowException.java
akarnokd: In case 3rd party users would want to have a custom message.

vanniktech: Alright!

PR #7469, opened 2022-09-09
ci: add minimum GitHub token permissions for workflows

ashishkurmi:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using https://github.com/step-security/secure-workflows.

GitHub Actions workflows have a GITHUB_TOKEN with write access to multiple scopes.
Here is an example of the permissions in one of the workflows:
https://github.com/ReactiveX/RxJava/runs/8243359652?check_suite_focus=true#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflows:

- gradle_release.yml
- gradle_snapshot.yml

The following workflow files already have the least privileged token permission set:

- gradle-wrapper-validation.yml
- gradle_branch.yml
- gradle_jdk11.yml
- gradle_pr.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- GitHub recommends defining minimum GITHUB_TOKEN permissions.
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/gradle_release.yml
+++ b/.github/workflows/gradle_release.yml
@@ -10,6 +10,9 @@ on:
     tags:
       - 'v3.*.*'   
 
+permissions:
+  contents: read
+
 jobs:
   build:
 
--- a/.github/workflows/gradle_snapshot.yml
+++ b/.github/workflows/gradle_snapshot.yml
@@ -7,6 +7,9 @@ on:
   push:
     branches: [ '3.x' ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
 


akarnokd:
It doesn't work. We need to write back to the repo.

https://github.com/ReactiveX/RxJava/pull/7410#issuecomment-1129719916

Issue #7481, opened 2022-11-04
3.x: Add onDropped callback to throttleFirst operator


Desislav-Petrov:
This PR adds an overload to the throttleFirst operator which takes an onDropped consumer.

This consumer is called when we're still within the gated time window when any new item isn't passed to downstream.

If the onDropped crashes, the exception is either delivered to the downstream (if the sequence is still live at that point) or to the global error handler (sequence is stopped).

akarnokd:
Not sure why you opened this issue. Just link to #7458 in your PRs.

Issue #7458, opened 2022-08-08
3.x: Add onDropped callbacks to operators


akarnokd:
Add `onDropped` callback overloads to most operators that drop items that can't be recovered by other means.

- [x] `throttleLatest` [PR #7457](https://github.com/ReactiveX/RxJava/issues/7457)
- [x] `throttleFirst` [PR #7482](https://github.com/ReactiveX/RxJava/issues/7482)
- [x] `throttleLast` / `sample` [PR #7488](https://github.com/ReactiveX/RxJava/issues/7488)
- [x] `throttleWithTimeout` / `debounce` [PR #7510](https://github.com/ReactiveX/RxJava/issues/7510)
- [x] `onBackpressureLatest` ([PR #7458](https://github.com/ReactiveX/RxJava/pull/7542))
- [x] `onBackpressureBuffer` ([PR #7567](https://github.com/ReactiveX/RxJava/pull/7567))
  - 📓 The current overloads with `Action` are somewhat unhelpful, however, we'll have to add overloads in a way that avoids lambda ambiguity.

Desislav-Petrov:
Hi - do you need any help with some of these?

akarnokd:
@Desislav-Petrov you can have a stab at them. I won't have the time for weeks although there is no rush either.

Desislav-Petrov:
Thanks, starting this week and will follow your PR per method strategy - I've started with throttleFirst

Desislav-Petrov:
hi @akarnokd - i'm looking at throttleLast. From what I see, we should change the  public void onNext(T t) method in the *SampleTimed class. More precisely, when we do the lazySet() we should check if a value exists there - if so we need to call the onDropped with this value as this is what we're dropping. 
Does this make sense?

akarnokd:
@Desislav-Petrov `lazySets` have to be made `getAndSet`: https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableThrottleLatest.java#L130

Desislav-Petrov:
Thx - leave this with me, will do a PR shortly.

Desislav-Petrov:
taking a look at throttleWithTimeout / debounce next

Desislav-Petrov:
Looking at onBackpressureLatest next

PR #7482, opened 2022-11-04
Add onDropped callback for throttleFirst - addresses #7481

Desislav-Petrov:
This PR adds an overload to the throttleFirst operator which takes an onDropped consumer.

This consumer is called when we're still within the gated time window when any new item isn't passed to downstream.

If the onDropped crashes, the exception is either delivered to the downstream (if the sequence is still live at that point) or to the global error handler (sequence is stopped).

Related: #7458

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -17096,7 +17096,48 @@ public abstract class Flowable<@NonNull T> implements Publisher<T> {
     public final Flowable<T> throttleFirst(long skipDuration, @NonNull TimeUnit unit, @NonNull Scheduler scheduler) {
         Objects.requireNonNull(unit, "unit is null");
         Objects.requireNonNull(scheduler, "scheduler is null");
-        return RxJavaPlugins.onAssembly(new FlowableThrottleFirstTimed<>(this, skipDuration, unit, scheduler));
+        return RxJavaPlugins.onAssembly(new FlowableThrottleFirstTimed<>(this, skipDuration, unit, scheduler, null));
+    }
+
+    /**
+     * Returns a {@code Flowable} that emits only the first item emitted by the current {@code Flowable} during sequential
+     * time windows of a specified duration, where the windows are managed by a specified {@link Scheduler}.
+     * <p>
+     * This differs from {@link #throttleLast} in that this only tracks the passage of time whereas
+     * {@link #throttleLast} ticks at scheduled intervals.
+     * <p>
+     * <img width="640" height="305" src="https://raw.github.com/wiki/ReactiveX/RxJava/images/rx-operators/throttleFirst.s.v3.png" alt="">
+     * <dl>
+     *  <dt><b>Backpressure:</b></dt>
+     *  <dd>This operator does not support backpressure as it uses time to control data flow.</dd>
+     *  <dt><b>Scheduler:</b></dt>
+     *  <dd>You specify which {@code Scheduler} this operator will use.</dd>
+     * </dl>
+     *
+     * @param skipDuration
+     *            time to wait before emitting another item after emitting the last item
+     * @param unit
+     *            the unit of time of {@code skipDuration}
+     * @param scheduler
+     *            the {@code Scheduler} to use internally to manage the timers that handle timeout for each
+     *            event
+     * @param onDropped
+     *            called when an item doesn't get delivered to the downstream
+     *
+     * @return the new {@code Flowable} instance
+     * @throws NullPointerException if {@code unit} or {@code scheduler} or {@code onDropped} is {@code null}
+     * @see <a href="http://reactivex.io/documentation/operators/sample.html">ReactiveX operators documentation: Sample</a>
+     * @see <a href="https://github.com/ReactiveX/RxJava/wiki/Backpressure">RxJava wiki: Backpressure</a>
+     */
+    @CheckReturnValue
+    @NonNull
+    @BackpressureSupport(BackpressureKind.ERROR)
+    @SchedulerSupport(SchedulerSupport.CUSTOM)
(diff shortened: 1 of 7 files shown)


src/main/java/io/reactivex/rxjava3/core/Flowable.java
```diff
@@ -17096,7 +17096,48 @@ public final Flowable<T> throttleFirst(long windowDuration, @NonNull TimeUnit un
...
+     *
+     * @param skipDuration
+     *            time to wait before emitting another item after emitting the last item
+     * @param unit
+     *            the unit of time of {@code skipDuration}
+     * @param scheduler
+     *            the {@code Scheduler} to use internally to manage the timers that handle timeout for each
+     *            event
+     * @param onDropped
+     *            called when an item doesn't get delivered to the down stream
```
akarnokd, line 17125: `downstream`

src/main/java/io/reactivex/rxjava3/core/Observable.java
```diff
@@ -14163,7 +14163,44 @@ public final Observable<T> throttleFirst(long windowDuration, @NonNull TimeUnit
...
+     *
+     * @param skipDuration
+     *            time to wait before emitting another item after emitting the last item
+     * @param unit
+     *            the unit of time of {@code skipDuration}
+     * @param scheduler
+     *            the {@code Scheduler} to use internally to manage the timers that handle timeout for each
+     *            event
+     * @param onDropped
+     *            called when an item doesn't get delivered to the down stream
```
akarnokd: `downstream`

src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableThrottleFirstTimed.java
```diff
@@ -106,6 +117,14 @@ public void onNext(T t) {
                 }
 
                 timer.replace(worker.schedule(this, timeout, unit));
+            } else if (onDropped != null) {
+                try {
+                    onDropped.accept(t);
+                } catch (Throwable ex) {
+                    Exceptions.throwIfFatal(ex);
+                    upstream.cancel();
+                    done = true;
```
akarnokd, line 128: This will hang the sequence because the error is swallowed. Also there is no cleanup of the worker.

src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableThrottleFirstTimed.java
```diff
@@ -83,6 +97,13 @@ public void onNext(T t) {
                     d.dispose();
                 }
                 DisposableHelper.replace(this, worker.schedule(this, timeout, unit));
+            } else if (onDropped != null) {
+                try {
+                    onDropped.accept(t);
+                } catch (Throwable ex) {
+                    Exceptions.throwIfFatal(ex);
+                    upstream.dispose();
```
akarnokd, line 107: This will hang the sequence because the error is swallowed. Also there is no cleanup of the worker.

src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableThrottleFirstTest.java
```diff
@@ -44,6 +44,39 @@ public void before() {
...
+        inOrder.verify(subscriber, times(0)).onNext("two");
+        dropCallbackOrder.verify(dropCallbackObserver, times(1)).onNext("two");
+        inOrder.verify(subscriber, times(1)).onNext("three");
+        inOrder.verify(subscriber, times(0)).onNext("four");
+        dropCallbackOrder.verify(dropCallbackObserver, times(1)).onNext("four");
+        inOrder.verify(subscriber, times(1)).onComplete();
+        inOrder.verifyNoMoreInteractions();
+        dropCallbackOrder.verifyNoMoreInteractions();
+    }
+
```
akarnokd, line 118: Please add a test that verifies a crashing `onDropped` callback triggers the correct cleanup and notifications.

src/test/java/io/reactivex/rxjava3/internal/operators/observable/ObservableThrottleFirstTest.java
```diff
@@ -41,6 +41,39 @@ public void before() {
...
+        inOrder.verify(observer, times(0)).onNext("two");
+        dropCallbackOrder.verify(dropCallbackObserver, times(1)).onNext("two");
+        inOrder.verify(observer, times(1)).onNext("three");
+        inOrder.verify(observer, times(0)).onNext("four");
+        dropCallbackOrder.verify(dropCallbackObserver, times(1)).onNext("four");
+        inOrder.verify(observer, times(1)).onComplete();
+        inOrder.verifyNoMoreInteractions();
+        dropCallbackOrder.verifyNoMoreInteractions();
+    }
+
```
akarnokd, line 114: Please add a test that verifies a crashing `onDropped` callback triggers the correct cleanup and notifications.

akarnokd (changes requested):

src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableThrottleFirstTimed.java
Desislav-Petrov: So we have to do downstream.onError(t) to propagate this, right? And .dispose() of the worker too?

akarnokd: Yes.

Desislav-Petrov: Should be addressed now, thx for the review

akarnokd (approved):

PR #7490, opened 2022-12-03
3.x: Fix experimental, signatures of throttleLast

akarnokd:
- Add experimental annotations
- Fix `Observable.sample` wrong overload
- Fix contravariance on the `onDropped` consumer.


vanniktech (commented):
👍

PR #7493, opened 2022-12-09
Use Comparator.naturalOrder().

Isira-Seneviratne:

----
Use `Comparator.naturalOrder()` in `Functions.naturalComparator()`.


akarnokd:
Thanks but we keep jdk8-related API calls separate, just in case. The current implementation does the job just fine.

PR #7495, opened 2022-12-10
Use lambdas in Functions.

Isira-Seneviratne:

----
Use lambdas in the `Functions` class to reduce the lines of code needed. On Android, lambdas are [desugared](https://jakewharton.com/androids-java-8-support/) into static classes.


akarnokd:
Hi. Again, please no Java 8 things outside the jdk8 packages and unit tests. Also there is no need to tidy up any of the helper methods in RxJava.

PR #7515, opened 2023-01-16
3.x: Fix Single.timeout race condition

### LINKED ISSUE #7514: .timeout with small value on Single does not emit error. RxJava 2.2.21
Recently I notices a problem when using Single.timeout with a value really small. So in the snippet bellow you will see 0 nanoseconds, but it also happens with 10 nanoseconds. So also values > 0.
The problem is, that in such a case, the error is no always emitted.

```java
		for (int i = 0; i < 10000000; i++) {
			final int y = i;
			final CountDownLatch latch = new CountDownLatch(1);
			Single.never()
					.timeout(0, TimeUnit.NANOSECONDS, Schedulers.computation())
					.subscribe(v -> {}, e -> {
						System.out.println("timeout " + y);
						latch.countDown();
					});
			if (!latch.await(1, TimeUnit.SECONDS)) {
				throw new IllegalStateException("Timeout was not happening!");
			}
		}
```

The workaround so far for me is to use a Single.timer instead.

```java
		for (int i = 0; i < 10000000; i++) {
			final int y = i;
			final CountDownLatch latch = new CountDownLatch(1);
			Single.amb(Arrays.asList(Single.never(), Single.timer(0, TimeUnit.NANOSECONDS, Schedulers.computation()).doOnSuccess(l -> {
				throw new TimeoutException();
			})))
					.subscribe(v -> {}, e -> {
						System.out.println("timeout " + y);
						latch.countDown();
					});
			if (!latch.await(1, TimeUnit.SECONDS)) {
				throw new IllegalStateException("Timeout was not happening!");
			}
		}
```

The above snippets take like 2-3 minute to run in the success case. Most of the time the first one fails during that time, but sometimes also succeeds.
I tested it also with Observable, but could not notice the problematic behavior.

Do you have an idea where this comes from?

akarnokd:
The PR fixes a race condition in `Single.timeout` where when the time was really small, the timeout logic in `run` would race with the `onSubscribe` logic and not handle the the timeout case.

The race happened between `Disposable d = get()` and the later `compareAndSet(d, DisposableHelper.DISPOSED)`. When the `get` executes, it sees `null`. Then comes `onSubscribe` and atomically sets some `Disposable` in `this`. Now `compareAndSet` fails because the `null -> Disposed` transition can't be made.

The fix is to use `DisposableHelper.dispose` which returns true iff the current thread managed to transition from a non-disposed state into a disposed state.

Fixes #7514

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/single/SingleTimeout.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/single/SingleTimeout.java
@@ -113,11 +113,7 @@ public final class SingleTimeout<T> extends Single<T> {
 
         @Override
         public void run() {
-            Disposable d = get();
-            if (d != DisposableHelper.DISPOSED && compareAndSet(d, DisposableHelper.DISPOSED)) {
-                if (d != null) {
-                    d.dispose();
-                }
+            if (DisposableHelper.dispose(this)) {
                 SingleSource<? extends T> other = this.other;
                 if (other == null) {
                     downstream.onError(new TimeoutException(timeoutMessage(timeout, unit)));
--- a/src/test/java/io/reactivex/rxjava3/internal/operators/single/SingleTimeoutTest.java
+++ b/src/test/java/io/reactivex/rxjava3/internal/operators/single/SingleTimeoutTest.java
@@ -28,7 +28,7 @@ import io.reactivex.rxjava3.exceptions.TestException;
 import io.reactivex.rxjava3.functions.Action;
 import io.reactivex.rxjava3.observers.TestObserver;
 import io.reactivex.rxjava3.plugins.RxJavaPlugins;
-import io.reactivex.rxjava3.schedulers.TestScheduler;
+import io.reactivex.rxjava3.schedulers.*;
 import io.reactivex.rxjava3.subjects.*;
 import io.reactivex.rxjava3.testsupport.TestHelper;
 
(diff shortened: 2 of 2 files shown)


src/main/java/io/reactivex/rxjava3/internal/operators/single/SingleTimeout.java
```diff
@@ -113,11 +113,7 @@ public void onError(Throwable e) {
 
         @Override
         public void run() {
-            Disposable d = get();
-            if (d != DisposableHelper.DISPOSED && compareAndSet(d, DisposableHelper.DISPOSED)) {
```
vanniktech, line 117: Are there more patterns in the codebase like this which would also need to be fixed?

vanniktech (approved):

src/main/java/io/reactivex/rxjava3/internal/operators/single/SingleTimeout.java
akarnokd: That's the only one with this pattern I could find (i.e., no loop).

akarnokd: This is fine because it loops:
https://github.com/ReactiveX/RxJava/blob/5bb119b1489afe4fbc5bb51b828be95971ee8f1d/src/main/java/io/reactivex/rxjava3/internal/observers/FutureObserver.java#L50-L56

This is fine too:
https://github.com/ReactiveX/RxJava/blob/5bb119b1489afe4fbc5bb51b828be95971ee8f1d/src/main/java/io/reactivex/rxjava3/internal/observers/FutureMultiObserver.java#L49-L55

vanniktech: :shipit:

PR #7520, opened 2023-01-26
add copyright and license to jar

tobi5775:
### Add copyright and license information to Jar.

For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar file. Therefore it would be great to have the copyright and license information included in the jar File as COPYRIGHT and LICENSE files.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,0 +1,13 @@
+Copyright (c) 2016-present, RxJava Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


akarnokd:
I don't know. I don't like putting files into the root of the jar. Maybe under `META-INF` instead?

What's the alternative for your tool?

tobi5775:
`META-INF` would be fine for us. I have adapted it!

PR #7535, opened 2023-03-07
3.x: Try using https to access the reactive streams javadoc

akarnokd:
You know, instead of just disabling all javadoc external links.


akarnokd:
Nope, didn't work.

akarnokd:
Seems to work (no warnings). Will only really know once the snapshot javadoc is posted after merging.

PR #7541, opened 2023-03-14
Add minimum GitHub token permissions for workflows

### LINKED ISSUE #7540: Add minimum GitHub token permissions for workflows
@akarnokd can I bring up this topic (#7469) again and perhaps suggest a PR with the permission that were missing in this one? It is really a relevant topic on supply-chain security that prevents any threats and attacks. 

But before I would like to understand what didn't work on https://github.com/ReactiveX/RxJava/actions/runs/2343714818/jobs/3517587271 because for me it shows as success.

Looking into gradle_snapshot.yml and gradle_release.yml I've noticed that perhaps the permission that were missing was a `contents: write` to run this command https://github.com/ReactiveX/RxJava/blob/ed674be204210fe0804d64e28fdd356745915048/push_javadoc.sh#L118

Let me know if a PR is welcome and I'll try testing it before submit it, although I think it will be difficult considering sonatype access used.

joycebrum:
Closes #7540 

### Changes

Set top level as contents: read and job level permission as contents: write to the following workflows

- gradle_release.yml
- gradle_snapshot.yml

Let me know if anything fail and I'll try to guess which permission may still be missing, but looking at the script that had failed, it seems it was only the contents: write indeed.

### DIFF
--- a/.github/workflows/gradle_release.yml
+++ b/.github/workflows/gradle_release.yml
@@ -10,10 +10,15 @@ on:
     tags:
       - 'v3.*.*'   
 
+permissions:
+  contents: read
+
 jobs:
   build:
 
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     env:
       CI_BUILD_NUMBER: ${{ github.run_number }}
     steps:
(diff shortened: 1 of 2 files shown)


akarnokd:
Thanks. The javadoc pushback still works. :+1:

PR #7574, opened 2023-06-23
Make inner class static for CompletableOnErrorComplete

chintando:

After taking a heap dump for a Rx chain that was completed, we saw evidence of inner class showing up on the heap dump analysis. It apears that CompletableOnErrorComplete#onError is leaking the predicate

For instance

class Foo {

fun test(): Completable {
Completable.fromCallable {
println("hello world")
}.onErrorComplete { --> leaks after chain completion
true
}

}

}

Verified with kotlin bytecode dump and heap analysis

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/completable/CompletableOnErrorComplete.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/completable/CompletableOnErrorComplete.java
@@ -35,7 +35,7 @@ public final class CompletableOnErrorComplete extends Completable {
         source.subscribe(new OnError(observer));
     }
 
-    final class OnError implements CompletableObserver {
+    static final class OnError implements CompletableObserver {
 
         private final CompletableObserver downstream;
 


akarnokd:
Your change makes no sense as the `OnError` loses the reference altogether and causes a compile error.

Usually a leak is caused by your application holding references, for example, in a `CompositeDisposable` which is not cleaned up. There are workarounds such as `onTerminateDetach` or using the 4 argument `subscribe` method that removes itself from a `CompositeDisposable` upon termination.

chintando:
i believe the `OnError` should have been taking the predicate as part of constructor, won't the onError hold on to predicate thereby not being able to clean up the predicate object? Just curious

akarnokd:
Predicates are never cleared in RxJava.

> i believe the OnError should have been taking the predicate as part of constructor

Yes, that could be done. If you want, you can resubmit the PR with this change.

chintando:
sure thing will do

PR #7590, opened 2023-08-21
Remove `@NonNull` annotation from `Collector` container type parameter.

cpovirk:
`collect` delegates to `FlowableCollectWithCollectorSingle`, which
is not annotated to require a non-nullable type for its corresponding
type parameter and whose implementation works fine with a null
container.

This PR lets Kotlin code pass a `Collector<T, A, R>`, where `A` is
declared as `<A: Any?>`.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -20394,7 +20394,7 @@ public abstract class Flowable<@NonNull T> implements Publisher<T> {
     @BackpressureSupport(BackpressureKind.UNBOUNDED_IN)
     @SchedulerSupport(SchedulerSupport.NONE)
     @NonNull
-    public final <@NonNull R, @NonNull A> Single<R> collect(@NonNull Collector<? super T, A, R> collector) {
+    public final <@NonNull R, @Nullable A> Single<R> collect(@NonNull Collector<? super T, A, R> collector) {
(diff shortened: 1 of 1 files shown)


cpovirk:
Wait, did I completely misread this? I need to have another look. Sorry....

cpovirk:
OK, I might have it:

I think that this change still makes sense. But I got confused because I was looking at the _other_ overload of `collect`. But... the other overload of `collect` appears to be the one that I was seeing the problem with in our code. And that overload [_does_ need a non-nullable value from its `Supplier`](https://github.com/ReactiveX/RxJava/blob/9975460e46116d1cac36e4a8b7ff9bb34df7c24a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableCollectSingle.java#L46). So our code is buggy, and the current annotations caught it!

Next, I'm going to see if our code could get away with calling the overload that I'm changing in this PR. If it can, then I will still be interested in the PR, since I do think the annotations for _this_ overload are unnecessarily strict.

cpovirk:
Yes: Our code _can_ be changed to call this overload. (The code was originally written for RxJava 2, when the `Collector` overload wasn't yet present.) That would eliminate our current nullness bug. We'd just need the annotation change of this PR so that kotlinc can tell that our old code is unsafe but our new code is safe.

cpovirk:
Thanks, and thank you for taking care of Observable, too.

PR #7603, opened 2023-09-29
In ofType methods, cast outer Rx construct directly instead of casting each element

derektom14:
- [X] Please give a description about what and why you are contributing, even if it's trivial.

Currently, `ofType` works by combining a `filter` and a `cast`. After the `filter` is done, we already know that the resulting Rx construct only emits elements of the desired type. We can cast the resulting construct to indicate this known type, instead of performing a cast on the element itself, which is an unnecessary ceremony that creates an additional segment of the Rx chain and all of the overhead that entails.

  - [X] Please include the issue list number(s) or other PR numbers in the description if you are contributing in response to those.

None, a direct proposal as a PR.

  - [X] Please include a reasonable set of unit tests if you contribute new code or change an existing one. If you contribute an operator, (if applicable) please make sure you have tests for working with an `empty`, `just`, `range` of values as well as an `error` source, with and/or without backpressure and see if unsubscription/cancellation propagates correctly.

Does not add any new behavior, existing unit tests suffice.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/core/Flowable.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Flowable.java
@@ -12388,7 +12388,7 @@ public abstract class Flowable<@NonNull T> implements Publisher<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <@NonNull U> Flowable<U> ofType(@NonNull Class<U> clazz) {
         Objects.requireNonNull(clazz, "clazz is null");
-        return filter(Functions.isInstanceOf(clazz)).cast(clazz);
+        return (Flowable<U>) filter(Functions.isInstanceOf(clazz));
     }
 
     /**
--- a/src/main/java/io/reactivex/rxjava3/core/Maybe.java
+++ b/src/main/java/io/reactivex/rxjava3/core/Maybe.java
@@ -4455,7 +4455,7 @@ public abstract class Maybe<@NonNull T> implements MaybeSource<T> {
     @SchedulerSupport(SchedulerSupport.NONE)
     public final <@NonNull U> Maybe<U> ofType(@NonNull Class<U> clazz) {
         Objects.requireNonNull(clazz, "clazz is null");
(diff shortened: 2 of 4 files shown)


akarnokd:
No. We need the hard cast as it is.

derektom14:
What's the reason that the hard cast is required?

akarnokd:
Soft casts are erased at compile time so if someone misuses the generics, ClassCastExceptions can happen at arbitrary stages.

derektom14:
The generics are all erased at compile-time, though. Before and after this change, the returned `Observable<U>` will only ever emit elements of type `U`. In what circumstance would someone be able to call `ofType(U.class)` and further subscribe to it in a way that leads to `ClassCastException`, but only after this change?

akarnokd:
What is your use case that would need this change? Can't you just use the "simplification" in your own project?

derektom14:
It's not a strictly needed change, just an optimization. The project I'm primarily using Rx for is large enough that applying this change in RxJava itself is far simpler than replacing every `ofType` method call with something else, and then adding linter rules to make sure everyone continued to use the optimized version.

akarnokd:
I see. Still no.

Why do you have so many `ofType`s? Are you using an event bus? Instead of dumping every object into one subject, have separate subjects per type (or the most common types) -> no cast required whatsoever.

derektom14:
Why still no? Logically, after type erasure, the `cast` call makes an assertion that we know will always be true (because it's the same predicate as the `filter` before it) and emits the same value it received. It cannot provide any benefit at runtime.

The `ofType`s have a variety of uses, often it's a case where a component is emitting its new states, which use a strongly-typed hierarchy of sealed classes (sometimes nested), and a particular observer at times cares only about a specific subset of those state changes, sometimes after first processing all state changes for a more general step. Splitting everything into separate `Subject`s would have even more overhead, as many objects would be emitted to multiple `Subject`s. Either way, the codebase's `ofType` calls number in the thousands, so any refactor attempt would be far more complicated than the change I'm proposing here.

PR #7645, opened 2024-02-01
Fix of small Javadoc mistake

GunniBusch:
Fixed a wrong placed dot in the Javadoc of the create method in the Observable class.


akarnokd:
Please try your suggestion.

GunniBusch:
@akarnokd Could you please tell me what you mean? I fixed only one line of JavaDoc.

akarnokd:
Your fix is wrong.

https://stackoverflow.com/a/47266244

Issue #7720, opened 2024-06-11
LambdaObserver fails to push exception to error channel and fails with NPE


eciuca:
https://github.com/ReactiveX/RxJava/blob/28679a3a047f46af3027a99041c77e45b7058106/src/main/java/io/reactivex/rxjava3/internal/observers/LambdaObserver.java#L66C17-L66C22

get() returns null so get().dispose() throws NPE instead of going to the next line

akarnokd:
It can only happen if `onSubscribe` was not called, which is a protocol violation.

How are you using the `LambdaObserver`?

eciuca:
@akarnokd This is how I am using it. I am not touching any internal classes, I got to it by debugging.

![Screenshot 2024-06-11 at 15 16 57](https://github.com/ReactiveX/RxJava/assets/3948642/1c9bd1e0-96a7-4465-bb87-21479d99a4a5)
![Screenshot 2024-06-11 at 15 17 27](https://github.com/ReactiveX/RxJava/assets/3948642/c121415f-35c8-4dce-84c3-826360ee53bd)

And whenever an error is thrown from any subscriber created with a lambda (I guess this is why a LambdaObservable is created) I get a NPE instead of the actual error itself.

So the general case should be I have an observable. I subscribe to it and in the onNext handler an error is thrown, instead of getting the error on the onError channel it blows up in the try-catch with NPE. Do you still think I am using it wrong because I don't see how.

EDIT: I saw your answer here after this post: https://github.com/ReactiveX/RxJava/pull/7721#issuecomment-2160216805
EDIT 2: I am continuing the discussion on the issue

akarnokd:
If you implement `Observable`s like this, you have to follow the protocol, i.e., call `onSubscribe` with a `Subscription` before calling `onNext`.

However, this is an advanced use case and not recommended in most cases. You haven't shown what `readItemPublisher` does but you could probably get by with simply using `Observable.create`.

eciuca:
Thank you very much! Using Observable.create instead of fromPublisher really does the trick (in my head I needed a Publisher to push values so I went for it instead of Observable.create which receives an ObservableOnSubscribe as parameter which is what I actually need anyway probably)

PR #7721, opened 2024-06-11
Prevent NPE when catching exception from LambdaObservable#onNext (#7720)

eciuca:
Closes (#7720)

I added a check to prevent NullPointerException when handling exception caught when trying to execute the code from the subscriber and trying to push to the Observer error handler

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/observers/LambdaObserver.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/observers/LambdaObserver.java
@@ -63,7 +63,10 @@ public final class LambdaObserver<T> extends AtomicReference<Disposable>
                 onNext.accept(t);
             } catch (Throwable e) {
                 Exceptions.throwIfFatal(e);
-                get().dispose();
+                Disposable d = get();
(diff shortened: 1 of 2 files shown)


akarnokd:
Based on the test case, you are misusing the LambdaObserver, which is a protocol violation, hence the NPE.

Also it is an internal class so direct usage is discouraged.

The fix for your case is simply do this in your code:

```java
o.onSubscribe(Disposable.empty());
o.onNext(1);
```

Or if you have created a custom `Observable`, make sure it correctly calls `onSubscribe`.

eciuca:
So I still cannot fix my issue. I am creating an Observable using Observable.fromPublisher and I don't seem to have access to the Observable#onSubscribe(Disposable) method. I am leaving some screenshots here:

Screenshot 2024-06-11 at 15 16 57
Screenshot 2024-06-11 at 15 17 27

eciuca:
Fixed with this suggestion: https://github.com/ReactiveX/RxJava/issues/7720#issuecomment-2160654014

Thank you again, @akarnokd !

PR #7747, opened 2024-08-10
Try fixing Jacoco - codecov errors

akarnokd:
Looking at similar issues, 0.8.9+ changed something that breaks codecov. Others have success downgrading to 0.8.8


akarnokd:
Nope. Since Jacoco is part of gradle, it can't be downgraded anymore? Bummer.

PR #7789, opened 2024-11-05
Update BlockingFlowableIterable.onNext() to set error before cancel

kaioni17:
To avoid race with hasNext(), which checks for cancel first before checking for error. For example, in the following case, hasNext() may return false to the caller, making the caller assume the iterable finished successfully.
1. onNext() called cancel
2. hasNext() found the iterable is cancelled
3. hasNext() found that error is null thus returned false to the caller, without throwing the error
4. onNext() set error

It's a bit hard to test this race condition in unit tests. Existing tests can make sure no regression is introduced.


akarnokd (approved):
Thanks for catching this.

kaioni17:
Thanks for the review!

PR #7801, opened 2024-11-26
Fix a possible issue introduced with #7799

akarnokd:
In JCTools, the current node's next is set to itself as null could mean trouble.

https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/BaseLinkedQueue.java#L232

Related #7790

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/queue/MpscLinkedQueue.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/queue/MpscLinkedQueue.java
@@ -92,7 +92,7 @@ public final class MpscLinkedQueue<T> implements SimplePlainQueue<T> {
             final T nextValue = nextNode.getAndNullValue();
             spConsumerNode(nextNode);
             // unlink previous consumer to help gc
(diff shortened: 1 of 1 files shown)


src/main/java/io/reactivex/rxjava3/internal/queue/MpscLinkedQueue.java
```diff
@@ -92,7 +92,7 @@ public T poll() {
             final T nextValue = nextNode.getAndNullValue();
             spConsumerNode(nextNode);
             // unlink previous consumer to help gc
-            currConsumerNode.soNext(null);
+            currConsumerNode.soNext(currConsumerNode);
```
carlcyang, line 95: currConsumerNode改成currConsumerNode2

Jasonw74:
> In JCTools, the current node's next is set to itself as null could mean trouble.
> 
> https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/BaseLinkedQueue.java#L232
> 
> Related #7790

PR #7815, opened 2024-12-29
chore(build): upgrade to gradle 8.12

sullis:
upgrade to latest version of Gradle:  8.12

https://docs.gradle.org/8.12/release-notes.html


akarnokd:
We don't want to update Gradle to any newer version as it breaks plugins. Also we don't accept upgrades to Gradle from outside.

PR #7817, opened 2025-01-11
3.x: Fix thread-safety issue reported by infer

dzmitrykashlach:
* Fix BufferBoundarySubscriber.emmited

   The issue was discovered during code static analysis with [infer](https://github.com/facebook/infer).  
Full issue text:  
```
src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableBufferBoundary.java:138: warning: Thread Safety Violation
 Unprotected write. Non-private method `FlowableBufferBoundary$BufferBoundarySubscriber.onError(...)` indirectly writes to field `this.emitted` outside of synchronization.
Reporting because this access may occur on a background thread.  
 136.                   }
 137.                   done = true;
 138. >                 drain();
 139.               }
 140.           }
```

   I think it's not a false positive and suggest to make `BufferBoundarySubscriber.emmited` field `AtomicLong`.  This change solves issue and shouldn't create new problems per my understanding.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableBufferBoundary.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableBufferBoundary.java
@@ -82,7 +82,7 @@ extends AbstractFlowableWithUpstream<T, U> {
 
         Map<Long, C> buffers;
 
-        long emitted;
+        AtomicLong emitted = new AtomicLong();
 
         BufferBoundarySubscriber(Subscriber<? super C> actual,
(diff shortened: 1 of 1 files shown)


akarnokd:
It is a false positive. The synchronization is ensured by the pair
```java
if (getAndIncrement() != 0) {
    return;
}
```
and
```java
missed = addAndGet(-missed);
if (missed == 0) {
    break;
}
```
the atomic increment and decrement ensure only 1 thread can enter the protected region, in which `emitted` is accessed. `emitted` is not accessed anywhere else so it is safe to have it as a plain field access.

PR #7855, opened 2025-04-09
3.x: Fix delay-cancellation race producing random subsequent emissions

### LINKED ISSUE #7851: Observable.delay(delayErrors=true) drops onNext signals at random when disposed of concurrently
When a subscription to an `Observable.delay(delayErrors=true)` operator is disposed of concurrently to the subscription's emissions, the `onNext` signals it issues to its downstream can be dropped at random.

See the following repro:

```java
Scheduler scheduler = Schedulers.computation();
int maxValue = 10;
int maxAttempts = 1000;
for (int attempt = 0; attempt < maxAttempts; attempt++) {
    ConcurrentLinkedDeque<Notification<Integer>> sink = new ConcurrentLinkedDeque<>();
    SequentialDisposable firstDisposable = new SequentialDisposable();

    Observable<Integer> integers = Observable.range(0, maxValue);
    Observable<Notification<Integer>> first = integers.delay(5, TimeUnit.MICROSECONDS, scheduler, true)
            .materialize()
            .doOnNext(sink::add)
            .doOnSubscribe(firstDisposable::replace);
    Observable<?> second = integers.delay(5, TimeUnit.MICROSECONDS, scheduler)
            .doOnNext($ -> firstDisposable.dispose())
            .doOnTerminate(latch::countDown);

    first.subscribe();
    second.subscribe();
    latch.await();

    boolean foundTermination = false;
    int nextExpectedValue = 0;
    List<String> errors = new ArrayList<>();
    List<Notification<Integer>> notifications = List.copyOf(sink);
    for (int i = 0; i < notifications.size(); i++) {
        Notification<Integer> notification = notifications.get(i);
        if (notification.isOnComplete() || notification.isOnError()) {
            foundTermination = true;
        }
        if (notification.isOnNext()) {
            if (foundTermination) {
                errors.add("Found onNext after termination at index " + i);
            }
            int value = notification.getValue();
            if (value != nextExpectedValue) {
                errors.add((errors.size() + 1) + ". Wrong value at index " + i + " (expected " + nextExpectedValue +
                        ", was " + value + ")");
            }
            nextExpectedValue = value + 1;
        }
    }
    if (!errors.isEmpty()) {
        throw new AssertionError(new StringBuilder()
                .append("Attempt ")
                .append(attempt)
                .append(" failed.\n\t\tError(s):\n\t\t\t")
                .append(String.join("\n\t\t\t", errors))
                .append("\n\t\tNotifications: ")
                .append(notifications)
                .toString());
    }
}
```

Here is an example failure:

```
java.lang.AssertionError: Attempt 11 failed.
            Error(s):
        1. Wrong value at index 2 (expected 2, was 3)
        2. Wrong value at index 3 (expected 4, was 5)
        3. Wrong value at index 6 (expected 8, was 9)
        Notifications: [OnNextNotification[0], OnNextNotification[1], OnNextNotification[3], OnNextNotification[5], OnNextNotification[6], OnNextNotification[7], OnNextNotification[9]]
```

I understand that a subscription that is actively producing signals may continue to propagate those signals to the downstream when disposed of concurrently (i.e. disposal is "eventually consistent"). But I would expect that any signals emitted during or after the race would preserve their original ordering. In other words, once `delay()` has caught up to its newly-disposed state and does _not_ emit its `N`th `onNext` signal, it seems wrong for it to go on to emit its `N+1`th signal.

akarnokd:
When a sequence using `delay` is cancelled asynchronously, there is a race between cancelling each individual `Future` inside the executor and them running, causing skips in the `onNext` emissions.

For example, let's say there are 3 futures waiting to be executed. The Future-1 triggers the cancellation asynchronously and emits `1`. This asynchronous routine will start cancelling Future-2 and Future-3 one by one. Concurrently, the executor will look at Future-2, see it cancelled and moves onto Future-3, which is still executable as the asynchronous routine hasn't gotten to cancel Future-3. Now Future-3 runs and emits `3`. This will appear to produce a sequence of `[1, 3]` where `2` is missing.

The fix is to hard-check for a disposed worker and prevent the `onNext` emissions in the operator.

Resolves #7851

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableDelay.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableDelay.java
@@ -111,7 +111,9 @@ public final class FlowableDelay<T> extends AbstractFlowableWithUpstream<T, T> {
 
             @Override
             public void run() {
-                downstream.onNext(t);
+                if (!w.isDisposed()) {
+                    downstream.onNext(t);
+                }
             }
         }
 
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableDelay.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableDelay.java
@@ -111,7 +111,9 @@ public final class ObservableDelay<T> extends AbstractObservableWithUpstream<T, 
 
             @Override
             public void run() {
-                downstream.onNext(t);
+                if (!w.isDisposed()) {
+                    downstream.onNext(t);
+                }
             }
         }
 
--- a/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableDelayTest.java
+++ b/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableDelayTest.java
@@ -20,6 +20,7 @@ import static org.mockito.Mockito.*;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
+import java.util.concurrent.locks.LockSupport;
 
 import org.junit.*;
 import org.mockito.InOrder;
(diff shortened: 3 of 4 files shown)


vanniktech (commented):
👍

PR #7856, opened 2025-04-14
Javadoc: Overveable.takeWhile vs. takeUntil: the condition is inverted

ePaul:
The current description of the difference between takeWhile and takeUntil just points to the evaluation time (before or after the element is emitted), but the (IMHO) more important fact is that the predicate is inverted between both: `takeWhile` emits just elements where the predicate is true, `takeUntil` just elements where the predicate is false (and the first true one).


ePaul:
The same sentence appears already in the [1.x](https://reactivex.io/RxJava/javadoc/rx/Observable.html#takeUntil-rx.functions.Func1-) and [2.x](https://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#takeUntil-io.reactivex.functions.Predicate-) docs, but I suppose there is no point in fixing that?

akarnokd:
The naming of `stopPredicate` was supposed to convey that.

I prefer not to change the docs.

Issue #7763, opened 2024-09-14
RxJava on JDK 21's virtual threads


abersnaze:
I'm exploring implementing a new branch of RxJava, where the internals are rewritten to leverage JDK 21's virtual threads. I hope this will significantly simplify Rx's internal implementation and debuggability while maintaining the same interface as 3.x.

This issue is to discuss the idea positively to address architectural problems with compatibility and performance.

I haven't yet set a time frame for when or if this will be completed. I will only submit this with the community's approval. If a community of developers likes this idea and runs with it faster than I keep up, I'm also okay with stepping aside.

@akarnokd et al., Before we discuss the merits or implementation of the idea, are there any additional ground rules for this discussion?

akarnokd:
> are there any additional ground rules for this discussion?

None I can think of right now.

cj-marius-duna:
Out of context: All I hope for this library is to die

josephtarbit:
What would you envision reactive programming's new role to be with the introduction of virtual threads? I can imagine virtual threads and structured concurrency would replace most non-stream based reactive usage (e.g. `Single` and `Mono`); but for streams of data I suppose reactive streams are still useful, albeit processing of individual stream elements could again be offloaded to virtual threads and structured concurrency.

abersnaze:
Much of the internals are devoted to non-blocking pressure. I was specifically thinking about the [drainLoop](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableFlatMap.java#L313)

I think the debugging stack traces and internals would be much simpler virtual blocking queue were used.

abersnaze:
I've been using claude to take on such a large task. Generated this plan

# RxJava 4: Virtual Thread Architecture Design

## Overview

RxJava 4 is a virtual-thread-first reactive library. It keeps the composable reactive API that makes RxJava powerful but replaces the thread-pool-based execution model with Java virtual threads. Same expressive operator chains, dramatically simpler mental model for concurrency.

### Motivations

- **Relevance** — Virtual threads change the landscape. RxJava must embrace them or become a legacy choice.
- **Simplicity** — Users shouldn't need to choose between `io()`, `computation()`, `newThread()`, and `single()` for most work.
- **Performance** — Virtual threads eliminate the thread-pool sizing game for I/O-bound workloads.

### Java Target

[...]

PR #8014, opened 2026-03-18
4.x: Add virtualCreate and virtualTransform to Flowable

akarnokd:
:information_source: You need the special operators below to make RxJava's non-blocking backpressure into virtually blocked backpressure.

### create

Creates a `Flowable` from a generator callback, that can emit via `VirtualEmitter`, run on an `ExecutorService` provided by the user and
is suspended automatically upon backpressure. The callback is executed inside the virtual thread thus you can call the usual blocking APIs and get suspensions the same way.

The created `Flowable` will complete once the callback returns normally or with an error if the callback throws an exception.

```java
try (var scope = Executors.newVirtualThreadPerTaskExecutor()) {
    Flowable.virtualCreate(emitter -> {
        for (int i = 1; i <= 5; i++) {
             emitter.emit(1);
        }
    }, scope)
    .test()
    .assertResult(1, 2, 3, 4, 5);
}
```

### transform

Transforms each upstream value via a callback that can emit zero or more values for each of those upstream values, run on an `ExecutorService` provided by the user and is suspended automatically upon backpressure. The callback is executed inside the virtual thread thus you can call the usual blocking APIs and get suspensions the same way.

```java
try (var scope = Executors.newVirtualThreadPerTaskExecutor()) {
    Flowable.range(1, 5)
    .virtualTransform((value, emitter) -> {
        emitter.emit(value);
        emitter.emit(value + 1);
    }, scope)
    .test()
    .assertResult(1, 2, 2, 3, 3, 4, 4, 5, 5, 6);
}
```

### blockingXXX

RxJava uses `java.util.concurrent` locks and `CountDownLatches` via its `blockingXXX` which will automatically work within a virtual thread. Therefore, there is no need for a separate interop operator. Just block.

```java
try (var scope = Executors.newVirtualThreadPerTaskExecutor()) {
    scope.submit(() -> {
        var v = Flowable.just(1)
        .delay(1, TimeUnit.SECONDS)
        .blockingLast();
        System.out.println(v);
    });
}
```


akarnokd:
Related: #7763

The scheduler upgrade / rewamp will happen in a separate PR.

PR #8056, opened 2026-03-25
Update gradle_release.yml, limit write capabilities of tools

akarnokd:
Improving our scorecard


akarnokd:
Who approves the approvers 😆 ?

PR #8062, opened 2026-03-25
gradle_snapshot.yml run build -> prerelease sequentially

akarnokd:
Fix the default behavior running build and *release in parallel.


akarnokd:
Will address test failures separately.

PR #8100, opened 2026-04-12
Refactor: migrate internal function handling from Function3 to FunctionRecord (record-based)

MNedzad:
Refactored internal function system to use FunctionRecord with record instead of Function3, simplifying argument handling and reducing the need for multiple function interfaces.


akarnokd:
No. We break minimal APIs. What worker with functional lambdas should work in 4.0.

MNedzad:
> No. We break minimal APIs. What worker with functional lambdas should work in 4.0.

Can this be added instead of replacing the existing implementation?

akarnokd:
No. We want to reduce the API surface. Config records are not the same as messing up the lambda types on a whim.

The aim of RxJava 4 development is: reduce the bulk and API surface, reduce the entropy, increase the beauty, conformally scale things.

PR #8159, opened 2026-06-22
Fix IllegalStateException in SchedulerToExecutorService.invokeAny

vasiliy-mikhailov:
## Problem

`SchedulerToExecutorService.invokeAny()` checks each future with:

```java
if (f.isDone() && !f.isCancelled() && f.exceptionNow() == null) {
```

But `Future.exceptionNow()` is specified to throw `IllegalStateException` when the task **completed normally** (there is no exception to return). So as soon as a task finishes successfully, this guard calls `exceptionNow()` on a successfully-completed future and throws `IllegalStateException` instead of returning the result. Both `invokeAny` overloads contain the same probe.

## Fix

Use the intended state query, which never throws:

```java
if (f.state() == Future.State.SUCCESS) {
```

## Test

Adds `SchedulerToExecutorServiceTest` covering the success paths. The new `invokeAnyShouldReturnResultOfCompletedTask` and `invokeAnyWithSingleTask` tests fail on `master` with `IllegalStateException` and pass with this fix.

Note: the timed `invokeAny(tasks, timeout, unit)` overload has a separate pre-existing issue in its time accounting (the `FIXME`-marked spin loop) that this PR intentionally does not touch; it only addresses the `exceptionNow()` misuse.

### DIFF
--- a/src/main/java/io/reactivex/rxjava4/internal/schedulers/SchedulerToExecutorService.java
+++ b/src/main/java/io/reactivex/rxjava4/internal/schedulers/SchedulerToExecutorService.java
@@ -189,7 +189,7 @@ public record SchedulerToExecutorService(@NonNull Scheduler scheduler,
 
         while (!isTerminated()) {
             for (var f : result) {
-                if (f.isDone() && !f.isCancelled() && f.exceptionNow() == null) {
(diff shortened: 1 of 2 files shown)


akarnokd:
Will fix the ParallelScheduler test leak in a separate PR

PR #8174, opened 2026-06-25
[Banned, AI Slop, Undisclosed AI Use, Had to Manually Correct PR] Fix NullPointerException in AppendOnlyLinkedArrayList.forEachWhile on a full last chunk

vasiliy-mikhailov:
When `forEachWhile(S, BiPredicate)` is called on an `AppendOnlyLinkedArrayList` whose last chunk is exactly full — i.e. the number of elements is a multiple of the chunk capacity — the next-chunk pointer stored at `a[capacity]` is still `null` (a new chunk is only allocated on the next `add`). After traversing the full chunk, `a` becomes `null` and the next loop iteration throws `NullPointerException` when reading `a[0]`.

This only affects the `BiPredicate` overload; it stops the traversal when the next chunk is `null`.

Reproducer (capacity 2, two elements, predicate never terminates):

```java
AppendOnlyLinkedArrayList<Integer> list = new AppendOnlyLinkedArrayList<>(2);
list.add(1);
list.add(2);
list.forEachWhile(null, (state, value) -> false); // NPE before this change
```

A regression test is added (`MiscUtilTest.appendOnlyLinkedArrayListForEachWhileBiFullChunkNoNext`) that fails before the change with `NullPointerException` and passes after it.

### DIFF
--- a/src/test/java/io/reactivex/rxjava4/internal/util/MiscUtilTest.java
+++ b/src/test/java/io/reactivex/rxjava4/internal/util/MiscUtilTest.java
@@ -235,4 +235,26 @@ public class MiscUtilTest extends RxJavaTest {
     public void queueDrainHelperUtility() {
         TestHelper.checkUtilityClass(QueueDrainHelper.class);
     }
+
+        @Test
+    public void appendOnlyLinkedArrayListForEachWhileBiFullChunkNoNext() throws Throwable {
+        // Fill exactly capacity elements in one chunk; no next chunk exists (a[c] == null).
+        // The predicate returns false for all elements so the loop must traverse past the full chunk.
+        AppendOnlyLinkedArrayList<Integer> list = new AppendOnlyLinkedArrayList<>(2);
+
+        list.add(1);
+        list.add(2);
+
+        final List<Integer> out = new ArrayList<>();
+
+        list.forEachWhile(null, new BiPredicate<Object, Integer>() {
+            @Override
+            public boolean test(Object state, Integer value) throws Throwable {
+                out.add(value);
+                return false; // never terminate early
+            }
+        });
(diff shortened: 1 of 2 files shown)


akarnokd (changes requested):
Please use lambdas.

src/test/java/io/reactivex/rxjava4/internal/util/MiscUtilTest.java
```diff
@@ -235,4 +235,26 @@ public void appendOnlyLinkedArrayListForEachWhileBiAll() throws Throwable {
...
+        AppendOnlyLinkedArrayList<Integer> list = new AppendOnlyLinkedArrayList<>(2);
+
+        list.add(1);
+        list.add(2);
+
+        final List<Integer> out = new ArrayList<>();
+
+        list.forEachWhile(null, new BiPredicate<Object, Integer>() {
+            @Override
+            public boolean test(Object state, Integer value) throws Throwable {
```
akarnokd: Please use lambdas.

akarnokd (changes requested):

akarnokd:
@copilot can you update PR so that the test method uses a lambda instead of the anonymous inner class?

akarnokd (approved):

akarnokd:
After some discussion with Grok, we came to the conclusion this PR is a result of an AI agent and as such is not disclosed.

In addition, your AI failed to respond to correction requests, which made me fix your AI's bad output.

[We clearly declared our stance on AI contributions](https://github.com/ReactiveX/RxJava/blob/4.x/CONTRIBUTING.md). You violated them.

You are hereby banned from this repository and will shame you for your failure too.

PR #8185, opened 2026-06-26
4.x cleanup: spinwait, grammar, minor syntax simplifications

akarnokd:
- Use `Thread.onSpinWait()`
- Javadoc insert `<p>` to correct styling.
- Some `signalled` -> `signaled`
- Some `cancelled` -> `canceled`


akarnokd:
Another weird crash on CI. Probably that spinwait introduced a runaway allocation somewhere?

PR #8206, opened 2026-06-30
4.x: Streamable + range, fromArray, take, error, defer

akarnokd:
Also

- Fix missing `onAssembly` calls
- Update `RxJavaPlugins` to support `Streamable` `onAssembly` calls
- Inlined a few `TestHelper.withXXX` into `RxJavaTest`.
- Added missing `extends RxJavaTest`s.
- Introduce `ThrowableWrapper` as a special class indicating it holds some exception that needs to be propagated.
- Make sure `Streamable` unwraps the `ThrowableWrapper`s the right time

### DIFF
--- a/src/main/java/io/reactivex/rxjava4/core/CompletionStageDisposable.java
+++ b/src/main/java/io/reactivex/rxjava4/core/CompletionStageDisposable.java
@@ -16,13 +16,14 @@ package io.reactivex.rxjava4.core;
 import java.io.Serial;
 import java.lang.ref.Cleaner;
 import java.util.Objects;
-import java.util.concurrent.CompletionStage;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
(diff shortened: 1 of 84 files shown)


hiSandog:
The `ThrowableWrapper` part is the one I would want especially well covered. It would be useful to have tests showing `Streamable.error`, `defer`, and plugin `onAssembly` paths unwrap exactly once and do not double-wrap fatal exceptions. That keeps the new API promotion from subtly changing error identity or plugin hooks compared with the existing reactive types.

akarnokd:
> The `ThrowableWrapper` part is the one I would want especially well covered. It would be useful to have tests showing `Streamable.error`, `defer`, and plugin `onAssembly` paths unwrap exactly once and do not double-wrap fatal exceptions. That keeps the new API promotion from subtly changing error identity or plugin hooks compared with the existing reactive types.

How are you using RxJava?

Are you a drive-by commenter without stakes or did you ask an AI to write a comment?

Issue #7912, opened 2025-10-02
Cache/Replay operators to be more GC friendly


akarnokd:
The current implementations of the various cache and replay operators and hot sources can hold onto effectively unreachable elements longer than necessary, causing GC pressure with certain collectors.

The following operators need to be checked and updated.

- [x] `ObservableCache`
- [ ] `FlowableCache`
- [ ] `ObservableReplay`
- [ ] `FlowableReplay`
- [ ] `ReplaySubject`
- [ ] `ReplayProcessor`

akarnokd:
`ObservableCache` - #7911

DragonFSKY:
I'd like to work on the FlowableCache item in #7912, using the ownership split from ObservableCache #7911 while preserving Flowable's per-subscriber request, cancellation, and serialized replay semantics. I'll add the same focused MemoryMXBean/System.gc regression test. Please let me know if someone is already working on this item.

akarnokd:
No one is working on these currently so go ahead 👍

PR #8247, opened 2026-07-27
[Problem] Fix FlowableCache GC issue

DragonFSKY:
## Summary

- Move the source-facing cache state and subscriber registry into a static `Multicaster`, so the upstream connection and active `CacheSubscription`s no longer retain the outer `FlowableCache`.
- Keep per-subscriber request accounting, cancellation, and serialized replay behavior unchanged.
- Retain the cache head only from `FlowableCache`, retain the current tail from `Multicaster`, and clear the tail on termination.
- Ignore post-terminal source signals now that termination clears the tail.
- Add a focused `MemoryMXBean` / `System.gc()` regression test matching the approach used for `ObservableCache` in #7911.

## Tests

- `./gradlew build --stacktrace`
- `./gradlew javadoc --stacktrace`
- `FlowableCacheTest.valuesAreReclaimable` repeated five times
- The regression test fails against the unmodified implementation because the cached payload remains retained.

## References

- Addresses the `FlowableCache` item in #7912.
- Follows the ownership split used for `ObservableCache` in #7911 while preserving Flowable-specific backpressure semantics.

## Checklist

- [x] The description explains what changed and why.
- [x] The related issue and reference PR are included.
- [x] A focused regression test is included; existing `FlowableCacheTest` coverage continues to exercise empty, range, error, backpressure, cancellation, and race behavior.

### DIFF
--- a/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableCache.java
+++ b/src/main/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableCache.java
@@ -28,23 +28,12 @@ import io.reactivex.rxjava3.plugins.RxJavaPlugins;
  *
  * @param <T> the source element type
  */
-public final class FlowableCache<T> extends AbstractFlowableWithUpstream<T, T>
-implements FlowableSubscriber<T> {
+public final class FlowableCache<T> extends AbstractFlowableWithUpstream<T, T> {
 
     /**
      * The subscription to the source should happen at most once.
      */
     final AtomicBoolean once;
-
-    /**
-     * The number of items per cached nodes.
-     */
-    final int capacityHint;
-
-    /**
-     * The current known array of subscriber state to notify.
-     */
-    final AtomicReference<CacheSubscription<T>[]> subscribers;
 
     /**
      * A shared instance of an empty array of subscribers to avoid creating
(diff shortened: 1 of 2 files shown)


akarnokd:
Looks okay, but how much do you need this in 3.x?

I'd prefer you primarily target 4.x first and then backport if someone does have a need for in 3.x. 

I don't want to release more 3.x versions that aren't bug or crash fixes. This issue is more like GC pedantics than a grave mistake.

akarnokd:
Also if you used AI, you have to disclose it, even for mechanistical changes.

I'm sorry but I have to do this on a high visibility attractive project like RxJava.

If you do not respond to change request in a reasonable time, indicated by your public github activity profile, or your contribution causes an effort amplification (i.e., I now have to port your changes over to 4.x and do all sorts of extra steps for no apparent reason or benefit), you will be banned and publicly shamed.

akarnokd (changes requested):
Please target 4.x first and we can discuss how much you, or anyone else needs this GC pedantic change backported into 3.x

DragonFSKY:
Thank you for the guidance. I have opened #8249 as a clean 4.x version of this change, targeting the current 4.x branch as requested. I am closing this PR as superseded, and I will only revisit a 3.x backport if there is a concrete need after the 4.x change is reviewed.

PR #8297, opened 2026-08-24
[Who asked for this? How is he still employed?] Upgrade actions/setup-java to v6

brunoborges:
Upgrade all active workflow references from actions/setup-java v5.7.0 to v6.0.0, pinned to the release commit SHA.

Testing: verified every active workflow reference uses the v6.0.0 SHA and ran git diff --check.

AI disclosure: GitHub Copilot was used to prepare this mechanical workflow update.

### DIFF
--- a/.github/workflows/gradle_branch.yml
+++ b/.github/workflows/gradle_branch.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
     - name: Set up JDK 26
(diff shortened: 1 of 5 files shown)


akarnokd:
Why?  Isn't Dependabot supposed to do it?

akarnokd:
Why on earth would you post a PR when we have dependabot and no urge to upgrade this action.

You are now banned from ever interacting with this repo.

brunoborges:
@akarnokd hey, yes dependabot is likely going to alert. This was more of an experiment on my end to alert repositories. Granted, I missed checking dependabot configuration. It was only meant to hit those that had config preventing major version alerts.

